PR #11, opened 2019-04-02
LT-906: I want to see the basic contents of the 'New Cases' work queue

jfaracik:
- Added model for Case and Queue
- Initial queue is created for 'New Cases'
- Cases are now created upon successful application submission
- Cases are then added to the default queue


cases/models.py
```diff
@@ -0,0 +1,10 @@
+import uuid
+
+from django.db import models
+
```
cstobart: White space

cases/serializers.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import serializers
+
```
cstobart: gaps in the imports

cases/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+
```
cstobart: gaps in the imports

applications/views.py
```diff
@@ -33,6 +35,15 @@ def post(self, request):
         draft.draft = False
         draft.save()
 
+        # Create a case
```
anirbanlahiri-dit: This needs to be in the form of a single atomic transaction. Otherwise the two models would become inconsistent

conf/urls.py
```diff
@@ -2,7 +2,9 @@
 
 urlpatterns = [
     path('applications/', include('applications.urls')),
+    path('cases/', include('cases.urls')),
     path('drafts/', include('drafts.urls')),
     path('organisations/', include('organisations.urls')),
     path('users/', include('users.urls')),
+    path('queues/', include('queues.urls')),
```
anirbanlahiri-dit: In alphabetic order the 'queues' need to be before 'users'

queues/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db import migrations, models
+import uuid
+
+
+def init(apps, schema_editor):
+    # We can't import the Queue model directly as it may be a newer
+    # version than this migration expects. We use the historical version.
+    Queue = apps.get_model('queues', 'Queue')
+    if not Queue.objects.all():
+        queue = Queue(id='00000000-0000-0000-0000-000000000000',
```
anirbanlahiri-dit: preferable to use '0000.....001' rather than '000....0000'. This would leave the UUID '000...000' reserved for any future special use.

queues/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
anirbanlahiri-dit: Include unit test for the queue model and remove the comment

applications/views.py
anirbanlahiri-dit: Use the following:

from django.db import transaction

@transaction.atomic
def post(...)

PR #15, opened 2019-04-09
fixes syntax error

cstobart:


conf/settings.py
```diff
@@ -46,6 +46,7 @@
     'cases.apps.CasesConfig',
     'queues.apps.QueuesConfig',
     'reversion',
+    ]
```
jfaracik: Spacing on the left of the bracket.

PR #20, opened 2019-04-24
fix: add easy start up support using docker

mh-hmcts:


conf/settings.py
```diff
@@ -27,9 +28,12 @@
 
 # Please use this to Enable/Disable the Admin site
 ADMIN_ENABLED = True
-
+ 
 ALLOWED_HOSTS = json.loads(env('ALLOWED_HOSTS')) if env('ALLOWED_HOSTS') else []
 
+if DEBUG:
+    ALLOWED_HOSTS = ['*']
```
jakub-kozlowski: I would remove this bit to keep `DEBUG` and `ALLOWED_HOSTS` uncoupled.

mh-hmcts:
fix: add easy start up support using docker

PR #29, opened 2019-05-03
LT-933-add-sites

cstobart:


organisations/serializers.py
```diff
@@ -1,16 +1,31 @@
...
+
+        class Meta:
+            model = Album
+            fields = ('album_name', 'artist', 'tracks')
+
+"""
+
+
+class OrganisationViewSerializer(serializers.ModelSerializer):
+    primary_site = PrimaryKeyRelatedField(queryset=Site.objects.all())
```
anirbanlahiri-dit: instead of objects.all() a filter would be more appropriate

PR #32, opened 2019-05-11
feat: LT-911 - Add notes to cases

jfaracik:
- Added endpoint for viewing case notes
- Added endpoint for creating case notes
- Add test for viewing and creating case notes using Parameterized


cases/models.py
```diff
@@ -8,3 +8,12 @@
 class Case(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     application = models.ForeignKey(Application, related_name='case', on_delete=models.CASCADE)
+
+
+@reversion.register()
+class CaseNote(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    case = models.ForeignKey(Case, related_name='case_note', on_delete=models.CASCADE)
+    # user = models.ForeignKey(User, related_name='case_note', on_delete=models.CASCADE)
```
jakub-kozlowski: we should be leaving commented out code - is there a special reason to have it here?

PR #37, opened 2019-05-29
LT-930 add departments

cstobart:


lukesa:
squash commits before merging

departments/tests/tests_create.py
```diff
@@ -0,0 +1,39 @@
...
+    def tests_create_department(self):
+        data = {
+            'name': 'new department'
+        }
+        response = self.client.post(self.url, data)
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(Department.objects.get().name, 'new department')
+
+    def tests_department_name_must_be_unique(self):
```
lukesa: parametrize it and combine with ```tests_department_name_must_be_unique_case_insensitive```
Should we ```trim()``` the department name from input? That would add a test case with spaces in front of the name

PR #57, opened 2019-06-18
feat: LT-1125 - Move cases across queues

lukesa:


cases/serializers.py
```diff
@@ -14,6 +14,14 @@ class Meta:
         fields = ('id', 'application')
 
 
+class CaseDetailSerializer(CaseSerializer):
```
cstobart: incorporate this into a single serialiser if you can, you can use SerializerMethodFields and reverse lookups for more return fields

queues/serializers.py
```diff
@@ -6,9 +6,14 @@
 
 class QueueSerializer(serializers.ModelSerializer):
     cases = CaseSerializer(many=True, read_only=True)
+    # team = serializers.SlugRelatedField(many=False,
```
cstobart: team_name = serializers.SerializerMethodField()

    def get_team_name(self, instance):
        return instance.team.name

jfaracik: The code is commented out until LT-919 is merged in

cases/serializers.py
jfaracik: Could you provide an example of how that'd be helpful in this situation?

cases/tests/tests_moving_cases_between_queues.py
```diff
@@ -0,0 +1,49 @@
...
+from cases.models import Case
+from teams.models import Team
+from test_helpers.clients import DataTestClient
+
+
+class MoveCasesTests(DataTestClient):
+
+    def setUp(self):
+        super().setUp()
+        self.draft = self.test_helper.create_draft_with_good_end_user_and_site('Example Application', self.test_helper.organisation)
```
anirbanlahiri-dit: please text wrap it

cases/tests/tests_moving_cases_between_queues.py
```diff
@@ -0,0 +1,49 @@
...
+        self.application = self.test_helper.submit_draft(self, self.draft)
+        self.case = Case.objects.get(application=self.application)
+        self.url = reverse('cases:case', kwargs={'pk': self.case.id})
+        self.team = Team.objects.get()
```
anirbanlahiri-dit: prefer to have a select clause in get() - otherwise it will fail if there are more than one queues

PR #63, opened 2019-06-27
feat: LT-947 - Assign users to cases

jfaracik:
- Cleaned up cases and queues relationship
- Added relationship between cases and gov users
- Added filtering for gov users
- Added appropriate tests


cases/views.py
```diff
@@ -26,38 +25,27 @@ def get(self, request, pk):
         serializer = CaseDetailSerializer(case)
         return JsonResponse(data={'case': serializer.data})
 
-    @swagger_auto_schema(
-        responses={
-            400: 'Input error, "queues" should be an array with at least one existing queue'
```
lukesa: we still return 400 in some cases so it should be defined in decorator for swagger

queues/tests/tests_case_assignment.py
```diff
@@ -0,0 +1,207 @@
...
+        self.default_team = Team.objects.get(id='00000000-0000-0000-0000-000000000001')
+
+        # Cases
+        self.case = Case.objects.get(application=self.application)
+        self.case2 = Case.objects.get(
+            application=self.test_helper.submit_draft(
+                self, self.test_helper.create_draft_with_good_end_user_and_site(
+                    'Example Application',
+                    self.test_helper.organisation)))
+        self.case3 = Case.objects.get(
```
lukesa: should it be a helper function?

addresses/serializers.py
```diff
@@ -34,7 +34,7 @@ def update(self, instance, validated_data):
         instance.postcode = validated_data.get('postcode', instance.postcode)
         instance.city = validated_data.get('city', instance.city)
         instance.region = validated_data.get('region', instance.region)
-        instance.country = validated_data.get('country', instance.country)
+        # instance.country = validated_data.get('country', instance.country)
```
lukesa: add explaining comment

cases/views.py
```diff
@@ -30,34 +30,27 @@ def get(self, request, pk):
         responses={
             400: 'Input error, "queues" should be an array with at least one existing queue'
         })
+    @transaction.atomic
     def put(self, request, pk):
```
lukesa: add CaseDetailSerializer to swagger decorator

PR #73, opened 2019-07-15
feat: LT-1127 - Control who can record the final decision

jfaracik:
* Views for getting all roles and permissions
* Views for adding and editing roles
* Permission class for certain roles


applications/views.py
```diff
@@ -115,6 +117,12 @@ def put(self, request, pk):
         Update an application instance.
         """
         with reversion.create_revision():
+            request_data = json.loads(request.body)
+
+            if request_data.get('status') == 'approved' or request_data.get('status') == 'declined':
+                if not has_permission(request.user, Permissions.MAKE_FINAL_DECISIONS):
+                    return JsonResponse(data={'application': 'Permission not found'}, status=403)
```
markm5A8Z9F43913A: Minimise the number of returns in each method.

cstobart: each return is distinct so needs to be kept separate

gov_users/views/roles_views.py
```diff
@@ -0,0 +1,95 @@
...
+    def get(self, request, pk):
+        """
+        Get the details of a specific role
+        """
+        role = get_role_by_pk(pk)
+        serializer = RoleSerializer(role)
+
+        return JsonResponse(data={'role': serializer.data})
+
+    def put(self, request, pk):
```
lukesa: need to add swagger decorator to bind RoleSerializer

teams/serializers.py
```diff
@@ -5,18 +5,14 @@
...
 class TeamSerializer(serializers.ModelSerializer):
-    name = serializers.CharField(
-        max_length=50,
-        validators=[UniqueValidator(queryset=Team.objects.all(), lookup='iexact',
-                                    message='Enter a name which is not already in use by another team')],
-        error_messages={'blank': 'Team name may not be blank'})
+    name = serializers.CharField(max_length=50,
+                                 validators=[
+                                     UniqueValidator(queryset=Team.objects.all(), lookup='iexact',
+                                                     message='Enter a name which is not already in use by another team')
```
lukesa: get_strings

teams/serializers.py
```diff
@@ -5,18 +5,14 @@
...
+                                 ],
+                                 error_messages={'blank': 'Team name may not be blank'})
```
lukesa: get_strings

PR #79, opened 2019-07-24
LT-1123 - Added All Cases and Open Cases system queues

markm5A8Z9F43913A:


queues/tests/tests_system_queues.py
```diff
@@ -0,0 +1,95 @@
...
+        self.queue1 = self.create_queue('queue1', self.team)
+        self.queue2 = self.create_queue('queue2', self.team)
+        self.case1 = self.create_application_case('case1 case1 for queue1')
+        self.case2 = self.create_application_case('case2 case2 case2 for queue2')
+        self.case3 = self.create_application_case('case3 case3 case3 for queue2')
+        self.case3.application.status = 'approved'
+        self.case3.application.save(update_fields=['status'])
+
+    def test_get_all_case_assignments(self):
+        case_assignment = CaseAssignment(queue=self.queue1, case=self.case1)
```
andytaylorkainos: Can we add the test scenario as comments please

andytaylorkainos: And also separate `Arrange` `Act` and `Assert` code with comments please

5pence: super().setUp() ... this can be super() and it will still go to the parent's setup (I believe)

efahy: @5pence As discussed, that `setup()` seems to be necessary

queues/tests/tests_system_queues.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def test_get_open_cases_system_queue(self):
+        """
+        Given that a number of cases exist and are assigned to different user defined queues
+        When a user the open cases system queue
+        Then only open cases are returned regardless of which user defined queues they are assigned to
+        """
+
+        # Arrange
+        url = '/queues/' + OPEN_CASES_SYSTEM_QUEUE_ID + '/'
```
jfaracik: User reverse() instead of hardcoding addresses

PR #82, opened 2019-07-30
fix notification for list pages

5pence:


goods/serializers.py
```diff
@@ -18,6 +18,7 @@ class GoodSerializer(serializers.ModelSerializer):
     status = serializers.ChoiceField(choices=GoodStatus.choices)
     not_sure_details_details = serializers.CharField(allow_blank=True, required=False)
     clc_query_case_id = serializers.SerializerMethodField()
+    clc_query = serializers.SerializerMethodField()
```
andytaylorkainos: i know this will require a front-end change, but can we use the name `clc_query_id` since its only the `id` we're returning and not the whole query object 😇

5pence: have renamed this to clc_query_id

PR #91, opened 2019-08-10
feat: LT-1192: Raise ECJU Queries

andytaylorkainos:


cases/libraries/get_ecju_queries.py
```diff
@@ -0,0 +1,17 @@
...
+    Returns all the ECJU queries from a case
+    """
+    return EcjuQuery.objects.filter(case=case).order_by('-created_at')
+
+
+def get_ecju_query(pk):
+    try:
+        return EcjuQuery.objects.get(pk=pk)
+    except EcjuQuery.DoesNotExist:
+        raise Http404
```
5pence: add message to 404 : raise Http404("ECJU Query does not exist")

andytaylorkainos: Done!

PR #92, opened 2019-08-13
test-coverage sh

lukesa:


bin/test-coverage.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+echo //////
+echo Note: Run from the main app directory "bin/test-coverage.sh"
+echo //////
+pipenv run coverage run --source='.' manage.py test
+pipenv run coverage report -m
+pipenv run coverage html
+open htmlcov/index.html
```
andytaylorkainos: new line please :)

bin/test-coverage.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+echo //////
+echo Note: Run from the main app directory "bin/test-coverage.sh"
+echo //////
+pipenv run coverage run --source='.' manage.py test
+pipenv run coverage report -m
```
andytaylorkainos: can we not parameterize your script to run only 1 report of `type`?

andytaylorkainos: for example `sh bin/test-coverage.sh html` would pas `html` as a parameter to the script.

you can grab that paremeter inside your script using `$1`
So line 6 could become `pipenv run coverage report $1` which evaluates as `pipenv run coverage report html`

bin/test-coverage.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+echo //////
+echo Note: Run from the main app directory "bin/test-coverage.sh"
+echo //////
+pipenv run coverage run --source='.' manage.py test
+pipenv run coverage report
```
andytaylorkainos: Parameterize the script please.
For example `sh bin/test-coverage.sh html` would pass `html` as a parameter to the script.

you can grab that paremeter inside your script using `$1`
So line 6 could become `pipenv run coverage report $1` which evaluates as `pipenv run coverage report html`

PR #95, opened 2019-08-16
feat: LT-1175 - Administer applications for multiple companies

5pence:
* Decoupled ExporterUser from Organisation
* Added a relationship model between them - this should contain all user/organisation centric data
* Added organisation header
* Finally _decimated_ AddressCountrylessSerializer - no more dodgy workarounds for sites
* Reworked how organisations and sites are returned from an organisation, far cleaner views and tests


gov_users/tests/tests_authenticate.py
```diff
@@ -52,13 +53,13 @@ def test_a_deactivated_user_cannot_log_in(self):
...
+    # @parameterized.expand([
+    #     [{'headers': {'HTTP_GOV_USER_EMAIL': str('test@mail.com')}, 'response': status.HTTP_200_OK}],
+    #     [{'headers': {}, 'response': status.HTTP_403_FORBIDDEN}],
+    #     [{'headers': {'HTTP_GOV_USER_EMAIL': str('sadkjaf@asdasdf.casdas')}, 'response': status.HTTP_403_FORBIDDEN}],
+    # ])
+    # def test_authorised_valid_email_in_header(self, data):
+    #     url = reverse('gov_users:gov_users')
+    #     headers = data['headers']
+    #     response = self.client.get(url, **headers)
+    #     self.assertEqual(response.status_code, data['response'])
```
efahy: Can this be removed?

conf/authentication.py
```diff
@@ -2,92 +2,83 @@
...
-            else:
-                user = GovUser.objects.get(email=email)
-        except GovUser.DoesNotExist:
-            raise exceptions.PermissionDenied(USER_DOES_NOT_EXIST_ERROR)
-
-        if user.status == GovUserStatuses.DEACTIVATED:
-            raise exceptions.PermissionDenied(USER_DEACTIVATED_ERROR)
+        """
+        When given a user token and an organisation id, validate that the user belongs to the
+        organisation and that they're allowed to access that organisation
```
efahy: Comment needs updated, because this doesn't validate the org bit

users/libraries/get_user.py
```diff
@@ -13,3 +13,34 @@ def get_user_by_pk(pk):
             return GovUser.objects.get(pk=pk)
         except GovUser.DoesNotExist:
             raise NotFoundError({'user': 'User not found - ' + str(pk)})
+
+
+def get_user_by_email(email):
```
efahy: Personally speaking I'd make separate library methods to get internal users by e-mail and exporter users by e-mail and do the orchestration between the two wherever they're used

users/serializers.py
```diff
@@ -47,31 +65,74 @@ class Meta:
...
-    organisation = serializers.PrimaryKeyRelatedField(queryset=Organisation.objects.all(), required=False)
+    organisation = serializers.PrimaryKeyRelatedField(queryset=Organisation.objects.all(),
+                                                      required=False,
+                                                      write_only=True)
 
     class Meta:
         model = ExporterUser
         fields = ('id', 'email', 'first_name', 'last_name', 'organisation')
 
+    def validate_email(self, email):
```
efahy: It seems odd that you're doing the org validation in a method that is supposedly validating e-mail addresses.  Can you move it to a `validate_organisation` method?

users/tests/tests_deactivate.py
```diff
@@ -1,59 +1,59 @@
...
+    #     }
+    #     self.client.put(url, data, **self.exporter_headers)
+    #     url = reverse('users:authenticate')
+    #     data = {
+    #         'email': user.email,
+    #         'first_name': user.first_name,
+    #         'last_name': user.last_name
+    #     }
+    #     response = self.client.post(url, data)
+    #     self.assertEqual(response.status_code, status.HTTP_200_OK)
```
efahy: Can you remove all this?

users/models.py
```diff
@@ -73,14 +78,20 @@ def __str__(self):
...
 
 class ExporterUser(BaseUser):
-    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE, default=None, null=True)
-
     def send_notification(self, case):
         from cases.models import Notification
         Notification.objects.create(user=self, note=case)
 
 
+class UserOrganisationRelationship(models.Model):
```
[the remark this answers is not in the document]
jfaracik: I get where you're coming from - but I'd rather do the reverse, for example instead of GoodOnApplication it should be GoodApplicationRelationship. We can talk more about this when we get to the Application model refactor :)

users/libraries/get_user.py
jfaracik: Good suggestion

users/serializers.py
jfaracik: I did it this way so that the serialiser returns the correct key (in this case _email_) - This is used on the frontend to display the correct styling on the email field.

PR #96, opened 2019-08-17
bugfix: LT-1444: Sort cases on system queues

andytaylorkainos:


queues/tests/tests_filter_and_sorting_cases.py
```diff
@@ -127,7 +127,7 @@ def test_get_all_cases_queue_submitted_status_and_clc_type_cases(self):
         case_status = get_case_status_from_status(CaseStatusEnum.SUBMITTED)
         clc_submitted_cases = list(filter(lambda case: case.clc_query.status == case_status, self.clc_cases))
         url = reverse('queues:queue', kwargs={'pk': 'de13c40a-b330-4d77-8304-57ac12326e5a'}
-                      ) + '?case_type=clc_query&status=' + case_status.status
+                      ) + '?case_type=CLC%20query&status=' + case_status.status + '&sort={"status":"asc"}'
```
5pence: why are we using a space in the query parameters - can this be changed to be avoided

5pence: also there are CircleCI errors that need to be resolved

andytaylorkainos: It's not white space, although i know what you mean.
Unforunately the `case_type` is now defined this way - so unless we do something like use `clc-query` as the query parameter and map it to `clc_query` on the back-end, there's no other way to avoid it

5pence: fair enough :)

PR #101, opened 2019-08-24
feat: LT-1185 - Goods flags

cstobart:
* Gov users can add flags to goods which persist between cases, this includes standard goods, goods on clc queries and goods types on open licenses.

* Pagination added to the queue detail views and the data payload largely cut down resulting in much faster load times

* A system queue to view all cases which are within my team's queues


goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,177 @@
...
+        Then the correct Good with an empty Flag list is returned
+        """
+
+        # Arrange
+
+        # Act
+        response = self.client.get(self.good_url, **self.gov_headers)
+
+        # Assert
+        self.assertEqual([], response.json()['good']['flags'])
```
efahy: Nicely commented throughout these tests and the setup reads easily - nice work :)

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,177 @@
...
+        When a user requests the Good
+        Then the correct Good with all assigned Flags are returned
+        """
+
+        # Arrange
+        self.good.flags.set(self.all_flags)
+
+        # Act
+        response = self.client.get(self.good_url, **self.gov_headers)
+        returned_good = response.json()['good']
```
efahy: Personally, only the get API call is the Act.  Getting the returned_good out of the response is part of the Assert... but not a big deal!

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,177 @@
...
+        flags_to_add = {'goods': [self.good.pk], 'flags': [self.other_team_good_flag.pk], 'note': 'A reason for changing the flags'}
+
+        # Act
+        response = self.client.put(self.good_flag_url, flags_to_add, **self.gov_headers)
+
+        # Assert
+        self.assertEquals(0, len(self.good.flags.all()))
+        self.assertEquals(status.HTTP_400_BAD_REQUEST, response.status_code)
+
+    def test_user_cannot_assign_flags_that_are_not_good_level(self):
```
efahy: If you exclude deactivated flags based on my other comment, you might consider adding another test here

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Assert
+        self.assertEquals(len(self.all_flags), len(self.good.flags.all()))
+        for flag in self.all_flags:
+            self.assertTrue(flag in self.good.flags.all())
+
+    def test_given_good_has_been_modified_then_appropriate_audit_is_in_place(self):
+        """
+        Given a Good with no Flags assigned
+        When a user attempts to add a non-good-level Flag owned by their Team to the Good
```
efahy: `non-good-level` should be `good-level`, I think

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,177 @@
...
+        self.client.put(self.good_flag_url, flags, **self.gov_headers)
+        response = self.client.get(self.audit_url, **self.gov_headers)
+
+        # Assert
+        response_data = response.json()
+        activity = response_data['activity']
+        self.assertEquals(len(flags['flags']), len(activity))
+        self.assertEquals([self.team_good_flag_1.__dict__['name']], activity[0]['data']['flags']['added'])
+
+    def test_setting_flags_on_two_goods(self):
```
efahy: Add comments as per other tests please

goods/views.py
```diff
@@ -212,3 +222,67 @@ def delete(self, request, pk, doc_pk):
...
+        remove_good_flags = [flag.name for flag in previously_assigned_team_flags if flag not in validated_data]
+
+        with reversion.create_revision():
+            reversion.set_comment(
+                ('{"flags": {"removed": ' + str(remove_good_flags) + ', "added": ' + str(add_good_flags) + ', "note": "' + str(note) + '"}}')
+                .replace('\'', '"')
+            )
+            reversion.set_user(user)
+
+            good.flags.set(validated_data + list(previously_assigned_not_team_flags))
```
efahy: I take it the fact that you are effectively replacing the whole list of flags on the good including other teams' flags, rather than taking a more precise approach, doesn't affect reversion/audit?  It would not be good if doing this gives us incorrect audit records

cases/views.py
```diff
@@ -131,47 +132,21 @@ def get(self, request, pk):
         for ecju_query in ecju_queries:
             activity.append(convert_ecju_query_to_activity(ecju_query))
 
+        for good in goods:
```
jfaracik: Method

flags/views.py
```diff
@@ -88,3 +91,38 @@ def put(self, request, pk):
...
+
+
+class AssignFlags(APIView):
+    authentication_classes = (GovAuthentication,)
+    def put(self, request):
+        data = JSONParser().parse(request)
+        level = data.get('level')
+        response_data = []
+
+        for pk in data.get('objects'):
```
jfaracik: Have the serialiser handle the loop

eg flag_assignment_serializer(many=True)

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,179 @@
...
+        self.audit_url = reverse('goods:activity', kwargs={'pk': self.good.id})
+
+    def test_no_flags_for_good_are_returned(self):
+        """
+        Given a Good with no Flags assigned
+        When a user requests the Good
+        Then the correct Good with an empty Flag list is returned
+        """
+
+        # Arrange
```
jfaracik: Remove comments

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        Given a Good with no Flags assigned
+        When a user attempts to add a good-level Flag owned by their Team to the Good
+        Then the Flag is successfully added
+        """
+
+        # Arrange
+        data = {
+            'level': 'goods',
+            'objects': [self.good.pk],
```
jfaracik: Would prefer if this was 'goods' instead of 'objects'

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def test_user_cannot_assign_flags_that_are_not_owned_by_their_team(self):
+        """
+        Given a Good with no Flags assigned
+        When a user attempts to add a good-level Flag not owned by their Team to the Good
+        Then the Flag is not added
+        """
+
+        # Arrange
+        flags_to_add = {'level': 'goods', 'objects': [self.good.pk], 'flags': [self.other_team_good_flag.pk], 'note': 'A reason for changing the flags'}
```
jfaracik: not all on one line

goods/tests/tests_managing_flags_on_good.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Act
+        response = self.client.put(self.good_flag_url, flags_to_add, **self.gov_headers)
+
+        # Assert
+        self.assertEquals(0, len(self.good.flags.all()))
```
jfaracik: Invert that

cases/libraries/activity_helpers.py
```diff
@@ -82,3 +83,30 @@ def convert_case_reversion_to_activity(version: Version):
...
                           data)
+
+
+def convert_good_reversion_to_activity(version: Version):
+    """
+    Converts an audit item to a dict suitable for the case activity list
+    """
+    revision_object = Revision.objects.get(id=version.revision_id)
+    user = get_user_by_pk(revision_object.user.id)
+    activity_type = CHANGE
```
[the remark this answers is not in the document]
cstobart: It is needed so there isn't an unset variable in the return value in the case of the try block failing

flags/views.py
cstobart: it screamed

PR #110, opened 2019-09-06
fix: LT-1485 Correct notification counts to be specific to user and org

efahy:


users/tests/tests_notifications.py
```diff
@@ -42,3 +47,42 @@ def tests_create_both_clc_and_application_notifications(self):
...
         self.assertEqual(Notification.objects.filter(case_note__case__clc_query_id__isnull=True).count(), 4)
         self.assertEqual(Notification.objects.filter(case_note__case__application_id__isnull=True).count(), 3)
+
+    def tests_get_notifications_for_user_in_multiple_orgs(self):
+        """
+        Given an exporter user in multiple orgs
+        When an API user gets notifications for the exporter user and one of their orgs
+        Then the notifications specific to that user and org combination are returned
+        """
+        # Assemble
```
jfaracik: No assemble/act/assert comments please

users/tests/tests_notifications.py
```diff
@@ -42,3 +47,42 @@ def tests_create_both_clc_and_application_notifications(self):
...
+        org2 = self.create_organisation_with_exporter_user('Org 2')
```
jfaracik: (Unwritten) variable name convention is org_2 and application_1, case_1 etc

users/tests/tests_notifications.py
```diff
@@ -1,8 +1,13 @@
+from django.urls import reverse
+from rest_framework import status
+
 from cases.models import Case, Notification
 from test_helpers.clients import DataTestClient
+from users.libraries.user_to_token import user_to_token
 
 
 class NotificationTests(DataTestClient):
+    url = '/users/notifications/'
```
jfaracik: Use reverse_lazy, don't hardcore urls

users/tests/tests_notifications.py
efahy: Why?

users/tests/tests_notifications.py
```diff
@@ -42,3 +47,42 @@ def tests_create_both_clc_and_application_notifications(self):
...
+        org2 = self.create_organisation_with_exporter_user('Org 2')
+        self.add_exporter_user_to_org(org2, self.exporter_user)
+
+        application1 = self.create_standard_application(self.organisation)
+        case1 = Case.objects.get(application=application1)
+
+        self.create_case_note(case1, 'This is a test note 1', self.gov_user, True)
+        self.create_case_note(case1, 'This is a test note 2', self.gov_user, True)
+
+        self.exporter_headers = {'HTTP_EXPORTER_USER_TOKEN': user_to_token(self.exporter_user),
```
jfaracik: (Unwritten) dict layout is:

{
       'key': 'value'
}

Only do one line if you only have one key in the dict

users/tests/tests_notifications.py
```diff
@@ -42,3 +47,42 @@ def tests_create_both_clc_and_application_notifications(self):
...
+
+        application2 = self.create_standard_application(org2)
+        case2 = Case.objects.get(application=application2)
+
+        self.create_case_note(case2, 'This is a test note 3', self.gov_user, True)
+        self.create_case_note(case2, 'This is a test note 4', self.gov_user, True)
+
+        # Act
+        response = self.client.get(self.url, **self.exporter_headers)
+
```
jfaracik: response_data goes under response, it doesn't belong in the assert section

users/tests/tests_notifications.py
efahy: This is hard coded in the front end, so I chose not to change the URL definition and potentially affect front end code and hard coded it here too

users/tests/tests_notifications.py
efahy: Will do.  Will also fix in `clients.py` where I got the original from

users/tests/tests_notifications.py
efahy: I disagree.  The thing being tested (the Act) is the API get.  Looking at what it returns, or preparing to do so, is not part of what's being tested

users/tests/tests_notifications.py
efahy: Especially since below you comment on what should be in the Act vs the Assemble

jfaracik: I think it should be obvious by the code and also layout of said code

Adding a single word comment doesn't particularly help anything

users/tests/tests_notifications.py
jfaracik: All frontend URLs are hardcoded not out of choice, and even then we have a constants file for them to maintain consistency.

API we have the choice to not hardcode, therefore we shouldn't.

`reverse_lazy('users:notifications')`

users/tests/tests_notifications.py
jfaracik: Cheers

users/tests/tests_notifications.py
efahy: The above results in `django.urls.exceptions.NoReverseMatch: Reverse for 'notifications' not found. 'notifications' is not a valid view function or pattern name.`.  Will look into how to get this working.  I assume its cause the URLs are set up like so: `router.register(r'notifications', views.NotificationViewset)`

jfaracik:
```
    # ex: /notifications/
    path('/clc_notifications/', views.ClcNotificationViewset.as_view(), name='notifications'),
    # ex: /clc_notifications/
    path('/clc_notifications/', views.ClcNotificationViewset.as_view(), name='clc_notifications'),```

Add this, then you can remove the router stuff at the top and bottom of the file

File: users/tests/tests_notifications.py (hunk 1 above)
Comment (efahy) [reply users/tests/tests_notifications.py]: I think it makes the tests easier to read at a glance for not a lot of effort.  And we've used it in quite a lot of API tests now.  I'm reluctant to remove them

File: users/tests/tests_notifications.py (hunk 5 above)
Comment (jfaracik) [reply users/tests/tests_notifications.py]: I get that, but you're not asserting anything with that line (nor are you acting upon it yada yada yada). I think its cleaner to have it under response, and also more consistent with the 200 other odd tests that we have written that way.

I was looking at writing a client wrapper (self.get), which returns the response, status_code - which could possibly make both of us happy.

File: users/tests/tests_notifications.py (hunk 1 above)
Comment (jfaracik) [reply users/tests/tests_notifications.py]: I see newlines as being simple enough to separate assemble/act/assert - considering it's already a consistent approach across the rest of the tests. I see it as just needless lines which don't provide value or insight into those sections

File: users/tests/tests_notifications.py (hunk 3 above)
Comment (efahy) [reply users/tests/tests_notifications.py]: `TypeError: The `actions` argument must be provided when calling `.as_view()` on a ViewSet. For example `.as_view({'get': 'list'})``

Comment (jfaracik) [reply users/tests/tests_notifications.py]:
```
    # ex: /notifications/
    path('notifications/', views.NotificationViewset.as_view(), name='notifications'),
    # ex: /clc_notifications/
    path('clc_notifications/', views.ClcNotificationViewset.as_view(), name='clc_notifications'),
```

Comment (jfaracik) [reply users/tests/tests_notifications.py]: And then in views change the classes to extend from generics.ListAPIView

Comment (efahy) [reply users/tests/tests_notifications.py]: Done.  Thanks for the pointers

Comment (jfaracik) [reply users/tests/tests_notifications.py]: No problem

File: users/tests/tests_notifications.py (hunk 1 above)
Comment (efahy) [reply users/tests/tests_notifications.py]: I still prefer tests written with these comments but I take your consistency argument and will remove them

Comment (jfaracik) [reply users/tests/tests_notifications.py]: Cheers

File: users/tests/tests_notifications.py (hunk 5 above)
Comment (efahy) [reply users/tests/tests_notifications.py]: Irrelevant as we are removing these comments now

Comment (efahy) [reply users/tests/tests_notifications.py]: However, I have moved the `response_data` line up as suggested

File: users/urls.py [hunk 6]
```diff
@@ -18,4 +13,8 @@
     path('<uuid:pk>/', views.UserDetail.as_view(), name='user'),
     # ex: /users/me/
     path('me/', views.UserMeDetail.as_view(), name='me'),
-] + router.urls
+    # ex: /notifications/
```
Mark-Spiers-kainos: I believe these comments should be "/users/...."

efahy: Changed to `/users/notifications/` and `/users/clc_notifications/`

PR #121, opened 2019-09-18
feat: LT-1483 - Detail view for End User Advisories

jfaracik:


queries/end_user_advisories/tests.py
```diff
@@ -19,6 +19,14 @@ def test_view_end_user_advisory_query(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEquals(len(response_data), 1)
 
+    def test_view_end_user_advisory_query(self):
+        query = self.create_end_user_advisory('a note', 'because I\'m unsure', self.organisation)
```
lukesa: should it come from contend strings?

queries/end_user_advisories/views.py
```diff
@@ -44,3 +44,12 @@ def post(self, request):
         except serializers.ValidationError as e:
             return JsonResponse(data={'errors': e},
                                 status=status.HTTP_400_BAD_REQUEST)
+
+
+class EndUserAdvisoryDetail(APIView):
+    authentication_classes = (ExporterAuthentication,)
+
+    def get(self, request, pk):
```
lukesa: docstring!

queries/end_user_advisories/tests.py
jfaracik: This is a test, so nope.

queries/end_user_advisories/views.py
```diff
@@ -12,6 +12,9 @@ class EndUserAdvisoriesList(APIView):
     authentication_classes = (ExporterAuthentication,)
 
     def get(self, request):
+        """
+        View a single end user advisory's details
```
Mark-Spiers-kainos: I believe this should say "list" rather than single

queries/end_user_advisories/tests.py
```diff
@@ -19,6 +19,14 @@ def test_view_end_user_advisory_query(self):
...
+
+        response = self.client.get(reverse('queries:end_user_advisories:end_user_advisory',
```
stephanieboyd: Could we make these 2 tests in `EndUserAdvisoryViewTests` consistent in how they do the get? It will also clean up the class variable `url` that's only used in 1 of the 2 tests.

jfaracik: Good suggestion

queries/end_user_advisories/views.py
jfaracik: Done

queries/end_user_advisories/tests.py
```diff
@@ -19,6 +19,14 @@ def test_view_end_user_advisory_query(self):
...
+                                           kwargs={'pk': query.id}), **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
stephanieboyd: Are we asserting enough here? It looks like we should really check that the data back is as we expect

jfaracik: Added

PR #123, opened 2019-09-19
LT-1465-optional-advice

BenAndrew-Kainos:


cases/views/views.py
```diff
@@ -145,9 +146,11 @@ def post(self, request, pk):
         for advice in data:
             advice['case'] = str(self.case.id)
             advice['user'] = str(request.user.id)
+            if advice['type'].lower() == 'refuse' and not advice['text']:
+                return JsonResponse({'errors': [{'text': [ErrorDetail(string='Enter some advice', code='blank')]}]},
```
markm5A8Z9F43913A: "You need to enter advice when refusing a case" might be better. Also should we put it into lite-content instead?

BenAndrew-Kainos: Yeah good idea

cases/views/views.py
```diff
@@ -145,9 +147,13 @@ def post(self, request, pk):
         for advice in data:
             advice['case'] = str(self.case.id)
             advice['user'] = str(request.user.id)
+            if advice['type'].lower() == 'refuse' and not advice['text']:
+                return JsonResponse(
+                    {'errors': [
+                        {'text': [ErrorDetail(string=get_string('cases.advice_refusal_error'), code='blank')]}
+                    ]}, status=status.HTTP_400_BAD_REQUEST)
```
lukesa: please add swagger decorator to add 400 response code

PR #130, opened 2019-09-24
feat: LT-1143 - Valid control list entry

jfaracik:


static/control_list_entries/models.py
```diff
@@ -0,0 +1,39 @@
+import uuid
+
+from django.db import models
+
+
+class ControlListEntry(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    rating = models.CharField(max_length=100, null=True, blank=True)
```
a-bek: Should uniqueness be enforced here?

static/control_list_entries/models.py
```diff
@@ -0,0 +1,39 @@
...
+                                 text=text,
+                                 parent=parent,
+                                 is_decontrolled=is_decontrolled)
+        control_list_entry.save()
+        return control_list_entry
+
+    @classmethod
+    def create_or_update(cls, rating, text, parent, is_decontrolled):
+        try:
+            control_list_entry = ControlListEntry.objects.get(rating=rating)
```
a-bek: Nothing is stopping someone from adding another ControlListEntry with a duplicate rating, in which case this would fail.

static/control_list_entries/models.py
```diff
@@ -0,0 +1,39 @@
...
+    def create(cls, rating, text, parent, is_decontrolled):
+        control_list_entry = cls(rating=rating,
+                                 text=text,
+                                 parent=parent,
+                                 is_decontrolled=is_decontrolled)
+        control_list_entry.save()
+        return control_list_entry
+
+    @classmethod
+    def create_or_update(cls, rating, text, parent, is_decontrolled):
```
a-bek: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#update-or-create

static/control_list_entries/helpers.py
```diff
@@ -0,0 +1,21 @@
+from conf.exceptions import NotFoundError
```
a-bek: Should these functions not be embedded in the model class through the use of a manager or something?

https://docs.djangoproject.com/en/2.2/topics/db/managers/

static/control_list_entries/models.py
jfaracik: Some control list entries don't have ratings, so unless we can have unique=True and blank/null=True at the same time sadly not

jfaracik: Scratch that, we're getting rid of null ratings - it can be unique now :)

PR #132, opened 2019-09-24
feat: LT-1417 - Individual Exporter Account

cstobart:


organisations/models.py
```diff
@@ -5,17 +5,19 @@
...
 @reversion.register()
 class Organisation(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     name = models.TextField(default=None, blank=True)
-    eori_number = models.TextField(default=None, blank=True)
-    sic_number = models.TextField(default=None, blank=True)
-    vat_number = models.TextField(default=None, blank=True)
-    registration_number = models.TextField(default=None, blank=True)
+    sub_type = models.CharField(choices=SubType.choices, default=SubType.COMMERCIAL, max_length=20)
+    eori_number = models.TextField(default=None, blank=True, null=True)
```
BenAndrew-Kainos: Why do fields need to be blank and null? (Just curious)

organisations/tests/tests_main.py
```diff
@@ -90,3 +93,91 @@ def test_cannot_create_organisation_with_invalid_data(self):
...
+        self.assertEqual(exporter_user.first_name, data['user']['first_name'])
+        self.assertEqual(exporter_user.last_name, data['user']['last_name'])
+
+        self.assertEqual(site.name, data['site']['name'])
+        self.assertEqual(site.address.address_line_1, data['site']['address']['address_line_1'])
+        self.assertEqual(site.address.address_line_2, data['site']['address']['address_line_2'])
+        self.assertEqual(site.address.region, data['site']['address']['region'])
+        self.assertEqual(site.address.postcode, data['site']['address']['postcode'])
+        self.assertEqual(site.address.city, data['site']['address']['city'])
+        self.assertEqual(str(site.address.country.id), data['site']['address']['country'])
```
andytaylorkainos: new line please

BenAndrew-Kainos: newline

organisations/serializers.py
```diff
@@ -85,11 +88,32 @@ class Meta:
                   'user',
                   'site')
 
+    standard_blank_error_message = 'This field may not be blank'
+
+    def validate_eori_number(self, attrs):
```
a-bek: Maybe change `attrs` to `value` as `attrs` suggests a list / dict.

organisations/models.py
cstobart: They can either be blank or not existent when saving the model

organisations/serializers.py
cstobart: value does not work, tried it yesterday, it needs the parameter to be called attrs for some reason (and it keeps it consistent with the base method)

organisations/serializers.py
```diff
@@ -85,11 +88,32 @@ class Meta:
...
+        if self.initial_data.get('sub_type') != SubType.INDIVIDUAL and not attrs:
+            raise serializers.ValidationError(self.standard_blank_error_message)
+        return attrs
```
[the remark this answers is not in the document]
cstobart: They can't be batched otherwise the entire class method validate will be overridden

organisations/views/main.py
```diff
@@ -35,6 +35,13 @@ def post(self, request):
         """
         with reversion.create_revision():
             data = JSONParser().parse(request)
+            if data['sub_type'] == 'individual':
+                try:
+                    data['name'] = data['user']['first_name'] + " " + data['user']['last_name']
```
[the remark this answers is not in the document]
a-bek: Oh it should get caught in the serializer validation below...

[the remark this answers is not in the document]
cstobart: We need to continue as we got though this end point several times and the order in which the data comes through changes depending on the flow, we don't have first name and last name until the final run through in the case of the commercial organisation being registered, the serialiser will catch anything bad going through

organisations/serializers.py
a-bek: It might have been failing for another reason.

https://github.com/encode/django-rest-framework/blob/e39d8410de26003421f7e644f9206833a3b9bbe0/rest_framework/serializers.py#L394

The source doesn't seem to pass the value using key-value pairing ie `func(data)` vs `func(attrs=data)`.

cstobart: 👍

organisations/serializers.py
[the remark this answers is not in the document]
cstobart: And they will have specific validation at a later date, e.g. for other exporter types

PR #135, opened 2019-09-25
Feature/cleardb

andytaylorkainos:


cases/migrations/0003_auto_20190919_1037.py
```diff
@@ -19,6 +18,11 @@ def init(apps, schema_editor):
         Permission.objects.get(id='MAKE_FINAL_DECISIONS').delete()
 
 
+def destroy(apps, schema_editor):
```
efahy: Please add comments with stars and lots of attention grabbing things saying that `initialize` and `destroy` need to be manually preserved in the case of full re-make-migrations (we are aware that we're not meant to do full re-make-migrations but in practical terms we all know it happens occasionally)

static/management/commands/cleardb.py
```diff
@@ -0,0 +1,34 @@
...
+        if not options['safe'] or options['safe'] != 'False':
+            print('\nSafely removing database..\n')
+            execute_bash_command('./manage.py flush --no-input', shell=True)
+            for app in apps.get_app_configs():
+                if app.get_models():
+                    app_name_index = app.name.rfind('.')
+                    app_name = app.name[app_name_index + 1:] if app_name_index > -1 else app.name
+                    execute_bash_command('./manage.py migrate ' + app_name.lower() + ' zero', shell=True)
+        else:
+            print('\nForcefully removing database..\n')
```
efahy: Forcefully dropping all database tables

andytaylorkainos: Done!

PR #136, opened 2019-09-25
[NEXT IN MERGE QUEUE AFTER BUG] feat: LT-1363 - Specific Countries For Open Goods

cstobart:


goodstype/views.py
```diff
@@ -49,3 +52,34 @@ def get(self, request, pk):
...
+        """
+        Accepts dict in style of {goddstype_id: [country_id, country_id]}
+        """
+
+        data = JSONParser().parse(request)
+        if data.get('csrfmiddlewaretoken'):
+            data.pop('csrfmiddlewaretoken')
+
+        # validation
+        for assignment in data:
```
a-bek: When iterating dicts you should use the builtin dict items method:

For example:
```
for pk, country_ids in data.items():
    get_goods_type(pk)
    for country_id in country_ids:
        get_country(country_id)
```

goodstype/views.py
```diff
@@ -49,3 +52,34 @@ def get(self, request, pk):
...
+            get_goods_type(assignment)
+            values = data.get(assignment)
+            for country_code in values:
+                get_country(country_code)
+
+        # persist
```
a-bek: Can the validation and persist blocks be merged? Otherwise you are hitting the database twice for all goods and countries!

goodstype/views.py
```diff
@@ -49,3 +52,34 @@ def get(self, request, pk):
...
+    authentication_classes = (ExporterAuthentication,)
+
+    @transaction.atomic
+    def put(self, request):
+        """
+        Accepts dict in style of {goddstype_id: [country_id, country_id]}
+        """
+
+        data = JSONParser().parse(request)
+        if data.get('csrfmiddlewaretoken'):
```
a-bek: Shouldn't data be retrievable by `request.data`? How come it is at the same level as `csrfmiddlewaretoken`?

goodstype/views.py
```diff
@@ -49,3 +52,25 @@ def get(self, request, pk):
...
+    @transaction.atomic
+    def put(self, request):
+        """
+        Accepts {goodstype_id: [country_id, country_id], goodstype_id: [etc...], etc...}
+        """
+        data = JSONParser().parse(request)
+
+        for good, countries in data.items():
+            good = get_goods_type(good)
+            for country_code in countries:
```
a-bek: What do you think of, as an optimisation:

`Country.objects.filter(pk__in=countries).count() == len(countries)` to validate they all exist?

This is 1 query vs N queries and doesn't load any country data.

Also, is there ever a case where a country may not exist, does the check even need to happen?

goodstype/serializers.py
```diff
@@ -38,7 +38,8 @@ class Meta:
                   'content_type',
                   'content_type_name',
                   'object_id',
-                  'content_object'
+                  'content_object',
+                  'countries'
                   )
```
jfaracik: )

goodstype/tests/tests_managing_countries_on_goods_type.py
```diff
@@ -0,0 +1,139 @@
...
+        # Add a country to the draft
+        self.country_1 = get_country('ES')
+        self.country_2 = get_country('US')
+        self.country_3 = get_country('FR')
+
+        self.all_countries = [self.country_1, self.country_2, self.country_3]
+        for country in self.all_countries:
+            CountryOnDraft(draft=self.open_draft, country=country).save()
+
+        self.good_url = reverse('goodstype:goodstypes-detail', kwargs={'pk': self.goods_type_1.id})
```
jfaracik: goodstypes_detail - underscore please :)

PR #139, opened 2019-09-28
enhance: Response Serialisers

cstobart:


goods/views.py
```diff
@@ -35,84 +34,51 @@ def get(self, request):
...
-            return JsonResponse(data={'good': serializer.data},
-                                status=status.HTTP_201_CREATED)
-
-        return JsonResponse(data={'errors': serializer.errors},
-                            status=status.HTTP_400_BAD_REQUEST)
+        return response_serializer(GoodSerializer,
+                                   data=data,
+                                   object_class=Good,
+                                   request=request,
+                                   pre_validation_actions=[add_organisation_to_data])
```
efahy: I'll be honest Chris and say that this enhancement reduces the readability of our view code quite a bit.  I get that it reduces the number of lines of code somewhat but for me you're trying to solve a problem that doesn't exist and furthermore I'd worry that the `response_serializer` won't cope with all our use cases in future without being extended

PR #140, opened 2019-09-30
Lt 1334 goods country decision matrix

cstobart:


cases/models.py
```diff
@@ -283,3 +283,23 @@ def create(cls, activity_type, case, user, additional_text=None, created_at=None
 class CaseActivity(BaseActivity):
     case = models.ForeignKey(Case, on_delete=models.CASCADE, null=False)
     activity_types = CaseActivityType
+
+
+class CaseGoodCountryDecision(models.Model):
```
a-bek: Would `Decision` be a more suitable name for the model rather than embedding all the fields in the name?

cases/models.py
```diff
@@ -283,3 +283,23 @@ def create(cls, activity_type, case, user, additional_text=None, created_at=None
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    case = models.ForeignKey(Case, on_delete=models.CASCADE)
+    good = models.ForeignKey(GoodsType, on_delete=models.CASCADE)
+    country = models.ForeignKey(Country, on_delete=models.CASCADE)
+    advice_type = models.CharField(choices=AdviceType.choices, max_length=30)
+
+    def save(self, *args, **kwargs):
+        try:
+            existing_object = CaseGoodCountryDecision.objects.get(case=self.case,
```
a-bek: I think `Decision.objects.filter(case=self.case, good=self.good, country=self.country).delete()` would optimise this and prevents the need to use a try/block.

cases/serializers.py
```diff
@@ -343,3 +343,14 @@ class CaseActivitySerializer(serializers.ModelSerializer):
     class Meta:
         model = CaseActivity
         fields = '__all__'
+
+
+class CaseGoodCountryDecisionSerializer(serializers.ModelSerializer):
```
a-bek: `DecisionSerializer`?

cases/views/views.py
```diff
@@ -397,3 +399,31 @@ def put(self, request, pk, ecju_pk):
 
         return JsonResponse(data={'errors': serializer.errors},
                             status=status.HTTP_400_BAD_REQUEST)
+
+
+class GoodsCountriesDecisions(APIView):
```
a-bek: `CaseDecisionView`?

cases/models.py
a-bek: Is it also worth considering using Django's concept of `unique_together`

https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together

I feel like a lot of the models in the project could make use of this. There are a lot of `get()` calls around on models that don't enfore uniqueness, so if somehow someone has added rows with certain duplicates, those `get()` calls will fail.

cases/models.py
cstobart: compromised

cases/models.py
cstobart: We need to be able to overwrite these, we do however use unique together in other instances

cases/tests/tests_create_goods_countries_decisions.py
```diff
@@ -0,0 +1,73 @@
...
+                {'good': str(self.goods_type_1.id), 'country': 'AL', 'decision': 'no_licence_required', 'case': str(self.case.id)},
+                {'good': str(self.goods_type_2.id), 'country': 'BW', 'decision': 'approve', 'case': str(self.case.id)},
+                {'good': str(self.goods_type_2.id), 'country': 'DE', 'decision': 'approve', 'case': str(self.case.id)},
+                {'good': str(self.goods_type_2.id), 'country': 'SC', 'decision': 'approve', 'case': str(self.case.id)},
+            ]}
+
+        response = self.client.post(self.goods_countries_url, data, **self.gov_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(GoodCountryDecision.objects.count(), 6)
```
BenAndrew-Kainos: Rather than hard-coding 6, could you do len(data['good_countries'])

cases/tests/tests_create_goods_countries_decisions.py
```diff
@@ -0,0 +1,73 @@
...
+            ]}
+
+        response = self.client.post(self.goods_countries_url, data, **self.gov_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(GoodCountryDecision.objects.count(), len(data['good_countries']))
+        self.assertEqual(len(response.json()['data']), len(data['good_countries']))
+
+    def test_saving_overwrites_previous_assignment(self):
+        data = {'good_countries':
```
jfaracik: Move to create_good_country_decision()

cases/tests/tests_create_goods_countries_decisions.py
BenAndrew-Kainos: doesn't seem to have any changes?

cstobart: there's trouble with pushing to GitHub in London at the moment but they're there now

PR #141, opened 2019-09-30
LT-1365-too-technical-messages

Mark-Spiers-kainos:


gov_users/views/gov_user_views.py
```diff
@@ -62,6 +62,15 @@ def post(self, request, *args, **kwargs):
         return JsonResponse(data={'token': token, 'lite_api_user_id': str(user.id)})
 
 
+def replace_default_string(data, fields=None):
```
efahy: Please consider a more informative function name and a docstring

conf/helpers.py
```diff
@@ -22,3 +22,16 @@ def ensure_x_items_not_none(data, x):
 def convert_pascal_case_to_snake_case(name):
     s1 = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
     return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s1).lower()
+
+
+def replace_default_string_for_form_select(data, fields=None, replacement_value=None):
```
a-bek: `fields` parameter should not have a default value. This will break if a list is not passed from the caller as below you'll be trying to iterate over a `None` value which isn't a python iterable.

Should the replacement value be `None` or an empty string? And it might be worth having `default_select_value` parametrised as well so the caller from the view can define the default value? Or default values should be assigned as a constant somewhere.

BenAndrew-Kainos: I agree although I would of thought your default would be an empty list.

Mark-Spiers-kainos: The fields parameter shouldn't have the possibility of None, will remove that now, and there are situations where replacement value = 'None' such as the post and put I'm fixing currently.

a-bek: ok, also it might be worth mentioning using an empty list in a default parameter introduces strange behavior in python because of the way the function objects are generated. You can read more about it here!

https://docs.python-guide.org/writing/gotchas/

Unless you are intending on using the default list as a cache or something, if you want the behavior of a default list, you should do something like:

```
def func(my_list=None):
     if not my_list:
         my_list = []
```

Mark-Spiers-kainos: The default select value is based on the value in lite_forms for this select type as well, should this function need refactored in the future to introduce this, it is left in a good position.

conf/helpers.py
```diff
@@ -22,3 +22,16 @@ def ensure_x_items_not_none(data, x):
 def convert_pascal_case_to_snake_case(name):
     s1 = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
     return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s1).lower()
+
+
+def replace_default_string_for_form_select(data, fields, replacement_value=None):
+    '''
```
a-bek: Docstring should use three double quote quotes. `"`

Mark-Spiers-kainos: fixed

conf/helpers.py
```diff
@@ -22,3 +22,16 @@ def ensure_x_items_not_none(data, x):
...
+    lite_forms submodule contains a possible default value for select questions, with a value of 'blank'
+    this replaces that default value to None(or expected value) so that the test may give correct serializer errors.
+    '''
+    default_select_value = 'blank'
```
a-bek: Can't you move it into the function signature so the caller can set the default?

`def replace_default_string_for_form_select(data, fields, replacement_value=None, default_value='blank'):`

Also I think a better function name is `def replace_default_dict_fields`

PR #145, opened 2019-10-02
feat: LT-1008 - HMRC Organisation

cstobart:


parties/enums.py
```diff
@@ -12,6 +12,18 @@ class PartyType:
...
 
 
+class OrganisationType:
+    HMRC = 'hmrc'
+    COMMERCIAL = 'commercial'
+    INDIVIDUAL = 'individual'
+
+    choices = [
+        (HMRC, 'Hmrc'),
+        (COMMERCIAL, 'Commercial Organisation'),
```
markm5A8Z9F43913A: Capitalisation of HMRC and Commercial organisation?

markm5A8Z9F43913A:
Apart from that one comment, I'm willing to give this an approval as it is.

parties/enums.py
cstobart: I'll fully capitalise HMRC I think that meets the pattern

organisations/views/main.py
```diff
@@ -35,7 +47,7 @@ def post(self, request):
         """
         with reversion.create_revision():
             data = JSONParser().parse(request)
-            if data.get('sub_type') and data['sub_type'] == 'individual':
+            if data.get('type') and data['type'] == 'individual':
```
andytaylorkainos: i know all you did was change the key from `sub_type` to `type`. but can you get rid of one of the conditions in this statement? Theyre both the same thing

organisations/views/main.py
```diff
@@ -1,27 +1,39 @@
...
-        view_serializer = OrganisationViewSerializer(organisations, many=True)
-        return JsonResponse(data={'organisations': view_serializer.data})
+        org_type = self.request.query_params.get('org_type', None)
+        name = self.request.query_params.get('name', '')
+
+        query = [Q(name__icontains=name)]
+
+        if org_type:
+            query.append(Q(type=org_type))
+        return Organisation.objects.filter(reduce(operator.and_, query)).order_by('name')
```
andytaylorkainos: Can you not filter using this approach:
 `filter(**args)` where `args = {'name__contains': name, 'type': org_type}`? 
just curious

efahy:
When creating a HMRC query, I get a CSRF exception when I click continue, having found and selected an org.  I realise that's somewhat beyond the scope of the story but for demoing, etc I'd prefer if you just disable the button or something

efahy:
I don't see an option to enter details of a new exporter, which is part of the story.  Have I missed something?

organisations/views/main.py
cstobart: type being optional throws a spanner in that

organisations/views/main.py
cstobart: 👍

PR #146, opened 2019-10-03
feat: LT-1029 - Letter Templates

jfaracik:
* Added ids filter to picklist items


.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    -   id: end-of-file-fixer
```
andytaylorkainos: amazing

Pipfile
```diff
@@ -23,6 +23,7 @@ django-background-tasks = ">=1.2"
 packaging = "==19.0"
 coverage = ">=4.5.4"
 openpyxl = "*"
+pre-commit = "*"
```
andytaylorkainos: 👍

letter_templates/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from letter_templates import views
+
+app_name = 'letter_templates'
+
+urlpatterns = [
+    path('', views.LetterTemplatesList.as_view(), name='letter_templates'),
```
andytaylorkainos: can we get an example comment please?

static/letter_layouts/models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+
+
+class LetterLayout(models.Model):
+    id = models.CharField(primary_key=True, editable=False, max_length=30)  # letter file name minus extension
```
andytaylorkainos: im a sucker for using uuids (i think theyre quicker to index for)
Can we not use the filename (without extension) as the name?

static/letter_layouts/tests.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class LetterLayoutsTests(DataTestClient):
+
+    def setUp(self):
+        super().setUp()
+        self.letter_layout = LetterLayout.objects.create(id='siel', name='SIEL')
+        self.url = reverse('static:letter_layouts:letter_layouts')
+
+    def test_get_letter_layouts(self):
```
andytaylorkainos: can we either get a doc string or append the test name with `failure`/`success`?

static/letter_layouts/tests.py
```diff
@@ -0,0 +1,37 @@
...
+        response = self.client.get(self.url)
+        response_data = response.json()['results'][0]
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response_data['id'], self.letter_layout.id)
+        self.assertEqual(response_data['name'], self.letter_layout.name)
+
+
+class LetterLayoutTests(DataTestClient):
```
andytaylorkainos: Can we not just put the individual layout test in the same class?

test_helpers/clients.py
```diff
@@ -325,7 +325,7 @@ def create_goods_type(self, application):
         goods_type.save()
         return goods_type
 
-    def create_picklist_item(self, name, team: Team, picklist_type, status):
+    def create_picklist_item(self, name: object, team: object, picklist_type: object, status: object):
```
andytaylorkainos: any reason why we dont want to be type-specific?

jfaracik: Honestly not too sure why that happened, must have clicked the wrong thing in pycharm

queries/end_user_advisories/views.py
```diff
@@ -101,4 +101,4 @@ def put(self, request, pk):
 
 
 class EndUserAdvisorysDetail(object):
```
Mark-Spiers-kainos: do you mind removing this function, since it does not get used anywhere?

static/letter_layouts/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path
+
+from static.letter_layouts import views
+
+app_name = 'letter_layouts'
+
+urlpatterns = [
+    # ex: /static/letter-layouts/
+    path('', views.LetterLayoutsList.as_view(), name='letter_layouts'),
+    # ex: /static/control-list-entries/<str:rating>/
```
Mark-Spiers-kainos: rating in comment and pk in url

letter_templates/migrations/0002_sort_letter_paragraphs.py
```diff
@@ -0,0 +1,19 @@
+from django.db import migrations
```
a-bek: Can you just have the initial migration as its a new model?

letter_templates/models.py
```diff
@@ -0,0 +1,24 @@
...
+    layout = models.ForeignKey(LetterLayout, on_delete=models.CASCADE, null=False)
+    letter_paragraphs = SortedManyToManyField(PicklistItem)
+    restricted_to = models.TextField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    last_modified_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        ordering = ['name']
+
+    def __str__(self):
```
a-bek: Don't think this is being used.

`str(letter_template.id)` seems to be used everywhere. With this function `str(letter_template)` would work too, but I would just remove this function as the former is explicit

letter_templates/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class LetterTemplateSerializer(serializers.ModelSerializer):
+    name = serializers.CharField(max_length=35,
+                                 validators=[UniqueValidator(queryset=LetterTemplate.objects.all(), lookup='iexact',
+                                                             message='The name of your letter template has to be unique')],
+                                 error_messages={'blank': 'Enter a name for the letter template'})
+    letter_paragraphs = serializers.PrimaryKeyRelatedField(queryset=PicklistItem.objects.all(),
+                                                           many=True)
+    restricted_to = CommaSeparatedListField(error_messages={'required': 'Select which types of case this letter template can apply to'},
```
a-bek: Just a style comment,

What do you think of wrapping like this:
```
function(
    arg1=...,
    arg2=...,
)
super_long_function_name(
    arg1=...,
    arg2=...,
)
```
rather than the args being indented based on the length of the function/class name? That way all the indentation is standardised too.

static/letter_layouts/models.py
```diff
@@ -0,0 +1,12 @@
...
+    name = models.CharField(max_length=30)  # friendly name
+
+    class Meta:
+        ordering = ['name']
+
+    def __str__(self):
```
a-bek: Don't think this is being used either

static/letter_layouts/models.py
jfaracik: Would be a big change to make right now but since there are going to many more doc stories in the future I think we could make the switch then if that's cool?

conf/serializers.py
```diff
@@ -88,3 +90,22 @@ class ControlListEntryField(CharField):
...
         super(ControlListEntryField, self).__init__(**kwargs)
         self.validators.append(ControlListEntryValidator())
+
+
+class CommaSeparatedListField(CharField):
+    def __init__(self, **kwargs):
+        super(CommaSeparatedListField, self).__init__(**kwargs)
+
+    def to_internal_value(self, data):
+        if isinstance(data, list):
```
a-bek: What if data passed contains a list? Would it be better to use some other special character to do the internal work?

cases/enums.py
```diff
@@ -9,6 +9,10 @@ class CaseType:
         (END_USER_ADVISORY_QUERY, 'End User Advisory Query'),
     ]
 
+    @classmethod
+    def get_text(cls, choice):
+        return [x for x in cls.choices if x[0] == choice][0][1]
```
Mark-Spiers-kainos: this will iterate through all choices, consider just making a for loop that returns the first value that equals.

static/letter_layouts/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 2.2.4 on 2019-10-07 09:10
+
+from django.db import migrations, models
+
+
+def initialize(apps, schema_editor):
+    LetterLayout = apps.get_model('letter_layouts', 'LetterLayout')
+    if not LetterLayout.objects.all():
+        team = LetterLayout(id='licence',
```
a-bek: `LetterLayout(id='', name='').save()` works

letter_templates/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+from conf.serializers import CommaSeparatedListField, PrimaryKeyRelatedSerializerField
+from letter_templates.models import LetterTemplate
+from picklists.models import PicklistItem
+from static.letter_layouts.models import LetterLayout
+from static.letter_layouts.serializers import LetterLayoutSerializer
+
+
+class LetterTemplateSerializer(serializers.ModelSerializer):
+    name = serializers.CharField(max_length=35,
+                                 validators=[UniqueValidator(queryset=LetterTemplate.objects.all(), lookup='iexact',
```
a-bek: Should `name` being unique also be enforced on the model?

letter_templates/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+    restricted_to = CommaSeparatedListField(error_messages={'required': 'Select which types of case this letter template can apply to'},
+                                            required=True,
+                                            allow_blank=False,
+                                            allow_null=False)
+    layout = PrimaryKeyRelatedSerializerField(queryset=LetterLayout.objects.all(),
+                                              serializer=LetterLayoutSerializer,
+                                              error_messages={'required': 'Select the layout you want to use for this letter template'})
+
+    def validate_restricted_to(self, attrs):
+        if len(attrs) == 0:
```
a-bek: `if not attrs:` is the pythonic way of checking a list is empty

craiga:
Screenshot 2019-11-01 at 09 57 11

The checkbox labels on this screen aren't clickable; should they be?

```
<input id="this" type="checkbox"><label for="this">The for attribute makes this clickable.</label>
<label><input type="checkbox">Wrapping the input in a label works too.</label>
```

craiga:
Screenshot 2019-11-01 at 10 03 37

Reordering paragraphs on the left of this page isn't reflected in the preview on the right.

craiga:
@BenAndrew-Kainos Did you still want `restricted_to` renamed?

letter_templates/migrations/0003_letter_template_name_unique.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.4 on 2019-11-01 10:14
```
efahy: Perhaps delete these three migration files and re-make migrations to get everything consolidated into one file?

static/letter_layouts/models.py
efahy: I think this is a cleaner change to make now, when we're creating the table for the first time.  And while there isn't going to be a large volume of data in this table and performance won't be a realistic issue, its not in line with our model pattern in LITE where everything has a GUID pk, so can we make the change now please?

letter_templates/views.py
```diff
@@ -0,0 +1,44 @@
...
+from letter_templates.serializers import LetterTemplateSerializer
+
+
+class LetterTemplatesList(generics.ListCreateAPIView):
+    """
+    Returns list of all letter templates or creates a letter template
+    """
+    queryset = LetterTemplate.objects.all()
+    serializer_class = LetterTemplateSerializer
+    authentication_classes = (GovAuthentication,)
```
efahy: We tend to have the setting of authentication classes as the first line in our View classes and people (or at least me) tend to look for in there so maybe move this up and same for the class below

letter_templates/migrations/0003_letter_template_name_unique.py
craiga: Yeah, a good idea. I will do this once we're sure there are no further changes to be made.

static/letter_layouts/models.py
craiga: Using `CharField` for a primary key is a pretty bad idea; good pick up. This could be set to something like `foo/../bar` (or even an empty string) which might wreak havoc in all kinds of places.

What's the rationale behind using GUIDs for primary keys? Is it worth the hassle given that Django gives us integer-based primary keys for zero effort?

efahy: That decision predates me so I can't give you an answer, I'm afraid.  Do you think that the hassle saving we'd see by changing to integers would be worth the upheaval of making the change?

craiga: Nah, UUID IDs are fine.

Changing the primary key of a table is a PITA though.

letter_templates/migrations/0003_letter_template_name_unique.py
craiga: Actually, I need to restart these migrations as part of the work in changing the primary key of letter layouts anyway. :+1:

static/letter_layouts/models.py
craiga: [This is coming from a separate PR into this branch.](https://github.com/uktrade/lite-api/pull/176)

letter_templates/migrations/0003_letter_template_name_unique.py
craiga: [This is coming from a separate PR into this branch.](https://github.com/uktrade/lite-api/pull/176)

BenAndrew-Kainos:
> Why are we creating two apps, LetterLayout and LetterTemplates when they belong together?

They do not. LetterLayouts refer to HTML layouts that are static in the system. LetterTemplates are the user created templates that use Layouts as well as Paragraphs and other data. 

So they are very different although their names can be misleading

craiga:
@BenAndrew-Kainos Not sure I agree. It doesn't make sense to have a letter layout without a letter template or vice-versa, does it?

BenAndrew-Kainos:
> @BenAndrew-Kainos Not sure I agree. It doesn't make sense to have a letter layout without a letter template or vice-versa, does it?

You can & do have Layouts without any Templates as they are static and non-user modifiable. That's why I believe they should be a separate static resource (see the `static` folder for other examples of entities that are placed here for the same reason)

craiga:
@BenAndrew-Kainos  I understand that they're different things, but they only make sense when used together.

Put it this way—what uses does a `LetterLayout` serve? AFAIK, it's **only** to provide a layout to a `LetterTemplate`.

There has been a lot of debate about when things should be in separate apps (some believe in striving for reusable apps, some believe in monolithic apps), but it's generally agreed that having lots of tiny apps is a bad thing—it causes migration headaches and makes it unnecessarily difficult to trace what links to what throughout the app.

Unless there's a good reason to keep them separate, I agree with @a-bek that these should be combined.

PR #149, opened 2019-10-05
 feat: LT-998 - Application edit logic testing

andytaylorkainos:
**This tests the functionality of the initial step for editing an application.**
(see `_copy_standard_application()` on `applications/tests/test_edit_standard_application::21`)

This functionality will be used when we hit the "edit" button:
- A  copy of the application will be created and a pointer to this copy will be placed on the original application (see `applications/models.py`)
- This pointer will be used when getting/updating an individual application i.e.:
   ```
   if application.application_copied_to is not None:
        return StandardApplication.objects.get(pk=application.application_copied_to)
   ```
- This means we can retain the application's pk on the front-end and we have a mechanism for deleting the original application when we click "save" or deleting the new (copied version) when we click "cancel"
  
------------------------------------------------------------------------------------------------------------
When we update the parties/goods etc. on the copied application (the instance being edited):
- **DO NOT DELETE THE PREVIOUSLY SET PARTY/GOOD IF IT'S THE SAME AS THE ONE ON THE ORIGINAL APPLICATION (THE APPLICATION WE COPIED FROM)**

If we click "save":
- iterate through all of the original application's items deleting the ones which are not equal to the copied application's items
- Then delete the original application
- Then set the copied application's pk to equal the original application's pk

If we click "cancel":
- iterate through all of the copied application's items and delete the ones which are not equal to the original application's items
- Then reset the `application_copied_to` field to None


applications/models.py
```diff
@@ -19,8 +20,8 @@ class BaseApplication(models.Model):
     activity = models.TextField(default=None, blank=True, null=True)
     usage = models.TextField(default=None, blank=True, null=True)
     organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE, default=None, null=True)
-    created_at = models.DateTimeField(auto_now_add=True, blank=True)
-    last_modified_at = models.DateTimeField(auto_now_add=True, blank=True)
+    created_at = models.DateTimeField(default=timezone.now, blank=True)
+    last_modified_at = models.DateTimeField(default=timezone.now, blank=True)
```
andytaylorkainos: The reason for this: if we copy an application, it would change the dates based on when it was copied. (prior to this change)

By doing this change, we can overwrite the value (`auto_add_now=True` does not let you manually update it). It still automatically sets a value if you to today's time (in utc) if you dont supply one.

applications/models.py
```diff
@@ -29,6 +30,7 @@ class BaseApplication(models.Model):
     reference_number_on_information_form = models.TextField(blank=True, null=True)
     have_you_been_informed = models.CharField(choices=ApplicationExportLicenceOfficialType.choices, default=None,
                                               max_length=50)
+    application_copied_to = models.ForeignKey('self', on_delete=models.DO_NOTHING, blank=True, null=True)
```
andytaylorkainos: Can be used as a flag, and to retrieve the `copied` version of the application or delete the copied/old version depending on if we `save` or `cancel` changes.

applications/serializers.py
```diff
@@ -302,7 +301,7 @@ def update(self, instance, validated_data):
         instance.export_type = validated_data.get('export_type', instance.export_type)
         instance.reference_number_on_information_form = validated_data.get(
             'reference_number_on_information_form', instance.reference_number_on_information_form)
-        instance.last_modified_at = datetime.now(timezone.utc)
+        instance.last_modified_at = timezone.now()
```
andytaylorkainos: django recommends using this (it uses the `USE_TZ` setting in `conf.settings`)

documents/av_scan.py
```diff
@@ -61,7 +61,7 @@ def _process_document(document_pk: str):
 
     is_file_clean = _scan_s3_object(doc.name, settings.AWS_STORAGE_BUCKET_NAME, doc.s3_key)
     if is_file_clean is not None:
-        doc.virus_scanned_at = now()
+        doc.virus_scanned_at = timezone.now()
```
andytaylorkainos: just being consistent with other implementations

applications/drafts/tests/tests_ultimate_end_users.py
```diff
@@ -8,7 +8,8 @@ class UltimateEndUsersOnDraft(DataTestClient):
 
     def setUp(self):
         super().setUp()
-        self.draft = self.create_standard_draft(self.organisation)
+        self.draft = self.create_standard_draft_with_incorporated_good(self.organisation)
```
andytaylorkainos: Updating draft to be in the correct state for handling UEUs

applications/drafts/tests/tests_third_parties.py
```diff
@@ -41,7 +40,7 @@ def test_set_and_remove_third_parties_on_draft_successful(self):
         response = self.client.delete(url, **self.exporter_headers)
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(self.draft.ultimate_end_users.count(), 0)
+        self.assertEqual(self.draft.third_parties.count(), 0)
```
andytaylorkainos: fixing test

applications/serializers.py
markm5A8Z9F43913A: Does this store the time in UTC regardless of the machine's current time zone?

PR #154, opened 2019-10-09
refactor: LT-998 integrate /drafts/ into /applications/

markm5A8Z9F43913A:


applications/views/application_parties.py
```diff
@@ -47,21 +46,25 @@ def post(self, request, draft):
...
 
-    @only_draft_types(ApplicationLicenceType.STANDARD_LICENCE)
-    def get(self, request, draft):
+    def get(self, request, pk):
         """
         Get ultimate end users associated with a draft
         """
-        serializer = UltimateEndUserSerializer(draft.ultimate_end_users, many=True)
-        return JsonResponse(data={'ultimate_end_users': serializer.data})
+        draft = get_application(pk)
```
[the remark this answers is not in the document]
andytaylorkainos: Youre right, done!

applications/views/views.py
```diff
@@ -24,40 +25,77 @@
...
+class ApplicationList(ListAPIView):
     authentication_classes = (ExporterAuthentication,)
 
-    def get(self, request):
+    def get(self, request, *args, **kwargs):
         """
         List all applications
         """
+        try:
+            submitted = optional_str_to_bool(request.GET.get('submitted', None))
```
craiga: There's no need to provide the second argument here. `None` is the default.

applications/views/views.py
```diff
@@ -24,40 +25,77 @@
...
+        except ValueError as e:
+            return JsonResponse(data={'errors': str(e)}, status=status.HTTP_400_BAD_REQUEST)
```
craiga: Would just using something like the following be bad?

```python
submitted = request.GET.get("submitted").lower() == "true"
```

applications/libraries/application_helpers.py
```diff
@@ -2,8 +2,19 @@
...
+
+def optional_str_to_bool(optional_string: str):
+    if optional_string is None:
+        return None
+    elif optional_string == 'true':
+        return True
+    elif optional_string == 'false':
+        return False
+    else:
+        raise ValueError('You provided ' + optional_string + ', while the allowed values are None, "true" or "false"')
```
craiga: What about `"True"`, `"0"`, `True`, etc.?

applications/views/views.py
markm5A8Z9F43913A: That would only allow for two values. We need three.

applications/views/views.py
markm5A8Z9F43913A: 👍

applications/libraries/application_helpers.py
markm5A8Z9F43913A: I'll convert the strings to lower case when comparing which would allow more flexibility. We don't need to cater for other values though.

applications/libraries/get_applications.py
```diff
@@ -2,87 +2,60 @@
...
         else:
-            return OpenApplication.objects.get(pk=pk, submitted_at__isnull=True)
-    except (StandardApplication.DoesNotExist, OpenApplication.DoesNotExist):
-        raise Http404
+            raise TypeError('object "organisation" provided is not an instance of Organisation model')
 
 
-def get_application_for_organisation(pk, organisation):
-    application = get_application(pk)
+def get_base_applications(organisation=None, submitted=None):
```
a-bek: I'm not sure about the need for this function, its only used once so maybe this logic can live in the view to make it more readbale? This is going three functions deep when it could be much simpler. 

Also, it may be worth reading about Django model managers and creating custom querysets.
https://docs.djangoproject.com/en/2.2/topics/db/managers/

For example in the view:

```
 if organisation:
    applications = BaseApplication.objects.filter(organisation=organisation)  
If submitted is not None:
    applications = applications.objects.filter(submitted=convert_to_bool(submitted_request_value)
```

Rather than nesting a bunch of functions to achieve the same.

applications/libraries/get_applications.py
```diff
@@ -2,87 +2,60 @@
...
-    return BaseApplication.objects.filter(organisation=organisation, submitted_at__isnull=True)
-
-
-def get_draft_application_for_organisation(pk, organisation):
-    draft = get_draft(pk=pk)
-
-    if draft.organisation.pk != organisation.pk:
+def get_application_licence_type(pk):
+    try:
+        return BaseApplication.objects.get(pk=pk).licence_type
```
a-bek: `return BaseApplication.objects.values_list('licence_type', flat=True).get(pk=pk)` should work the same and not load the whole Application from db

applications/views/application_external_locations_views.py
```diff
@@ -34,7 +34,7 @@ def post(self, request, pk):
         organisation = get_organisation_by_user(request.user)
```
a-bek: Why does this function exist:

```
def get_organisation_by_user(user):
    return user.organisation
```

Can't user.organisation be used everywhere?

applications/libraries/get_applications.py
andytaylorkainos: Ahh, i didnt realise that we could just add more filters, one after the other. Good to know!

applications/libraries/get_applications.py
```diff
@@ -2,87 +2,60 @@
...
-
-
-def get_draft(pk):
-    draft_type = get_draft_type(pk)
-    try:
-        if draft_type == ApplicationLicenceType.STANDARD_LICENCE:
-            return StandardApplication.objects.get(pk=pk, submitted_at__isnull=True)
+def _add_organisation_filter(kwargs, organisation):
+    if organisation:
+        if isinstance(organisation, Organisation):
```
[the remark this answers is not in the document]
andytaylorkainos: Done!

applications/libraries/get_applications.py
andytaylorkainos: done!

applications/views/application_external_locations_views.py
andytaylorkainos: Switched to using `user.organisation` in areas we have introduced new code

applications/libraries/get_applications.py
```diff
@@ -2,87 +2,60 @@
...
 
-    if application.organisation.pk != organisation.pk:
-        raise Http404
-
-    return application
+    If param submitted is None, all applications are returned
+    If it is true (and consequently submitted_at__isnull is false), only submitted applications are returned
+    If it is false (and consequently submitted_at__isnull is true), only draft applications are returned
+    """
+    kwargs = dict()
```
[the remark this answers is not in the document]
andytaylorkainos: To be honest i was in favour of `dict()` and `list()` but unfortunately we had a discussion and everyone preferred `{}` and `[]` so im only trying to comply with that - im MORE than happy to have the discussion again tho :D :D :D

applications/libraries/get_applications.py
```diff
@@ -2,87 +2,50 @@
...
-def _get_application_from_model(pk, model):
-    try:
-        return model.objects.get(pk=pk, submitted_at__isnull=False)
-    except model.DoesNotExist:
-        raise Http404
-
-
-def get_applications():
-    return BaseApplication.objects.filter(submitted_at__isnull=False)
+def _add_submitted_filter(kwargs, submitted: bool):
```
a-bek:
```
from typing import Optional

def _add_submitted_filter(kwargs, submitted: Optional[bool]):
```

applications/libraries/get_applications.py
markm5A8Z9F43913A: We got rid of this completely

applications/libraries/get_applications.py
markm5A8Z9F43913A: This is not relevant anymore.

applications/views/views.py
```diff
@@ -4,53 +4,89 @@
...
         return JsonResponse(data={'applications': serializer.data})
 
+    def post(self, request):
+        data = request.data
+        data['organisation'] = str(request.user.organisation.id)
+
+        # Use generic serializer to validate all types of application as we may not yet know the application type
+        serializer = DraftApplicationCreateSerializer(data=data)
+
+        if serializer.is_valid():
```
a-bek: Rather than having the application in a block, what about switching the logic here:

```
if not serializer.is_valid():
    return ErrorResponse
...
```
and keeping the main logic in the main code block of the view.

markm5A8Z9F43913A: 👍

PR #156, opened 2019-10-15
feat: LT-998 - Edit application

markm5A8Z9F43913A:


applications/creators.py
```diff
@@ -102,18 +104,20 @@ def validate_open_licence(draft, errors):
     return errors
 
 
-def check_application_for_errors(draft):
+def validate_application_ready_for_submission(application):
     errors = {}
```
cstobart: Direct assignment `errors = dict()`

applications/views/application_goods.py
```diff
@@ -5,10 +5,12 @@
...
+from applications.libraries.get_goods_on_applications import get_good_on_application
 from applications.models import GoodOnApplication
 from applications.serializers import GoodOnApplicationViewSerializer, GoodOnApplicationCreateSerializer
 from conf.authentication import ExporterAuthentication
-from conf.decorators import only_application_types
+from conf.decorators import only_application_type
+from goods.enums import GoodStatus
 from goods.libraries.get_goods import get_good_with_organisation
 from goods.models import GoodDocument
 from goodstype.models import GoodsType
```
cstobart: class ApplicationGoodsType(APIView - > ListView)

with a get_queryset

goods/tests/tests_submitting_goods.py
```diff
@@ -15,7 +15,7 @@ def test_submitted_good_changes_status(self):
         """
         draft = self.create_standard_draft(self.organisation)
         self.assertEqual(Good.objects.get().status, 'draft')
-        url = reverse('applications:application_submit', kwargs={'pk': draft.id})
+        url = reverse('applications:application_submit', kwargs={'pk':  draft.id})
```
cstobart: remove extra white space

goodstype/views.py
```diff
@@ -1,17 +1,17 @@
...
+from conf.decorators import only_application_type, authorised_user_type
 from goodstype.helpers import get_goods_type
 from goodstype.models import GoodsType
 from goodstype.serializers import GoodsTypeSerializer, FullGoodsTypeSerializer
 from static.countries.models import Country
-from users.models import GovUser
+from users.models import GovUser, ExporterUser
 
 
 class GoodsTypeList(APIView):
```
cstobart: Use a ListView instead of an APIView

organisations/views/sites.py
```diff
@@ -6,8 +6,10 @@
 from rest_framework.views import APIView
 
 from conf.authentication import SharedAuthentication
+from conf.decorators import authorised_user_type
 from organisations.models import Organisation, Site
 from organisations.serializers import SiteViewSerializer, SiteSerializer
+from users.models import ExporterUser
 
 
 class SitesList(APIView):
```
cstobart: Could use ListCreateAPIView with a `def put` overriding

applications/creators.py
andytaylorkainos: Im happy to do it, but i think we need to, as a team, explicity state this is how we assign empty sets/dicts/types - i know that a few people prefer it as `{}` and others prefer `dict()`

cstobart: Looks like this is resolved!

applications/creators.py
```diff
@@ -102,18 +104,20 @@ def validate_open_licence(draft, errors):
...
 
+    if application.status and application.status != get_case_status_from_status_enum(CaseStatusEnum.APPLICANT_EDITING):
+        errors['status'] = 'This application cannot be submitted.'
     # Generic errors
-    if SiteOnApplication.objects.filter(application=draft).count() == 0 and \
-            ExternalLocationOnApplication.objects.filter(application=draft).count() == 0:
+    if SiteOnApplication.objects.filter(application=application).count() == 0 and \
```
a-bek: If you're checking to see if a queryset is empty, `.exists()` is much more efficient than `count()`. `.exists()` will only fetch one record even if there are many, `count()` needs to check the whole table.

static/statuses/libraries/get_case_status.py
```diff
@@ -4,7 +4,7 @@
 
 def get_case_status_from_status_enum(case_status):
     # if passed tuple
-    if len(case_status) == 2:
+    if case_status and len(case_status) == 2:
```
a-bek: Why would a tuple ever be passed? This is a good example of why typing should be used :)

applications/libraries/application_helpers.py
```diff
@@ -18,3 +24,26 @@ def optional_str_to_bool(optional_string: str):
...
+                               user: BaseUser) -> Optional[str]:
+    try:
+        get_case_status_from_status_enum(new_status)
+    except NotFoundError:
+        return 'Status not found.'
+
+    if isinstance(user, ExporterUser):
+        if original_status != CaseStatusEnum.SUBMITTED and new_status == CaseStatusEnum.APPLICANT_EDITING:
+            return f'Setting application status to "{str(new_status)}" when application status is ' \
+                f'"{str(original_status)}" is not allowed.'
```
a-bek: Is it necessary to cast to string `str(original_status)`. Won't f strings invoke the __str__ dunder function for you?

applications/creators.py
andytaylorkainos: Done!

applications/libraries/application_helpers.py
andytaylorkainos: Fixed!

static/statuses/migrations/0003_auto_20191014_1046.py
```diff
@@ -0,0 +1,38 @@
...
+    CaseStatus.objects.all().delete()
+
+    for choice in CaseStatusEnum.choices:
+        case_status = CaseStatus(status=choice[0], priority=CaseStatusEnum.priorities[choice[0]])
+        case_status.save()
+
+
+def destroy(apps, schema_editor):
+    CaseStatus = apps.get_model('statuses', 'CaseStatus')
+    CaseStatus.objects.all().delete()
```
Mark-Spiers-kainos: not sure if we should delete all casestatus's if reverse migrated. may affect the system. Also when initialising, we should consider what happens to Cases, if the casestatus's are deleted.

conf/decorators.py
```diff
@@ -1,28 +1,74 @@
...
+        application = kwargs['application']
+    else:
+        return HttpResponseNotFound()
+
+    kwargs['application'] = application
+
+    return application
+
+
+def application_licence_type(type):
```
a-bek: https://docs.python.org/3/library/functions.html#type

Generally not a good idea to override built-in names.

Maybe call it `licence_type` which is what you are passing.

organisations/libraries/get_external_location.py
```diff
@@ -1,22 +1,24 @@
...
+from applications.models import BaseApplication, ExternalLocationOnApplication
+from conf.exceptions import NotFoundError
 from organisations.models import ExternalLocation
 
 
-def get_external_location_by_pk(pk):
+def get_external_location(pk, organisation=None):
+    kwargs = {'pk': pk}
+
+    if organisation:
```
a-bek: If you know the `pk`, why would you ever need to verify another field value?

organisations/libraries/get_site.py
```diff
@@ -1,21 +1,24 @@
+from applications.models import BaseApplication, SiteOnApplication
 from conf.exceptions import NotFoundError
 from organisations.models import Site
 
 
-def get_site_by_pk(pk):
+def get_site(pk, organisation=None):
+    kwargs = {'pk': pk}
```
a-bek: Same here, you know the `pk`

applications/libraries/application_helpers.py
```diff
@@ -18,3 +24,26 @@ def optional_str_to_bool(optional_string: str):
         return False
     else:
         raise ValueError('You provided ' + optional_string + ', while the allowed values are None, "true" or "false"')
+
+
+def validate_status_can_be_set(original_status: CaseStatusEnum,
+                               new_status: CaseStatusEnum,
+                               user: BaseUser) -> Optional[str]:
+    try:
+        get_case_status_from_status_enum(new_status)
```
a-bek: Is this necessary? CaseStatus is generated from `CaseStatusEnum` so it will always exist?

applications/libraries/application_helpers.py
```diff
@@ -18,3 +24,26 @@ def optional_str_to_bool(optional_string: str):
         return False
     else:
         raise ValueError('You provided ' + optional_string + ', while the allowed values are None, "true" or "false"')
+
+
+def validate_status_can_be_set(original_status: CaseStatusEnum,
```
a-bek: Maybe split this in 2 functions? One for export users and one for interanl?

organisations/libraries/get_external_location.py
andytaylorkainos: This is a security measure - users should not be able to locations from another organisation

organisations/libraries/get_site.py
andytaylorkainos: This is a security measure - users should not be able to add sites from another organisation

applications/libraries/application_helpers.py
andytaylorkainos: Although `CaseStatusEnum` is the type for `new_status` - youre still annoyingly able to pass in any old random string to the function (one which may not be an actual enum that exists in CaseStatusEnum)

conf/decorators.py
andytaylorkainos: 👍

applications/views/application_external_locations_views.py
```diff
@@ -18,20 +19,20 @@ class ApplicationExternalLocations(APIView):
     """
     authentication_classes = (ExporterAuthentication,)
 
-    def get(self, request, pk):
-        draft = get_application(pk)
+    @authorised_users(ExporterUser)
```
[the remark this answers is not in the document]
andytaylorkainos: The Authentication is to verify user credentials. 
Authorisation is to verify that the the user is allowed to perform that operation on the application

applications/views/views.py
```diff
@@ -78,50 +81,52 @@ class ApplicationDetail(APIView):
...
-                            case=application.case.get(),
-                            user=request.user,
-                            status=data.get('status'))
-
         serializer.save()
-        return JsonResponse(data={'application': serializer.data})
 
-    def delete(self, request, pk):
+        if application.case.get():
+            kwargs = {
```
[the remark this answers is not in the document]
andytaylorkainos: 👍

applications/views/views.py
```diff
@@ -130,29 +140,84 @@ class ApplicationSubmission(APIView):
...
+
+        data = request.data
+        new_status_enum = data.get('status')
+
+        # Only allow the final decision if the user has the MANAGE_FINAL_ADVICE permission
+        # This can return 403 forbidden
+        if new_status_enum == CaseStatusEnum.FINALISED:
+            assert_user_has_permission(request.user, Permissions.MANAGE_FINAL_ADVICE)
+
+        if isinstance(request.user, ExporterUser) and request.user.organisation.id != application.organisation.id:
```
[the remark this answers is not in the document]
andytaylorkainos: 👍

static/statuses/migrations/0003_auto_20191014_1046.py
andytaylorkainos: sort of fixed

static/statuses/libraries/get_case_status.py
andytaylorkainos: I think some of the tests have been built around this assumption

applications/libraries/application_helpers.py
a-bek: You could assert that is in an instance of `CaseStatusEnum`

users/libraries/token_to_user.py
```diff
@@ -3,5 +3,8 @@
 
 def token_to_user_pk(token):
     data = Token.decode_to_json(token)
-    pk = data['id']
-    return pk
+    if 'id' in data:
```
a-bek: `return data.get('id')` is the pythonic way of doing this.

`.get(key)` returns `None` if `key` doesn't exists.

andytaylorkainos: 🤣 i have no idea why i didnt do it that way

goods/tests/tests_submitting_goods.py
andytaylorkainos: 👍

static/statuses/libraries/get_case_status.py
andytaylorkainos: Updated tests and removed this logic

applications/libraries/application_helpers.py
andytaylorkainos: Okay so i've had to think about this one - unfortunately our CaseStatusEnum is a class so we cant do anything clever in terms of asserting the type being passed in. I've refeactored the arguements to be of type `str`.
We `could` assert that the status being passed in is in of CaseStatusEnum.choices which would save the trip to the DB.
The problem i have with this, is that the enum very may well contain the item in `choices` but it may not exist in the DB.
If it does not exist in the DB then it'd fall over when trying to assign the a foreign key to the status from the case

applications/libraries/application_helpers.py
andytaylorkainos: Done!

a-bek:
e2e tests pass and nothing breaks?? :)

PR #159, opened 2019-10-16
feat: LT-1609 Store letter paragraphs in order.

craiga:


letter_templates/tests/tests_create.py
```diff
@@ -130,3 +132,23 @@ def test_create_letter_templates_no_restricted_to_failure(self):
         response = self.client.post(self.url, data, **self.gov_headers)
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_create_letter_templates_order_is_saved(self):
+        """Check the order of letter paragraphs is saved."""
+        for i, picklist_items in enumerate(
```
a-bek: The iteration is for two different test cases generated by permutations right? Maybe use pytest parametrize?

craiga: Good idea, but pytest isn't in use in this repository. It seems like most tests are written using good ol' unittest, which I don't mind.

a-bek: You might be able to use `parametrized` package.

https://pypi.org/project/parameterized/

craiga: Unfortunately, `parameterized` or anything of its ilk won't work as I need to refer to `self`.

```
    @parameterized((self.picklist_item_1, self.picklist_item_2),
                   (self.picklist_item_2, self.picklist_item_1))
    def test_create_letter_templates_order_is_saved(self, picklist_items):
```

PR #160, opened 2019-10-17
LT-1580-assess-goods

Mark-Spiers-kainos:


users/migrations/0008_assess_goods_permission.py
```diff
@@ -0,0 +1,33 @@
...
+
+from django.db import migrations, models
+
+from goods.models import Good
+from queries.control_list_classifications.models import ControlListClassificationQuery
+
+
+def create_permission(apps, schema_editor):
+    Permission = apps.get_model('users', 'Permission')
+    if not Permission.objects.filter(id='ASSESS_GOODS'):
```
cstobart: This supersedes needing a remove permission?

users/migrations/0008_assess_goods_permission.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Mark S on 2019-10-16 14:08
```
stephanieboyd: Should this be removed, don't see them left in other places?

PR #164, opened 2019-10-24
LT-1612: Cases API changes

a-bek:
Sub task of: https://uktrade.atlassian.net/browse/LT-951
Front end: https://github.com/uktrade/lite-internal-frontend/pull/199

Changes:

1) Endpoint change for retrieving case list:
      New endpoint:
          `/cases/`
      Old endpoint:
          `/queues/<pk>/cases/`

      (Queues is passed as a filter parameter)
     The change also simplifies logic in `internal-frontend`
2) Returns all data required by cases internal fronted view.
3) Refactor search and filter functions into model Manager
4) Update tests
5) Enforce unique together in CaseStatus model [status, priority]


cases/managers.py
```diff
@@ -0,0 +1,58 @@
...
+    def order_by_date(self, order=''):
+        """
+        :param order: ('', '-')
+        :return:
+        """
+        order = order if order in ['', '-'] else ''
+
+        return self.annotate(
+            created_at=Coalesce('application__submitted_at', 'query__submitted_at'),
+        ).order_by(f'{order}created_at')
```
andytaylorkainos: Is there any reason why we need to coalesce by `submitted_at` when we order by `created_at`? 
(unless the front end needs both on the cases search/filter page?)

a-bek: `created_at` is the value of `Coalesce('application__submitted_at', 'query__submitted_at')`. 

`annotate` creates the `created_at` field for all objects in the queryset, otherwise we could not order it. Honestly, we should just add a `created_at` field to the Case model.

andytaylorkainos: Would it be too much work to move it?

static/statuses/migrations/0004_auto_20191031_1212.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.4 on 2019-10-31 12:12
```
andytaylorkainos: Can you move the custom migration steps from the previous one into here? (it just makes them easier to find)

cases/managers.py
a-bek: https://github.com/uktrade/lite-api/pull/178

efahy:
The branch name being named for a subtask (LT-1612) and not the story confused me.  I assume this is a PR for all of LT-951, since the PR is to pull it into master?  Can you name your "main" branch using the story number in future?

cases/serializers.py
```diff
@@ -70,6 +73,40 @@ class TinyCaseSerializer(serializers.Serializer):
...
+
+    def __init__(self, *args, **kwargs):
+        self.team = kwargs.pop('team', None)
+        super().__init__(*args, **kwargs)
+
+    def get_flags(self, instance):
+        """
+        Gets flags for a case and returns in sorted order by team.
+        """
+        org = instance.organisation()
```
andytaylorkainos: does this need the parenthesis?

cases/serializers.py
```diff
@@ -70,6 +73,40 @@ class TinyCaseSerializer(serializers.Serializer):
...
+
+        case_flag_data = FlagSerializer(instance.flags.order_by('name'), many=True).data
+        org_flag_data = FlagSerializer(org.flags.order_by('name'), many=True).data
+
+        if instance.application:
+            goods = GoodOnApplication.objects.filter(application=instance.application).select_related('good')
+            goods_flags = list(itertools.chain.from_iterable([g.good.flags.order_by('name') for g in goods]))
+            good_flag_data = FlagSerializer(goods_flags, many=True).data
+
```
andytaylorkainos: remove blank line

cases/serializers.py
```diff
@@ -70,6 +73,40 @@ class TinyCaseSerializer(serializers.Serializer):
...
+        else:
+            good_flag_data = []
+
+        flag_data = case_flag_data + org_flag_data + good_flag_data
+
+        if not self.team:
+            return flag_data
+
+        # Sort flags by user's team.
+        team_flags = list(filter(lambda x: x['team']['id'] == str(self.team.id), flag_data))
```
andytaylorkainos: can we just use one loop and append the flag to `team_flags` or `non_team_flags` instead of using two lambdas to iterate the flags twice?

cases/serializers.py
```diff
@@ -70,6 +73,40 @@ class TinyCaseSerializer(serializers.Serializer):
...
+        non_team_flags = list(filter(lambda x: x['team']['id'] != str(self.team.id), flag_data))
+
+        return sorted(team_flags, key=lambda x: x['name']) + sorted(non_team_flags, key=lambda x: x['name'])
```
andytaylorkainos: you could sort the entire flags list before its split to team + non team. 
Then you could divide the list out using my suggestion in the previous comment above (itll still maintain the order)

cases/views/search/views.py
```diff
@@ -0,0 +1,44 @@
...
+            status=request.GET.get('status'),
+            case_type=request.GET.get('case_type'),
+            sort=request.GET.get('sort'),
+            date_order='-' if queue_id in SYSTEM_QUEUES else '',
+        ))
+        queues = SearchQueueSerializer(service.get_search_queues(team=request.user.team), many=True).data
+        cases = TinyCaseSerializer(page, context=context, team=request.user.team, many=True).data
+        statuses = service.get_case_status_list()
+        case_types = service.get_case_type_list()
+        queue = list(filter(lambda q: q['id'] == queue_id, queues)).pop()
```
andytaylorkainos: why would there be two queues with the same id?

cases/serializers.py
a-bek: Changed it to a property.

cases/views/search/views.py
a-bek: What do you mean? This just gets the queue that is currently being looked at.

cases/views/search/queue.py
```diff
@@ -0,0 +1,63 @@
+from typing import List
+
+from cases.models import Case
+from queues.constants import ALL_CASES_SYSTEM_QUEUE_ID, MY_TEAMS_QUEUES_CASES_ID, OPEN_CASES_SYSTEM_QUEUE_ID
+from queues.models import Queue
+from teams.models import Team
+
+
+class SearchQueue:
```
efahy: Docstring comment please

cases/views/search/views.py
```diff
@@ -0,0 +1,44 @@
...
+from cases.models import Case
+from cases.serializers import TinyCaseSerializer
+from cases.views.search import service
+from cases.views.search.serializers import SearchQueueSerializer
+from conf.authentication import GovAuthentication
+from conf.pagination import MaxPageNumberPagination
+from queues.constants import SYSTEM_QUEUES, ALL_CASES_SYSTEM_QUEUE_ID
+
+
+class CasesSearchView(generics.ListAPIView):
```
efahy: Docstring comment please

cases/views/search/views.py
andytaylorkainos: So we're filtering for the queue_id by iterating through the queues. Creating a list. then taking the last element from that list? Seems a bit overkill.

Can we not just (pseudo)
```
for q in queues:
   if q.id == queue_id
      queue = q
      break
```
Saves us going through all the queues as soon as we find the the one were looking for

a-bek: You're right, queues will always be quite small so I didnt think too much of it.

Ive fixed it using a generator.

PR #165, opened 2019-10-25
fix: LT-1029 - Allow for correct saving of many case types.

craiga:


.circleci/config.yml
```diff
@@ -6,13 +6,16 @@ orbs:
...
-      PIPENV_VENV_IN_PROJECT: 'enabled'
+        environment:
+          DEBUG: 'True'
+          PIPENV_VENV_IN_PROJECT: 'enabled'
+          DATABASE_URL: postgresql://lite-api:security_is_important@localhost/lite-api
+      - image: circleci/postgres:11
+        environment:
+          POSTGRES_USER: lite-api
+          POSTGRES_PASSWORD: security_is_important
+          POSTGRES_DB: lite-api
```
craiga: Using PostgreSQL in CircleCI lets us use [PostgreSQL specific model fields](https://docs.djangoproject.com/en/stable/ref/contrib/postgres/fields/).

PR #166, opened 2019-10-25
Supporting changes for fixes to adding multiple case types to letter templates

craiga:


conf/serializers.py
```diff
@@ -90,22 +88,3 @@ class ControlListEntryField(CharField):
...
-        if isinstance(data, list):
-            data = ','.join(data)
-        return data
-
-    def to_representation(self, value):
-        if isinstance(value, set):
-            value = value.pop()
-
-        if '{' in value or '[' in value:
-            return ast.literal_eval(value)
```
craiga: Use of eval here was allowing some very strange data to be saved. Ideally this would be replaced with [an ArrayField in the model](https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/fields/#arrayfield), but the tests don't work with Postgres at the moment.

letter_templates/serializers.py
```diff
@@ -16,10 +16,15 @@ class LetterTemplateSerializer(serializers.ModelSerializer):
                                  error_messages={'blank': 'Enter a name for the letter template'})
     letter_paragraphs = serializers.PrimaryKeyRelatedField(queryset=PicklistItem.objects.all(),
                                                            many=True)
-    restricted_to = CommaSeparatedListField(error_messages={'required': 'Select which types of case this letter template can apply to'},
-                                            required=True,
-                                            allow_blank=False,
-                                            allow_null=False)
+
+    # restricted_to is backed by a text field containing comma delimited data.
+    restricted_to = serializers.MultipleChoiceField(error_messages={'required': 'Select which types of case this letter template can apply to'},
```
BenAndrew-Kainos: We (Eugene and I) believe the correct way to do this is to introduce a new `CaseType` model and have a many to many relationship between `Template` and `CaseType`.  For the implementation of the `CaseType` model, you can use the `CaseStatus` model and Enum as an example.  In doing this you will also need to change the `type` field on the `Case` model to also refer to the new model.

letter_templates/tests/tests_view.py
```diff
@@ -18,8 +18,8 @@ def setUp(self):
                                                        PickListStatus.ACTIVE)
         self.letter_layout = LetterLayout.objects.first()
         self.letter_template = LetterTemplate.objects.create(name='SIEL',
-                                                             restricted_to=[CaseType.CLC_QUERY,
-                                                                            CaseType.END_USER_ADVISORY_QUERY],
+                                                             restricted_to=",".join([CaseType.CLC_QUERY,
```
BenAndrew-Kainos: We should stop treating this a string once we update the serializer

conf/serializers.py
craiga: Yep, it was only used in once place.

letter_templates/serializers.py
craiga: OK, that's going to require a pretty substantial refactor of document generation and cases in general (as far as I understand it).

Given that `CaseTyoe` isn't something introduced by this PR, can we spin this off into a separate issue?

craiga: FWIW, I 100% agree that `cases.enums.CaseType` isn't a great implementation.

letter_templates/serializers.py
```diff
@@ -34,18 +39,25 @@ def validate_letter_paragraphs(self, attrs):
...
-        fields = '__all__'
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        data["restricted_to"] = instance.restricted_to.split(",")
+        return data
+
+    def to_internal_value(self, data):
+        data = super().to_internal_value(data)
+        if "restricted_to" in data:
+            data["restricted_to"] = ",".join(data["restricted_to"])
```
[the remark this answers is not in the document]
craiga: Yeah, this is why I was trying to make use of the Postgres-specific `ArrayField`.

letter_templates/serializers.py
```diff
@@ -36,18 +41,13 @@ def validate_letter_paragraphs(attrs):
...
-        Convert restricted_to to list of key value entries
+        Provide display values for restricted_to.
         """
-        repr_dict = super(LetterTemplateSerializer, self).to_representation(value)
-
-        repr_dict['restricted_to'] = [{'key': x, 'value': CaseType.get_text(x)} for x in
-                                      repr_dict['restricted_to']]
-        repr_dict['restricted_to'].sort(key=lambda x: x['value'])
+        display_names = dict(CaseType.choices)
+        return [display_names.get(r) for r in instance.restricted_to]
```
BenAndrew-Kainos: Change variable name from `r` to something more meaningful

PR #172, opened 2019-10-30
feat: LT-982 - Audit anchor

andytaylorkainos:


users/models.py
```diff
@@ -86,6 +86,13 @@ def send_notification(self, case_note=None, query=None, ecju_query=None):
             Notification.objects.create(user=self, query=actual_query)
         elif ecju_query:
             Notification.objects.create(user=self, ecju_query=ecju_query)
+        elif case_activity:
+            try:
+                notification = Notification.objects.get(user=self)
+                notification.case_activity = case_activity
+                notification.save()
+            except Notification.DoesNotExist:
+                Notification.objects.create(user=self, case_activity=case_activity)
```
craiga: Why are case activities handled so differently from from the other kinds of notifications?

Is it worth including a comment explaining this?

andytaylorkainos: Yeah that's not a bad idea - its because a gov user only has one notification per case

users/tests/tests_gov_user_case_notifications.py
```diff
@@ -0,0 +1,56 @@
...
+            'new_name': 'new app name'
+        }
+        case_activity = CaseActivity.create(case=case, user=self.exporter_user, **case_activity)
+        self.gov_user.send_notification(case_activity=case_activity)
+        url = reverse_lazy('users:case_notification') + '?case=' + str(case.id)
+
+        response = self.client.get(url, **self.gov_headers)
+
+        self.assertEqual(len(response.json()['notification']), 1)
+        self.assertEqual(response.json()['notification']['case_activity'], case_activity.id)
```
markm5A8Z9F43913A: Looks like `response.json()['notification']` might look better in its own variable.

andytaylorkainos: Done

users/models.py
```diff
@@ -111,3 +115,19 @@ def unassign_from_cases(self):
...
         self.case_assignments.clear()
+
+    def send_notification(self, case_activity=None):
+        from cases.models import Notification
+        # circular import prevention
+
+        if case_activity:
+            # There can only be one notification per gov user's case
+            # If a notification for that gov user's case already exists, update the case activity it points to
+            try:
```
a-bek: Seems like a good use case for djangos `update_or_create()`.

users/tests/tests_gov_user_case_notifications.py
```diff
@@ -0,0 +1,57 @@
...
+        case_activity = {
+            'activity_type': CaseActivityType.UPDATED_APPLICATION_NAME,
+            'old_name': 'old app name',
+            'new_name': 'new app name'
+        }
+        case_activity = CaseActivity.create(case=case, user=self.exporter_user, **case_activity)
+        self.gov_user.send_notification(case_activity=case_activity)
+        prev_notification_count = Notification.objects.filter(user=self.gov_user, case_activity__case=case).count()
+        url = reverse('applications:application', kwargs={'pk': case.application.id})
+        data = {'name': 'even newer app name!'}
```
a-bek: Maybe assert the value of the newer name?

andytaylorkainos: Done!

users/models.py
andytaylorkainos: Tried to use this but because we need to filter the notification by the case_activity's `case` not on the entire `case_activity`

PR #175, opened 2019-11-01
feat: LT-886 - add good to application

Mark-Spiers-kainos:


goods/models.py
```diff
@@ -15,9 +15,9 @@ class Good(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     description = models.TextField(default=None, blank=True, null=True, max_length=280)
     is_good_controlled = models.CharField(choices=GoodControlled.choices, default=GoodControlled.UNSURE, max_length=20)
-    control_code = models.TextField(default=None, blank=True, null=True)
+    control_code = models.TextField(default='', blank=True, null=True)
     is_good_end_product = models.BooleanField(default=None, blank=True, null=True)
-    part_number = models.TextField(default=None, blank=True, null=True)
+    part_number = models.TextField(default='', blank=True, null=True)
     organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE, default=None)
```
cstobart: changes

goods/models.py
```diff
@@ -13,12 +13,12 @@
...
-    description = models.TextField(default=None, blank=True, null=True, max_length=280)
+    description = models.TextField(max_length=280)
     is_good_controlled = models.CharField(choices=GoodControlled.choices, default=GoodControlled.UNSURE, max_length=20)
-    control_code = models.TextField(default=None, blank=True, null=True)
-    is_good_end_product = models.BooleanField(default=None, blank=True, null=True)
-    part_number = models.TextField(default=None, blank=True, null=True)
-    organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE, default=None)
+    control_code = models.TextField(default='', blank=True, null=True)
+    is_good_end_product = models.BooleanField()
+    part_number = models.TextField(default='', blank=True, null=True)
```
jfaracik: Why was this changed?

Mark-Spiers-kainos: This was changed since it we pass Null/None as values for part_number or control_code, then when we filter the goods, it wasn't finding the new goods being added. 
Also the goods model doesn't reflect the current system and how we want it to work, so I made some minor adjustments to bring it closer to our current system.

applications/views/application_goods.py
```diff
@@ -41,25 +41,44 @@ def get(self, request, application):
...
+            if serializer.is_valid():
+                serializer.save()
+
+                set_application_goods_case_activity(CaseActivityType.ADD_GOOD_TO_APPLICATION, good.description,
+                                                    request.user,
+                                                    application)
+
+                return JsonResponse(data={'good': serializer.data}, status=status.HTTP_201_CREATED)
+
+        return JsonResponse(data={'errors': serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
```
andytaylorkainos: can we invert the logic, so that the final `JsonResponse` is the success? (think we're trying to take the 'failure first' approach)

content_strings/strings.py
```diff
@@ -9,4 +11,11 @@ def get(d, keys):
...
 
-    return get(constants, value)
+    return get(values, value)
+
+
+def get_const_string(value):
+    try:
+        return getattr(constants, value)
+    except AttributeError:
+        return ""
```
andytaylorkainos: single quotes

goods/views.py
```diff
@@ -102,10 +102,13 @@ def post(self, request):
...
         if serializer.is_valid():
-            serializer.save()
+            if 'validate_only' in data and data['validate_only'] is True:
+                return HttpResponse(status=200)
+            else:
+                serializer.save()
 
-            return JsonResponse(data={'good': serializer.data},
-                                status=status.HTTP_201_CREATED)
+                return JsonResponse(data={'good': serializer.data},
```
andytaylorkainos: same again, can we invert the success/failure logic please

applications/views/application_goods.py
Mark-Spiers-kainos: By putting the error response at the end, we can use the same error for bother sections of code. We could seperate these out into both with validate_only and without, if you prefer.

Mark-Spiers-kainos: Discussed with andy, going to keep it the way it is

goods/views.py
```diff
@@ -101,15 +101,18 @@ def post(self, request):
         data['status'] = GoodStatus.DRAFT
         serializer = GoodSerializer(data=data)
 
-        if serializer.is_valid():
+        if not serializer.is_valid():
+            return JsonResponse(data={'errors': serializer.errors},
+                                status=status.HTTP_400_BAD_REQUEST)
+
+        if 'validate_only' in data and data['validate_only'] is True:
+            return HttpResponse(status=200)
```
andytaylorkainos: `status.HTTP_200_OK`

efahy: Done

PR #176, opened 2019-11-04
Make letter layout primary key a UUID.

craiga:
Note that this deletes all existing migrations in `letter_layouts` and `letter_templates`.

Changing primary keys while maintaining existing migrations is very, very difficult (maybe even impossible).


static/letter_layouts/serializers.py
```diff
@@ -6,4 +6,4 @@
 class LetterLayoutSerializer(serializers.ModelSerializer):
     class Meta:
         model = LetterLayout
-        fields = '__all__'
```
craiga: I'd argue that using `__all__` is bad practice.

We might add an extra fields to a model down the road which contains sensitive information. If we use `__all__` on serializers like this, that data might accidentally be leaved via the API.

craiga:
FYI, before switching to this branch, be sure to run the following:

```bash
pipenv run ./manage.py migrate letter_templates zero
pipenv run ./manage.py migrate letter_layouts zero
```

static/letter_layouts/models.py
```diff
@@ -1,9 +1,12 @@
+import uuid
+
 from django.db import models
 
 
 class LetterLayout(models.Model):
-    id = models.CharField(primary_key=True, editable=False, max_length=30)  # letter file name minus extension
-    name = models.CharField(max_length=30)  # friendly name
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    name = models.TextField(help_text="Friendly name")
```
BenAndrew-Kainos: What is `"Friendly name"`

static/letter_layouts/serializers.py
BenAndrew-Kainos: That is quite a good point (it's a bit like using `SELECT * FROM ...`). Something to raise with the devs tomorrow I'd say

static/letter_layouts/models.py
craiga: The name that's shown in the UI.

I was just moving the comment into [`help_text`](https://docs.djangoproject.com/en/2.2/ref/models/fields/#help-text) as it seemed like a more useful place for it.

BenAndrew-Kainos: Ok can we move this to lite-content please

BenAndrew-Kainos:
> FYI, before switching to this branch, be sure to run the following:
> 
> ```shell
> pipenv run ./manage.py migrate letter_templates zero
> pipenv run ./manage.py migrate letter_layouts zero
> ```

Why does it include the word zero?

PR #178, opened 2019-11-04
feat: Timestamp abstract base model

a-bek:


common/models.py
```diff
@@ -0,0 +1,17 @@
...
+    updated_at = models.DateTimeField(db_index=True, null=True)
+
+    class Meta:
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        if not self.id:
+            self.created_at = timezone.now()
+        self.updated_at = timezone.now()
+        super().save(*args, **kwargs)
```
craiga: Did you consider using [`TimeStampedModel` from django-model-utils](https://django-model-utils.readthedocs.io/en/latest/models.html#timestampedmodel)? Or at least [`AutoCreatedField` and `AutoLastModifiedField`](https://github.com/jazzband/django-model-utils/blob/master/model_utils/models.py#L30)?

a-bek: Nice suggestion, will update now :)

cases/migrations/0004_case_timestamp.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('cases', '0003_notification_case_activity'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='case',
+            name='created_at',
+            field=models.DateTimeField(db_index=True, editable=False, null=True),
```
craiga: This migration is out-of-date. It should reference `model_utils.fields.AutoCreatedField` and `model_utils.fields.AutoLastModifiedField`.

PR #182, opened 2019-11-06
Task: Remove usages & tests of all PUT's & POST's in static

BenAndrew-Kainos:


BenAndrew-Kainos:
This change is coming in because no static resource should ever be modified and so it shouldn't have PUT & POST endpoints. I've verified with @jfaracik that these are unused

markm5A8Z9F43913A:
Fix your imports though

PR #187, opened 2019-11-13
bugfix: LT-1762 - Stop gov user changes from creating notification

andytaylorkainos:


cases/models.py
```diff
@@ -292,7 +292,7 @@ def create(cls, activity_type, case, user, additional_text=None, created_at=None
         activity = super(CaseActivity, cls).create(activity_type, case, user, additional_text, created_at,
                                                    save_object, **kwargs)
 
-        if save_object:
+        if save_object and isinstance(user, ExporterUser):
```
stephanieboyd: Move this check to the first line of function logic. No point building the activity if it can't be sent - also, it will help the function visually

stephanieboyd: add a test

andytaylorkainos: Done!

PR #189, opened 2019-11-13
feat: LT-1412 - Case Documents Audit Trail

andytaylorkainos:


cases/views/views.py
```diff
@@ -115,6 +115,11 @@ def post(self, request, pk):
         serializer = CaseDocumentCreateSerializer(data=data, many=True)
         if serializer.is_valid():
             serializer.save()
+
+            for document in serializer.data:
+                case_activity = {'activity_type': 'upload_case_document', 'file_name': document['name']}
```
markm5A8Z9F43913A: Wrong quotes

andytaylorkainos: dont do this to me mark

PR #194, opened 2019-11-15
Task - Add Case Status On Case Type

stephanieboyd:


static/statuses/models.py
```diff
@@ -2,11 +2,19 @@
...
+    status = models.CharField(null=False, blank=False, max_length=50)
+    priority = models.IntegerField(null=False, blank=False)
     is_read_only = models.BooleanField(blank=False, null=True)
+
+
+class CaseStatusOnType(models.Model):
+    class Meta:
+        unique_together = (("type", "status"),)
+
+    type = models.CharField(choices=CaseType.choices, null=False, max_length=35)
```
Mark-Spiers-kainos: I don't like the word type here

BenAndrew-Kainos: case_type better?

Mark-Spiers-kainos: yeah

stephanieboyd: @Mark-Spiers-kainos A different change made, following Eugene's inout

PR #195, opened 2019-11-15
feat: LT-1665 - Control Code Audit Log

andytaylorkainos:


goods/views.py
```diff
@@ -48,18 +48,21 @@ def post(self, request, case_pk):
...
                     serializer = ClcControlGoodSerializer(good, data=data)
                     if serializer.is_valid():
                         serializer.save()
 
-                    control_code = data.get('control_code')
-                    if control_code == "":
-                        control_code = "No control code"
+                    new_control_code = data.get('control_code')
+                    if new_control_code == "":
+                        new_control_code = "No control code"
```
markm5A8Z9F43913A: I'd say "none" would fit better in that sentence.

andytaylorkainos: changed it a little

PR #199, opened 2019-11-18
feat: LT-1028 - Document Generation

BenAndrew-Kainos:


applications/tests/tests_consignee.py
```diff
@@ -4,7 +4,7 @@
 from parameterized import parameterized
 from rest_framework import status
 
-from parties.document.models import PartyDocument
+from parties.models import PartyDocument
```
Mark-Spiers-kainos: could you tidy up the imports a touch? two models from parties.models now in this file and a few others.

cases/models.py
```diff
@@ -337,3 +340,8 @@ class Notification(models.Model):
     ecju_query = models.ForeignKey(EcjuQuery, on_delete=models.CASCADE, null=True)
     case_activity = models.ForeignKey(CaseActivity, on_delete=models.CASCADE, null=True)
     viewed_at = models.DateTimeField(null=True)
+
+
+class CaseType(models.Model):
```
a-bek: Why do we need a CaseType model? CaseTypeEnum achieves the exact same thing while also not adding complexity to our database and queries.

cases/generated_documents/tests/test_generate_document.py
```diff
@@ -0,0 +1,100 @@
...
+        response = self.client.get(url, **self.gov_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        body = response.json()
+        self.assertTrue("errors" in body)
+        self.assertEqual(body["errors"], ["Failed to get preview"])
+
+    @mock.patch("cases.generated_documents.views.html_to_pdf")
+    @mock.patch("cases.generated_documents.views.DocumentOperation.upload_bytes_file")
+    def test_generate_document_success(self, upload_bytes_file_func, html_to_pdf_func):
```
markm5A8Z9F43913A: Can we assert that the mock functions have or have not been called depending on the test? Ideally we'd do this even for the unhappy paths.

cases/generated_documents/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cases.generated_documents import views
+
+app_name = "generated_documents"
+
+urlpatterns = [
+    # ex: /cases/<uuid:pk>/generated-documents/
```
markm5A8Z9F43913A: I think we don't add examples anymore.

efahy:
Please include instructions in the README or the docs folder for installing weasy print or at least a link to the appropriate weasy docs

letter_templates/helpers.py
```diff
@@ -1,9 +1,85 @@
...
+        libraries={"static": "django.templatetags.static", "sass_tags": "sass_processor.templatetags.sass_tags"},
+    )
+
+
+def get_paragraphs_as_html(paragraphs: list):
+    return "\n\n".join([Markdown().convert(paragraph.text) for paragraph in paragraphs])
+
+
+def get_css_location(filename):
+    return BASE_DIR + "/" + CSS_LOCATION + filename + ".css"
```
markm5A8Z9F43913A: Probably should use `os.path.join` and other useful commands instead of manually trying to construct the path ourselves.

letter_templates/helpers.py
```diff
@@ -1,9 +1,85 @@
...
+
+
+def load_css(filename):
+    with open(get_css_location(filename)) as css_file:
+        css = css_file.read()
+    return "<style>\n" + css + "</style>\n"
```
markm5A8Z9F43913A: Does this ever need to be human-readable?

cases/generated_documents/urls.py
BenAndrew-Kainos: Has this been discussed?

cases/models.py
andytaylorkainos: We've done this because we agreed to remove the `ArrayField` from the template model which contained a list of case-types that the template could be used for.

The relationship between case-type and template is `many-to-many` and I know we could create our own many-to-many relationship table that points to the enum. 
But, there was a long discussion (i believe you were off at the time) and we opted to model the case-types so that they could be reflected in the database and re-used across all models.

We're happy to re-open the discussion with the team if you feel strongly about it :)

letter_templates/helpers.py
BenAndrew-Kainos: Done

cases/enums.py
```diff
@@ -1,4 +1,4 @@
-class CaseType:
+class CaseTypeEnum:
```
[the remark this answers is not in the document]
BenAndrew-Kainos: To distinguish from the CaseType model

letter_templates/helpers.py
```diff
@@ -1,9 +1,83 @@
...
 
 
 def get_letter_template(pk):
     try:
         return LetterTemplate.objects.get(pk=pk)
     except LetterTemplate.DoesNotExist:
         raise NotFoundError({"letter_template": "LetterTemplate not found - " + str(pk)})
+
+
+class InvalidVarException(Exception):
```
markm5A8Z9F43913A: Feels like this doesn't belong here. Might be worth putting exceptions in a separate file.

BenAndrew-Kainos: Done

applications/tests/tests_consignee.py
BenAndrew-Kainos: Done

cases/generated_documents/views.py
```diff
@@ -0,0 +1,96 @@
...
+        if "error" in self.html:
+            return JsonResponse(data={"errors": [self.html["error"]]}, status=status.HTTP_400_BAD_REQUEST)
+
+        return None
+
+    def get(self, request, pk):
+        """
+        Get a preview of the document to be generated
+        """
+        error_response = self._fetch_generated_document_data(request.GET, pk)
```
markm5A8Z9F43913A: If you're fetching document data, I'd expect it to be stored in `document_data` rather than `error_response`. Is the method named incorrectly?

documents/helpers.py
```diff
@@ -0,0 +1,37 @@
...
+    @staticmethod
+    def _get_bucket_name():
+        return env("AWS_STORAGE_BUCKET_NAME")
+
+    @staticmethod
+    def generate_s3_key(document_name, file_extension):
+        return f"{document_name}-{uuid.uuid4()}.{file_extension}"
+
+    def upload_bytes_file(self, raw_file, s3_key):
+        bucket = self.get_client()
```
a-bek: Its a client object, not a bucket object so the name is misleading.

letter_templates/views.py
```diff
@@ -48,3 +67,19 @@ def update(self, request, *args, **kwargs):
             return JsonResponse(serializer.data)
 
         return JsonResponse({"errors": serializer.errors})
+
+
+class TemplatePreview(APIView):
```
jfaracik: Instead of APIView could you extend RetrieveAPIView https://www.django-rest-framework.org/api-guide/generic-views/#retrieveapiview

documents/helpers.py
```diff
@@ -0,0 +1,37 @@
+import logging
+import uuid
+
+import boto3
+
+from conf.settings import env
+from lite_content.lite_api.documents import DocumentsEndpoint
+
+
+class DocumentOperation:
```
a-bek: Id make this functional and instantiate the client at module level.

SO when you import it:

```
from documents import s3
# API
s3.generate_key(...)
s3.delete_file(...)
s3.upload_bytes_file(...)
```

Rather than having to instantiate an object everywhere.

a-bek: Or even name the module `s3_client`

letter_templates/serializers.py
```diff
@@ -23,37 +24,31 @@ class LetterTemplateSerializer(serializers.ModelSerializer):
...
+    class Meta:
+        model = LetterTemplate
+        fields = "__all__"
+
     @staticmethod
-    def validate_restricted_to(attrs):
+    def validate_case_types(attrs):
         if not attrs:
-            raise serializers.ValidationError("Select at least one case restriction for your letter template")
+            raise serializers.ValidationError("You need at least one case type for your letter template")
```
jfaracik: Should this be in content? (I genuinely don't know)

letter_templates/serializers.py
```diff
@@ -23,37 +24,31 @@ class LetterTemplateSerializer(serializers.ModelSerializer):
     )
     letter_paragraphs = serializers.PrimaryKeyRelatedField(queryset=PicklistItem.objects.all(), many=True)
 
-    restricted_to = serializers.ListField(
-        child=serializers.CharField(),
-        error_messages={"required": "Select which types of case this letter template can apply to",},
+    case_types = PrimaryKeyRelatedSerializerField(
+        queryset=CaseType.objects.all(),
+        serializer=CaseTypeSerializer,
+        error_messages={"required": "Select the case types you want to use for this letter template"},
```
jfaracik: Content?

conf/settings.py
```diff
@@ -118,6 +118,26 @@
...
 TEST_OUTPUT_DIR = "test-results/unittest/"
 
+STATIC_URL = "/assets/"
+
+# SCSS
+STATIC_ROOT = os.path.join(BASE_DIR, "assets")
+CSS_ROOT = os.path.join(STATIC_ROOT, "css")
+SASS_ROOT = STATIC_ROOT
+SASS_PROCESSOR_ROOT = STATIC_ROOT
+COMPRESS_PRECOMPILERS = (("text/x-scss", "django_libsass.SassCompiler"),)
```
jfaracik: Does SASS stuff still need to be here?

cases/generated_documents/urls.py
BenAndrew-Kainos: Done

cases/generated_documents/views.py
BenAndrew-Kainos: Improved this so it now makes more sense

documents/helpers.py
BenAndrew-Kainos: Done

documents/helpers.py
BenAndrew-Kainos: Done

Pipfile
```diff
@@ -13,6 +13,7 @@ django-environ = "==0.4.5"
 django-model-utils = "*"
 django-reversion = "==3.0.3"
 djangorestframework = "==3.9.2"
+django-sass-processor = "==0.7.2"
```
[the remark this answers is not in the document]
BenAndrew-Kainos: Technically we could just use CSS files but this has not been the pattern in the past. We did consider removing CSS from the gitignore and doing away with SASS on the API. Do you know if there is any implications/requirements meaning we couldn't do that?

README.md
```diff
@@ -26,6 +26,14 @@ Service for handling backend calls in LITE.
     * `docker-compose up`
 * Go to the index page (e.g. `http://localhost:8100`)
 
+## Installing WeasyPrint for document generation
+To produce PDF documents you will also need to install WeasyPrint. 
+Do this after installing the python packages in the Pipfile;
+
+> MacOS: https://weasyprint.readthedocs.io/en/stable/install.html#macos
+
+> Linux: https://weasyprint.readthedocs.io/en/stable/install.html#debian-ubuntu
```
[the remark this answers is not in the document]
BenAndrew-Kainos: This is more for OPs if they need to manually fix an environment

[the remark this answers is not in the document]
BenAndrew-Kainos: I can remove if it's upsetting

conf/settings.py
andytaylorkainos: These packages are used to compress the SASS into css when generating the views :) (i suppose we could just have raw css if thats what you mean?)

cases/enums.py
```diff
@@ -38,3 +38,10 @@ class AdviceType:
         (NOT_APPLICABLE, "Not Applicable"),
         (CONFLICTING, "Conflicting"),
     ]
+
+
+class CaseDocumentType:
```
efahy: I think the name `CaseDocumentType` is misleading.  With that name, I'd expect "ECJU Letter" or something to be defined in here.  How about `CaseDocumentStatus` or `CaseDocumentState` or something?

cases/generated_documents/helpers.py
```diff
@@ -0,0 +1,33 @@
...
+from lite_content.lite_api.letter_templates import LetterTemplatesPage
+
+
+def html_to_pdf(request, html: str, template_name: str):
+    html = HTML(string=html, base_url=request.build_absolute_uri())
+    css = CSS(filename=get_css_location(template_name))
+    return html.write_pdf(stylesheets=[css])
+
+
+def get_letter_templates_for_case(case):
```
efahy: Since its possible that none may exist for the case type, how about renaming to `find.....`

cases/generated_documents/helpers.py
```diff
@@ -0,0 +1,33 @@
...
+    return LetterTemplate.objects.filter(case_types__id=case.type)
+
+
+def get_letter_template_for_case(id, case):
+    try:
+        return LetterTemplate.objects.get(pk=id, case_types__id=case.type)
```
efahy: Docstring please, explaining why we're searching by pk and case type, i.e. the template with the id may not be applicable to the case type, i.e. we're sneaking a bit of filtering in here!

cases/generated_documents/views.py
```diff
@@ -0,0 +1,91 @@
...
+
+        try:
+            pdf = html_to_pdf(request, document_html, self.template.layout.filename)
+        except Exception:  # noqa
+            return JsonResponse(
+                {"errors": [GeneratedDocumentsEndpoint.PDF_ERROR]}, status=status.HTTP_500_INTERNAL_SERVER_ERROR
+            )
+
+        s3_key = s3_operations.generate_s3_key(self.template.name, "pdf")
+        document_name = f"{s3_key[:len(self.template.name) + 6]}.pdf"
```
markm5A8Z9F43913A: Probably a good idea to explain this in a comment

cases/generated_documents/views.py
```diff
@@ -0,0 +1,91 @@
...
+            return JsonResponse(data={"errors": [document_html["error"]]}, status=status.HTTP_400_BAD_REQUEST)
+
+        return JsonResponse(data={"preview": document_html}, status=status.HTTP_200_OK)
+
+    @transaction.atomic
+    def post(self, request, pk):
+        """
+        Create a generated document
+        """
+
```
efahy: Remove blank line please, for consistency (and cause I prefer it that way)

cases/generated_documents/views.py
```diff
@@ -0,0 +1,91 @@
...
+            return JsonResponse({"errors": [LetterTemplatesPage.MISSING_TEMPLATE]}, status=status.HTTP_400_BAD_REQUEST)
+
+        tpk = request_params["template"]
+        self.case = get_case(pk)
+        self.template = get_letter_template_for_case(tpk, self.case)
+        return get_preview(template=self.template, case=self.case)
+
+    def get(self, request, pk):
+        """
+        Get a preview of the document to be generated
```
efahy: I don't love that the get in `GeneratedDocuments` doesn't return a generated document.  Perhaps create a new `DocumentPreview` View for this get?

static/management/commands/seedcasetypes.py
```diff
@@ -0,0 +1,23 @@
+from django.db import transaction
+
+from cases.models import CaseType
+from static.management.SeedCommand import SeedCommand
+
+CASE_TYPES_FILE = "lite_content/lite-api/case_types.csv"
```
efahy: Its a bit rubbish that we have both an enum and a CSV where we define case types - can we have just one place, e.g. seed from the enum?

letter_templates/serializers.py
andytaylorkainos: Done!

letter_templates/serializers.py
andytaylorkainos: Done!

letter_templates/views.py
andytaylorkainos: Done!

cases/generated_documents/views.py
andytaylorkainos: Done!

cases/generated_documents/models.py
```diff
@@ -0,0 +1,8 @@
+from django.db import models
+
+from cases.models import CaseDocument
+from letter_templates.models import LetterTemplate
+
+
+class GeneratedDocument(CaseDocument):
```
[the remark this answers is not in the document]
andytaylorkainos: Done!

cases/enums.py
andytaylorkainos: Done!

cases/generated_documents/views.py
andytaylorkainos: moved to a different endpoint

cases/generated_documents/views.py
andytaylorkainos: Done!

cases/generated_documents/views.py
```diff
@@ -0,0 +1,103 @@
...
+            "file_name": document_name,
+            "template": template.name,
+        }
+        case_activity = CaseActivity.create(case=case, user=request.user, **case_activity)
+
+        try:
+            s3_operations.upload_bytes_file(raw_file=pdf, s3_key=s3_key)
+        except Exception:  # noqa
+            case_activity.delete()
+            generated_doc.delete()
```
Mark-Spiers-kainos: function is transaction atomic, don't need these here.

cases/generated_documents/helpers.py
andytaylorkainos: Resolved in discussion

cases/generated_documents/helpers.py
andytaylorkainos: Docstring added

cases/generated_documents/views.py
andytaylorkainos: Done!

conf/settings.py
BenAndrew-Kainos: We've changed to CSS :)

Pipfile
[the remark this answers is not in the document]
BenAndrew-Kainos: We've switched to CSS :)

static/management/commands/seedcasetypes.py
BenAndrew-Kainos: This seeding vs. enums debate has caused a lot of issues recently so it will be good to get a clear strategy outlined. 

As discussed this morning we're currently going with a model rather than enum approach for static data. This is a change from previous implementation where all static data was defined in enums.

To do this we can either use the enums to seed data or simply do away with enums all together. I would suggest the latter as otherwise we run the risk of data inconsistency (the same data is stored in two places) and will still run into some of the enum issues we've discussed before.

However, we haven't agreed to all change over currently so code is still being used that relies on enums. Additionally changing from enum to models may take a lot of work and pose a high risk to potentially breaking the system. Therefore I propose we produce tasks on a case by case basis to tackle these enums so that we can all verify that we will no longer use it, and produce less risk with it being coupled with a story

andytaylorkainos: I completely agree. In addition to that, I think we should remove `type` from the model which is validated against the enum.
I suggest that any enum class we use is strictly for easily referencing the `id` `name` quickly in code e.g.
```
CaseType.objects.filter(id=CaseTypeEnum.APPLICATION).
```
i.e. the enum is solely for ease-of-use rather than what drives the data being inserted to the DB.

cases/serializers.py
```diff
@@ -475,3 +478,15 @@ class GoodCountryDecisionSerializer(serializers.ModelSerializer):
...
+class CaseTypeSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CaseType
+        fields = (
+            "id",
+            "name",
+        )
+
+    def to_representation(self, instance):
+        return dict(key=instance.id, value=instance.name)
```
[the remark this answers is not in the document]
andytaylorkainos: This is just to follow the convention of all other key-value pairs (front-end approach) :)

PR #200, opened 2019-11-18
Task: Update seed operations to update or create based on ID

BenAndrew-Kainos:


static/management/SeedCommand.py
```diff
@@ -33,10 +33,17 @@ def handle(self, *args, **options):
     @staticmethod
     def read_csv(filename):
         with open(filename, newline="") as csvfile:
-            reader = csv.reader(csvfile)
-            next(reader)  # skip the headers
+            reader = csv.DictReader(csvfile)
             return list(reader)
 
+    @staticmethod
+    def update_or_create(model: models.Model, rows: list):
```
efahy: Docstring comment please, that includes information about what the `rows` parameter needs to be

static/management/SeedCommand.py
```diff
@@ -33,10 +33,17 @@ def handle(self, *args, **options):
...
+        for row in rows:
+            if model.objects.filter(id=row['id']).exists():
```
efahy: Since you're querying by ID and expecting one row, shouldn't you use `get` instead of `filter`?  If you change, then instead of an if/else, it becomes a try/except

static/management/SeedCommand.py
```diff
@@ -33,10 +33,17 @@ def handle(self, *args, **options):
     @staticmethod
     def read_csv(filename):
```
efahy: Docstring comment here explaining how `csv.DictReader` works - specifically how the CSV needs to have a header row so it can make a dict with the headings as keys

static/management/commands/seedlayouts.py
```diff
@@ -13,11 +13,9 @@ def operation(self, *args, **options):
         """
         pipenv run ./manage.py seedlayouts
         """
-        # Add layouts
-        reader = self.read_csv(FILE)
-        for row in reader:
-            LetterLayout.objects.get_or_create(filename=row[0], name=row[1])
-            print("Seeded %s layout" % row[1])
+        csv = self.read_csv(FILE)
```
markm5A8Z9F43913A: I like how you've done `STATUS_ON_TYPE_FILE`, but I don't like how in some places you just use `FILE`. I'd be more explicit in naming it.

BenAndrew-Kainos: Done :)

PR #204, opened 2019-11-21
Refactor: Case inheritance

Mark-Spiers-kainos:


applications/migrations/0002_baseapplication_countryonapplication_externallocationonapplication_goodonapplication_hmrcquery_opena.py
```diff
@@ -0,0 +1,176 @@
...
+                    "unit",
+                    models.CharField(
+                        choices=[
+                            ("GRM", "Gram(s)"),
+                            ("KGM", "Kilogram(s)"),
+                            ("NAR", "Number of articles"),
+                            ("MTK", "Square metre(s)"),
+                            ("MTR", "Metre(s)"),
+                            ("LTR", "Litre(s)"),
+                            ("MTQ", "Cubic metre(s)"),
```
markm5A8Z9F43913A: Strings

applications/libraries/case_activity.py
```diff
@@ -4,7 +4,7 @@
 
 def _get_case_from_application(application):
     try:
-        return application.case.get()
+        return application.status != None
```
a-bek: The function name suggests you are retrieving a case, this returns a boolean, or `None` if it doesnt exist

cases/views/search/views.py
```diff
@@ -29,7 +29,7 @@ def get(self, request, *args, **kwargs):
                 case_type=request.GET.get("case_type"),
                 sort=request.GET.get("sort"),
                 date_order="-" if queue_id in SYSTEM_QUEUES else "",
-            )
+            ).filter(status__isnull=False)
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: It can have a null status cause it's an application that is not submitted.

applications/serializers/generic_application.py
```diff
@@ -57,11 +56,7 @@ def get_status(self, instance):
         return None
 
     def get_case(self, instance):
-        try:
-            return Case.objects.get(application=instance).id
-        except Case.DoesNotExist:
-            # Case will only exist if application has been submitted
-            return None
+        return instance.pk
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: The function is just updated so we don't need to look at stripping the front-ends and rebuilding within this refactor. Now that the main purpose of this refactor is achieved, there are many sub points that could be refactored and made cleaner, but this can happen during stories so that this doesn't drag on and block us.
You will see similar examples of code like this in the other serializers, which can be tidied up now, with minor changes.

users/serializers.py
```diff
@@ -187,13 +183,9 @@ def get_parent(self, obj):
 
     def get_parent_type(self, obj):
         if obj.case_note:
-            object = next(
-                item for item in [obj.case_note.case.application, obj.case_note.case.query] if item is not None
-            )
+            object = next(item for item in [obj.case_note.case.type] if item is not None)
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: existing code

goods/serializers.py
```diff
@@ -32,11 +32,10 @@ def get_documents(self, instance):
...
 
     def get_query_id(self, instance):
-        try:
-            clc_query = ControlListClassificationQuery.objects.get(good=instance)
+        clc_query = ControlListClassificationQuery.objects.filter(good=instance)
+        if clc_query:
             return clc_query.id
-        except ControlListClassificationQuery.DoesNotExist:
-            return None
+        return None
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: good catch

goods/libraries/get_goods.py
```diff
@@ -27,7 +27,7 @@ def get_good_document(good: Good, pk):
 
 def get_goods_from_case(case):
     if case.query:
-        query = get_exporter_query(case.query.id)
+        query = get_exporter_query(case.id)
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: This function wasn't being used. Case doesn't contain a query anymore, so it would have thrown an error. Have removed it.

applications/libraries/case_activity.py
Mark-Spiers-kainos: Have removed the function since it was used in one place, and turned into if statement

goods/serializers.py
```diff
@@ -32,11 +32,10 @@ def get_documents(self, instance):
         return None
 
     def get_query_id(self, instance):
-        try:
-            clc_query = ControlListClassificationQuery.objects.get(good=instance)
+        clc_query = ControlListClassificationQuery.objects.filter(good=instance)
+        if clc_query:
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: have updated the `clc_query.id `to be `clc_query.first().id` should fix this issue

users/serializers.py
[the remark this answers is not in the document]
a-bek: Ok, but i'd still change it you're editing the lines so it was existing code, now its new. Also can't we just return the value rather than assign to `object`?

cases/views/search/views.py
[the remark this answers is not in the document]
a-bek: Ok. I still think adding a filter here in the view is going against the design pattern. Perhaps add a new manager method called `submitted`?

cases/libraries/get_case.py
```diff
@@ -8,7 +8,7 @@ def get_case(pk):
     Returns a case or returns a 404 on failure
     """
     try:
-        return Case.objects.get(pk=pk)
+        return Case.objects.filter(status__isnull=False).get(pk=pk)
```
a-bek: If you add the `submitted` to the manager, here you can just do Case.objects.submitted().get(pk=pk). Reusing the function and more explicit.

cases/serializers.py
```diff
@@ -182,6 +177,16 @@ class Meta:
...
 
+    def get_application(self, instance):
+        # The case has a reference to a BaseApplication but
+        # we need the full details of the application it points to
+        if instance.type in [CaseType.APPLICATION, CaseType.HMRC_QUERY]:
+            application = get_application(instance.id)
+            serializer = get_application_view_serializer(application)
+            return serializer(application).data
+
+        return None
```
a-bek: Delete `return None`. It is the default behaviour at the end of any python function block.

cases/views/search/views.py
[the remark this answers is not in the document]
Mark-Spiers-kainos: yeah I have moved this to the manager that we currently have for cases.

users/serializers.py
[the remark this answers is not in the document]
Mark-Spiers-kainos: changed the names for the variables, let me know if they are acceptable. I wasn't 100% what to name them, since they are so generic.

organisations/models.py
```diff
@@ -46,6 +43,13 @@ def get_users(self):
         return [x.user for x in user_organisation_relationships]
 
 
+class UserOrganisationRelationship(models.Model):
```
a-bek: Can you inhereit from the abstract class Timestampable? Under `/common/models.py`

cases/models.py
```diff
@@ -3,51 +3,41 @@
...
 from teams.models import Team
-from users.models import BaseUser, ExporterUser, GovUser, UserOrganisationRelationship
+from users.models import BaseUser, ExporterUser, GovUser
 
 
 @reversion.register()
 class Case(models.Model):
     """
-    Wrapper for application and query model intended for internal users.
+    Base model for applications and queries
```
a-bek: Can this base class also inherit from `Timestampable` model under `/common/models.py`

applications/serializers/hmrc_query.py
```diff
@@ -81,15 +82,11 @@ def __init__(self, **kwargs):
...
 
     class Meta:
         model = HmrcQuery
-        fields = (
-            "reasoning",
-            "application_type",
-            "organisation",
-            "hmrc_organisation",
-        )
+        fields = ("reasoning", "application_type", "organisation", "hmrc_organisation", "type")
```
andytaylorkainos: can you put a comma at the end of the tuple pls? :)

cases/serializers.py
```diff
@@ -182,6 +175,14 @@ class Meta:
             "has_advice",
         )
 
+    def get_application(self, instance):
+        # The case has a reference to a BaseApplication but
+        # we need the full details of the application it points to
+        if instance.type in [CaseType.APPLICATION, CaseType.HMRC_QUERY]:
+            application = get_application(instance.id)
+            serializer = get_application_view_serializer(application)
```
andytaylorkainos: the function name for retrieving the serializer could be improved (since `get_application` is suggesting its only application types being returned)
Maybe something like `get_serializer_for_case_type`?

goods/views.py
```diff
@@ -100,13 +100,15 @@ def get(self, request):
         """
         description = request.GET.get("description", "")
         part_number = request.GET.get("part_number", "")
-        control_rating = request.GET.get("control_rating", "")
+        control_rating = request.GET.get("control_rating", None)
```
andytaylorkainos: This should default to `None` if it's not there, so we dont need to supply it as a fallback value:
`control_rating = request.GET.get("control_rating")`

users/models.py
```diff
@@ -85,9 +83,9 @@ def send_notification(self, **kwargs):
 
 class ExporterUser(BaseUser):
     def send_notification(self, case_note=None, query=None, ecju_query=None):
-        from cases.models import Notification
-
-        # circular import prevention
+        # getting a reference to models by name to avoid circular imports
+        Query = get_model("queries.Query")
+        Notification = get_model("cases.Notification")
```
andytaylorkainos: nice!

users/models.py
```diff
@@ -120,9 +111,9 @@ def unassign_from_cases(self):
 
     # pylint: disable=W0221
     def send_notification(self, case_activity=None):
-        from cases.models import Notification
+        Notification = get_model("cases.Notification")
 
-        # circular import prevention
+        # getting models due to circular imports
```
andytaylorkainos: can you put the comment above the code pls :)

cases/serializers.py
Mark-Spiers-kainos: It does just return Application types, since I never updated the function inside it. return StandardApplication, OpenApplication or HMRCQuery instance

users/models.py
Mark-Spiers-kainos: sure

goods/views.py
Mark-Spiers-kainos: done

applications/serializers/hmrc_query.py
Mark-Spiers-kainos: done

applications/views/applications.py
```diff
@@ -153,7 +153,7 @@ def delete(self, request, application):
         """
         Deleting an application should only be allowed for draft applications
         """
-        if application.submitted_at:
+        if application.status:
```
andytaylorkainos: should this not be `.status == DRAFT` (asuming we have a draft status now)

Mark-Spiers-kainos: yeah, and it's all implemented now.

static/statuses/libraries/case_status_validate.py
```diff
@@ -0,0 +1,8 @@
+from static.statuses.enums import CaseStatusEnum
+from static.statuses.models import CaseStatus
+
+
+def case_status_draft(status):
```
andytaylorkainos: can you prepend `is_` to this method name please

Mark-Spiers-kainos: sure

applications/views/external_locations.py
```diff
@@ -53,15 +54,20 @@ def post(self, request, application):
...
+        ):
             return JsonResponse(
                 data={
                     "errors": {"external_locations": [f"Application status {application.status.status} is read-only."]}
                 },
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
-        if not application.status or application.status.status == CaseStatusEnum.APPLICANT_EDITING:
+        if application.status.status in [
```
andytaylorkainos: howcome you check if the status is draft/editing using this approach, but in `applications/views/countries.py` you use ```            if (
                case_status_draft(application.status.status)
                or application.status.status == CaseStatusEnum.APPLICANT_EDITING
            ):```

applications/views/external_locations.py
```diff
@@ -146,13 +152,18 @@ class ApplicationRemoveExternalLocation(APIView):
...
+        if not case_status_draft(application.status.status) and application.status.status in get_case_statuses(
+            read_only=True
+        ):
             return JsonResponse(
                 data={"error": f"Application status {application.status.status} is read-only."},
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
-        if application.status and application.status.status != CaseStatusEnum.APPLICANT_EDITING:
+        if not case_status_draft(application.status.status) and application.status.status not in [
```
andytaylorkainos: seems like this logic could be wrapped into one function to handle it (expanding upon my previous comment)

cases/managers.py
```diff
@@ -113,3 +99,7 @@ def search(
             case_qs = case_qs.order_by_status(order="-" if sort.startswith("-") else "")
 
         return case_qs
+
+    def submitted(self):
+        draft = get_case_status_by_status(CaseStatusEnum.DRAFT)
```
andytaylorkainos: starting to think `get_case_status_by_status` needs a better name:
```
   get_case_status_pk_from_enum
```

goods/views.py
```diff
@@ -154,12 +154,12 @@ def get(self, request, pk):
             serializer = GoodSerializer(good)
 
             # If there's a query with this good, update the notifications on it
-            try:
-                query = ControlListClassificationQuery.objects.get(good=good)
-                request.user.notification_set.filter(case_note__case__query=query).update(viewed_at=timezone.now())
+
+            query = ControlListClassificationQuery.objects.filter(good=good)
```
andytaylorkainos: why did we change it to this approach instead of `get`?

queries/control_list_classifications/views.py
```diff
@@ -40,16 +42,20 @@ def post(self, request):
 
         good.status = GoodStatus.CLC_QUERY
         good.control_code = data["not_sure_details_control_code"]
-        good.save()
 
         clc_query = ControlListClassificationQuery.objects.create(
-            details=data["not_sure_details_details"], good=good, organisation=data["organisation"],
+            details=data["not_sure_details_details"],
```
andytaylorkainos: can we do `dat.get()` on all of these (if data doesnt contain these keys, it will fall over)

queries/control_list_classifications/views.py
```diff
@@ -40,16 +42,20 @@ def post(self, request):
...
+            good=good,
+            organisation=data["organisation"],
+            type=CaseTypeEnum.CLC_QUERY,
+            status=get_case_status_by_status(CaseStatusEnum.SUBMITTED),
+            submitted_at=datetime.now(timezone.utc),
```
andytaylorkainos: can we use django's timezone.utc for this?


```from django.utils import timezone

now = timezone.now()
```

goods/views.py
Mark-Spiers-kainos: I feel we throw exceptions too commonly. To me Exceptions should be cases our application does not expect. In the case of a good's details. It may or may not have a query, so we should treat it as such, and program to accept both cases.

queries/control_list_classifications/views.py
Mark-Spiers-kainos: have updated all but the organisation, since it will fall over anyway if an organisation is not given.

queries/control_list_classifications/views.py
Mark-Spiers-kainos: done

PR #211, opened 2019-11-29
Bug: LT-1803 URL error message

BenAndrew-Kainos:


parties/serializers.py
```diff
@@ -29,6 +30,16 @@ class Meta:
             "document",
         )
 
+    @staticmethod
+    def validate_website(value):
+        if value:
+            val = URLValidator()
+            url = f"https://{value}"
+            val(url)
```
markm5A8Z9F43913A: If the URL is invalid, will this not throw an exception? Please add some tests for a `None`, a valid and an invalid URL.

parties/serializers.py
```diff
@@ -29,6 +30,16 @@ class Meta:
...
+            return url
+        else:
+            return value
```
markm5A8Z9F43913A: The `else` statement is probably not necessary the way the code is written now. But that's less important. Why do you want to return the invalid original value if it's empty?

parties/serializers.py
BenAndrew-Kainos: The string will not be verified if None/empty (the opening if statement) and it does throw exceptions for invalid URLs

parties/serializers.py
BenAndrew-Kainos: Because if it's empty we don't want to validate it (field is optional). The return could instead be `return ''` if you want

parties/serializers.py
```diff
@@ -29,6 +30,16 @@ class Meta:
             "document",
         )
 
+    @staticmethod
+    def validate_website(value):
+        if value:
+            val = URLValidator()
```
stephanieboyd: Call this `validator`, val is an abbreviation of value, and we've also a value argument

BenAndrew-Kainos: done

parties/serializers.py
```diff
@@ -29,6 +30,16 @@ class Meta:
             "document",
         )
 
+    @staticmethod
+    def validate_website(value):
+        if value:
+            validator = URLValidator()
+            url = f"https://{value}"
```
andytaylorkainos: Might need to account for when the user actually did provide `https` (only inject `https` if it has not been provided)

parties/serializers.py
```diff
@@ -29,6 +30,16 @@ class Meta:
             "document",
         )
 
+    @staticmethod
+    def validate_website(value):
+        if value:
```
stephanieboyd: Gimme a docstring

parties/serializers.py
BenAndrew-Kainos: Just thought of that too :)

parties/serializers.py
BenAndrew-Kainos: done

parties/serializers.py
markm5A8Z9F43913A: Where is the exception being caught?

BenAndrew-Kainos: It's doesn't need to be caught. You let the exception happen and that's how it sends the error back to the frontend

BenAndrew-Kainos: So the behaviour with URlValidator() is the same and we're just making http/https optional

parties/serializers.py
BenAndrew-Kainos: Changed it to return '' for clarity

PR #213, opened 2019-12-02
Bug: LT-1690 don't show deactivated flags on assign

Mark-Spiers-kainos:


flags/views.py
```diff
@@ -41,6 +42,9 @@ def get(self, request):
         if team:
             flags = flags.filter(team=request.user.team.id)
 
+        if include_deactivated == "False":
```
stephanieboyd: Could we not just use the falsey value? ``` if not include_deactivated:```

andytaylorkainos: There's also a helper function `str_to_bool`

Mark-Spiers-kainos: done

PR #215, opened 2019-12-03
chore: SSO refactor

jfaracik:
Update the first and last name of exporter users from SSO


users/views.py
```diff
@@ -46,9 +42,16 @@ def post(self, request, *args, **kwargs):
         except ParseError:
             return JsonResponse(data={"errors": "Invalid Json"}, status=status.HTTP_400_BAD_REQUEST)
         email = data.get("email")
+        first_name = data.get("user_profile").get("first_name")
+        last_name = data.get("user_profile").get("last_name")
```
BenAndrew-Kainos: Could you just use these lines on line 52 & 53

users/views.py
```diff
@@ -45,10 +41,14 @@ def post(self, request, *args, **kwargs):
             data = JSONParser().parse(request)
         except ParseError:
             return JsonResponse(data={"errors": "Invalid Json"}, status=status.HTTP_400_BAD_REQUEST)
-        email = data.get("email")
 
         try:
-            user = ExporterUser.objects.get(email=email)
+            user = ExporterUser.objects.get(email=data.get("email"))
```
stephanieboyd: The tests assert on a status code 200, this is never explicitly returned. Could you improve this and set it explicitly?

PR #217, opened 2019-12-04
LT-1761 Gov user to change status when in applicant editing state

stephanieboyd:


applications/tests/tests_application_denial.py
```diff
@@ -0,0 +1,82 @@
+from parameterized import parameterized
```
stephanieboyd: I pulled these out, didn't feel like they should have been where they were.

applications/libraries/application_helpers.py
```diff
@@ -25,12 +25,6 @@ def validate_status_can_be_set_by_exporter_user(original_status: str, new_status
...
-    if new_status == CaseStatusEnum.APPLICANT_EDITING:
-        return f'Setting application status to "{new_status}" is not allowed for GovUsers.'
-
-    if original_status == CaseStatusEnum.APPLICANT_EDITING:
-        return (
-            f"Setting application status when its existing status is "
-            f'"{original_status}" is not allowed for GovUsers.'
-        )
+def validate_status_can_be_set_by_gov_user(status: str) -> Optional[str]:
+    if status == CaseStatusEnum.APPLICANT_EDITING:
```
andytaylorkainos: Have you got a test to show that gov users `can`  change the status when the applicant is editing?

applications/tests/tests_application_denial.py
stephanieboyd: Update, these are deleted. They were for unnecessary logic.

applications/serializers/generic_application.py
```diff
@@ -147,23 +146,5 @@ def update(self, instance, validated_data):
         if validated_data.get("status") == get_case_status_by_status(CaseStatusEnum.FINALISED):
             ApplicationDenialReason.objects.filter(application=get_application(instance.id)).delete()
 
-        # If the status has been set to under final review, add reason_details to application
```
stephanieboyd: This has been removed after chatting with Anne - we're not sure why it was added. Some tidying happened around it's removal (removing tests, removing unused serializers)

PR #218, opened 2019-12-04
feat: LT-1522 - Exporter Generated Docs

andytaylorkainos:


static/statuses/models.py
```diff
@@ -2,7 +2,7 @@
 
 from django.db import models
 
-from cases.enums import CaseTypeEnum
+from static.case_types.enums import CaseType
```
jfaracik: Maybe.. maybe.. call this CaseTypes? 👀

applications/views/documents.py
```diff
@@ -109,3 +113,29 @@ def delete(self, request, application, goods_type_pk):
             return JsonResponse(data={"error": "No such goods type"}, status=status.HTTP_400_BAD_REQUEST)
 
         return delete_goods_type_document(goods_type)
+
+
+class GeneratedDocuments(APIView):
```
jfaracik: ListAPIView please :)

cases/generated_documents/models.py
```diff
@@ -2,7 +2,17 @@
 
 from cases.models import CaseDocument
 from letter_templates.models import LetterTemplate
+from users.models import UserOrganisationRelationship
 
 
 class GeneratedCaseDocument(CaseDocument):
     template = models.ForeignKey(LetterTemplate, on_delete=models.DO_NOTHING)
+
```
jfaracik:
```class Meta:
        db_table = "temp_user"```

File: static/statuses/models.py (hunk 1 above)
Comment (andytaylorkainos) [reply static/statuses/models.py]: reverted back to the enum name

File: applications/views/documents.py (hunk 2 above)
Comment (andytaylorkainos) [reply applications/views/documents.py]: Cant do it because i need the decorator on the `get` :(

File: cases/generated_documents/models.py (hunk 3 above)
Comment (andytaylorkainos) [reply cases/generated_documents/models.py]: we aspire to do this as one big fix for all tables

**Comment by andytaylorkainos** [comment]
I actually agree - i was only moving CaseType over because its the same approach that was taken for the other lookup tables. i'll revert it back into `cases/` and we can have a plan to move the other ones into where they should be :)

File: users/views.py [hunk 4]
```diff
@@ -156,6 +156,8 @@ def get_queryset(self):
             | Q(query__organisation__id=organisation_id)
             | Q(ecju_query__case__application__organisation_id=organisation_id)
             | Q(ecju_query__case__query__organisation_id=organisation_id)
+            | Q(generated_case_document__case__application__organisation__id=organisation_id)
+            | Q(generated_case_document__case__application__organisation__id=organisation_id)
```
Mark-Spiers-kainos: duplicate lines.

andytaylorkainos: Woops! nice find

andytaylorkainos: Done

users/serializers.py
```diff
@@ -181,14 +181,24 @@ class NotificationSerializer(serializers.ModelSerializer):
     def get_object(self, obj):
         return next(
             item
-            for item in [getattr(obj, "case_note"), getattr(obj, "query"), getattr(obj, "ecju_query"),]
+            for item in [
```
a-bek: Perhaps add a `get_item` method on the Notification class?

users/models.py
```diff
@@ -107,6 +107,8 @@ def send_notification(self, case_note=None, query=None, ecju_query=None):
             Notification.objects.create(user=self, query=actual_query)
         elif ecju_query:
             Notification.objects.create(user=self, ecju_query=ecju_query)
+        elif generated_case_document:
+            Notification.objects.create(user=self, generated_case_document=generated_case_document)
         else:
             raise Exception("ExporterUser.send_notification: objects expected have not been added.")
```
stephanieboyd: This error message could be improved. `No notification created due to yada yada`

cases/models.py
```diff
@@ -330,6 +330,9 @@ class Notification(models.Model):
     case_note = models.ForeignKey(CaseNote, on_delete=models.CASCADE, null=True)
     query = models.ForeignKey("queries.Query", on_delete=models.CASCADE, null=True)
     ecju_query = models.ForeignKey(EcjuQuery, on_delete=models.CASCADE, null=True)
+    generated_case_document = models.ForeignKey(
```
stephanieboyd: Is cascade really the behaviour we want here?

applications/views/documents.py
```diff
@@ -109,3 +113,29 @@ def delete(self, request, application, goods_type_pk):
...
+        return JsonResponse(data={"generated_documents": generated_documents_data}, status=status.HTTP_200_OK,)
+
+
+class GeneratedDocument(APIView):
+    authentication_classes = (ExporterAuthentication,)
+
+    @authorised_users(ExporterUser)
+    def get(self, request, application, gcd_pk):
+        """
+        Gets a generated documents for the application's case
```
stephanieboyd: Typo in docstring

applications/views/applications.py
```diff
@@ -123,8 +123,11 @@ def put(self, request, application):
         else:
             application_old_name = application.name
 
-            if application.application_type == ApplicationType.STANDARD_LICENCE:
-                application_old_ref_number = application.reference_number_on_information_form
+            application_old_ref_number = (
```
stephanieboyd: Might flow better if you placed this under the check that uses it. Line 142 or something. Will only need to check if it's a standard license once

applications/tests/test_get_generated_case_documents.py
```diff
@@ -0,0 +1,109 @@
...
+            "#1", self.team, PicklistType.LETTER_PARAGRAPH, PickListStatus.ACTIVE
+        )
+        self.letter_layout = LetterLayout.objects.first()
+        self.letter_template = LetterTemplate.objects.create(name="SIEL", layout=self.letter_layout,)
+        self.letter_template.case_types.add(CaseTypeEnum.APPLICATION)
+        self.letter_template.letter_paragraphs.add(self.picklist_item)
+        self.case = self.create_standard_application_case(self.organisation)
+        self.generated_case_document = self.create_generated_case_document(self.case, template=self.letter_template)
+
+    def test_get_generated_documents_success(self):
```
stephanieboyd: Is there a test for multiple documents returned?

users/serializers.py
andytaylorkainos: Done!

applications/views/documents.py
andytaylorkainos: Done!

users/serializers.py
```diff
@@ -241,7 +236,25 @@ def get_parent_type(self, obj):
...
-        fields = ("object", "object_type", "parent", "parent_type")
+        fields = (
+            "object",
+            "object_type",
+            "parent",
+            "parent_type",
+        )
+
+
+def _get_notification_object_item(obj):
```
a-bek: I meant as a method on `class Notification(models.Model):`

That way you just do,

`notification.get_item()`

and is reusable on notification objects anywehere

applications/tests/test_get_generated_case_documents.py
andytaylorkainos: Done!

applications/views/applications.py
andytaylorkainos: done :)

applications/libraries/document_helpers.py
```diff
@@ -146,3 +147,8 @@ def delete_goods_type_document(goods_type):
         document.delete()
 
     return HttpResponse(status=status.HTTP_204_NO_CONTENT)
+
+
+def get_generated_case_document(generated_case_document):
+    documents = GeneratedCaseDocument.objects.filter(pk=generated_case_document)
```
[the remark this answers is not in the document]
andytaylorkainos: Sadly the entire helper has been set up to work with using multiple

users/serializers.py
andytaylorkainos: Think ive done it, will check with you

cases/models.py
andytaylorkainos: Yep! :)

cases/models.py
```diff
@@ -336,6 +336,22 @@ class Notification(models.Model):
     case_activity = models.ForeignKey(CaseActivity, on_delete=models.CASCADE, null=True)
     viewed_at = models.DateTimeField(null=True)
 
+    def get_notification_object_item(self):
```
a-bek: I would just call it `get_item` for simplification. 

And `get_notification_parent_item` to `get_case` because that is what you are getting.

And you can just make use of the `get_item` method in `get_case`

```
def get_case(self):
     item = self.get_item()
     return item.case if item else None
```

a-bek: Or even shorter just, `return getattr(self.get_item(), 'case', None)`

andytaylorkainos: Awesome! done

applications/views/applications.py
```diff
@@ -139,6 +136,12 @@ def put(self, request, application):
                 request.data.get("reference_number_on_information_form")
                 and application.application_type == ApplicationType.STANDARD_LICENCE
             ):
+                application_old_ref_number = (
+                    application.reference_number_on_information_form
+                    if application.application_type == ApplicationType.STANDARD_LICENCE
```
stephanieboyd: Can't you remove this check? We're only in here cause we're standard

andytaylorkainos: Nicely spotted, done

PR #225, opened 2019-12-10
test enhance: removing the av scan check when running e2e tests

phlptt:


conf/constants.py
```diff
@@ -29,3 +29,6 @@ class Roles:
     IMMUTABLE_ROLES = [INTERNAL_SUPER_USER_ROLE_ID, EXPORTER_SUPER_USER_ROLE_ID, EXPORTER_DEFAULT_ROLE_ID]
 
     EXPORTER_PRESET_ROLES = [EXPORTER_SUPER_USER_ROLE_ID, EXPORTER_DEFAULT_ROLE_ID]
+
+
+skip_av_for_end_to_end_testing = 0
```
jfaracik: This should be ALL_CAPS

static/upload_document_for_tests/views.py
```diff
@@ -5,6 +5,7 @@
 from rest_framework import status
 from rest_framework.views import APIView
 
+from conf import constants
```
stephanieboyd: This import looks to be unused now, please remove if it is

static/upload_document_for_tests/views.py
```diff
@@ -16,20 +17,22 @@ def get(self, request):
             return JsonResponse(
                 data={"errors": "This endpoint is not enabled"}, status=status.HTTP_405_METHOD_NOT_ALLOWED,
             )
+        # if "static.upload_document_for_tests" in str(request.__dict__):
```
stephanieboyd: I don't like that we're commenting this out and merging. If we want to keep it commented out, can we at have a detailed comment above it?

static/upload_document_for_tests/views.py
```diff
@@ -16,20 +17,22 @@ def get(self, request):
...
 
-        bucket_name = env("AWS_STORAGE_BUCKET_NAME")
-        s3 = boto3.client(
-            "s3", aws_access_key_id=env("AWS_ACCESS_KEY_ID"), aws_secret_access_key=env("AWS_SECRET_ACCESS_KEY"),
-        )
+        # bucket_name = env("AWS_STORAGE_BUCKET_NAME")
+        # s3 = boto3.client(
+        #     "s3", aws_access_key_id=env("AWS_ACCESS_KEY_ID"), aws_secret_access_key=env("AWS_SECRET_ACCESS_KEY"),
+        # )
         s3_key = "lite-e2e-test-file.txt"
```
andytaylorkainos: this is literally the problem we tried to solve by using this endpoint. what happens if the file gets deleted?

static/upload_document_for_tests/views.py
```diff
@@ -16,20 +17,22 @@ def get(self, request):
...
+        # file_to_upload_abs_path = os.path.abspath(
+        #     os.path.join(os.path.dirname(__file__), "resources", "lite-e2e-test-file.txt")
+        # )
+        #
+        # try:
+        #     s3.upload_file(file_to_upload_abs_path, bucket_name, s3_key)
+        # except Exception as e:  # noqa
+        #     return JsonResponse(data={"errors": str(e)}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
         return JsonResponse(data={"s3_key": s3_key}, status=status.HTTP_200_OK)
```
andytaylorkainos: this is literally the problem we tried to solve by using this endpoint. what happens if the file gets deleted?

andytaylorkainos:
This is not a solution, or a good temporary fix. this endpoint is pointless with the other code comment out, as all it will do is return an s3 key.

PR #226, opened 2019-12-11
bug: LT-1759 Template Pagination

BenAndrew-Kainos:


letter_templates/views.py
```diff
@@ -21,12 +21,18 @@ class LetterTemplatesList(generics.ListCreateAPIView):
...
-    def get_queryset(self):
+    def get(self, request, *args, **kwargs):
         case = self.request.GET.get("case")
+        queryset = get_letter_templates_for_case(get_case(pk=case)) if case else self.queryset
 
-        return get_letter_templates_for_case(get_case(pk=case)) if case else LetterTemplate.objects.all()
+        results = self.paginator.paginate_queryset(queryset, request)
+        serializer = LetterTemplateSerializer(results, many=True)
+        return self.get_paginated_response(
+            {"templates": serializer.data, "total_pages": len(self.paginator.page.paginator.page_range)}
```
jfaracik: Is there any reason why you're overriding get?

jfaracik: The built in pagination class should handle all this.

BenAndrew-Kainos: Done

BenAndrew-Kainos: Done

PR #229, opened 2019-12-13
feat: LT-1262 Good Document Sensitivity

BenAndrew-Kainos:
https://github.com/uktrade/lite-api/pull/229
https://github.com/uktrade/lite-exporter-frontend/pull/258
https://github.com/uktrade/lite-content/pull/42
https://github.com/uktrade/lite-tests-common/pull/94


static/missing_document_reasons/enums.py
```diff
@@ -0,0 +1,10 @@
+class GoodMissingDocumentReasons:
+    NO_DOCUMENT = "NO_DOCUMENT"
+    OFFICIAL_SENSITIVE = "OFFICIAL_SENSITIVE"
+    COMMERCIALLY_SENSITIVE = "COMMERCIALLY_SENSITIVE"
+
+    choices = [
+        (NO_DOCUMENT, "No document available for the good"),
```
andytaylorkainos: is "good" the correct word? (might be worth checking the strings briefly with anthony :)

BenAndrew-Kainos: Sure. About to move these to content too

applications/tests/tests_adding_goods.py
```diff
@@ -158,3 +158,33 @@ def test_adding_good_validate_only(self, data):
...
+            "value": 50000.45,
+        }
+
+        url = reverse("applications:application_goods", kwargs={"pk": self.draft.id})
+        response = self.client.post(url, data, **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json()["error"], GoodsOnApplication.DOCUMENT_ERROR)
+
+    def test_adding_good_with_reason_success(self):
```
efahy: Can you be specific in the test names that its document missing reason, not just reason please

applications/tests/tests_adding_goods.py
```diff
@@ -158,3 +158,33 @@ def test_adding_good_validate_only(self, data):
         response = self.client.post(url, data, **self.exporter_headers)
 
         self.assertEquals(response.status_code, data["response"])
+
+    def test_adding_good_without_document_or_reason_failure(self):
```
efahy: What about a test where there is both a document and a missing reason on a good?  I assume the API should reject that on the basis that it is illogical

goods/models.py
```diff
@@ -21,6 +22,9 @@ class Good(models.Model):
     organisation = models.ForeignKey(Organisation, on_delete=models.CASCADE)
     status = models.CharField(choices=GoodStatus.choices, default=GoodStatus.DRAFT, max_length=20)
     flags = models.ManyToManyField(Flag, related_name="goods")
+    missing_document_reason = models.CharField(
+        choices=GoodMissingDocumentReasons.choices, blank=True, null=True, max_length=20
```
efahy: I'm not sure if we should allow blank.  I would think null yes; blank no

static/missing_document_reasons/enums.py
```diff
@@ -0,0 +1,13 @@
+from lite_content.lite_api.static import GoodMissingDocumentReasonsOptions
+
+
+class GoodMissingDocumentReasons:
+    NO_DOCUMENT = "NO_DOCUMENT"
+    OFFICIAL_SENSITIVE = "OFFICIAL_SENSITIVE"
+    COMMERCIALLY_SENSITIVE = "COMMERCIALLY_SENSITIVE"
```
efahy: I wonder if you could give a bit of context to these values with some comments, while we've got the story in mind, i.e. given what the scenarios are telling us, can we pass on some of that detail, that will otherwise be forgotten, in comments?

goods/models.py
BenAndrew-Kainos: Done :)

applications/tests/tests_adding_goods.py
BenAndrew-Kainos: The same result is expected for all reasons but I've renamed to specify which reason I'm testing now anyway

applications/tests/tests_adding_goods.py
BenAndrew-Kainos: I've updated the code to remove the `missing_document_reason` if a document is uploaded (thus removing the possibility of both existing) and a unit test for this :)

static/missing_document_reasons/enums.py
BenAndrew-Kainos: Added a docstring to explain

PR #232, opened 2019-12-13
feat: LT-1186 - Destination Flags

cstobart:


flags/views.py
```diff
@@ -211,5 +226,44 @@ def _set_case_activity_for_goods(self, added_flags, removed_flags, case, user, n
...
+            Q(consignee=party) | Q(end_user=party) | Q(ultimate_end_users__in=[party]) | Q(third_parties__in=[party])
+        )
+        if not qs:
+            qs = HmrcQuery.objects.filter(
+                Q(consignee=party)
+                | Q(end_user=party)
+                | Q(ultimate_end_users__in=[party])
+                | Q(third_parties__in=[party])
             )
+        if qs:
```
andytaylorkainos: is this if needed?

flags/helpers.py
```diff
@@ -15,3 +17,9 @@ def get_object_of_level(level, pk):
         return get_case(pk)
     elif level == "organisation":
         return get_organisation_by_pk(pk)
+    elif level == "destination":
+        try:
+            destination = Country.objects.get(pk=pk)
+        except Country.DoesNotExist:
+            destination = Party.objects.get(pk=pk)
```
andytaylorkainos: can we try-except this also please

cases/urls.py
```diff
@@ -7,6 +7,7 @@
 
 urlpatterns = [
     path("", search_views.CasesSearchView.as_view(), name="search"),
+    path("destination/<uuid:pk>", views.Destination.as_view(), name="destination"),
```
andytaylorkainos: change to `destinations/<uuid:pk>/` please

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+        destination = Country.objects.get(pk=pk)
+    except Country.DoesNotExist:
+        destination = Party.objects.get(pk=pk)
+    return destination
+
+
+def get_standard_application_destination_flags(application):
+    flags = []
+    [flags.append(flag) for flag in application.end_user.flags.order_by("name")]
+    [flags.append(flag) for flag in application.consignee.flags.order_by("name")]
```
andytaylorkainos: Can we make this a little more efficient? 
```
end_user_flags = application.end_user.flags.order_by("name")
consignee_flags = application.consignee.flags.order_by("name")

ultimate_end_users = [
        ultimate_end_user_id for ultimate_end_user_id in application.ultimate_end_users.values_list("id", flat=True)
]
ultimate_end_user_flags = [
    get_destination(str(ultimate_end_user)).flags.order_by("name")) 
    for ultimate_end_user in ultimate_end_users
]

third_parties = [third_party_id for third_party_id in application.third_parties.values_list("id", flat=True)]
third_party_flags = [get_destination(third_party).flags.order_by("name") for third_party in third_parties]

return end_user_flags + consignee_flags + ultimate_end_user_flags + third_party_flags
```

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    return flags
+
+
+def get_destination_flags(instance):
+    application = get_application(instance.id)
+    countries = CountryOnApplication.objects.filter(application=instance).select_related("country")
+    flags = list(itertools.chain.from_iterable([c.country.flags.order_by("name") for c in countries]))
+    if isinstance(application, StandardApplication):
+        flags += get_standard_application_destination_flags(application)
+    deduplicated_flags = list(set(flags))
```
andytaylorkainos: distinct_flags

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+        goods_flags = []
+        destination_flags = []
+
+    flag_data = (
+        FlagSerializer(case_flags, many=True).data
+        + FlagSerializer(org_flags, many=True).data
+        + FlagSerializer(goods_flags, many=True).data
+        + FlagSerializer(destination_flags, many=True).data
+    )
+    flag_data = sorted(flag_data, key=lambda x: x["name"])
```
andytaylorkainos: Remove the ordering from `get_destination_flags` if you're going to order the entire list here

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+
+    if not team:
+        return flag_data
+
+    # Sort flags by user's team.
```
andytaylorkainos: can you change the comment to: `Group flags by user's team`

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    try:
+        destination = Country.objects.get(pk=pk)
+    except Country.DoesNotExist:
+        destination = Party.objects.get(pk=pk)
+    return destination
+
+
+def get_standard_application_destination_flags(application):
+    flags = []
+    [flags.append(flag) for flag in application.end_user.flags.order_by("name")]
```
stephanieboyd: We're generating a lot of lists of `[None, None, None]` here. List comp is only useful when you use the generated list. One suggestion would be to generate the lists in comps, and concat them.

flags/helpers.py
```diff
@@ -15,3 +17,9 @@ def get_object_of_level(level, pk):
         return get_case(pk)
     elif level == "organisation":
         return get_organisation_by_pk(pk)
+    elif level == "destination":
+        try:
```
stephanieboyd: Use the helper function you created

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+from applications.libraries.get_applications import get_application
+from applications.models import GoodOnApplication, CountryOnApplication, StandardApplication
+from cases.enums import CaseTypeEnum
+from flags.serializers import FlagSerializer
+from parties.models import Party
+from static.countries.models import Country
+
+
+def get_destination(pk):
+    try:
```
stephanieboyd: We know this has to be one or the other. I'd prefer any exception raised to be when we have neither

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    for third_party in third_parties:
+        [flags.append(flag) for flag in get_destination(third_party).flags.order_by("name")]
+
+    return flags
+
+
+def get_destination_flags(instance):
+    application = get_application(instance.id)
+    countries = CountryOnApplication.objects.filter(application=instance).select_related("country")
+    flags = list(itertools.chain.from_iterable([c.country.flags.order_by("name") for c in countries]))
```
stephanieboyd: Do we need to use itertools? We've only one iterable

efahy: Don't think you need to use `itertools.chain.from_iterable` here, given that you've only got one iterable in the first place.  The comprehension on its own should work fine.  There is one other usage in here and when you ditch it as well you can ditch the itertools import

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    return deduplicated_flags
+
+
+def get_ordered_flags(instance, team):
+    case_flags = instance.flags.order_by("name")
+    org_flags = instance.organisation.flags.order_by("name")
+
+    if instance.type in [CaseTypeEnum.APPLICATION, CaseTypeEnum.HMRC_QUERY]:
+        goods = GoodOnApplication.objects.filter(application=instance).select_related("good")
+        goods_flags = list(itertools.chain.from_iterable([g.good.flags.order_by("name") for g in goods]))
```
stephanieboyd: Not sure about itertools here either, or the casting?

cases/views/views.py
```diff
@@ -423,3 +425,11 @@ def post(self, request, pk):
             return JsonResponse(data={"data": data})
 
         return JsonResponse(data={"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+
+class Destination(APIView):
+    def get(self, request, pk):
+        destination = get_destination(pk)
```
stephanieboyd: Could we just: `serializer = PartyWithFlagsSerializer(get_destination(pk))`cause we don't use the destination again

cases/libraries/get_destination.py
efahy: Do you need to turn the `set` back into a `list`?  Both are iterable right?  Could maybe save a CPU cycle or two?

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    for ultimate_end_user in ultimate_end_users:
+        [flags.append(flag) for flag in get_destination(str(ultimate_end_user)).flags.order_by("name")]
+    third_parties = [third_party_id for third_party_id in application.third_parties.values_list("id", flat=True)]
+    for third_party in third_parties:
+        [flags.append(flag) for flag in get_destination(third_party).flags.order_by("name")]
+
+    return flags
+
+
+def get_destination_flags(instance):
```
stephanieboyd: instance? can we call this something better

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    application = get_application(instance.id)
+    countries = CountryOnApplication.objects.filter(application=instance).select_related("country")
```
stephanieboyd: Nice, this got me to read up on `select_related` :)

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    flags = list(itertools.chain.from_iterable([c.country.flags.order_by("name") for c in countries]))
+    if isinstance(application, StandardApplication):
+        flags += get_standard_application_destination_flags(application)
+    deduplicated_flags = list(set(flags))
+
+    return deduplicated_flags
+
+
+def get_ordered_flags(instance, team):
```
efahy: Could you consider specifying types for these parameters?

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+        [flags.append(flag) for flag in get_destination(third_party).flags.order_by("name")]
+
+    return flags
+
+
+def get_destination_flags(instance):
+    application = get_application(instance.id)
+    countries = CountryOnApplication.objects.filter(application=instance).select_related("country")
+    flags = list(itertools.chain.from_iterable([c.country.flags.order_by("name") for c in countries]))
+    if isinstance(application, StandardApplication):
```
stephanieboyd: Do we need to be concerned about people on end user advisories?

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+        flags += get_standard_application_destination_flags(application)
+    deduplicated_flags = list(set(flags))
+
+    return deduplicated_flags
+
+
+def get_ordered_flags(instance, team):
+    case_flags = instance.flags.order_by("name")
```
stephanieboyd: Gimme a constant, `NAME_FIELD = "name"`. And any other heavily repeated ones, please

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+    org_flags = instance.organisation.flags.order_by("name")
+
+    if instance.type in [CaseTypeEnum.APPLICATION, CaseTypeEnum.HMRC_QUERY]:
+        goods = GoodOnApplication.objects.filter(application=instance).select_related("good")
+        goods_flags = list(itertools.chain.from_iterable([g.good.flags.order_by("name") for g in goods]))
+        good_ids = {}
+        goods_flags = [good_ids.setdefault(g, g) for g in goods_flags if g.id not in good_ids]  # dedup
+        destination_flags = get_destination_flags(instance)
+    else:
+        goods_flags = []
```
stephanieboyd: Instead of declaring these as empty here - can you declare them under case and org flags, and remove the else

audit_trail/payload.py
```diff
@@ -10,6 +10,11 @@ class AuditType(Enum):
     GOOD_ADD_FLAGS = "added flags: {added_flags} to good: {good_name}"
     GOOD_REMOVE_FLAGS = "removed flags: {removed_flags} from good: {good_name}"
     GOOD_ADD_REMOVE_FLAGS = "added flags: {added_flags}, and removed: {removed_flags} from good: {good_name}"
+    DESTINATION_ADD_FLAGS = "added flags: {added_flags} to destination: {destination_name}"
+    DESTINATION_REMOVE_FLAGS = "removed flags: {removed_flags} from destination: {destination_name}"
+    DESTINATION_ADD_REMOVE_FLAGS = (
```
stephanieboyd: I don't see this being used - should it exist or should there be logic in the `_set_case_activity_for_destinations` function?

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+
+    return deduplicated_flags
+
+
+def get_ordered_flags(instance, team):
+    case_flags = instance.flags.order_by("name")
+    org_flags = instance.organisation.flags.order_by("name")
+
+    if instance.type in [CaseTypeEnum.APPLICATION, CaseTypeEnum.HMRC_QUERY]:
+        goods = GoodOnApplication.objects.filter(application=instance).select_related("good")
```
stephanieboyd: Does this work for open applications too?

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,77 @@
...
+        goods_flags = list(itertools.chain.from_iterable([g.good.flags.order_by("name") for g in goods]))
+        good_ids = {}
```
stephanieboyd: Can we remove `good_ids`? It looks to only be used for the deduplication - but we should be able to just `set(goods_flags)`

cases/libraries/get_destination.py
cstobart: is this more efficient?

cases/libraries/get_destination.py
cstobart: If you don't order them, they become uniterable, I will try and find a work around which isn't list processing within a list processing

cases/libraries/get_destination.py
cstobart: done this 👍

cases/libraries/get_destination.py
cstobart: done 👍

cases/libraries/get_destination.py
cstobart: done 👍

cases/libraries/get_destination.py
cstobart: no idea, not my code, just moved it out of the serialiser into a function

cases/libraries/get_destination.py
cstobart: Haven't touched this stuff, assume it was there for a reason

cases/libraries/get_destination.py
cstobart: This approach does not work as you are left with query_sets, the list operations with an order_by are the easiest (cleanest) way to get the data into iterable form

cases/libraries/get_destination.py
cstobart: Andy raised a similar concern, can't really see a more elegant way around it. Do you want to have a call about it tomorrow?

flags/helpers.py
cstobart: done 👍

cases/libraries/get_destination.py
cstobart: tried an approach to make it more informative

cases/libraries/get_destination.py
andytaylorkainos: surely applying `.values()` will make it iterable and avoid ordering them (where ordering happens again a few lines below in the lambda)

cstobart: I'll give it a go

cases/libraries/get_destination.py
cstobart: In order to replace that one line we need to do this ```        goods_flags = []
        for good in goods:
            [goods_flags.append(flag) for flag in good.good.flags.order_by("name")]```

cases/libraries/get_destination.py
cstobart: good point

cases/libraries/get_destination.py
cstobart: The alternative I wrote in reply to one of Eugene's comments

flags/views.py
cstobart: 👍

audit_trail/payload.py
a-bek: DESTINATION_ADD_REMOVE_FLAGS is redundant now, removed in the audit refactor.

The old activity logic used this but it duplicated entries from ADD_FLAGS and REMOVE_FLAGS

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,81 @@
...
+    for third_party in application.third_parties.values_list("id", flat=True):
+        flags += get_destination(third_party).flags.all()
+
+    return flags
+
+
+def get_destination_flags(instance):
+    flags = []
+
+    countries = CountryOnApplication.objects.filter(application=instance).select_related("country")
```
a-bek: Can variables be named as closely to the model? It's slightly misleading because `countries` implies a `Country` queryset, not `CountryOnApplication`

cases/libraries/get_destination.py
```diff
@@ -0,0 +1,81 @@
...
+        flags += country.country.flags.all()
+
+    application = get_application(instance.id)
+    if isinstance(application, StandardApplication):
+        flags += get_standard_application_destination_flags(application)
+
+    return flags
+
+
+def get_ordered_flags(instance: Case, team: Team):
```
a-bek: Rename `instance:case`

cases/libraries/get_destination.py
cstobart: Done

cases/libraries/get_destination.py
cstobart: done

cases/libraries/get_destination.py
cstobart: no longer required

cases/libraries/get_destination.py
cstobart: no longer required

cases/urls.py
```diff
@@ -7,6 +7,7 @@
 
 urlpatterns = [
     path("", search_views.CasesSearchView.as_view(), name="search"),
+    path("destinations/<str:pk>", views.Destination.as_view(), name="destination"),
```
andytaylorkainos: needs to be a trailing `/` at the end of this url

cases/views/views.py
```diff
@@ -423,3 +427,15 @@ def post(self, request, pk):
...
+
+        if isinstance(destination, Country):
+            serializer = CountryWithFlagsSerializer(destination)
+        else:
+            serializer = PartyWithFlagsSerializer(destination)
+
+        return JsonResponse(data={"destination": serializer.data})
```
andytaylorkainos: please return the status code aswell, i know it defaults to 200 but we're trying to be explicit in code

PR #234, opened 2019-12-16
feat: LT-1450 - Paginate and filter goods

Mark-Spiers-kainos:


goods/views.py
```diff
@@ -100,27 +102,46 @@ def post(self, request, case_pk):
...
-class GoodList(APIView):
+class GoodList(ListCreateAPIView):
+    model = Good
     authentication_classes = (ExporterAuthentication,)
-
-    def get(self, request):
+    serializer_class = GoodListSerializer
+    description = None
+    part_number = None
+    control_rating = None
```
jfaracik: Why is this not inside of get_queryset?

Mark-Spiers-kainos: get_queryset() doesn't have access to the request data, so I had to set those in the get() and then use them inside get_queryset().

a-bek: I don't think this is the right way to solve your problem. Why don't you just override the `list()` function and do the filtering in there? 

You want minimal state in your views besides what is intended by django.

Mark-Spiers-kainos: The list function is where the queryset is collected, then paginated by the class. I wouldn't want to move the override to this function when the get function just calls it currently as well.

jfaracik: get_queryset can access request data - `submitted = optional_str_to_bool(self.request.GET.get("submitted"))`

Mark-Spiers-kainos: Oh I must of missed that, perfect will do.

PR #235, opened 2019-12-17
task: API strings refactor

BenAndrew-Kainos:


applications/serializers/good.py
```diff
@@ -45,18 +46,22 @@ class GoodOnApplicationCreateSerializer(serializers.ModelSerializer):
     good = PrimaryKeyRelatedField(queryset=Good.objects.all())
     application = PrimaryKeyRelatedField(queryset=StandardApplication.objects.all())
     quantity = DecimalField(
-        max_digits=256, decimal_places=6, error_messages={"invalid": strings.Goods.ErrorMessages.INVALID_QTY},
+        max_digits=256,
+        decimal_places=6,
+        error_messages={"invalid": lite_content.lite_api.goods.Goods.ErrorMessages.INVALID_QTY},
```
efahy: Can you change it up a bit so that we don't have to qualify every constant usage with `lite_content.lite_api` and the module name, `good` in this case.  Makes for annoyingly long lines

conf/decorators.py
```diff
@@ -98,7 +98,7 @@ def inner(request, *args, **kwargs):
 
             if application.status.status in get_case_statuses(read_only=True):
                 return JsonResponse(
-                    data={"errors": [strings.System.READ_ONLY_CASE_CANNOT_PERFORM_OPERATION_ERROR]},
+                    data={"errors": [lite_content.lite_api.cases.System.READ_ONLY_CASE_CANNOT_PERFORM_OPERATION_ERROR]},
```
efahy: Why is `System` in `cases.py`?  Can you either move it or rename it or something?

applications/tests/tests_standard_application_submit.py
```diff
@@ -164,7 +173,7 @@ def test_status_code_post_with_infected_document_failure(self):
 
         self.assertContains(
             response,
-            text=strings.Applications.Standard.END_USER_DOCUMENT_INFECTED,
+            text=lite_content.lite_api.applications.Standard.END_USER_DOCUMENT_INFECTED,
```
efahy: Can you make a root `Application` class in `applications.py` please, to be consistent with other const modules, e.g. `goods.py`?  If you manage to adjust the imports and shorten the lines per my other comment, it would be odd for this line to just contain `Standard.END_USER_DOCUMENT_INFECTED`

applications/serializers/good.py
BenAndrew-Kainos: Fixed all the imports

conf/decorators.py
BenAndrew-Kainos: Renamed :)

applications/tests/tests_standard_application_submit.py
BenAndrew-Kainos: Excellent point but I'm going to leave this for another day if that's OK

PR #239, opened 2019-12-18
task: API string cleanup

BenAndrew-Kainos:


goods/serializers.py
```diff
@@ -73,7 +68,7 @@ class GoodSerializer(serializers.ModelSerializer):
         choices=GoodMissingDocumentReasons.choices,
         allow_blank=True,
         required=False,
-        error_messages={"invalid_choice": goods.Good.INVALID_MISSING_DOCUMENT_REASON},
+        error_messages={"invalid_choice": Goods.INVALID_MISSING_DOCUMENT_REASON},
```
efahy: Following discussion in the room, we think its more intuitive if the usage in the code is `strings.ClassName.CONST_NAME`

BenAndrew-Kainos: Done all of them

PR #246, opened 2019-12-30
ci: Workflows

jfaracik:
This pull request introduces workflows for our CircleCI - this means that our tests and linting can run in parallel and they can also report separate statuses. No more having to dig in to CircleCI to see why your branch is red :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
   jq: circleci/jq@1.8.0
 
 jobs:
-  build:
+  tests:
     docker:
       - image: circleci/python:3.7
       - image: circleci/postgres:11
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -87,3 +79,42 @@ jobs:
...
+            - ./venv
+          key: dependencies-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+
+      - run:
+          name: Prospector
+          command: pipenv run prospector
+
+      - run:
+          name: Black
+          command: pipenv run black . --check --diff
```
efahy: Do you need to give this the `-l 120` argument?

jfaracik: Nope, it's in pyproject.toml

PR #249, opened 2020-01-06
hotfix: Flags on Cases List Page

andytaylorkainos:


cases/libraries/get_destination.py
```diff
@@ -41,11 +43,13 @@ def get_destination_flags(case):
 
     countries_on_application = CountryOnApplication.objects.filter(application=case).select_related("country")
     for country_on_application in countries_on_application:
-        flags += country_on_application.country.flags.all()
+        if country_on_application.country:
```
cstobart: this one isn't needed

PR #253, opened 2020-01-07
Timestampable, migrations and reversion deprecation

a-bek:


efahy:
Hi Asmund, I really like this PR - I love standardising of things! 

There are still a lot of models that are not time stamp-able.  I went through them and here is a short list of the ones I think would be the most worthwhile to make timestamp-able:

BaseUser/AbstractUser
Party
GoodsType
Site
ExternalLocation
SiteManager

picklists/views.py
```diff
@@ -41,7 +41,7 @@ def get(self, request):
             query.append(Q(id__in=ids))
 
         picklist_items = PicklistItem.objects.filter(reduce(operator.and_, query))
-        picklist_items = picklist_items.order_by("-last_modified_at")
+        picklist_items = picklist_items.order_by("-updated_at")
```
efahy: If updated_at and creted_at are mandatory, you'll need to do some manual migration stuff to avoid needing to clear the DB.  If you need to clear the DB you'll need to get pre-sign of to do so from Mark H & Anne

a-bek: As discussed, the test data with the old timestamps will lose that particular information and have the new timestamp fields with the date set to the migration time. Happy to proceed in all development environments.

PR #254, opened 2020-01-07
feat: LT-1239 - Incorporated application

jfaracik:
* Rename is_good_end_product to is_good_incorporated
* Removed is_good_end_product from good, moved to GoodOnApplication
* Made is_good_incorporated on goods_type optional depending on type of app, required on Open licences, not used on HMRC queries


applications/models.py
```diff
@@ -99,6 +99,7 @@ class GoodOnApplication(models.Model):
     quantity = models.FloatField(null=True, blank=True, default=None)
     unit = models.CharField(choices=Units.choices, default=Units.GRM, max_length=50)
     value = models.DecimalField(max_digits=256, decimal_places=2)
+    is_good_incorporated = models.BooleanField(default=False)
```
Mark-Spiers-kainos: I don't think this field should contain a default

goodstype/serializers.py
```diff
@@ -15,30 +17,38 @@ class GoodsTypeSerializer(serializers.ModelSerializer):
...
-        )
+        ]
 
     def __init__(self, *args, **kwargs):
         """
         Initializes serializer for Goods Type
         """
         super(GoodsTypeSerializer, self).__init__(*args, **kwargs)
 
+        # Only add is_good_incorporated if application is of type OPEN_LICENCE
```
efahy: Please append "and not if its a HMRC query"

goodstype/serializers.py
```diff
@@ -15,30 +17,38 @@ class GoodsTypeSerializer(serializers.ModelSerializer):
...
+        application = self.get_initial().get("application")
+        if application:
```
Mark-Spiers-kainos: Is there situations where we contain a GoodsType without an application?

Mark-Spiers-kainos: in what situations is application not given?

goodstype/serializers.py
```diff
@@ -15,30 +17,38 @@ class GoodsTypeSerializer(serializers.ModelSerializer):
...
     is_good_controlled = serializers.BooleanField()
-    is_good_end_product = serializers.BooleanField()
     application = serializers.PrimaryKeyRelatedField(queryset=BaseApplication.objects.all())
     countries = serializers.SerializerMethodField()
     document = serializers.SerializerMethodField()
 
     class Meta:
         model = GoodsType
-        fields = (
+        fields = [
```
Mark-Spiers-kainos: why is this changed from parentheses, when we do that everywhere else?

applications/models.py
jfaracik: It needs a default as its a migration

goodstype/serializers.py
jfaracik: It fails on the country matrix without it

applications/tests/tests_goods_type_on_application.py
```diff
@@ -69,7 +87,7 @@ def test_create_goodstype_on_standard_application_as_exporter_user_failure(self)
             "description": "Widget",
             "is_good_controlled": True,
             "control_code": "ML1a",
-            "is_good_end_product": True,
+            "is_good_incorporated": True,
```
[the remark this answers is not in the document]
jfaracik: is_good_incorporated is a field for GoodsType.

applications/tests/tests_view.py
```diff
@@ -151,6 +151,7 @@ def test_view_draft_open_application_as_exporter_success(self):
         self.assertIsNotNone(retrieved_application["updated_at"])
         self.assertIsNone(retrieved_application["submitted_at"])
         self.assertEqual(retrieved_application["status"]["key"], CaseStatusEnum.DRAFT)
+        self.assertIn("is_good_incorporated", str(retrieved_application))
```
andytaylorkainos: why did you cast the dictionary to a string to assert this?

applications/serializers/good.py
```diff
@@ -31,33 +31,22 @@ class GoodOnApplicationViewSerializer(serializers.ModelSerializer):
...
         model = GoodOnApplication
-        fields = (
-            "id",
-            "good",
-            "application",
-            "quantity",
-            "unit",
-            "value",
-        )
+        fields = ("id", "good", "application", "value", "quantity", "unit", "is_good_incorporated")
```
andytaylorkainos: Same as above

applications/serializers/good.py
```diff
@@ -31,33 +31,22 @@ class GoodOnApplicationViewSerializer(serializers.ModelSerializer):
...
         model = GoodOnApplication
-        fields = (
-            "id",
-            "good",
-            "application",
-            "quantity",
-            "unit",
-            "value",
-        )
+        fields = ("id", "good", "application", "quantity", "unit", "value", "is_good_incorporated")
```
andytaylorkainos: The reason we leave the tuple exploded across multiple lines is to remain consistent with black formatting.

If the tuple exceeds 120 line length, it gets exploded to be multi-line.
So its a bit annoying when some are exploded and others aren't.

Add the comma back at the end of the tuple for consistency, please.

applications/tests/tests_view.py
jfaracik: Solely because I didn't want to have to dive into the dictionary to find the string. Less code this way to ensure that the key is in the dict.

andytaylorkainos: It might be less code, but its less performant and since we know the key we should specify where the key should exist in the dictionary (within another object inside `retrieved_application`)

applications/tests/tests_adding_goods.py
```diff
@@ -199,6 +202,7 @@ def test_adding_good_with_reason_official_sensitive_success(self):
             "quantity": 1200.098896,
             "unit": Units.NAR,
             "value": 50000.45,
+            "is_good_incorporated": True,
```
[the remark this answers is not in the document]
jfaracik: Done

applications/tests/tests_view.py
jfaracik: Are we worried about a test being a microsecond slower? This way is far more readable and I think that's more beneficial than that.

andytaylorkainos: Not really worried about the time; more so that we specify where the data should exist in the response object (if it moved for example, this test would fail and stop a bug appearing on the frontend)

jfaracik: That's already handled in the view goods_type details. But anyways, I'll change it.

applications/serializers/good.py
```diff
@@ -38,26 +38,31 @@ class Meta:
...
+            "is_good_incorporated",
         )
 
 
 class GoodOnApplicationCreateSerializer(serializers.ModelSerializer):
     good = PrimaryKeyRelatedField(queryset=Good.objects.all())
     application = PrimaryKeyRelatedField(queryset=StandardApplication.objects.all())
-    quantity = DecimalField(max_digits=256, decimal_places=6, error_messages={"invalid": strings.Goods.INVALID_QTY},)
-    value = (DecimalField(max_digits=256, decimal_places=2, error_messages={"invalid": strings.Goods.INVALID_VALUE},),)
+    value = DecimalField(max_digits=256, decimal_places=2, error_messages={"invalid": strings.Goods.INVALID_VALUE})
```
Mark-Spiers-kainos: I think we should perhaps look at capping this at a lower amount, since I think this is more money than in the world. (although I know it's not your story)

PR #257, opened 2020-01-09
feat: LT-955 - User Assigned Cases Queue

andytaylorkainos:


queues/helpers.py
```diff
@@ -68,7 +90,9 @@ def get_queues(include_system_queues=False, user: GovUser = None):
         queues.append(_all_cases_queue())
         queues.append(_open_cases_queue())
         if user:
```
stephanieboyd: I read the whole function - is a user, with a default of None valid? The optional return, seems to be, according to the docstring, only system queues. When would this function get called without a user?

queues/tests/tests_system_queues.py
```diff
@@ -5,10 +5,12 @@
...
 from cases.models import CaseAssignment
 from queues.constants import (
-    ALL_CASES_SYSTEM_QUEUE_ID,
-    OPEN_CASES_SYSTEM_QUEUE_ID,
+    ALL_CASES_QUEUE_ID,
+    OPEN_CASES_QUEUE_ID,
     MY_TEAMS_QUEUES_CASES_ID,
     UPDATED_CASES_QUEUE_ID,
+    MY_ASSIGNED_CASES_QUEUE_ID,
+    MY_CASE_OFFICER_CASES_QUEUE_ID,
```
stephanieboyd: ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_ID?

Long, I know - but currently, it reads a little like the cases of their case officer? Which isn't a thing, but I think the name could be clearer

queues/helpers.py
andytaylorkainos: This function is called by `get_queue` on line 101 in the same file.

That function is sometimes used without a user parameter

queues/tests/tests_system_queues.py
andytaylorkainos: Not worried about the length, so im happy to change the name :)

queues/tests/tests_system_queues.py
```diff
@@ -17,22 +19,52 @@
...
+    case_assignment.users.set([self.gov_user])
+    return case_assigned_to_user
 
-        self.case = self.create_clc_query("Query", self.organisation)
-        self.case_2 = self.create_clc_query("Query", self.organisation)
-        self.case_3 = self.create_clc_query("Query", self.organisation)
 
-        self.case_3.query.status = get_case_status_by_status(CaseStatusEnum.FINALISED)
-        self.case_3.query.save()
+def _create_case_user_is_assigned_as_case_officer_to(self):
```
stephanieboyd: `create_case_where_user_is_case_officer`

stephanieboyd: that went through some edits, hopefully you didn't see them all hah

queues/tests/tests_system_queues.py
```diff
@@ -17,22 +19,52 @@
...
+
+    self.case = self.create_clc_query("Query", self.organisation)
+    self.case_2 = self.create_clc_query("Query", self.organisation)
+    self.case_3 = self.create_clc_query("Query", self.organisation)
+
+    self.case_3.query.status = get_case_status_by_status(CaseStatusEnum.FINALISED)
+    self.case_3.query.save()
+
+
+def _create_case_user_is_assigned_to(self):
```
stephanieboyd: `create_case_and_assign_user`

queues/tests/tests_system_queues.py
```diff
@@ -17,22 +19,52 @@
...
-        self.queue1 = self.create_queue("queue1", self.team)
-        self.queue2 = self.create_queue("queue2", self.team)
+def _setup_common(self):
+    self.queue1 = self.create_queue("queue1", self.team)
+    self.queue2 = self.create_queue("queue2", self.team)
+
+    self.case = self.create_clc_query("Query", self.organisation)
+    self.case_2 = self.create_clc_query("Query", self.organisation)
+    self.case_3 = self.create_clc_query("Query", self.organisation)
+
```
stephanieboyd: Remove newline

queues/tests/tests_system_queues.py
andytaylorkainos: Done!

queues/tests/tests_system_queues.py
andytaylorkainos: Done!

queues/tests/tests_system_queues.py
andytaylorkainos: Done!

queues/tests/tests_system_queues.py
andytaylorkainos: Done!

cases/managers.py
```diff
@@ -48,11 +48,15 @@ def is_updated(self, user):
 
     def assigned_to_user(self, user):
         assigned_to_user_case_ids = get_assigned_to_user_case_ids(user)
-        return self.filter(id__in=assigned_to_user_case_ids)
+        return self.filter(id__in=assigned_to_user_case_ids).exclude(
+            status__status__in=[CaseStatusEnum.WITHDRAWN, CaseStatusEnum.FINALISED]
```
stephanieboyd: As discussed, let's exclude all terminal cases

andytaylorkainos: 😎 Done

PR #260, opened 2020-01-13
bug: LT-1952 Update Third Party design for sub type & role

BenAndrew-Kainos:


parties/serializers.py
```diff
@@ -145,21 +131,11 @@ class Meta:
 
 
 class ThirdPartySerializer(PartySerializer):
-    sub_type = KeyValueChoiceField(choices=ThirdPartySubType.choices)
+    role = KeyValueChoiceField(choices=ThirdPartyRole.choices)
```
jfaracik: Could you write custom errors for if the data isn't set

BenAndrew-Kainos: For sure.

parties/migrations/0004_auto_20200113_1129.py
```diff
@@ -0,0 +1,45 @@
+# Generated by Django 2.2.8 on 2020-01-13 11:29
+
+from django.db import migrations, models
```
stephanieboyd: Combine migrations

BenAndrew-Kainos: Done

PR #265, opened 2020-01-14
feat: LT-1200 - Case reference codes

jfaracik:
* All cases will now have a unique, memorable reference code such as _P/GBOIE/2020/000012_
* Added a migration to add the reference code to existing cases
* Renamed some tests to be in the correct format


cases/libraries/reference_code.py
```diff
@@ -0,0 +1,68 @@
...
+        reference_code += "GQY" + SLASH
+
+    if case.type == CaseTypeEnum.END_USER_ADVISORY_QUERY:
+        reference_code += "EUA" + SLASH
+
+    if case.type == CaseTypeEnum.HMRC_QUERY:
+        reference_code += "CRE" + SLASH
+
+    # Year
+    reference_code += str(datetime.now().year) + SLASH
```
cstobart: UTC this please

cases/models.py
```diff
@@ -55,6 +64,14 @@ def get_case(self):
         return Case.objects.get(id=self.id)
 
 
+class CaseReferenceCode(models.Model):
```
cstobart: This feels like a large over-engineering. We could lookup the last case created_at, check if the years are the same, if so increment the reference by 1 else start from 1.

cstobart: That or put these fields on the case model itself as they are intrinsic and that way you can lose the year field as it can be derived from the submitted at date time

cases/migrations/0007_case_reference_code.py
```diff
@@ -0,0 +1,28 @@
...
+from django.db import migrations, models
+
+from cases.libraries.reference_code import generate_reference_code
+
+
+class Migration(migrations.Migration):
+    def initialize(apps, schema_editor):
+        # All cases need to have a reference code
+        Case = apps.get_model("cases", "Case")
+        for case in Case.objects.all():
```
cstobart: order_by created_at first

cases/libraries/reference_code.py
```diff
@@ -0,0 +1,74 @@
...
+
+    For clearance application cases: F680
+    Exhibition clearance: EXHC
+
+    End user advisory: EUA
+    Goods queries: GQY
+    HMRC / Border Force customs enquiry: CRE
+    For compliance site cases: COMP
+    For compliance visit cases: CVIS
+    """
```
efahy: Now that's my kind of comment!

cases/libraries/reference_code.py
```diff
@@ -0,0 +1,74 @@
...
+        # General or individual
+        reference_code += "I"
+
+        # Export, transhipment and trade control
+        if Site.objects.filter(sites_on_application__application=case):
+            reference_code += "E" + SLASH
+        if ExternalLocation.objects.filter(external_locations_on_application__application=case):
+            reference_code += "C" + SLASH
+
+    if case.type == CaseTypeEnum.CLC_QUERY:
```
efahy: I would personally make these else ifs.  I think it makes the code more easily digestible

cases/managers.py
```diff
@@ -170,3 +171,15 @@ def get_obj(self, case):
...
+
+
+class CaseReferenceCodeManager(models.Manager):
+    def create(self):
+        CaseReferenceCode = self.model
+
+        year = datetime.now().year
+        reference_number = CaseReferenceCode.objects.filter(year=year).count() + 1
+
+        case_reference_code = CaseReferenceCode(reference_number=reference_number, year=year)
```
efahy: This is not what I'd envisaged for the CaseReferenceCodes DB table.  I expected there to only be one row for each year, and on the one row, the number column contains the last case number used and is incremented each time a new case is created.  Having this table contain as many rows as there are cases is not the plan at all

cases/tests/tests_reference_code.py
```diff
@@ -0,0 +1,25 @@
+from datetime import datetime
+
+from test_helpers.clients import DataTestClient
+
+
+class ReferenceCode(DataTestClient):
```
efahy: Please include API tests for all permutations, temp/perm, export/trans shipment/trade control, etc etc

Please also include a test that creates at least two cases and checks that its correctly incrementing numbers

cases/libraries/reference_code.py
```diff
@@ -0,0 +1,69 @@
...
+    elif case.type == CaseTypeEnum.END_USER_ADVISORY_QUERY:
+        reference_code += "EUA" + SLASH
+    elif case.type == CaseTypeEnum.HMRC_QUERY:
+        reference_code += "CRE" + SLASH
+
+    # Year
+    reference_code += str(datetime.now().year) + SLASH
+
+    # Int
+    from cases.models import CaseReferenceCode
```
Mark-Spiers-kainos: I see you have 2 imports inside the function are you able to move these out?

cases/migrations/0007_case_reference_code.py
Mark-Spiers-kainos: ^

PR #267, opened 2020-01-15
feat: LT-1586 - Licence duration view and permissions

a-bek:
1) Introduce new permission MANAGE_LICENCE_DURATION
2) Add duration field to BaseApplication
3) Create a finalise view, and disallow status to be set to finalised in ManageStatusView
4) Default durations logic (not in db)
5) Added is_eu flag to lite_content countries.csv as a datapoint to be stored in Country model


applications/libraries/licence.py
```diff
@@ -0,0 +1,26 @@
...
+
+def get_default_duration(application):
+    """
+    Returns default duration in months
+
+    Rules defined in: https://uktrade.atlassian.net/browse/LT-1586
+    """
+
+    if application.export_type == ApplicationExportType.TEMPORARY:
+        return 1 * 12
```
BenAndrew-Kainos: Could you define these as constants at the top

applications/models.py
```diff
@@ -23,6 +24,12 @@ class BaseApplication(Case):
     application_type = models.CharField(choices=ApplicationType.choices, default=None, max_length=50)
     activity = models.TextField(default=None, blank=True, null=True)
     usage = models.TextField(default=None, blank=True, null=True)
+    duration = models.IntegerField(
```
BenAndrew-Kainos: Is duration applicable to an application or licence (should this not be the start of the licence model)?

applications/tests/tests_duration_view.py
```diff
@@ -0,0 +1,48 @@
...
+        url = reverse("applications:duration", kwargs={"pk": self.open_application.pk})
+
+        country = CountryOnApplication.objects.get(application=self.open_application).country
+
+        # Non eu
+        country.is_eu = False
+        country.save()
+        response = self.client.get(url, **self.gov_headers)
+
+        self.assertEqual(response.json()["duration"], 5 * 12)
```
BenAndrew-Kainos: When you move the duration to constants, you can update these values to point to those constants (so if the default changes the tests wont break)

applications/tests/tests_duration_view.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # eu
+        country.is_eu = True
```
BenAndrew-Kainos: I know it's fussy but could you split the EU & non EU bits into separate tests and add to the get statement to get EU/Non-EU where applicable rather than overwriting

applications/tests/tests_finalise_application.py
```diff
@@ -0,0 +1,29 @@
...
+        super().setUp()
+        self.standard_application = self.create_standard_application(self.organisation)
+        self.submit_application(self.standard_application)
+        self.url = reverse("applications:finalise", kwargs={"pk": self.standard_application.id})
+
+    def test_gov_super_user_finalise_application(self):
+        self.assertEqual(self.standard_application.duration, None)
+
+        # Give gov user super used role, to include reopen closed cases permission
+        self.gov_user.role = self.super_user_role
```
BenAndrew-Kainos: Culd you be more specific and just add the 'include reopen closed cases' permission (proves that it is this permission that allows them)

applications/tests/tests_finalise_application.py
```diff
@@ -0,0 +1,29 @@
...
+        self.gov_user.save()
+
+        data = {"duration": 12}
+        response = self.client.put(self.url, data=data, **self.gov_headers)
+
+        self.standard_application.refresh_from_db()
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(self.standard_application.status, get_case_status_by_status(CaseStatusEnum.FINALISED))
+        self.assertEqual(self.standard_application.duration, data["duration"])
```
BenAndrew-Kainos: Could you add a test for them not having the permission & assert 404

applications/views/applications.py
```diff
@@ -241,25 +244,29 @@ class ApplicationManageStatus(APIView):
     def put(self, request, pk):
         application = get_application(pk)
 
-        data = request.data
-        new_status = data.get("status")
+        data = deepcopy(request.data)
```
BenAndrew-Kainos: Not sure deepcopy is required here (could you just use .copy() ?)

applications/views/applications.py
```diff
@@ -241,25 +244,29 @@ class ApplicationManageStatus(APIView):
...
+
+        if data["status"] == CaseStatusEnum.FINALISED:
+            return JsonResponse(
+                data={"errors": ["Status cannot be set to finalised."]}, status=status.HTTP_400_BAD_REQUEST
```
BenAndrew-Kainos: Please move this error to content in the API section

applications/views/applications.py
```diff
@@ -241,25 +244,29 @@ class ApplicationManageStatus(APIView):
...
+            )
 
         if isinstance(request.user, ExporterUser):
             if request.user.organisation.id != application.organisation.id:
                 raise PermissionDenied()
 
-            if not can_status_can_be_set_by_exporter_user(application.status.status, new_status):
+            if not can_status_can_be_set_by_exporter_user(application.status.status, data.get("status")):
                 return JsonResponse(
                     data={"errors": ["Status cannot be set by Exporter user."]}, status=status.HTTP_400_BAD_REQUEST
```
BenAndrew-Kainos: Same as above

applications/views/applications.py
```diff
@@ -241,25 +244,29 @@ class ApplicationManageStatus(APIView):
...
                 )
         else:
-            if not can_status_can_be_set_by_gov_user(request.user, application.status.status, new_status):
+            if not can_status_can_be_set_by_gov_user(request.user, application.status.status, data.get("status")):
                 return JsonResponse(
                     data={"errors": ["Status cannot be set by Gov user."]}, status=status.HTTP_400_BAD_REQUEST
```
BenAndrew-Kainos: Same as above

applications/views/applications.py
```diff
@@ -241,25 +244,29 @@ class ApplicationManageStatus(APIView):
...
                 )
 
-        new_status = get_case_status_by_status(new_status)
-        request.data["status"] = str(new_status.pk)
+        case_status = get_case_status_by_status(data["status"])
```
BenAndrew-Kainos: You have done a mixture of data["status"] & data.get("status"). Could you make this consistent

applications/views/applications.py
```diff
@@ -273,7 +280,65 @@ def put(self, request, pk):
...
+    @transaction.atomic
+    def put(self, request, pk):
+        """
+        Finalise an application
+        """
+        application = get_application(pk)
+
+        if not can_status_can_be_set_by_gov_user(request.user, application.status.status, CaseStatusEnum.FINALISED):
+            return JsonResponse(
+                data={"errors": ["Status cannot be set by Gov user."]}, status=status.HTTP_400_BAD_REQUEST
```
BenAndrew-Kainos: Move string to content

cases/tests/tests_final_decisions.py
```diff
@@ -33,10 +33,10 @@ def test_can_record_final_decision_with_correct_permissions(self):
         self.gov_user.role = role
         self.gov_user.save()
 
-        data = {"status": CaseStatusEnum.FINALISED}
+        data = {"duration": 10}
```
BenAndrew-Kainos: Have you not just changed the purpose of this test (from checking FINALISED to a duration)

BenAndrew-Kainos:
I'm a bit concerned about the work done to 'finalised' statuses in this story. I don't think it's part of the scope

applications/models.py
a-bek: Duration is part of the application, like other things, that end up in the licence. I considered starting a Licence model, but I think its best to wait for a story that requires a Licence model.

applications/views/applications.py
a-bek: copy won't truly copy mutable objects. With copy: `id(data) != id(request.data)` but `id(data['mutable_list']) == id(request.data['mutable_list'])`

applications/models.py
stephanieboyd: During the tech discussion for this ticket, it was agreed that "duration" should be on a licence. It was also agreed (and looking at the lack of subtasks) that we wouldn't create the licence model or store the value anywhere just yet. Please remove from application, this is not the correct place for this property :)

cases/tests/tests_final_decisions.py
a-bek: No, previously finalising a case was using the `manage_status` view. `manage_status` view can no longer finalise a case, and finalising has a view of its own, which this is now testing and only expects a `duration` param

a-bek:
> I'm a bit concerned about the work done to 'finalised' statuses in this story. I don't think it's part of the scope

@BenAndrew-Kainos Could you elaborate?

applications/models.py
```diff
@@ -23,6 +19,7 @@ class BaseApplication(Case):
     application_type = models.CharField(choices=ApplicationType.choices, default=None, max_length=50)
     activity = models.TextField(default=None, blank=True, null=True)
     usage = models.TextField(default=None, blank=True, null=True)
+    duration = models.IntegerField(default=None, null=True, help_text="Set when application finalised")
```
stephanieboyd: Please rename to licence_duration

applications/views/applications.py
```diff
@@ -273,7 +283,72 @@ def put(self, request, pk):
...
+                actor=request.user,
+                verb=AuditType.UPDATED_STATUS,
+                target=application.get_case(),
+                payload={"status": CaseStatusEnum.human_readable(CaseStatusEnum.FINALISED)},
+            )
+
+        return JsonResponse(data={}, status=status.HTTP_200_OK)
+
+
+class DurationView(APIView):
```
stephanieboyd: Perhaps rename to:  ApplicationDurationView

applications/views/applications.py
```diff
@@ -273,7 +283,72 @@ def put(self, request, pk):
...
             verb=AuditType.UPDATED_STATUS,
             target=application.get_case(),
-            payload={"status": CaseStatusEnum.human_readable(new_status.status)},
+            payload={"status": CaseStatusEnum.human_readable(case_status.status)},
         )
 
         return JsonResponse(data={}, status=status.HTTP_200_OK)
+
+
+class FinaliseView(APIView):
```
stephanieboyd: Perhaps rename to: ApplicationFinaliseView

conf/constants.py
```diff
@@ -11,6 +11,7 @@ class GovPermissions(Enum):
     CONFIRM_OWN_ADVICE = "Confirm own advice"
     CONFIGURE_TEMPLATES = "Create and edit templates"
     REOPEN_CLOSED_CASES = "Can re-open closed cases"
+    MANAGE_LICENCE_DURATION = "Can edit licence duration"
```
stephanieboyd: Should this permission not reflect overriding like mentioned in the ticket. 

`Override_default_licence_duration`

applications/models.py
BenAndrew-Kainos: As per our call we are leaving this. However, please rename to `licence_duration` and rebuild migrations / update accordingly

applications/enums.py
```diff
@@ -28,3 +28,8 @@ class ApplicationExportLicenceOfficialType:
         (YES, "Yes"),
         (NO, "No"),
     ]
+
+
+class Duration:
```
BenAndrew-Kainos: Please rename to LicenceDuration

applications/models.py
BenAndrew-Kainos: ^^

applications/tests/tests_duration_view.py
BenAndrew-Kainos: Could you add `is_eu=True` to the get instead of overwriting

applications/tests/tests_finalise_application.py
```diff
@@ -0,0 +1,72 @@
...
+    def test_gov_super_invalid_role(self):
+        self.gov_user.role = self.role
+        self.gov_user.role.permissions.set([])
+        self.gov_user.save()
+
+        data = {"duration": 13}
+        response = self.client.put(self.url, data=data, **self.gov_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+        self.assertEqual(response.json(), {"errors": {"error": "You don't have permission to do that action"}})
```
BenAndrew-Kainos: Please point the error to the string in lite_content

applications/tests/tests_finalise_application.py
```diff
@@ -0,0 +1,72 @@
...
+        self.gov_user.role.permissions.set(
+            [GovPermissions.MANAGE_FINAL_ADVICE.name, GovPermissions.MANAGE_LICENCE_DURATION.name]
+        )
+        self.gov_user.save()
+
+        data = {"duration": Duration.MAX + 1}
+        response = self.client.put(self.url, data=data, **self.gov_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json(), {"errors": {"non_field_errors": ["Duration 1000 not in range [1-999]"]}})
```
BenAndrew-Kainos: Please point the error to the string in lite_content

applications/tests/tests_finalise_application.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_default_duration_can_be_set_without_role(self):
+        self.gov_user.role = self.role
+        self.gov_user.role.permissions.set([GovPermissions.MANAGE_FINAL_ADVICE.name])
+        self.gov_user.save()
+
+        data = {"duration": get_default_duration(self.standard_application)}
+        response = self.client.put(self.url, data=data, **self.gov_headers)
+        self.standard_application.refresh_from_db()
```
BenAndrew-Kainos: This test wouldn't catch if the duration wasn't updated. It's only asserting that the duration sent back matches what is in the object, not what was sent in the put request. 
Please change to `self.assertEqual(get_default_duration(self.standard_application), data["duration"])`

applications/views/applications.py
BenAndrew-Kainos: ok

cases/tests/tests_final_decisions.py
BenAndrew-Kainos: Could you write a docstring & rename the unit test to make this clear

applications/tests/tests_finalise_application.py
a-bek: The test asserts that a user can finalise the application using the default duration, without having manage_licence_duration permission.

Also, data["duration"] is `get_default_duration(self.standard_application)`

applications/views/applications.py
a-bek: It's already within the `application` namespace. It's better practice to avoid name smurfing.

conf/constants.py
a-bek: I think MANAGE_LICENCE_DURATION is more suitable, what if the user wants to edit a duration that the user set when reissuing a licence, ie, editing a non-default duration. Then `Override_default_licence_duration` wouldn't be a clear name.

applications/tests/tests_duration_view.py
a-bek: Just explicitly setting it for testing purposes.

applications/views/applications.py
BenAndrew-Kainos: I'm still against us having this extra view for when cases are finalised. Is there a reason we cannot add this logic to the existing update status view?

applications/tests/tests_finalise_application.py
BenAndrew-Kainos: Do not resolve a point without reason or action please

applications/tests/tests_finalise_application.py
BenAndrew-Kainos: Same here

BenAndrew-Kainos:
In future please leave a comment why it's resolved (i.e. done)

a-bek:
@BenAndrew-Kainos 
`But this means we don't test that we have any EU/ non-EU countries`
We're not testing data, we're testing the behaviour of the duration view for a country in EU.

`I'm still against us having this extra view for when cases are finalised. Is there a reason we cannot add this logic to the existing update status view?`
It's a seperation of concerns, manage status view is simply to edit a status. Finalising an application is a process in itself, I really don't think that belongs as and edge case in the manage_status view.

jfaracik:
> It's a seperation of concerns, manage status view is simply to edit a status. Finalising an application is a process in itself, I really don't think that belongs as and edge case in the manage_status view.

Agreed with this.

BenAndrew-Kainos:
> @BenAndrew-Kainos
> `But this means we don't test that we have any EU/ non-EU countries`
> We're not testing data, we're testing the behaviour of the duration view for a country in EU.

I know it's not the direct purpose of this test but it is important to verify EU countries are being loaded somewhere. If you don't want to do it here for some reason, update the seeding unit test to do so

BenAndrew-Kainos:
I'm still against us having this extra view for when cases are finalised. Is there a reason we cannot add this logic to the existing update status view?
It's a seperation of concerns, manage status view is simply to edit a status. Finalising an application is a process in itself, I really don't think that belongs as and edge case in the manage_status view.

That's a fair point but to me it's more important that you have a consistent API design. I would expect as a dev that all updates to status go via the updateStatus PUT request. Maintaining two endpoints for updating status seems unnecessary to me. 

@stephanieboyd you've done the most work with statuses before. What do you reckon?

a-bek:
```That's a fair point but to me it's more important that you have a consistent API design. I would expect as a dev that all updates to status go via the updateStatus PUT request. Maintaining two endpoints for updating status seems unnecessary to me.```

@BenAndrew-Kainos having `/finalise/` is actually consistent with API design. It is like `/submit/`as it edits the status to `SUBMITTED` and any other logic to do with submitting. Now if we went down the route of baking everything in manage_status, we would have to move the existing `/submit/` logic to manage_status as well, and eventually you end up with a giant manage_status view that handles edge cases.

Submit and Finalise are two actions that, yes, require managing status, but also require submit and finalise specific logic, so they warrant their own view.

File: applications/libraries/licence.py [hunk 22]
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Returns default duration in months
+
+    Rules defined in: https://uktrade.atlassian.net/browse/LT-1586
+    """
+
+    if application.export_type == ApplicationExportType.TEMPORARY:
+        return DefaultDuration.TEMPORARY
+
+    if (
```
Mark-Spiers-kainos: can you make these elifs?

applications/views/applications.py
stephanieboyd: You're inconsistent with how the team name things, and within this file already there is a pattern. Please rename

applications/urls.py
```diff
@@ -19,6 +19,8 @@
     path("", applications.ApplicationList.as_view(), name="applications"),
     path("<uuid:pk>/", applications.ApplicationDetail.as_view(), name="application"),
     path("<uuid:pk>/submit/", applications.ApplicationSubmission.as_view(), name="application_submit",),
+    path("<uuid:pk>/finalise/", applications.FinaliseView.as_view(), name="finalise"),
```
Mark-Spiers-kainos: can you change name to "finalise_application" or something similar? same with duration

applications/libraries/licence.py
a-bek: done, but it doesn't make a difference.

applications/urls.py
a-bek: I prefer 
`/applications/1/finalise/`
to
`/applications/1/finalise_application/`

https://devcards.io/smurf-naming-convention

applications/tests/tests_duration_view.py
a-bek: Added test to static

applications/serializers/generic_application.py
```diff
@@ -187,10 +191,23 @@ def update(self, instance, validated_data):
...
         return instance
+
+    def validate(self, data):
+        """
+        Check that the start is before the stop.
+        """
+        if data.get("licence_duration") is not None and (
+            data["licence_duration"] > LicenceDuration.MAX.value or data["licence_duration"] < LicenceDuration.MIN.value
+        ):
+            _min = LicenceDuration.MIN.value
```
BenAndrew-Kainos: Where are these used??

applications/views/applications.py
```diff
@@ -273,7 +283,72 @@ def put(self, request, pk):
...
+        data = deepcopy(request.data)
+
+        default_licence_duration = get_default_duration(application)
+
+        if (
+                data.get("licence_duration") is not None and
+                str(data["licence_duration"]) != str(default_licence_duration) and
+                not request.user.has_permission(GovPermissions.MANAGE_LICENCE_DURATION)
+        ):
+            #
```
BenAndrew-Kainos: Remove

applications/views/applications.py
```diff
@@ -273,7 +283,72 @@ def put(self, request, pk):
...
+        serializer.save()
+
+        audit_trail_service.create(
+            actor=request.user,
+            verb=AuditType.FINALISED_APPLICATION,
+            target=application.get_case(),
+            payload={"licence_duration": serializer.validated_data["licence_duration"]},
+        )
+
+        return JsonResponse(data={}, status=status.HTTP_200_OK)
```
BenAndrew-Kainos: This is a PUT endpoint so should return the updated object

applications/views/applications.py
BenAndrew-Kainos: Resolved without comment or action?

static/countries/enums.py
```diff
@@ -0,0 +1,5 @@
+from enum import Enum
+
+
+class Misc(Enum):
```
BenAndrew-Kainos: Don't like this at all. This enum will be left behind over time with countries being updated. Please remove

static/countries/models.py
```diff
@@ -1,10 +1,16 @@
...
 class Country(models.Model):
     id = models.CharField(primary_key=True, editable=False, max_length=10)  # Country Code
     name = models.CharField(max_length=100)
     type = models.CharField(max_length=100)
     flags = models.ManyToManyField(Flag, related_name="countries")
+    is_eu = models.BooleanField()
+
+    @classmethod
+    def eu_count(cls):
+        return Misc.EU_COUNT.value
```
BenAndrew-Kainos: Why do you need this count. (Also change to `Country.objects.filter(is_eu=True).count()`)

static/management/commands/seedcountries.py
```diff
@@ -26,4 +27,5 @@ def operation(self, *args, **options):
 class SeedCountriesTests(SeedCommandTest):
     def test_seed_countries(self):
         self.seed_command(Command)
-        self.assertTrue(Country.objects.count() == len(Command.read_csv(COUNTRIES_FILE)))
+        self.assertEqual(Country.objects.count(), len(Command.read_csv(COUNTRIES_FILE)))
+        self.assertEqual(Country.objects.filter(is_eu=True).count(), Misc.EU_COUNT.value)
```
BenAndrew-Kainos: Instead of this compare the number of occurrences in the csv to the number of objects. Also only read the csv once (every time you call `Command.read_csv(COUNTRIES_FILE))` you're unnecessarily re-reading the csv)

cases/tests/tests_final_decisions.py
BenAndrew-Kainos: Resolved without comment or action (again)

applications/views/applications.py
a-bek: It's already within the application namespace. It's better practice to avoid name smurfing.

static/countries/enums.py
a-bek: Without over-engineering, this is a simple way to validate the EU countries data like you asked. There is nothing wrong with using this as a miscellaneous value, if the actual number of EU countries are updated in the csv this will be caught.

static/countries/models.py
a-bek: Removed

static/management/commands/seedcountries.py
a-bek: Reading the CSV is only being called once in the test function block, and is used to validate `seed_command` does what it is supposed to do

applications/views/applications.py
a-bek: Yes I know there some kind of a pattern already, but even the existing pattern is inconsistent. Generally its recommended to avoid name smurfing, so that SHOULD be the practise going forward, but sure I'll add the Application prefix.

static/management/commands/seedcountries.py
BenAndrew-Kainos: It was being used twice when I reviewed but looks good now

static/countries/enums.py
BenAndrew-Kainos: The problem is content have the responsibility to keep CSV's up to date but not enums. If a country becomes EU or leaves then the tests will fail because this number is incorrect. I would replace all instances of this with a .count() on countries.

BenAndrew-Kainos:
Approving subject to the Misc Enum being deleted!!!

static/countries/enums.py
a-bek: Thats expected behaviour of the test? It validates content have added the right number of EU countries and it validates that EU countries exist? I thought this is what you wanted when you said:

`But this means we don't test that we have any EU/ non-EU countries`

Updated and removed again.

PR #268, opened 2020-01-15
bug: LT-1889 CLC Query Generated Documents

BenAndrew-Kainos:


queries/control_list_classifications/tests.py
```diff
@@ -191,3 +192,29 @@ def test_user_set_clc_status_success(self):
 
         self.assertEqual(query.status.status, CaseStatusEnum.WITHDRAWN)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+
+class ControlListClassificationsGeneratedDocumentsTests(DataTestClient):
+    def setUp(self):
```
stephanieboyd: What's going on with your tests?

documents/views.py
```diff
@@ -20,3 +29,17 @@ def get(self, request, pk):
             return JsonResponse({"document": serializer.data})
         except Document.DoesNotExist:
             raise NotFoundError({"document": "Document not found"})
+
+
+class ExporterCaseDocumentDownload(APIView):
+    authentication_classes = (ExporterAuthentication,)
+
+    def get(self, request, case_pk, file_pk):
```
stephanieboyd: minor - `document_pk` would be clearer and more consistent

queries/control_list_classifications/tests.py
BenAndrew-Kainos: This bug required a new endpoint to get the list of CLC documents (what this test covers) & a new endpoint to download documents (ExporterCaseDocumentDownload)

documents/views.py
BenAndrew-Kainos: Done

documents/libraries/s3_operations.py
```diff
@@ -24,8 +27,23 @@ def upload_bytes_file(raw_file, s3_key):
...
+# Delete
 def delete_file(s3_key):
     try:
         _client.delete_object(Bucket=_bucket_name, Key=s3_key)
     except Exception:  # noqa
         logging.warning("Failed to delete file")
+
+
+# Download
+def _generate_file(result):
```
efahy: Perhaps _stream_file would be a more accurate name?

cases/tests/tests_case_documents.py
```diff
@@ -25,16 +26,43 @@ def test_can_view_all_documents_on_a_case(self):
...
+    @mock.patch("documents.libraries.s3_operations.get_object")
+    def test_download_case_document_success(self, get_object_function):
+        url = reverse(self.path, kwargs={"case_pk": self.case.id, "document_pk": self.file.id})
+
+        response = self.client.get(url, **self.exporter_headers)
+
+        get_object_function.assert_called_once()
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertTrue(isinstance(response, StreamingHttpResponse))
+        self.assertTrue(f'filename="{self.file.name}"' in response._headers["content-disposition"][1])
```
efahy: Is there any chance we can actually stream the test doc into a byte array or something like that and assert that the correct number of bytes were downloaded?

documents/libraries/s3_operations.py
BenAndrew-Kainos: Done

cases/tests/tests_case_documents.py
BenAndrew-Kainos: Unfortunately we currently don't have any good ways of mocking s3 returns. It might be something I look into when we move more of the s3 content over

PR #269, opened 2020-01-16
bug: LT-1837 HMRC Query Letter Template

BenAndrew-Kainos:


static/case_types/views.py
```diff
@@ -0,0 +1,11 @@
+from django.http import JsonResponse
+from rest_framework.views import APIView
+from rest_framework.status import HTTP_200_OK
+
+from cases.enums import CaseTypeEnum
+
+
+class CaseTypes(APIView):
```
a-bek: This view belongs in `/cases/`?

BenAndrew-Kainos: It's static & doesn't relate to cases in any way (only CaseTypes)

static/case_types/views.py
```diff
@@ -0,0 +1,11 @@
...
+    def get(self, request):
+        case_types = {choice[0]: choice[1] for choice in CaseTypeEnum.choices}
```
stephanieboyd: In case you weren't aware and want to follow the approach done for statuses (and others), there is a function that does:
``` @classmethod
    def as_list(cls):
        return [{"status": choice[0], "priority": cls.priority[choice[0]]} for choice in cls.choices]```

BenAndrew-Kainos: It's more efficient on the frontend to pass as a dict as this list will be used to populate radio button selections. However if it's very important that it is passed in this format I can change it.

PR #270, opened 2020-01-17
task: add Seeding tests to CircleCI

BenAndrew-Kainos:


static/management/commands/seedgovusers.py
```diff
@@ -27,6 +27,8 @@ class Command(SeedCommand):
 
     @transaction.atomic
     def operation(self, *args, **options):
+        assert Role.objects.count(), "Role permissions must be seeded first!"
```
Mark-Spiers-kainos: This is looking at the Role mode, but seems to comment about Role permissions if assertion fails?

BenAndrew-Kainos: Yeah so this seed is dependent on the Role permissions seed. This assertion ensures that this seed command has been run first

PR #272, opened 2020-01-20
feat: LT-948 Show assigned users on the case screen

pkonopacki:
- changed CaseDetailSeralizer to return users if they are assigned to the case

- refactored the code in TinyCaseSerializer to the helper method


cases/serializers.py
```diff
@@ -114,18 +115,7 @@ def get_users(self, instance):
...
-            queue_users = [
-                {"first_name": first_name, "last_name": last_name, "email": email, "queue": case_assignment.queue.name,}
-                for first_name, last_name, email in case_assignment.users.values_list(
-                    "first_name", "last_name", "email"
-                )
-            ]
-
-            users.extend(queue_users)
-        return users
+        return get_users_assigned_to_case(case_assignments)
```
a-bek: Perhaps this can be optimised by using select_related on users, to avoid `n` additional queries.

pkonopacki: Changed, can you resolve @a-bek ?

PR #274, opened 2020-01-22
chore: LT-1658 - Refactor Party model

a-bek:
Squash Api

models.py
```
class Party
```
serializers.py
```
class PartySerializer
```

It used to be:

models.py
```
class Party
class Consignee
class EndUser
class ThirdParty
class UltimateEndUser
```
serializers.py
```
class PartySerializer
class EndUserSerializer
class EndUserWithFlagsSerializer
class UltimateEndUserSerializer
class UltimateEndUserWithFlagsSerializer
class ConsigneeSerializer
class ConsigneeWithFlagsSerializer
class ThirdPartySerializer
class ThirdPartyWithFlagsSerializer
```

Note:
- unit tests pass
- e2e tests pass : internal + exporter
- ignore deleted migration files


addresses/migrations/0001_initial.py
```diff
@@ -1,29 +0,0 @@
-# Generated by Django 2.2.4 on 2019-12-09 12:09
-
-from django.db import migrations, models
-import django.db.models.deletion
-import uuid
-
-
-class Migration(migrations.Migration):
```
efahy: Please make double and triple sure that you haven't deleted any custom migration logic.  I know Andrew mentioned something to do with permissions this morning.  I think it would be worth you asking the senior devs on the team if they are aware of any manual steps in migrations that you might have deleted

cases/serializers.py
```diff
@@ -302,10 +303,18 @@ class CaseAdviceSerializer(serializers.ModelSerializer):
     good = serializers.PrimaryKeyRelatedField(queryset=Good.objects.all(), required=False)
     goods_type = serializers.PrimaryKeyRelatedField(queryset=GoodsType.objects.all(), required=False)
     country = serializers.PrimaryKeyRelatedField(queryset=Country.objects.all(), required=False)
-    end_user = serializers.PrimaryKeyRelatedField(queryset=EndUser.objects.all(), required=False)
-    ultimate_end_user = serializers.PrimaryKeyRelatedField(queryset=UltimateEndUser.objects.all(), required=False)
-    consignee = serializers.PrimaryKeyRelatedField(queryset=Consignee.objects.all(), required=False)
-    third_party = serializers.PrimaryKeyRelatedField(queryset=ThirdParty.objects.all(), required=False)
+    end_user = serializers.PrimaryKeyRelatedField(
+        queryset=Party.objects.filter(type=PartyType.END_USER), required=False
```
efahy: We're doing a lot of filtering now on this party type field and its currently an unindexed char field, which is going to make this very inefficient as the data in the table grows over time.  I would like you to either put an index on the column/field or even better, have a PartyType model and foreign key to it

a-bek: What do you think of this approach?

```
class PartyOnApplication(Timestampable):
    party = models.ForeignKey(Party)
    application = models.ForeignKey(Application)
    is_active = models.BooleanField()
    party_type = models.TextField(choice=PartyType.choices)
```
Remove the end_user, third_party, etc, fields on The Application models that use it. 

We can also remove `party_type` from the Party model as that is actually dependant on the application its on.

To get a parties on an application is 1 query, no matter the party type: 

```PartyOnApplication.objects.filter(application=application)```

File: parties/models.py [hunk 3]
```diff
@@ -6,7 +6,7 @@
 from documents.models import Document
 from flags.models import Flag
 from organisations.models import Organisation
-from parties.enums import PartyType, SubType, ThirdPartyRole
+from parties.enums import PartyType, SubType, PartyRole
```
efahy: On the screen, this is called "Third Party Role", so it'd probably be more consistent to keep the old name

cases/serializers.py
efahy: As discussed, lets not go with this approach.  The `PartyOnApplication` table is not a good fit because
1. It suggests parties can exist outside the scope of any application
2. It suggests a party could exist on multiple applications (essentially its a many to many join table) when in fact we always make copies of parties and every party is only ever associated with one application
3. Anne doesn't want an address book type concept for parties
4. I think the refactor is big enough already especially considering none of it was pre-planned or discussed
5. Performance of the look ups on party type are easily solved with an index or separate table & foreign key

PR #282, opened 2020-01-29
feat: Lt 1874 remove duplicate people from previously referenced list

stephanieboyd:


applications/views/existing_parties.py
```diff
@@ -21,11 +24,36 @@ class ExistingParties(generics.ListCreateAPIView):
...
+        # Essentially, we group copied parties by name and copy_of_id and for each grouping, get the id
+        # of the newest party in the group.  Then use that set of ids to get the corresponding list of parties
+        copied_party_ids = (
+            Party.objects.filter(organisation=application.organisation, copy_of_id__isnull=False, **params)
+            .annotate(
+                first_party_id=Window(
+                    expression=FirstValue("id"), partition_by=["name", "copy_of_id"], order_by=F("created_at").desc(),
+                )
+            )
+            .values_list("first_party_id")
```
Mark-Spiers-kainos: if you supply flat=True this will equal an array of values rather than a dict of values.

applications/views/parties.py
```diff
@@ -259,21 +271,29 @@ def post(self, request, application):
...
-
-        audit_trail_service.create(
-            actor=request.user,
-            verb=AuditType.ADD_PARTY,
-            target=case,
-            payload={"party_type": third_party.type.replace("_", " "), "party_name": third_party.name},
-        )
-
-        return JsonResponse(data={"third_party": serializer.data}, status=status.HTTP_201_CREATED)
+        if "validate_only" in data and not isinstance(data["validate_only"], bool):
```
andytaylorkainos: there should be a `str_to_bool` helper you can use for this :)

applications/views/parties.py
```diff
@@ -259,21 +271,29 @@ def post(self, request, application):
...
+            return JsonResponse(data={"error": ""}, status=status.HTTP_400_BAD_REQUEST,)
+
+        if "validate_only" in data and data["validate_only"] is True:
```
andytaylorkainos: there should be a str_to_bool helper you can use for this :)

applications/views/parties.py
```diff
@@ -115,21 +115,32 @@ def post(self, request, application):
...
-
-        audit_trail_service.create(
-            actor=request.user,
-            verb=AuditType.ADD_PARTY,
-            target=application.get_case(),
-            payload={"party_type": ultimate_end_user.type.replace("_", " "), "party_name": ultimate_end_user.name,},
-        )
-
-        return JsonResponse(data={"ultimate_end_user": serializer.data}, status=status.HTTP_201_CREATED)
+        if "validate_only" in data and not isinstance(data["validate_only"], bool):
```
andytaylorkainos: there should be a str_to_bool helper you can use for this :)

applications/views/parties.py
```diff
@@ -115,21 +115,32 @@ def post(self, request, application):
...
+            return JsonResponse(data={"error": ""}, status=status.HTTP_400_BAD_REQUEST,)
+
+        if "validate_only" in data and data["validate_only"] is True:
```
andytaylorkainos: there should be a str_to_bool helper you can use for this :)

applications/views/existing_parties.py
```diff
@@ -21,11 +24,36 @@ class ExistingParties(generics.ListCreateAPIView):
...
         params = {f"{key}__contains": value[0] for key, value in dict(self.request.GET).items()}
         # Rename country to country__name for filter
         if "country__contains" in params:
             params["country__name__contains"] = params.pop("country__contains")
 
         application_id = self.kwargs["pk"]
         application = get_application(application_id)
-        return Party.objects.filter(organisation=application.organisation, **params)
+
+        # Stage 1: Build a query for uncopied parties, i.e. parties that are not copies of other
```
andytaylorkainos: can you please split out the stages to separate functions

applications/views/parties.py
```diff
@@ -259,21 +271,29 @@ def post(self, request, application):
...
-            target=case,
-            payload={"party_type": third_party.type.replace("_", " "), "party_name": third_party.name},
-        )
-
-        return JsonResponse(data={"third_party": serializer.data}, status=status.HTTP_201_CREATED)
+        if "validate_only" in data and not isinstance(data["validate_only"], bool):
+            return JsonResponse(data={"error": ""}, status=status.HTTP_400_BAD_REQUEST,)
+
+        if "validate_only" in data and data["validate_only"] is True:
+            serializer = ThirdPartySerializer(data=data)
```
Mark-Spiers-kainos: repeating code in both validate_only and not validate_only. perhaps do is_valid first?

applications/views/existing_parties.py
```diff
@@ -21,11 +24,36 @@ class ExistingParties(generics.ListCreateAPIView):
...
+            Party.objects.filter(organisation=application.organisation, copy_of_id__isnull=False, **params)
+            .annotate(
+                first_party_id=Window(
+                    expression=FirstValue("id"), partition_by=["name", "copy_of_id"], order_by=F("created_at").desc(),
+                )
+            )
+            .values_list("first_party_id")
+            .distinct()
+        )
+        copied_parties = Party.objects.filter(id__in=copied_party_ids)
```
Mark-Spiers-kainos: why does uncopied_parties filter on params and not copied_parties?

applications/views/existing_parties.py
```diff
@@ -21,11 +24,36 @@ class ExistingParties(generics.ListCreateAPIView):
...
+        # parties and have never been copied themselves
+        party_copy_of_ids = Party.objects.filter(copy_of_id__isnull=False).values("copy_of_id").distinct()
+        uncopied_parties = Party.objects.filter(
+            organisation=application.organisation, copy_of_id__isnull=True, **params
+        ).exclude(id__in=party_copy_of_ids)
+
+        # Stage 2: Build a query set using a Django window function to find the most recent parties that are copies.
+        # Essentially, we group copied parties by name and copy_of_id and for each grouping, get the id
+        # of the newest party in the group.  Then use that set of ids to get the corresponding list of parties
+        copied_party_ids = (
```
BenAndrew-Kainos: Could you rename as this isn't technically all the copied parties (it's the latest party in each group)

applications/views/parties.py
```diff
@@ -115,21 +115,32 @@ def post(self, request, application):
...
-        audit_trail_service.create(
-            actor=request.user,
-            verb=AuditType.ADD_PARTY,
-            target=application.get_case(),
-            payload={"party_type": ultimate_end_user.type.replace("_", " "), "party_name": ultimate_end_user.name,},
-        )
-
-        return JsonResponse(data={"ultimate_end_user": serializer.data}, status=status.HTTP_201_CREATED)
+        if "validate_only" in data and not isinstance(data["validate_only"], bool):
+            return JsonResponse(data={"error": ""}, status=status.HTTP_400_BAD_REQUEST,)
```
BenAndrew-Kainos: Is there a point passing an empty error (can we change to just a status code)

applications/views/parties.py
BenAndrew-Kainos: Yeah I think if you use that helper then it defaults to False by default so you don't need to validate it

applications/views/parties.py
```diff
@@ -115,21 +115,32 @@ def post(self, request, application):
...
+            if serializer.is_valid():
+                ultimate_end_user = serializer.save()
+                application.ultimate_end_users.add(ultimate_end_user.id)
+
+                audit_trail_service.create(
+                    actor=request.user,
+                    verb=AuditType.ADD_PARTY,
+                    target=application.get_case(),
+                    payload={
+                        "party_type": ultimate_end_user.type.replace("_", " "),
```
BenAndrew-Kainos: Does party type not have a human readable value (if not it should!)

applications/views/parties.py
```diff
@@ -259,21 +271,29 @@ def post(self, request, application):
...
-        return JsonResponse(data={"third_party": serializer.data}, status=status.HTTP_201_CREATED)
+        if "validate_only" in data and not isinstance(data["validate_only"], bool):
+            return JsonResponse(data={"error": ""}, status=status.HTTP_400_BAD_REQUEST,)
+
+        if "validate_only" in data and data["validate_only"] is True:
+            serializer = ThirdPartySerializer(data=data)
+            if serializer.is_valid():
+                return JsonResponse(data={"third_party": serializer.initial_data}, status=status.HTTP_200_OK)
+        else:
+            serializer = ThirdPartySerializer(data=data)
```
BenAndrew-Kainos: You could refactor this block of code for all parties passing the serializer, the objects for it to be added to (application.third_parties in this case) and the entity (third_party)

applications/views/parties.py
```diff
@@ -259,21 +271,29 @@ def post(self, request, application):
...
+            if serializer.is_valid():
+                third_party = serializer.save()
+                application.third_parties.add(third_party.id)
+
+                audit_trail_service.create(
+                    actor=request.user,
+                    verb=AuditType.ADD_PARTY,
+                    target=case,
```
BenAndrew-Kainos: Why is this one case but in ultimate end users it uses application.get_case()

applications/tests/tests_existing_parties.py
```diff
@@ -26,6 +26,24 @@ def test_get_existing_parties(self):
...
+            website="https://www.gov.py",
+            country=self.country,
+            sub_type="government",
+            organisation=second_organisation,
+        )
+
+        response_data = self.client.get(self.url, **self.exporter_headers).data["results"]
+        response_data_ids = [party["id"] for party in response_data]
+
+        self.assertEqual(Party.objects.count() - 1, len(response_data))
```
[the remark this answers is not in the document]
stephanieboyd: Sure, changed. It was the original organisation parties we want though

applications/tests/tests_existing_parties.py
```diff
@@ -59,3 +77,68 @@ def test_get_existing_party_with_all_filters(self):
...
+                country=self.country,
+                sub_type="government",
+                organisation=self.organisation,
+                copy_of_id=original_party.id,
+            )
+
+        response_data = self.client.get(self.url, **self.exporter_headers).data["results"]
+
+        response_data_ids = [party["id"] for party in response_data]
+        expected_copy_id = Party.objects.all().filter(name="Mr Original", address="456 abc st.").get().id
```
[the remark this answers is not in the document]
stephanieboyd: Changed

applications/tests/tests_existing_parties.py
```diff
@@ -59,3 +77,68 @@ def test_get_existing_party_with_all_filters(self):
...
+            sub_type="government",
+            organisation=self.organisation,
+        )
+
+        parties = [
+            {"name": "Mr Original", "address": "456 abc st.", "website": "https://www.gov.py"},
+            {"name": "Mr Copy", "address": "456 abc st.", "website": "https://www.gov.py"},
+        ]
+
+        for party in parties:
```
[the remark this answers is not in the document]
stephanieboyd: Yeah, the draft already has 3 users added as part of all the test setup - so I've added an assert to check for one of those original parties in the response data.

applications/views/existing_parties.py
```diff
@@ -21,11 +24,36 @@ class ExistingParties(generics.ListCreateAPIView):
...
         if "country__contains" in params:
             params["country__name__contains"] = params.pop("country__contains")
 
         application_id = self.kwargs["pk"]
         application = get_application(application_id)
-        return Party.objects.filter(organisation=application.organisation, **params)
+
+        # Stage 1: Build a query for uncopied parties, i.e. parties that are not copies of other
+        # parties and have never been copied themselves
+        party_copy_of_ids = Party.objects.filter(copy_of_id__isnull=False).values("copy_of_id").distinct()
```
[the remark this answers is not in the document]
stephanieboyd: It could contain them but it doesn't need to. This is just a simple, give me all the distinct copy of ids.

applications/views/existing_parties.py
stephanieboyd: Sure thing, changed to `newest_copied_party_ids_in_group`

applications/views/parties.py
andytaylorkainos: You shouldn't pass an empty error with a bad request (you should always try to explain why it was bad)

applications/views/parties.py
stephanieboyd: This was all existing, and just moved under the validate only check. On looking, there is a `case` variable that's assigned the `get_case()` call. I've removed the case variable and made them consistent for these and elsewhere in the file where it doesn't need to use it more than once.

applications/views/existing_parties.py
stephanieboyd: Yeah, done

applications/views/parties.py
stephanieboyd: Changed as discussed with Andy.

applications/views/parties.py
stephanieboyd: This has been removed as discussed with Andy

applications/views/existing_parties.py
stephanieboyd: Split out

applications/views/existing_parties.py
stephanieboyd: Both get filtered on org and params, but it happens at different steps of building our desired result (based on how we had to query to get it). copied_parties happens in the initial query to build the copied_party_ids and to include it in the second query is unnecessary.

applications/views/existing_parties.py
stephanieboyd: Yeah, it would - gonna leave it off though as it isn't necessary as we don't need to work with the returned results ourselves but just letting Django union them :)

applications/views/parties.py
stephanieboyd: Changed as discussed with Andy

applications/views/parties.py
stephanieboyd: I've moved `serializer = ThirdPartySerializer(data=data)` to outside the blocks, but leaving the is valid calls inside.

applications/views/parties.py
stephanieboyd: I could and I like your suggestion, but it's unrelated to the ticket and would need changed for the rest of them so going to leave it.

applications/views/parties.py
stephanieboyd: Looks like case status is the only enum that does?

applications/tests/tests_third_parties.py
```diff
@@ -242,10 +241,80 @@ def test_delete_third_party_when_application_read_only_failure(
...
+        }
+
+        # Delete existing third party to enable easy assertion of copied third party
+        delete_url = reverse(
+            "applications:remove_third_party",
+            kwargs={"pk": self.draft.id, "tp_pk": self.draft.third_parties.first().id},
+        )
+        self.client.delete(delete_url, **self.exporter_headers)
+
+        response = self.client.post(self.url, third_party, **self.exporter_headers)
```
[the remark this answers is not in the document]
stephanieboyd: As discussed, a full body check is added to an earlier existing test. Test specific copy of check added here

applications/tests/tests_third_parties.py
```diff
@@ -242,10 +241,80 @@ def test_delete_third_party_when_application_read_only_failure(
...
+        }
+
+        original_party_count = self.draft.third_parties.count()
+
+        response = self.client.post(self.url, third_party, **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(original_party_count, self.draft.third_parties.count())
+
+    def test_third_party_validate_only_failure(self):
```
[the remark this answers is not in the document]
stephanieboyd: As we tried to discuss, ahaha, I have added a parametrised input to `test_unsuccessful_add_third_party` instead of this test

applications/tests/tests_existing_parties.py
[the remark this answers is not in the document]
BenAndrew-Kainos: Still showing up on this line for me

applications/views/parties.py
BenAndrew-Kainos: This one should as well (if you can)

stephanieboyd: It was existing, and isn't to do with the story Benny - I will pass on changing it.

applications/tests/tests_existing_parties.py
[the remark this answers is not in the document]
stephanieboyd: I'd just missed one, both changed

applications/views/parties.py
```diff
@@ -122,22 +130,29 @@ def post(self, request, application):
...
+                ultimate_end_user = serializer.save()
+                application.ultimate_end_users.add(ultimate_end_user.id)
 
-        audit_trail_service.create(
-            actor=request.user,
-            verb=AuditType.ADD_PARTY,
-            target=application.get_case(),
-            payload={"party_type": ultimate_end_user.type.replace("_", " "), "party_name": ultimate_end_user.name,},
-        )
+                audit_trail_service.create(
```
BenAndrew-Kainos: Could we refactor these repeating party Create & Remove audit trail blocks and just pass the party object instead

andytaylorkainos: I feel like this is a little outside the scope of the story ben. I agree it would have been better if the code was done this way from the start, but from what I can see there's only been some logic added to the views and the code was indented - if we refactor every line of code that we touch we'll be spring cleaning for far too long

PR #284, opened 2020-01-30
feat: LT-1128 - Remove case from queues and users when terminal

andytaylorkainos:


cases/models.py
```diff
@@ -50,6 +50,12 @@ class Case(TimestampableModel):
     def save(self, *args, **kwargs):
         if not self.reference_code and self.status != get_case_status_by_status(CaseStatusEnum.DRAFT):
             self.reference_code = generate_reference_code(self)
+
+        if self.status.status in CaseStatusEnum.terminal_statuses():
```
stephanieboyd: You've got a function that does this check - it's called `is_terminal()`

andytaylorkainos: Done :)

cases/models.py
```diff
@@ -50,6 +50,12 @@ class Case(TimestampableModel):
...
+            self.case_officer = None
+            self.queues.clear()
```
stephanieboyd: I believe the case is to remain on All cases queue. Does this happen, or is clear from all queues?

andytaylorkainos: All cases queue is a system queue, you dont set it (its dynamically created)

cases/models.py
```diff
@@ -50,6 +50,12 @@ class Case(TimestampableModel):
...
+            CaseAssignment.objects.filter(case=self).delete()
```
stephanieboyd: Is this line necessary? Looks like deleting the queue might take care of it as the CaseAssignment model queue field is defined as `on_delete=models.CASCADE)`

cases/models.py
stephanieboyd: Okay - there are other system queues, aren't there? And they are removed from that okay?

cases/models.py
andytaylorkainos: you dont delete the queue, you just remove the link between the case and the queue

andytaylorkainos: The only queues that are set against a case are non-system (so the case will still appear only appear on `All cases` as it will no longer be an open case [another system queue])

cases/models.py
andytaylorkainos: System/sudo queues arent set in the DB, they're sudo queues which grab cases via a filterset
So the case will only appear on All cases as it will no longer be an open case (another system/sudo queue)

applications/tests/tests_application_status.py
```diff
@@ -62,6 +63,30 @@ def test_exporter_set_application_status_withdrawn_when_application_not_terminal
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(self.standard_application.status, get_case_status_by_status(CaseStatusEnum.WITHDRAWN))
 
+    @parameterized.expand(
+        [case_status for case_status in CaseStatusEnum.terminal_statuses() if case_status != CaseStatusEnum.FINALISED]
```
stephanieboyd: Can you please check if we need a test for `ApplicationFinaliseView` - as that is where the finalised terminal status gets set

andytaylorkainos: Awesome, didnt realise we had an endpoint for that!

andytaylorkainos: Done!

applications/tests/tests_finalise_application.py
```diff
@@ -46,6 +47,33 @@ def test_gov_user_finalise_application_success(self):
         self.assertEqual(self.standard_application.status, get_case_status_by_status(CaseStatusEnum.FINALISED))
         self.assertEqual(self.standard_application.licence_duration, data["licence_duration"])
 
+    def test_gov_user_finalise_application_removes_case_from_queues_and_users_from_case_success(self):
+        """
+        When a case is set to a terminal status, its assigned users, case officer and queues should be removed
```
stephanieboyd: Since we've testing finalised view, can you make the comment specific. sorry. thanks.

andytaylorkainos: Done :)

PR #286, opened 2020-01-30
feat: LT-1084 - Record details of when a customs enquiry is being performed

jfaracik:
* Added `have_goods_departed` field to HMRC Query
* Signals! - When HMRC Query is saved and `have_goods_departed` is set to True it clears all attached sites and locations 
* Added tests for this feature

PHG2de9


goodstype/serializers.py
```diff
@@ -49,9 +45,12 @@ def __init__(self, *args, **kwargs):
         if application:
             if get_application(application).application_type == ApplicationType.OPEN_LICENCE:
                 self.fields["is_good_incorporated"] = serializers.BooleanField()
-                self.Meta.fields = self.Meta.fields + ("is_good_incorporated",)
+                self.fields["is_good_controlled"] = serializers.BooleanField()
+                self.fields["control_code"] = serializers.CharField(required=False, allow_blank=True, allow_null=True)
```
andytaylorkainos: should it not be required? as i think the options are either "True" or "False" for this application type?

applications/tests/tests_hmrc_query.py
```diff
@@ -89,3 +101,36 @@ def test_status_code_post_with_infected_document_failure(self):
...
+        delete all existing sites and locations on that application
+        """
+        data = {"is_goods_departed": False}
+
+        response = self.client.put(
+            reverse("applications:application", kwargs={"pk": self.hmrc_query.id}), data, **self.hmrc_exporter_headers
+        )
+        self.hmrc_query.refresh_from_db()
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
andytaylorkainos: can you assert the response body for PUT/POST requests too

applications/models.py
```diff
@@ -67,6 +67,7 @@ class HmrcQuery(BaseApplication):
     )
     third_parties = models.ManyToManyField(ThirdParty, related_name="hmrc_query_third_parties")
     reasoning = models.CharField(default=None, blank=True, null=True, max_length=1000)
+    is_goods_departed = models.BooleanField(default=False)
```
andytaylorkainos: this is extremely pernickety but, should it not be `are_goods_departed`?

applications/signals.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@receiver(post_save, sender=HmrcQuery)
+def hmrc_query_save(instance: HmrcQuery, **kwargs):
+    """
+    Every time a HMRC Query is saved, run these automations
+    """
+    # If the goods are set to already departed, clear all existing sites and locations
+    if instance.is_goods_departed:
+        instance.application_sites.all().delete()
```
andytaylorkainos: careful of the use of `delete` (i may be wrong, but it might delete the actual site)
Safer to use `instance.application_sites.clear()` which just removes the M-M link

goodstype/serializers.py
jfaracik: I've went with whatever options were there before.

applications/tests/tests_hmrc_query.py
jfaracik: Doesn't return anything.

applications/models.py
jfaracik: Or `have_goods_departed` - depends on how much you want to stick to prefixing booleans with `is_`

andytaylorkainos: I'll let you decide. The grammar makes me shiver, but i understand why you went for that name

jfaracik: I've went with `have_goods_departed`

applications/signals.py
jfaracik: `Method 'clear' does not exist on manager when foreign key does not allow nulls. consider adding null=true to foreign key constructor`

jfaracik: Doesn't seem to happy with `.clear()` :/

goodstype/serializers.py
jfaracik: Ive hard set it to required=True, not too sure what the default value for the object is though

applications/signals.py
andytaylorkainos: Ahh, its becuase `application_sites` is a related_name lookup to one of our own manually created m-2-m tables (this means .delete() is only removing the `SiteOnApplication` relationship - awesome!)

applications/signals.py
```diff
@@ -0,0 +1,15 @@
...
+
+from applications.models import HmrcQuery
+
+
+@receiver(post_save, sender=HmrcQuery)
+def hmrc_query_save(instance: HmrcQuery, **kwargs):
+    """
+    Every time a HMRC Query is saved, run these automations
+    """
+    # If the goods are set to already departed, clear all existing sites and locations
```
efahy: The signal on its own is not enough to always ensure that a HMRC query is in a valid state within the API.  I realise you have validation in the UI but the API should also ensure its own integrity.  I think all you need to do is validate when adding sites or external locations to a HMRC query, that the have goods departed flag is false and if not return a 400 or something.

I managed to submit an app with the flag set and a site added by hacking the UI a little bit

applications/tests/tests_hmrc_query.py
```diff
@@ -89,3 +101,36 @@ def test_status_code_post_with_infected_document_failure(self):
...
+        self.assertTrue(self.hmrc_query.have_goods_departed)
+        self.assertEqual(SiteOnApplication.objects.filter(application=self.hmrc_query).count(), 0)
+        self.assertEqual(ExternalLocationOnApplication.objects.filter(application=self.hmrc_query).count(), 0)
```
efahy: Nice 👍

efahy: Can you add another test to check that a site or external location can only be added when `have_goods_departed` is false please?

PR #292, opened 2020-02-05
feat: LT-979 surrender a licence

BenAndrew-Kainos:


applications/libraries/application_helpers.py
```diff
@@ -22,6 +22,9 @@ def can_status_can_be_set_by_exporter_user(original_status: str, new_status: str
     if new_status == CaseStatusEnum.WITHDRAWN:
         if CaseStatusEnum.is_terminal(original_status):
             return False
+    elif original_status == CaseStatusEnum.FINALISED:
+        if new_status != CaseStatusEnum.SURRENDERED:
```
efahy: From reading your story, shouldn't this be (in pseudo code terms) `if new status = surrendered and original status != finalised return false`?   Furthermore, I would expect the check in the API to include the bit from the story about the case needing to have one or more approve or proviso decisions on it especially since this is exporter side functionality, unless you have agreement from Anne to the contrary

efahy: Also, I'd expect this to be covered in a new API test somewhere

applications/libraries/application_helpers.py
```diff
@@ -22,6 +22,9 @@ def can_status_can_be_set_by_exporter_user(original_status: str, new_status: str
     if new_status == CaseStatusEnum.WITHDRAWN:
```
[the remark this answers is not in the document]
BenAndrew-Kainos: It would be but we also have lots of helpers and I'm not sure how they've been split up

applications/libraries/application_helpers.py
BenAndrew-Kainos: Firstly yes that's how the statement was written before but the logic actually breaks down because the last statement will evaluate FINALISED -> SURRENDERED as invalid so we have to make sure it doesn't go to that statement for FINALISED cases.

Secondly we can check that it has a duration (which is only present on approved/proviso applications) which is how the frontend also decided whether to show the button

Thirdly tests in progress :)

BenAndrew-Kainos: We've added the tests & licence duration check so I'll resolve :)

applications/views/applications.py
```diff
@@ -255,16 +255,24 @@ def put(self, request, pk):
...
         else:
             if not can_status_can_be_set_by_gov_user(request.user, application.status.status, data["status"]):
                 return JsonResponse(
-                    data={"errors": [strings.Applications.Finalise.Error.SET_STATUS]},
+                    data={"errors": [strings.Applications.Finalise.Error.GOV_SET_STATUS]},
                     status=status.HTTP_400_BAD_REQUEST,
                 )
 
+        if data["status"] == CaseStatusEnum.SURRENDERED:
+            if not application.licence_duration:
```
efahy: I would have thought you should incorporate the `licence_duration` check into the `can_status_se_set_by_gov_user` method - Its all closely related logic and will have the added advantage of making this `put` method not quite so long

applications/tests/tests_application_status.py
```diff
@@ -141,13 +137,67 @@ def test_exporter_set_application_status_failure(self, new_status):
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(self.standard_application.status, get_case_status_by_status(CaseStatusEnum.SUBMITTED))
 
+    def test_exporter_set_application_status_surrendered_success(self):
+        self.standard_application.status = get_case_status_by_status(CaseStatusEnum.FINALISED)
+        self.standard_application.licence_duration = 24
+        self.standard_application.save()
+
+        data = {"status": CaseStatusEnum.SURRENDERED}
+        response = self.client.put(self.url, data=data, **self.exporter_headers)
```
andytaylorkainos: response = self.client.put(self.url, data=data, **self.exporter_headers)


please assert the response body

PR #294, opened 2020-02-05
feat: LT-972 - copy application

Mark-Spiers-kainos:


a-bek:
Hey, I think the view needs to be much simpler, and the view should ideally not store state of an individual request. along the lines of:
```
def get(self, request, pk):
     copy = application.copy()
     serializer = ApplicationSerializer(copy)
     return JsonResponse(data={"data": serializer.data}, status=status.HTTP_201_CREATED)
```
Also, I think its a bad idea to edit the existing instance of the application you're copying (overwriting ids).

Could you retrieve all `copyable` data as dicts from an application instance, and its associated objects, Then save it from there?

Similar to parties:

https://github.com/uktrade/lite-api/blob/master/parties/models.py#L14
https://github.com/uktrade/lite-api/blob/master/applications/views/parties.py#L132

applications/models.py
```diff
@@ -134,7 +134,7 @@ class BaseApplication(ApplicationPartyMixin, Case):
 
 class StandardApplication(BaseApplication):
     export_type = models.CharField(choices=ApplicationExportType.choices, default=None, max_length=50)
-    reference_number_on_information_form = models.TextField(blank=True, null=True)
+    reference_number_on_information_form = models.CharField(blank=True, null=True, max_length=255)
```
Mark-Spiers-kainos: Text Field was slightly too big, considering James reckons it will usually be less than 30 characters. Can tighten in the future more when there is a better understanding

applications/views/applications.py
```diff
@@ -369,3 +382,142 @@ def get(self, request, pk):
...
+        data = request.data
+
+        serializer = GenericApplicationCopySerializer(
+            data=data, context={"application_type": old_application.application_type}
+        )
+
+        if not serializer.is_valid():
+            return JsonResponse(data={"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        # deepcopy so new_application is not a pointer to old_application
```
efahy: This comment seems to contradict the thing you do next, which is to remove the pks and ids so Django doesn't consider the new application to be a duplicate of the old.  Its a deep copy just because we want everything copied, or am I mistaken?

applications/views/applications.py
```diff
@@ -369,3 +382,142 @@ def get(self, request, pk):
...
+        self.strip_id_for_application_copy()
+
+        # replace the reference and have you been informed (if required) with users answer
+        self.new_application.name = request.data["name"]
+        self.new_application.have_you_been_informed = request.data.get("have_you_been_informed")
+        self.new_application.reference_number_on_information_form = request.data.get(
+            "reference_number_on_information_form"
+        )
+        self.new_application.status = get_case_status_by_status(CaseStatusEnum.DRAFT)
+        self.new_application.copy_of_id = self.old_application_id
```
efahy: So just to double check, applications are always copies of their immediate parent, unlike parties?

applications/views/applications.py
```diff
@@ -369,3 +382,142 @@ def get(self, request, pk):
...
+        self.remove_data_from_application_copy()
+
+        # need to save here to create the pk/id for relationships
+        self.new_application.save()
+
+        # create new foreign key connection using data from old application (this is for tables pointing to the case)
+        self.create_foreign_relations_for_new_application()
+
+        # get all parties connected to the application and produce a copy (and replace reference for each one)
+        self.update_parties_with_copies()
```
efahy: Might I propose that `duplicate_parties_on_new_application` is a more representative name?

applications/views/applications.py
```diff
@@ -369,3 +382,142 @@ def get(self, request, pk):
...
+                result.pk = None
+                result.id = None
+                result.application = self.new_application
+                # Some models listed above are not inheriting timestampable models,
+                # as such we need to ensure created_at exists
+                if getattr(result, "created_at", False):
+                    result.created_at = now()
+                result.save()
+
+        # GoodsType has more logic than above, such as listing the countries on the goodstype, and flags
```
efahy: Same as with parties, you might consider a separate function for goods types, since as you say its completely different logic from the relationships copied in the above loop.  Not a huge deal though if you wanna keep it the way it is

applications/tests/tests_copy_application.py
```diff
@@ -0,0 +1,389 @@
...
+    def reset_data_test(self):
+        self.assertEqual(self.copied_application.copy_of.id, self.original_application.id)
+        self.assertEqual(self.copied_application.status, get_case_status_by_status(CaseStatusEnum.DRAFT))
+        self.assertGreater(self.copied_application.created_at, self.original_application.created_at)
+        self.assertGreater(self.copied_application.updated_at, self.original_application.updated_at)
+
+    def good_on_application_test(self):
+        new_goods_on_app = self.copied_application.goods.all()
+        original_goods_on_app = self.original_application.goods.all()
+        for good_on_app in new_goods_on_app:
```
efahy: This would pass if there were no goods on the new application - please also test that all goods from the old app are on the new app.  Please apply this comment to all other applicable methods here

applications/tests/tests_copy_application.py
```diff
@@ -0,0 +1,389 @@
...
+        self.response_data = self.response.json()["data"]
+
+        self.assertEqual(self.response.status_code, status.HTTP_201_CREATED)
+        self.assertNotEqual(self.response_data, self.original_application.id)
+
+        self.copied_application = HmrcQuery.objects.get(id=self.response_data)
+
+        self.hmrc_enquiry_test()
+
+    def standard_application_test(self):
```
efahy: Perhaps stick an underscore in front of all these methods to help distinguish them from the actual test methods.  You might also consider names such as `_validate_standard_application` - I think this would be a clearer indication of what they're all doing

applications/views/applications.py
Mark-Spiers-kainos: The deep copy means that we have a a complete new reference to the application. If we make changes on old application, we won't see them on new_application now and vice versa. Without this, if we make changes to new_application, we could view them in old_application. The actually new object is achieved once the id's are removed and we save.

applications/views/applications.py
Mark-Spiers-kainos: `When the case has been created by copying another submitted case
Then I should see a link to the case it was copied from in the application header`
The story asks for the case it was copied from.

applications/views/applications.py
Mark-Spiers-kainos: I have updated the comment to give a bit more detail as well.

applications/tests/tests_copy_application.py
Mark-Spiers-kainos: I have inversed the test, since we want to test that all goods on the original app are on the new app. And then added a length test to ensure they have the same amount of goods

applications/serializers/generic_application.py
```diff
@@ -233,3 +233,32 @@ def validate(self, data):
         ):
             raise serializers.ValidationError(strings.Applications.Finalise.Error.DURATION_RANGE)
         return data
+
+
+class GenericApplicationCopySerializer(serializers.ModelSerializer):
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: Done

applications/views/applications.py
Mark-Spiers-kainos: Done

applications/serializers/generic_application.py
```diff
@@ -233,3 +233,41 @@ def validate(self, data):
...
+    """
+    Serializer for copying applications that can handle any type
```
stephanieboyd: minor - prefix type with application, just to be clear

PR #295, opened 2020-02-05
chore: Set dependency versions

phlptt:
* Set dependency versions based on what was in the current master Pipfile.lock
* Moved some dependencies from packages to dev packages
* got rid of a couple of packages that are no longer used.


Pipfile
```diff
@@ -4,34 +4,33 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-parameterized = "==0.7.0"
 black = "==19.10b0"
+bandit = "==1.6.2"
+coverage = "==5.0.3"
```
efahy: As discussed on a call, I suggest we only fix the major version for all packages, to allow for potential minor fixes of those major versions, for for this example, `coverage = "==5.0.3"` becomes `coverage = "~=5.0"` (I believe Pipenv supports this but please double check the syntax etc!)

phlptt: Cheers that's updated other than black which seems to have b0 appended to their versions.

PR #296, opened 2020-02-05
LT-1658: refactor parties

a-bek:
Changes

1) models.py
```
Added: Party

Removed: BaseParty, Consignee, EndUser, ThirdParty, UltimateEndUser
```
2) serializers.py
```
Added: PartySerializer

Removed: PartySerializer, EndUserSerializer, EndUserWithFlagsSerializer, UltimateEndUserSerializer, UltimateEndUserWithFlagsSerializer ,ConsigneeSerializer ,ConsigneeWithFlagsSerializer, ThirdPartySerializer, ThirdPartyWithFlagsSerializer
```

All party urls squashed to: 
```
applications/<pk>/parties/<party_pk>/
```

3) introduction of PartyOnApplication model to store relationship between a party and application + meta information. All foreign key relationships have been moved from the application models.
4) Introduced application model and serializer mixins that allow management of parties
5) Deleting a party now timestamps `deleted_at` on PartyOnApplication model.
6) Party API squashed from ~15 urls to 3, reducing the communication complexity

Note:

- Migration files removed from PR to remove noise
- unit test pass
- exporter/internal e2e pass


applications/views/party_documents.py
```diff
@@ -35,138 +24,39 @@ class EndUserDocumentView(APIView):
...
-    @transaction.atomic
-    @allowed_application_types(
-        [ApplicationType.STANDARD_LICENCE, ApplicationType.HMRC_QUERY, ApplicationType.EXHIBITION_CLEARANCE]
-    )
-    @authorised_users(ExporterUser)
-    def delete(self, request, application, tp_pk):
-        third_party = get_third_party(tp_pk)
-        return delete_party_document(third_party, application, request.user)
+    def delete(self, request, application, party_pk):
+        try:
```
andytaylorkainos: can you move all of the try-except duplicated code into a method please

parties/serializers.py
```diff
@@ -33,9 +50,21 @@ class Meta:
...
             "sub_type",
+            "role",
+            "flags",
             "copy_of",
+            "deleted_at",
         )
 
+    def __init__(self, *args, **kwargs):
+        super(PartySerializer, self).__init__(*args, **kwargs)
+        if "data" in kwargs and "type" in kwargs["data"]:
```
andytaylorkainos:
```
data = kwargs.get("data")

if data:
    party_type = data.get("type)
    if party_type:
        ...
```

parties/models.py
```diff
@@ -22,34 +22,11 @@ class Party(TimestampableModel):
     )
     flags = models.ManyToManyField(Flag, related_name="parties")
     sub_type = models.CharField(choices=SubType.choices, default=SubType.OTHER, max_length=20)
+    role = models.CharField(choices=PartyRole.choices, default=PartyRole.OTHER, max_length=22, null=True)
```
Mark-Spiers-kainos: can there be a comment so we know this field is just for Third parties?

parties/serializers.py
```diff
@@ -58,125 +87,8 @@ def validate_website(value):
...
-    flags = serializers.SerializerMethodField()
-
-    def get_flags(self, instance):
-        return list(instance.flags.values("id", "name"))
-
-    class Meta:
-        model = ThirdParty
-        fields = "__all__"
+        docs = PartyDocument.objects.filter(party=instance)
+        return docs.values()[0] if docs.exists() else None
```
andytaylorkainos: docs.first().values()

applications/tests/tests_hmrc_query_submit.py
```diff
@@ -0,0 +1,93 @@
+from applications.models import PartyOnApplication
```
Mark-Spiers-kainos: a lot of the tests in this file are in `applications/tests/tests_hmrc_query.py` currently

parties/serializers.py
```diff
@@ -5,21 +5,38 @@
...
 from lite_content.lite_api.strings import Parties
 from organisations.models import Organisation
+from parties.enums import PartyType, SubType, PartyRole
+from parties.models import Party
 from parties.models import PartyDocument
-from parties.enums import PartyType, SubType, ThirdPartyRole
-from parties.models import Party, EndUser, UltimateEndUser, Consignee, ThirdParty
+
+
+class FlagSerializer(serializers.Serializer):
```
Mark-Spiers-kainos: could this be in the `flag/serializers` to keep it with the other Flag related serializers? We also have another class called FlagSerializer so perhaps another name relating to how if only gives minimalist data

parties/serializers.py
```diff
@@ -5,21 +5,38 @@
...
+    name = serializers.CharField()
+    id = serializers.CharField()
+
+
+class CountrySerializer(serializers.Serializer):
```
Mark-Spiers-kainos: as far as I can tell this serializer is already in `static/countries/serializers`

applications/models.py
```diff
@@ -1,21 +1,130 @@
...
+        old_poa = None
+
+        if PartyOnApplication.objects.filter(party=party).exclude(application=self).exists():
+            # Party cannot be used in multiple applications
+            raise ApplicationException({"error": "Party exists"})
+
+        if party.type == PartyType.ULTIMATE_END_USER:
+            # Rule: Append
+            pass
+
```
andytaylorkainos: please remove the blank lines between the `elif` conditionals

applications/tests/tests_hmrc_query_submit.py
```diff
@@ -0,0 +1,93 @@
...
+        self.assertEqual(case.id, self.draft.id)
+        self.assertIsNotNone(case.submitted_at)
+        self.assertEqual(case.status.status, CaseStatusEnum.SUBMITTED)
+        self.assertEqual(case.type, CaseTypeEnum.HMRC_QUERY)
+
+    def test_submit_hmrc_query_with_invalid_id_failure(self):
+        draft_id = "90D6C724-0339-425A-99D2-9D2B8E864EC7"
+        url = "applications/" + draft_id + "/submit/"
+
+        response = self.client.put(url, **self.hmrc_exporter_headers)
```
andytaylorkainos: we decided as a team to assert response data; please assert the body returned from the API

applications/views/applications.py
```diff
@@ -108,8 +108,11 @@ def get(self, request, application):
         Retrieve an application instance
         """
         serializer = get_application_view_serializer(application)
-        serializer = serializer(application, context={"exporter_user": request.user})
-        return JsonResponse(data=serializer.data, status=status.HTTP_200_OK)
+        data = serializer(application, context={"exporter_user": request.user}).data
+
+        data["is_major_editable"] = application.is_major_editable()
```
andytaylorkainos: why is this being injected in the view and not the serializer?

parties/enums.py
```diff
@@ -1,16 +1,32 @@
...
-        (THIRD, "Third Party"),
+        (END_USER, "End User"),
+        (ULTIMATE_END_USER, "Ultimate End User"),
+        (THIRD_PARTY, "Third Party"),
     ]
 
+    @classmethod
+    def plural(cls, key):
+        return {
+            PartyType.CONSIGNEE: "consignees",
```
andytaylorkainos: do we ever have multiple consignees/end users?

parties/enums.py
```diff
@@ -1,16 +1,32 @@
...
+            PartyType.END_USER: "end_users",
+            PartyType.ULTIMATE_END_USER: "ultimate_end_users",
+            PartyType.THIRD_PARTY: "third_parties",
+        }[key]
+
+    @classmethod
+    def api_compatible(cls, party_type):
+        is_plural = party_type in [PartyType.ULTIMATE_END_USER, PartyType.THIRD_PARTY]
```
andytaylorkainos: seems strange to check if the party type is in this list and then also try to index it into the `plural` method.

would it not be better to just return a static dict of pluralised applications from `plural` and use .get(party_type) from that returned dict?

parties/models.py
andytaylorkainos: Is there something in the serializer restricting the role being set for other party_types?

queries/end_user_advisories/models.py
```diff
@@ -11,7 +11,7 @@ class EndUserAdvisoryQuery(Query):
     Query into ensuring that an end user is valid
     """
 
-    end_user = models.ForeignKey(EndUser, on_delete=models.DO_NOTHING, null=False, related_name="eua_query")
+    end_user = models.ForeignKey(Party, on_delete=models.DO_NOTHING, null=False, related_name="eua_query")
```
Mark-Spiers-kainos: Is there any logic that enforces the correct Party is attached to each type of Case? That functionality was built-in before.

applications/models.py
```diff
@@ -1,21 +1,130 @@
...
 
-class BaseApplication(Case):
+class ApplicationException(Exception):
+    def __init__(self, data=None, *args, **kwargs):
+        super().__init__(data)
+        self.data = data
+
+
+class ApplicationPartyMixin:
+    def add_party(self, party):
```
efahy: There is nothing in this method or anywhere else in the PR that I can see that imposes the restrictions we used to have implicitly with the existing application models as regards which party types can exist on which application types, e.g. you shouldn't be able to add party of any type to an open application and you can only have one end_user on a standard application and so on.  This validation, for all application types, is something we have to have in the API.

efahy: And since these restrictions are now going to be in application code, you'll need to cover it all with API tests

applications/models.py
```diff
@@ -1,21 +1,130 @@
...
+        )
+
+    @property
+    def consignee(self):
+        """
+        Backwards compatible
+        Standard and HMRC Query applications
+        """
+        try:
+            return self.active_parties.get(party__type=PartyType.CONSIGNEE)
```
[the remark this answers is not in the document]
a-bek: DoesNotExist is an expected error.

However, we want it to fail if we have multiple consignees active on the application which this will.

parties/serializers.py
a-bek: I just retrieved the party_type in one line to avoid 4 nested ifs.

parties/models.py
a-bek: The PartySerializer requires the `type` field, but not `role`, upon initialisation if the `type  == third_party` then it dynamically updates the `role` field to required.

applications/models.py
a-bek: The restrictions are covered in this method.

Its done in the `party.type in [PartyType.END_USER, PartyType.CONSIGNEE]` if block

If an old consignee or end user exists. it deletes it before proceeding to add the new one.

parties/serializers.py
a-bek: `first()` returns the model instance, `values()` is a queryset.

applications/models.py
efahy: Ok I see.  Can you explain that with more that a 1 word comment please since its pretty central to the whole application?

And you still need a restriction preventing addition of parties to open applications, right?

applications/models.py
```diff
@@ -1,21 +1,130 @@
...
+    def all_parties(self):
+        return self.parties.prefetch_related("party__flags").select_related(
+            "party", "party__organisation", "party__country"
+        )
+
+    @property
+    def consignee(self):
+        """
+        Backwards compatible
+        Standard and HMRC Query applications
```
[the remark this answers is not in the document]
a-bek: Yep, works for exhibition clearances out of the box.

parties/enums.py
a-bek: No, I have changed it to just `api_compatible_name` to get rid of the notion of `plural`

queries/end_user_advisories/models.py
a-bek: The view for query explicitly sets the type to end_user.

applications/tests/tests_hmrc_query_submit.py
a-bek: Yes, I'm slightly confused as I'm not sure how this file got in, it must have snuck in somehow from a merge or something...

PR #299, opened 2020-02-10
Bug: LT-2040 manage organisation (user) page performance

Mark-Spiers-kainos:
Minimise the amount of data being passed forward to the page, increasing the speed greatly. Deployed on QA.

### DIFF
--- a/gov_users/serializers.py
+++ b/gov_users/serializers.py
@@ -27,7 +27,10 @@ class PermissionSerializer(serializers.ModelSerializer):
 class RoleNameSerializer(serializers.ModelSerializer):
     class Meta:
         model = Role
-        fields = ("id", "name")
+        fields = (
+            "id",
+            "name",
+        )
 
 
 class RoleSerializer(serializers.ModelSerializer):
--- a/gov_users/serializers.py
+++ b/gov_users/serializers.py
@@ -24,6 +24,12 @@ class PermissionSerializer(serializers.ModelSerializer):
         )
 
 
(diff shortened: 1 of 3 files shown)


gov_users/serializers.py
```diff
@@ -24,6 +24,12 @@ class Meta:
         )
 
 
+class RoleNameSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Role
+        fields = ("id", "name")
```
andytaylorkainos: 👀 comma at the end of the tuple please

organisations/serializers.py
```diff
@@ -213,3 +214,17 @@ class ExternalLocationSerializer(serializers.ModelSerializer):
...
         fields = ("id", "name", "address", "country", "organisation")
+
+
+class OrganisationUserListView(serializers.ModelSerializer):
+    status = serializers.CharField()
+    role = serializers.SerializerMethodField()
+
+    class Meta:
+        model = ExporterUser
+        fields = ("id", "first_name", "last_name", "email", "role", "status")
```
andytaylorkainos: comma

gov_users/serializers.py
jfaracik: Is there anyway we could get Black to do this? Purpose of Black was to avoid conversations like this

jfaracik:
> Minimise the amount of data being passed forward to the page, increasing the speed greatly. Deployed on QA.

How much faster will the user's page be after this change?

PR #301, opened 2020-02-10
Bug: LT-2066 - editing admin team

Mark-Spiers-kainos:


teams/helpers.py
```diff
@@ -7,3 +7,7 @@ def get_team_by_pk(pk):
         return Team.objects.get(pk=pk)
     except Team.DoesNotExist:
         raise NotFoundError({"team": "Team not found - " + str(pk)})
+
+
+def get_admin_team():
+    return Team.objects.get(pk="00000000-0000-0000-0000-000000000001")
```
efahy: Please create a constant for the pk in `conf/constants.py`

teams/views.py
```diff
@@ -74,6 +74,8 @@ def put(self, request, pk):
         """
         Update a team instance
         """
+        if pk == get_admin_team().id:
```
efahy: Correct me if I'm wrong, but you don't need to go to the DB for this, you can just do `if pk == "000...001"`, right?

teams/helpers.py
efahy: Also if you decide to keep this method (see other comment about not needing to go to the DB), there's a `get_team_by_pk` method you can call from here

efahy: In fact, its just above this one

conf/constants.py
```diff
@@ -32,3 +32,7 @@ class Roles:
     IMMUTABLE_ROLES = [INTERNAL_SUPER_USER_ROLE_ID, EXPORTER_SUPER_USER_ROLE_ID, EXPORTER_DEFAULT_ROLE_ID]
 
     EXPORTER_PRESET_ROLES = [EXPORTER_SUPER_USER_ROLE_ID, EXPORTER_DEFAULT_ROLE_ID]
+
+
+class Teams:
+    ADMIN_TEAM_ID = "00000000-0000-0000-0000-000000000001"
```
a-bek: This is also in `conf/settings`

Perhaps import the id from there?

PR #304, opened 2020-02-11
feat: LT-1229 - Anti-piracy 

jfaracik:
* Added goods_categories field for StandardApplications
* Added dependency `django-separatedvaluesfield` to store multi choice lists
* Added tests
* EXTRA: Added logo to readme


applications/tests/tests_create_application.py
```diff
@@ -39,6 +40,43 @@ def test_create_draft_standard_application_successful(self):
...
+            "name": "Test",
+            "application_type": ApplicationType.STANDARD_LICENCE,
+            "export_type": ApplicationExportType.TEMPORARY,
+            "goods_categories": [GoodsCategory.ANTI_PIRACY, GoodsCategory.FIREARMS],
+            "have_you_been_informed": ApplicationExportLicenceOfficialType.YES,
+            "reference_number_on_information_form": "123",
+        }
+
+        response = self.client.post(self.url, data, **self.exporter_headers)
+        standard_application = StandardApplication.objects.get()
```
Mark-Spiers-kainos: does the response not contain the details about this standard application? if so could test it against that

applications/models.py
```diff
@@ -138,6 +144,9 @@ class StandardApplication(BaseApplication):
     have_you_been_informed = models.CharField(
         choices=ApplicationExportLicenceOfficialType.choices, default=None, max_length=50,
     )
+    goods_categories = SeparatedValuesField(
```
Mark-Spiers-kainos: My understanding of this field is that, it takes a list of items, and joins them together by a separator, can you confirm?

jfaracik: Completely correct

applications/tests/tests_create_application.py
jfaracik: For the post only the ID is needed (to navigate to the correct task list) so I don't think it'd be necessary to check all the other values.

applications/serializers/standard_application.py
```diff
@@ -17,6 +18,16 @@ class StandardApplicationViewSerializer(PartiesSerializerMixin, GenericApplicati
     goods = GoodOnApplicationViewSerializer(many=True, read_only=True)
     destinations = serializers.SerializerMethodField()
     additional_documents = serializers.SerializerMethodField()
+    goods_categories = serializers.SerializerMethodField()
+
+    def get_goods_categories(self, instance):
+        if not instance.goods_categories:
```
a-bek: Could do the below and delete this condition?
```
return_value = [......... for x in instance.goods_categories or []]
```

PR #305, opened 2020-02-13
Lt 1177 users filter pagination

a-bek:
- Return gov_users filter from API rather than making a second request on internal frontend.
- Allow optional pagination on users endpoint
- Allow filter users by status.


users/tests/tests_exporter_roles_and_permissions.py
```diff
@@ -33,6 +33,8 @@ def test_get_list_of_all_roles_as_exporter_super_user(self):
         response = self.client.get(url, **self.exporter_headers)
         response_data = response.json()
 
+        print(response_data)
```
iichr: 👀

gov_users/views/gov_user_views.py
```diff
@@ -60,21 +60,19 @@ def post(self, request, *args, **kwargs):
         return JsonResponse(data={"token": token, "lite_api_user_id": str(user.id)})
 
 
-class GovUserList(APIView):
+class GovUserList(generics.ListCreateAPIView):
     authentication_classes = (GovAuthentication,)
 
     def get(self, request):
         """
         Fetches all government users
```
Mark-Spiers-kainos: update the docstring to be more representative of what the code is doing please

gov_users/views/gov_user_views.py
```diff
@@ -60,21 +60,19 @@ def post(self, request, *args, **kwargs):
         return JsonResponse(data={"token": token, "lite_api_user_id": str(user.id)})
 
 
-class GovUserList(APIView):
+class GovUserList(generics.ListCreateAPIView):
```
Mark-Spiers-kainos: A lot of this code below seems to be replicating the built-in functionality of the class get method. A more standard approach to using `generics.ListCreateAPIView` for paging of results is in `ExistingParties` view.

cases/views/search/views.py
```diff
@@ -38,11 +42,18 @@ def get(self, request, *args, **kwargs):
         case_types = service.get_case_type_list()
         queue = next(q for q in queues if q["id"] == queue_id)
 
+        gov_users = [
+            {"key": full_name.lower(), "value": full_name}
```
BenAndrew-Kainos: Is there a point making a key value pair where the key is essentially the same as the value? (If this needs to happen can it be done on the frontend)

gov_users/tests/tests_view.py
```diff
@@ -36,18 +36,30 @@ def test_filter_users_by_multiple_teams(self):
...
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
 
+        self.assertEqual(len(response_data["results"]["gov_users"]), self.gov_user_preexisting_count)
+
+    def test_get_deactivated_users(self):
+        self.user.status = UserStatuses.DEACTIVATED
+        self.user.save()
+
+        self.url = reverse("gov_users:gov_users") + "?status="
```
BenAndrew-Kainos: Does an empty status mean deactivated? (if so I don't think that's a good idea because it means if they don't submit a value for that filter it will come back with deactivated users)

gov_users/views/gov_user_views.py
```diff
@@ -84,8 +82,13 @@ def get(self, request):
...
 
-        serializer = GovUserViewSerializer(gov_users_qs.order_by("email"), many=True)
-        return JsonResponse(data={"gov_users": serializer.data})
+        if "no_page" in request.GET:
+            return JsonResponse(data={"results": {"gov_users": GovUserViewSerializer(gov_users_qs, many=True).data}})
+
+        page = self.paginate_queryset(gov_users_qs)
+        serializer = GovUserViewSerializer(page, many=True)
+
+        return self.get_paginated_response({"gov_users": serializer.data})
```
BenAndrew-Kainos: You shouldn't need to define the get for this only `get_queryset` and define the serializer in the class (see LetterTemplatesList as an example)

organisations/serializers.py
```diff
@@ -232,6 +232,9 @@ class Meta:
         )
 
     def get_role(self, instance):
+        if instance.role:
+            return RoleNameSerializer(instance.role).data
```
BenAndrew-Kainos: You can just point this field to the serializer like `role = RoleNameSerializer()`

organisations/views/users.py
```diff
@@ -21,20 +21,28 @@
...
+        user_relationships = get_user_organisation_relationships(org_pk, status)
+        page = self.paginate_queryset(user_relationships)
+
+        for p in page:
+            p.user.status = p.status
+            p.user.role = p.role
+
+        serializer = OrganisationUserListView([p.user for p in page], many=True, context=org_pk)
+
+        return self.get_paginated_response({"users": serializer.data})
```
BenAndrew-Kainos: As above use the standard pagination process and move most of this to `get_queryset`

users/enums.py
```diff
@@ -7,6 +7,13 @@ class UserStatuses:
         (DEACTIVATED, "Deactivated"),
     ]
 
+    @classmethod
+    def from_string(cls, s):
+        return {
+            UserStatuses.ACTIVE.lower(): UserStatuses.ACTIVE,
```
BenAndrew-Kainos: Another key value pair that probably isn't needed

cases/views/search/views.py
a-bek: Yes, as the frontend filters are using key/value pairs, this is consistent with current filters that are returned (case_type and statuses). We also then have all the filters returned from this view where it logically belongs (the gov_user filter was originally a separate call to the API on the frontend.)

gov_users/tests/tests_view.py
a-bek: An empty status means all users.

`status=active` is default
`status=` is for all
`status=deactivated` can be used in future out of the box, but outside scope in terms of testing and filtering in this PR.

gov_users/views/gov_user_views.py
a-bek: The View is there to be able to be extended, and as we have optional pagination, we need to override `get()`

gov_users/views/gov_user_views.py
a-bek: The View is there to be able to be extended, and as we have optional pagination, we need to override `get()`

organisations/serializers.py
a-bek: nice!

cases/views/search/views.py
BenAndrew-Kainos: If this is to make it consistent then that's OK and I agree multiple API calls isn't a good idea.

gov_users/tests/tests_view.py
BenAndrew-Kainos: Could we test this please. Although you may not use it yet if you make the endpoint support active then it is implied it will support deactivated and may catch out devs. Also your test is called `test_get_deactivated_users` which is misleading if what you are actually doing is getting all

gov_users/views/gov_user_views.py
BenAndrew-Kainos: For performance & standardisation of the return body from the API paginated endpoints should be consistent. Django page pagination (which we implement) defaults page number to 1 when not passed and requires that GET param to be called `page`. If you'd like me to explain further give me a call :)

users/enums.py
BenAndrew-Kainos: Could you explain why this was ignored?

gov_users/views/gov_user_views.py
a-bek: Yes, but we are making pagination optional on this view. `no_page` is a separate parameter which tells the view if we want to paginate or not. If `no_page` is passed, we return the full queryset:

```JsonResponse(data={"results": {"gov_users": GovUserViewSerializer(gov_users_qs, many=True).data}})```

If `no_page` is not passed, the view continues with the standard way of paginating using `page`:

```return self.get_paginated_response({"gov_users": serializer.data})```

users/enums.py
a-bek: I resolved it along with the other key value comment you made.

This function just handles capitalisation of the `status` parameter.

gov_users/views/gov_user_views.py
BenAndrew-Kainos: But what I'm saying is that the pagination should not be optional. 

> For performance & standardisation of the return body from the API, paginated endpoints > should be consistent.

organisations/views/users.py
a-bek: It is using the standard pagination process, but as I mentioned before, because we are making pagination optional, we override the `get`

users/enums.py
BenAndrew-Kainos: I didn't realise you needed this here too. I'm not a fan of this but if you say it's consistent then that's ok.

gov_users/views/gov_user_views.py
a-bek: But we need it to be optional because the frontend has paginated and non-paginated requirements

organisations/views/users.py
a-bek: Cheers Ben sorry for confusion.

gov_users/views/gov_user_views.py
a-bek: Filters now on frontend.

PR #306, opened 2020-02-13
LT-2045: Bug - fix case queue ordering

stephanieboyd:
* Add submitted_at to a goods query


queries/goods_query/views.py
```diff
@@ -76,6 +78,7 @@ def post(self, request):
             organisation=data["organisation"],
             type=CaseTypeEnum.GOODS_QUERY,
             status=get_case_status_by_status(CaseStatusEnum.SUBMITTED),
+            submitted_at=datetime.now(timezone.utc),
```
a-bek: https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.timezone.now

Could you use django datetime?

stephanieboyd: Sure, changed

PR #308, opened 2020-02-17
feat: LT-1249 - Add note to a draft application

jfaracik:
* Now possible to return draft cases
* _FINALLY_ Added case note content to strings.py
* `create_standard_application` renamed to `create_draft_standard_application`


jfaracik:
> pretty sure the intention of this helper method was for internal-side usage (i.e. dont return drafts cases)

Yeah Chris and I agree. He's adding a parameter which is just going to be used on the `case-notes` endpoint so that `get_case` works with drafts

PR #311, opened 2020-02-19
feat: LT-990 Case SLA

BenAndrew-Kainos:
Creates the SLA Updater background task & Case fields

**New fields;**
- sla_days: Number of days since case was created
- sla_remaining_days: Number of days left to process to meet SLA goal
- last_closed_at: Last time case was closed. Needed becuase previously closed cases don't update SLA
- sla_updated_at: Last time SLA was recalculated. used to ensure we don't recalculate twice a day

**New key functions;**
- update_cases_sla: Background task run once a day to update cases SLA.
- get_application_target_sla: Get the goal days for a given application type
- get_bank_holidays: Use the GOV API to get bank holidays & save a local backup


cases/models.py
```diff
@@ -59,6 +59,11 @@ class Case(TimestampableModel):
     )
     case_officer = models.ForeignKey(GovUser, null=True, on_delete=models.DO_NOTHING)
     copy_of = models.ForeignKey("self", default=None, null=True, on_delete=models.DO_NOTHING)
+    last_closed_at = models.DateTimeField(blank=True, null=True)
+
+    sla_days = models.IntegerField(null=False, blank=False, default=0)
```
Mark-Spiers-kainos: These are very big fields in terms of length. They also include negative numbers. Perhaps look at SmallIntegerField/PositiveSmallIntegerField
https://docs.djangoproject.com/en/3.0/ref/models/fields/#positivesmallintegerfield
https://docs.djangoproject.com/en/3.0/ref/models/fields/#smallintegerfield

BenAndrew-Kainos: Great suggestion!

cases/models.py
```diff
@@ -59,6 +59,11 @@ class Case(TimestampableModel):
     )
     case_officer = models.ForeignKey(GovUser, null=True, on_delete=models.DO_NOTHING)
     copy_of = models.ForeignKey("self", default=None, null=True, on_delete=models.DO_NOTHING)
+    last_closed_at = models.DateTimeField(blank=True, null=True)
+
+    sla_days = models.PositiveSmallIntegerField(null=False, blank=False, default=0)
+    sla_remaining_days = models.SmallIntegerField(null=True, blank=True)
+    sla_updated_at = models.DateTimeField(blank=True, null=True)
```
stephanieboyd: Is `blank=True` necessary?

cases/sla.py
```diff
@@ -0,0 +1,108 @@
...
+
+from cases.enums import CaseTypeSubTypeEnum
+from cases.models import Case
+
+SLA_UPDATE_TASK_TIME = time(0, 0, 0)
+SLA_UPDATE_CUTOFF_TIME = time(18, 0, 0)
+BANK_HOLIDAY_API = "https://www.gov.uk/bank-holidays.json"
+BACKUP_FILE_NAME = "bank-holidays.csv"
+
+STANDARD_APPLICATION_TARGET = 20
```
stephanieboyd: It could be clearer that this is a target in days

cases/sla.py
```diff
@@ -0,0 +1,108 @@
...
+def get_bank_holidays():
+    """
+    Uses the GOV bank holidays API.
+    If it can connect to the API, it extracts the list of bank holidays,
+    saves a backup of this list as a CSV and returns the list.
+    If it cannot connect to the service it will use the CSV backup and returns the list.
+    """
+    data = []
+    r = requests.get(BANK_HOLIDAY_API)
+    if r.status_code != 200:
```
stephanieboyd: HTTP status, please

cases/sla.py
```diff
@@ -0,0 +1,108 @@
...
+                        sla_days=F("sla_days") + 1, sla_remaining_days=F("sla_remaining_days") - 1, sla_updated_at=date
+                    )
+                )
+                logging.info(f"SLA Update Successful: Updated {results} cases")
+                return results
+        except Exception as e:  # noqa
+            logging.error(e)
+            return False
+
+    logging.info("SLA Update Ignored: Non-working day")
```
stephanieboyd: Super minor - not a fan that the message is  `ignored`. Perhaps `not performed`

cases/models.py
BenAndrew-Kainos: I don't think so

cases/sla.py
BenAndrew-Kainos: Done

cases/sla.py
BenAndrew-Kainos: Done

cases/sla.py
BenAndrew-Kainos: Done

PR #315, opened 2020-02-21
feat: LT-2149 - Add PV Grading and CLC to list of available goods query states

iichr:
- added two additional states for goods queries - PV review and CLC review

https://uktrade.atlassian.net/browse/LT-2149

### DIFF
--- a/cases/tests/tests_search_cases.py
+++ b/cases/tests/tests_search_cases.py
@@ -26,7 +26,7 @@ class FilterAndSortTests(DataTestClient):
         app_statuses = [
             app_status[0]
             for app_status in CaseStatusEnum.choices
-            if app_status[0] not in [CaseStatusEnum.CLC, CaseStatusEnum.PV]
+            if app_status[0] not in CaseStatusEnum.goods_query_statuses
         ]
         self.application_cases = []
         for app_status in app_statuses:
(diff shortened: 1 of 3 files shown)


cases/tests/tests_search_cases.py
```diff
@@ -22,10 +22,15 @@ def setUp(self):
         super().setUp()
         self.url = reverse("cases:search")
 
+        app_statuses = [
+            app_status[0]
+            for app_status in CaseStatusEnum.choices
+            if app_status[0] not in [CaseStatusEnum.CLC, CaseStatusEnum.PV]
```
iichr: Excluding the two new states which are only applicable to goods queries

static/statuses/enums.py
```diff
@@ -22,6 +22,8 @@ class CaseStatusEnum:
     UNDER_FINAL_REVIEW = "under_final_review"
     UNDER_REVIEW = "under_review"
     WITHDRAWN = "withdrawn"
+    CLC = "clc_review"
+    PV = "pv_review"
```
iichr: Not really case statuses but neither are closed/submitted/withdrawn, perhaps they don't warrant a good specific status enum

cases/tests/tests_search_cases.py
Mark-Spiers-kainos: could you add this as a comment in the code?

iichr: Yeah sure

cases/tests/tests_search_cases.py
```diff
@@ -22,10 +22,15 @@ def setUp(self):
         super().setUp()
         self.url = reverse("cases:search")
 
+        app_statuses = [
+            app_status[0]
+            for app_status in CaseStatusEnum.choices
```
BenAndrew-Kainos: Could we instead make a new list in CaseStatusEnum called something like ApplicationStatuses which has the correct list with PV & CLC removed?

BenAndrew-Kainos: This would be more typical in the system at the moment

iichr: @BenAndrew-Kainos Would it be worth duplicating the enum - 23 status for applications only, just to exclude the 2?

BenAndrew-Kainos: No I mean just define this in the enum instead

BenAndrew-Kainos: It'll be easier to maintain & reuse that way

PR #316, opened 2020-02-24
feat: LT-1594 - Application and party clearance levels

a-bek:
- F380 application requires clearance level
- F380 parties require clearance level


applications/views/parties.py
```diff
@@ -51,8 +51,9 @@ def post(self, request, application):
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
+        serializer = PartySerializer(data=data, application_type=application.case_type.sub_type)
+
         # Validate data
-        serializer = PartySerializer(data=data)
```
iichr: could replace the removed line with your new serialiser ?

cases/views/views.py
```diff
@@ -60,9 +60,9 @@ def get(self, request, pk):
         Retrieve a case instance
         """
         case = get_case(pk)
-        serializer = CaseDetailSerializer(case, user=request.user, team=request.user.team)
+        serializer = CaseDetailSerializer(case, user=request.user, team=request.user.team).data
 
-        return JsonResponse(data={"case": serializer.data}, status=status.HTTP_200_OK)
```
iichr: perhaps a bit inconsistent with the other `get/put`s in the file, then again they make of use of more than the data in the serialiser

applications/tests/tests_mod_clearance_submit.py
```diff
@@ -204,7 +204,6 @@ def setUp(self):
     def test_submit_F680_clearance_success(self):
         response = self.client.put(self.url, **self.exporter_headers)
         application = F680ClearanceApplication.objects.get()
-
```
iichr: I'm partial to a line separating the setup and the assertions

applications/tests/tests_edit_application.py
```diff
@@ -176,3 +178,95 @@ def test_edit_application_name_in_read_only_status_failure(self, read_only_statu
...
+    def test_add_party_no_clearance_to_f380_failure(self):
+        party = {
+            "type": PartyType.THIRD_PARTY,
+            "name": "Government of Paraguay",
+            "address": "Asuncion",
+            "country": "PY",
+            "sub_type": "government",
+            "website": "https://www.gov.py",
+            "role": "agent",
+        }
```
iichr: Since the party's reused, do you reckon it will be more concise to define it outside of the scope of this method and just append `"clearance_level"` to it in the previous test, if that is the only difference?

cases/views/views.py
a-bek: Changes reverted

applications/tests/tests_edit_application.py
a-bek: As discussed, fine in the tests :)

applications/creators.py
```diff
@@ -165,6 +165,16 @@ def _validate_has_goods(draft, errors, is_mandatory):
     return errors
 
 
+def _validate_has_clearance_level(draft, errors, is_mandatory):
```
cstobart: have an optional parameter for the level itself?

applications/creators.py
```diff
@@ -165,6 +165,16 @@ def _validate_has_goods(draft, errors, is_mandatory):
...
+    """ Checks draft has clearance level """
+
+    if is_mandatory:
+        if not draft.clearance_level:
```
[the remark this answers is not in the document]
a-bek: Yeah the clearance levels are based on the PV gradings.

applications/creators.py
a-bek: As discussed, not necessary!

PR #318, opened 2020-02-25
Various ui bugs

BenAndrew-Kainos:
LT-2205 (Assign case officer)
Lt-2206 (Hide deactivated users)


gov_users/views/gov_user_views.py
```diff
@@ -66,11 +66,16 @@ class GovUserList(generics.ListCreateAPIView):
 
     def get_queryset(self):
         teams = self.request.GET.get("teams")
-        status = self.request.GET.get("status")
+        activated = self.request.GET.get("activated")
```
a-bek: Why has this been changed to a boolean when we are filtering by the user status?

BenAndrew-Kainos: Because that's how it was being used in all places except one on the frontend. If it's True is shows activated only, if it's False it shows deactivated and shows all if None.

BenAndrew-Kainos: It makes sense to be honest. I'm even considering bringing up a discussion to make the field a nullable boolean (would be more optimal)

a-bek: Ok, but we are filtering by UserStatus, so would you not agree that using `activated` is just adding another level of indirection here as you are essentially just converting it to `UserStatus`. The existing logic is also cleaner.

iichr: Quite minor but showing deactivated only has not been in scope, it's either all or active - https://uktrade.atlassian.net/browse/LT-1177 .

gov_users/views/gov_user_views.py
```diff
@@ -66,11 +66,16 @@ class GovUserList(generics.ListCreateAPIView):
...
         full_name = self.request.GET.get("name")
         gov_users_qs = GovUser.objects.all().order_by("email")
-        if status:
-            gov_users_qs = gov_users_qs.filter(status=UserStatuses.from_string(status))
+        if activated is not None and activated != "":
+            activated = str_to_bool(activated)
```
andytaylorkainos: does `activated = str_to_bool(activated)` not cover the previous line also?

gov_users/views/gov_user_views.py
BenAndrew-Kainos: Ok will change it back.

BenAndrew-Kainos: Actually nevermind this basically voids everything I've done. I'll either start again or drop the bug

a-bek: Can't you just update the calls to the view to pass the status rather than a bool?

gov_users/views/gov_user_views.py
BenAndrew-Kainos: All of this code has been removed

BenAndrew-Kainos:
All changes to the users filter have been removed

PR #319, opened 2020-02-26
bug: LT-2200 - Finalising MOD licence throws an API error

iichr:
- reason for this being MOD clearance applications (F680, Exhibition and Gifting) not having an export type, so no default licence duration was being returned when the advice was finalised
  - added a check for whether the application type does not use an export type - assigning temporary duration if that's the case

_This is to unblock LT-1168._
Forms part of https://uktrade.atlassian.net/browse/LT-2200 
NOTE: this does not address the finalised advice not showing on the exporter's side bug


iichr:
Deployed successfully on devpopcorn

PR #321, opened 2020-02-28
chore: seed team, organisation and role with user

andytaylorkainos:
This PR is to allow seeding to be more accessible:

You can now specify what team/organisation/role combination each user has in the environment variable.


conf/constants.py
```diff
@@ -26,9 +26,13 @@ class ExporterPermissions(Enum):
 
 class Roles:
     INTERNAL_DEFAULT_ROLE_ID = UUID("00000000-0000-0000-0000-000000000001")
+    INTERNAL_DEFAULT_ROLE_NAME = "Default"
     INTERNAL_SUPER_USER_ROLE_ID = UUID("00000000-0000-0000-0000-000000000002")
+    INTERNAL_SUPER_USER_ROLE_NAME = "Super User"
     EXPORTER_SUPER_USER_ROLE_ID = UUID("00000000-0000-0000-0000-000000000003")
+    EXPORTER_SUPER_USER_ROLE_NAME = "Super User"
     EXPORTER_DEFAULT_ROLE_ID = UUID("00000000-0000-0000-0000-000000000004")
+    EXPORTER_DEFAULT_ROLE_NAME = "Default"
```
phlptt: just wondering if this needs a UID too

andytaylorkainos: This has a UID (its the line above :) )

local.env
```diff
@@ -1,26 +1,27 @@
 DEBUG=True
 DATABASE_URL=postgres://postgres:password@localhost:5432/postgres
 PORT=8100
-ALLOWED_HOSTS='["*"]'
+ALLOWED_HOSTS=["*"]
 
 # Seed
-SEED_USERS='["test-uat-user@digital.trade.gov.uk"]'
+INTERNAL_ADMIN_USER=[{"email": "", "role": ""}]
```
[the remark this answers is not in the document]
efahy: If I don't set `EXPORTER_USERS`, then I don't get any seeded data anywhere, even if I have `INTERNAL_ADMIN_USERS` set and the output said a user was created.  Edge case, but is misleading

static/management/SeedCommand.py
```diff
@@ -106,6 +106,12 @@ def delete_unused_objects(model: models.Model, rows: list):
                     if not settings.SUPPRESS_TEST_OUTPUT:
                         print(f"Object {id} could not be deleted due to foreign key constraint")
 
+    @staticmethod
+    def print_created_or_updated(obj, data, is_created: bool):
```
efahy: The word "updated" prompted me to change my role from "Super User" to "Default" in my .env but there was no console output when I re-seeded and my role wasn't updated in the DB, is this out of scope?

static/management/commands/seedadminusers.py
```diff
@@ -0,0 +1,54 @@
...
+from users.enums import UserType
+from users.models import Role, GovUser
+
+
+class Command(SeedCommand):
+    """
+    pipenv run ./manage.py seedadminusers
+    """
+
+    help = "Seeds a gov admin team users so that other users can be added"
```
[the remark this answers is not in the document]
efahy: Again, is the word "admin" still applicable throughout this?

static/management/commands/seedadminusers.py
```diff
@@ -0,0 +1,54 @@
...
+
+        if created:
+            cls.print_created_or_updated(
+                Team, dict(id=Teams.ADMIN_TEAM_ID, name=Teams.ADMIN_TEAM_NAME), is_created=True
+            )
+
+    @classmethod
+    def seed_internal_users(cls):
+        admin_users = env("INTERNAL_ADMIN_USERS")
+        admin_users = admin_users.replace("=>", ":")
```
efahy: Perhaps a quick comment explaining the thinking here

static/management/commands/seedadminusers.py
```diff
@@ -0,0 +1,54 @@
...
+        admin_users = serialize(admin_users)
+
+        for user in admin_users:
+            email = user.get("email")
+            role = Role.objects.get(name=user.get("role", Roles.INTERNAL_SUPER_USER_ROLE_NAME), type=UserType.INTERNAL)
```
efahy: I think this explains the earlier failure to change my user's role from Super User to Default - this is hard coded to always get the Super User role and not the one configured in `INTERNAL_ADMIN_USERS` (unless I've missed something)

static/management/commands/seedexporterusers.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @transaction.atomic
+    def operation(self, *args, **options):
+        assert (
+            Organisation.objects.filter(name=DEFAULT_DEMO_ORG_NAME).exists()
+            and Organisation.objects.filter(name=DEFAULT_DEMO_HMRC_ORG_NAME).exists()
+        ), "Demo organisations must be seeded first!"
+        assert Role.objects.count(), "Role permissions must be seeded first!"
+
+        # Seed admin users first, as they could be re-defined with a role in the exporter users environment variable
```
efahy: By way of additional explanation, can you comment what you just explained to me, i.e. as a convenience you seed the internal users as exporter users so that the same user can log into both sites.  Is this also worth a comment in `local.env`, because by my reckoning we can leave `EXPORTER_USERS` empty and still get seeded for both sites

local.env
```diff
@@ -1,26 +1,27 @@
...
+EXPORTER_USERS=[{"email": "", "organisation": "", "role": ""}}]
```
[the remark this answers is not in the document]
andytaylorkainos: Added some defaults for guidance

static/management/commands/seedadminusers.py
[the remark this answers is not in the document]
andytaylorkainos: "admin" is specifying the team, rather than the role/permissions

static/management/commands/seedexporterusers.py
andytaylorkainos: Done! :)

PR #322, opened 2020-02-28
feat: LT-1168 - Add clearance level in MOD clearance advice

iichr:


goods/enums.py
```diff
@@ -46,7 +46,7 @@ class PvGrading:
     choices = [
         (UK_UNCLASSIFIED, "UK UNCLASSIFIED"),
         (UK_OFFICIAL, "UK OFFICIAL"),
-        (UK_OFFICIAL_SENSITIVE, "UK OFFICAL - SENSITIVE"),
+        (UK_OFFICIAL_SENSITIVE, "UK OFFICIAL - SENSITIVE"),
```
iichr: typo, hence all the extra migrations

PR #325, opened 2020-03-02
feat: LT - 1593 - F680 Clearance Types

andytaylorkainos:


applications/serializers/f680_clearance.py
```diff
@@ -10,15 +10,26 @@
...
+    class Meta:
+        model = F680ClearanceType
+        fields = ("name",)
 
 
 class F680ClearanceViewSerializer(PartiesSerializerMixin, GenericApplicationViewSerializer):
     goods = GoodOnApplicationViewSerializer(many=True, read_only=True)
     destinations = serializers.SerializerMethodField()
     additional_documents = serializers.SerializerMethodField()
+    f680_clearance_types = F680ClearanceTypeSerializer(read_only=True, many=True)
```
jfaracik: Could this not just be `clearance_types`?

static/urls.py
```diff
@@ -7,6 +7,7 @@
     path("control-list-entries/", include("static.control_list_entries.urls")),
     path("private-venture-gradings/", include("static.private_venture_gradings.urls")),
     path("countries/", include("static.countries.urls")),
+    path("f680-clearance-types/", include("static.f680_clearance_types.urls")),
```
jfaracik: Could this not just be `clearance-types`?

andytaylorkainos: these types are only applicable to f680s so I think it makes sense here

applications/serializers/f680_clearance.py
andytaylorkainos: From the perspective of an F680 applicaton it ofcourse doesnt make sense to duplicate the `f680_clearance` information in the name if we already know.

However, we refer to our cases and applications from `case` or `application` where it makes more sense to have this information in the name as it only relates to cases/applications that are F680 instances:
`case.application.f680_clearance_details`

I honestly don't mind what we name it, but i dont think it's worth the effort at this point for a system-wide change (seeding to DB, updating all request and response data on two frontends and updating the ui/shared seeding scripts)

cstobart: duplication of class name as attribute, please change

static/urls.py
jfaracik: What about `f680/clearance-types/`

applications/creators.py
```diff
@@ -232,6 +232,9 @@ def _validate_f680_clearance(draft, errors):
     if SiteOnApplication.objects.filter(application=draft).exists():
         errors["location"] = strings.Applications.F680.LOCATIONS
 
+    if not draft.f680_clearance_types.exists():
+        errors["f680-details"] = "Cannot create an application without F680 details"
```
cstobart: Will people know what specifically constitutes a F680 detail?

applications/models.py
```diff
@@ -163,7 +164,7 @@ class GiftingClearanceApplication(BaseApplication):
 
 # F680 includes End User & Third parties
 class F680ClearanceApplication(BaseApplication):
-    pass
+    f680_clearance_types = models.ManyToManyField(F680ClearanceType, related_name="f680_clearance_type_application")
```
cstobart: duplication of class name as attribute

applications/creators.py
andytaylorkainos: i think this is just to match the section on the application-overview (which content has named as f680-details. good point tho, will rename to match the fields required)

cstobart: 👍

applications/serializers/f680_clearance.py
andytaylorkainos: Damn.

applications/models.py
andytaylorkainos: Done

applications/serializers/f680_clearance.py
andytaylorkainos: Done

andytaylorkainos: Looks like others agreed with you, done! :)

PR #327, opened 2020-03-03
feat: LT-1385 - Highlight cases that haven't been updated in the last five days

jfaracik:


audit_trail/service.py
```diff
@@ -45,3 +45,7 @@ def get_user_obj_trail_qs(user, obj):
     audit_trail_qs = audit_trail_qs.filter(obj_as_action_filter | obj_as_target_filter)
 
     return audit_trail_qs
+
+
+def get_case_object_trail(case):
+    return Audit.objects.filter(target_content_type=ContentType.objects.get_for_model(case), target_object_id=case.id,)
```
efahy: I believe you also need to filter out audits resulting from exporter updates and it would be best to do that here when fetching from the DB

cases/tests/tests_is_recently_updated.py
```diff
@@ -0,0 +1,45 @@
...
+        # GOV user updates the case (changes the status of it)
+        data = {"status": CaseStatusEnum.RESUBMITTED}
+        self.client.put(reverse("applications:manage_status", kwargs={"pk": case.id}), data=data, **self.gov_headers)
+
+        response = self.client.get(self.url, **self.gov_headers)
+        response_data = response.json()["results"]["cases"][0]
+
+        self.assertEqual(response_data["is_recently_updated"], True)
+
+    def test_recently_updated_only_if_case_is_older_than_five_days(self):
```
efahy: This test name is a little confusing, how about `recently_submitted_cases_are_flagged_as_updated`?

efahy: Also, for this test to add value, there should be no gov user audit on the case - is that true?

conf/settings.py
```diff
@@ -195,8 +196,9 @@
 else:
     LOGGING = {"version": 1, "disable_existing_loggers": True}
 
-ADMIN_TEAM_ID = "00000000-0000-0000-0000-000000000001"
-
+RECENTLY_UPDATED_DAYS = env(
+    "RECENTLY_UPDATED_DAYS"
+)  # Days that must have past until we indicate a case has not been updated recently
```
efahy: "past" should be "passed"

cases/serializers.py
```diff
@@ -141,6 +147,14 @@ def get_status(self, instance):
     def get_users(self, instance):
         return instance.get_users(queue=self.context["queue_id"] if not self.context["is_system_queue"] else None)
 
+    def get_is_recently_updated(self, instance):
+        submitted_at = instance.submitted_at
+        trail = get_case_object_trail(instance).filter(
+            created_at__range=[timezone.now() - timedelta(days=settings.RECENTLY_UPDATED_DAYS), timezone.now()]
+        )
+
+        return (timezone.now() - submitted_at).days < settings.RECENTLY_UPDATED_DAYS or trail.exists()
```
stephanieboyd: The ticket mentions criteria of working days. Need to add that logic?

cases/serializers.py
```diff
@@ -141,6 +147,14 @@ def get_status(self, instance):
     def get_users(self, instance):
         return instance.get_users(queue=self.context["queue_id"] if not self.context["is_system_queue"] else None)
 
+    def get_is_recently_updated(self, instance):
+        submitted_at = instance.submitted_at
+        trail = get_case_object_trail(instance).filter(
```
efahy: Thinking more about it, this is a problem because its running a DB query for every case on the queue page, presumably 25 DB queries if thats the pages size, just for this indicator on the page.  As Mark H showed, we're already running way too may DB queries on the queue page so we need to come up with a better way to do this.  Discussion in Belfast is that we may require another job that runs at midnight, similar to the SLA job (or we could extend that one) that increments a last_updated_by_gov days counter for all open cases.  Would also need logic to reset the counter to zero when a gov user updates a case. Happy to discuss

a-bek: The need for us to do daily jobs for simple problems is because of the limitations of the current way the application is implemented. There is business logic in every layer. We should never do queries inside a serialiser, which makes us susceptible to the `n+1` query problem. Ideally, we just use serializers for validation, and then use the orm separately to retrieve all the data required. Adding a daily job for this would just add tech debt, and also its adding business logic in a completely new component of the application. 

For now we should definitely leave this here, when we optimise this, which we have to do at some point, this problem will no longer exist.

efahy: I disagree for several reasons:
1. We're meant to have a complete, viable MVP solution in 3 months.  To give us a chance of doing that, we need to start writing efficient code now rather than writing code we know is unperformant and we know from day 1 needs to be refactored 
2. I completely agree with you that doing DB queries and other complex logic in serializers is bad, however it seems to me that to refactor the complete API to simplify all serializers is a very large task and I doubt that we'd ever get business case to do it, even after MVP
3. Even if we did refactor the code as you suggest and include the above query into the main case query for the queue page, it'll always be a subquery within it and the overall query would be complicated and slow.  For this reason I disagree that we're introducing tech debt by doing the overnight job.  The overnight job is adding value by making our queue page queries simpler, no matter how the API is structured.  It means when we render the queue page we don't need to touch audit data at all

For these reasons, please put the logic into the existing SLA overnight job or create a new one

cases/tests/tests_is_recently_updated.py
jfaracik: Yep there isn't any audit on it

cases/serializers.py
a-bek: We have just optimised it so it does 1 query no matter the pagination size. Rather than using the serializer and fetching data per case (`n+1` query problem), we now do one query to fetch all relevant audits for all cases in the page, and append that data to the dict returned by the serializer.

Also, it is very bad pratice to embed query logic in the serializers, it means for each case you are doing queries to fetch the relevant data (`n+1` query problem) rather than doing one query to fetch all relevant data for all cases. Serializers remove a lot of the querying flexibility that the django orm provides, which is why it is bad practice.

What we just did for this change can be applied to all serializer fields that make queries, taking all those queries from PAGINATION_SIZE -> 1, hence optimising the retrieval.

Also, it would not be a subquery if we refactor it as suggested per case.

andytaylorkainos: If this works as intended (2 queries - one for the 25 cases and another for `get_all_audits_on_these_25_cases`) then I think this is a fairly viable solution.

Are you able to compare the two solutions Asumd, for the team? I think it'd be beneficial to see the data.
The query count with having the serializer doing a `get` per case and another query count for this proposed solution.

cases/views/search/service.py
```diff
@@ -29,3 +37,31 @@ def get_gov_users_list():
...
+    If the case was not, check that it has audit activity less than settings.RECENTLY_UPDATED_DAYS
+    ago and return True, else return False
+    """
+    now = timezone.now()
+    recent_audits = (
+        Audit.objects.filter(
+            target_content_type=ContentType.objects.get_for_model(Case),
+            target_object_id__in=[case["id"] for case in cases],
+            actor_content_type=ContentType.objects.get_for_model(GovUser),
+            created_at__gt=now - timedelta(days=settings.RECENTLY_UPDATED_DAYS),
```
Mark-Spiers-kainos: this setting is still checking the pass 5 days, should be working days. perhaps calculate and cache it, and use the cache?

cases/views/search/service.py
```diff
@@ -29,3 +37,31 @@ def get_gov_users_list():
...
+        )
+        .values("target_object_id")
+        .annotate(Count("target_object_id"))
+    )
+
+    audit_dict = {a["target_object_id"]: a["target_object_id__count"] for a in recent_audits}
+
+    for case in cases:
+        case["is_recently_updated"] = (
+            working_days_in_range(case["submitted_at"], now) < settings.RECENTLY_UPDATED_DAYS
```
andytaylorkainos: do we need to do this? surely the audit_dict contains all the information needed;

Maybe change the logic so that the audit_query isn't executed UNLESS:             working_days_in_range(case["submitted_at"], now) `>` settings.RECENTLY_UPDATED_DAYS

cases/serializers.py
BenAndrew-Kainos: I agree with Eugene and would instead suggest adding logic to the SLA background task to update a counter on the case. This would have a few benefits;

1. As this task only runs once per day at an inactive time for the service, this would minimise the processing impact
2. With a simple counter field, both storing and reporting this information would be very efficient
3. Updating all the cases in this service also provides to ability to filter or sort all cases based on this property

cases/sla.py
```diff
@@ -72,6 +77,19 @@ def is_bank_holiday(date):
     return formatted_date in get_bank_holidays()
 
 
+def num_days_since(date, num_working_days):
```
efahy: Needs a good comment please

cases/views/search/service.py
a-bek: Is that necessary, we have already squashed to 1 query? Do you really want to add a condition to MAYBE remove that 1 query every now and then?

audit_trail/service.py
```diff
@@ -45,3 +45,11 @@ def get_user_obj_trail_qs(user, obj):
     audit_trail_qs = audit_trail_qs.filter(obj_as_action_filter | obj_as_target_filter)
 
     return audit_trail_qs
+
+
+def get_case_object_trail(case):
```
efahy: Not used any more, since the revised approach to DB query?

cases/views/search/service.py
a-bek: Done

audit_trail/service.py
jfaracik: Great spot!

cases/sla.py
jfaracik: Done

cases/sla.py
```diff
@@ -36,6 +36,11 @@ def is_weekend(date):
     return date.weekday() > 4
 
 
+def working_days_in_range(start_date, end_date):
+    dates_in_range = [start_date + timedelta(n) for n in range((end_date - start_date).days)]
+    return len([date for date in dates_in_range if not is_bank_holiday(date) or not is_weekend(date)])
```
BenAndrew-Kainos: You cannot use bank_holiday in this current format because it currently always queries the Bank Holiday API (see https://github.com/uktrade/lite-api/pull/328/files for how this has been resolved in the 3 pointer)

cases/serializers.py
```diff
@@ -129,6 +129,9 @@ def get_flags(self, instance):
         """
         return get_ordered_flags(instance, self.team)
 
+    def get_submitted_at(self, instance):
+        return instance.submitted_at
```
BenAndrew-Kainos: Why is this a method field if it just returns instance.submitted_at?

cases/views/search/service.py
```diff
@@ -29,3 +37,32 @@ def get_gov_users_list():
...
+            created_at__gt=now - timedelta(days=number_of_days_since(now, settings.RECENTLY_UPDATED_WORKING_DAYS)),
+        )
+        .values("target_object_id")
+        .annotate(Count("target_object_id"))
+    )
+
+    audit_dict = {audit["target_object_id"]: audit["target_object_id__count"] for audit in recent_audits}
+
+    for case in cases:
+        case["is_recently_updated"] = bool(
```
BenAndrew-Kainos: Is there any way we can do this would a for loop (like an update across the cases)?

cases/tests/tests_sla.py
```diff
@@ -181,3 +183,19 @@ def test_is_bank_holiday(self):
...
+            (date(2020, 2, 10), 1, 1),
+            (date(2020, 2, 11), 2, 2),
+            (date(2020, 2, 12), 3, 3),
+            (date(2020, 2, 13), 4, 4),
+            (date(2020, 2, 14), 5, 5),
+            (date(2020, 2, 15), 6, 9),
+            (date(2020, 2, 16), 7, 11),
+        ]
+    )
+    def test_num_working_days_since(self, test_date, num_working_days, expected_result):
```
BenAndrew-Kainos: Why is this in test_sla when it's not used by sla?

cases/tests/tests_is_recently_updated.py
```diff
@@ -0,0 +1,63 @@
...
+
+from static.statuses.enums import CaseStatusEnum
+from test_helpers.clients import DataTestClient
+
+
+@override_settings(RECENTLY_UPDATED_WORKING_DAYS=5)
+class TestIsRecentlyUpdated(DataTestClient):
+
+    url = reverse("cases:search")
+    past_date = timezone.now() - timedelta(days=10)
```
BenAndrew-Kainos: This assumes there are no more than 5 days of consecutive holidays & weekends (please comment this)

cases/views/search/service.py
```diff
@@ -29,3 +37,32 @@ def get_gov_users_list():
...
+    ago and return True, else return False
+    """
+    now = timezone.now()
+
+    recent_audits = (
+        Audit.objects.filter(
+            target_content_type=ContentType.objects.get_for_model(Case),
+            target_object_id__in=[case["id"] for case in cases],
+            actor_content_type=ContentType.objects.get_for_model(GovUser),
+            created_at__gt=now - timedelta(days=number_of_days_since(now, settings.RECENTLY_UPDATED_WORKING_DAYS)),
```
BenAndrew-Kainos: This doesn't account for non-working days

cases/views/search/service.py
```diff
@@ -29,3 +37,32 @@ def get_gov_users_list():
...
+        )
+        .values("target_object_id")
+        .annotate(Count("target_object_id"))
+    )
+
+    audit_dict = {audit["target_object_id"]: audit["target_object_id__count"] for audit in recent_audits}
+
+    for case in cases:
+        case["is_recently_updated"] = bool(
+            working_days_in_range(case["submitted_at"], now) < settings.RECENTLY_UPDATED_WORKING_DAYS
```
efahy: As mentioned on our call just now, this is currently calling out to the external API to get bank holidays for every case.  Can you please change it so that it either uses only the cached version or at most calls the external API once per queue page load?

cases/views/search/service.py
efahy: Ben and I discussed this and Ben is now happy that it does account for working days and weekends.  @BenAndrew-Kainos are you happy to resolve?

cases/views/search/service.py
efahy: I see Ben has already made this comment, so resolving this one

cases/views/search/service.py
BenAndrew-Kainos: Yeah sorry my bad

cases/serializers.py
jfaracik: As it returned a string before, now it returns a DateTimeField

cases/sla.py
jfaracik: We've added caching for this as part of this story.

cases/tests/tests_is_recently_updated.py
a-bek: What do you mean?

cases/views/search/service.py
a-bek: what do you mean?

cases/tests/tests_sla.py
a-bek: Its uses only existing date functions inside the sla module. Perhaps those should have been in a separate module to begin with.

cases/serializers.py
BenAndrew-Kainos: Please leave comment

cases/tests/tests_sla.py
BenAndrew-Kainos: Please move

PR #328, opened 2020-03-04
feat: LT-987 SLA excludes active ECJU queries

BenAndrew-Kainos:


cases/sla.py
```diff
@@ -72,7 +80,37 @@ def is_bank_holiday(date):
...
+def get_case_ids_with_active_ecju_queries(date):
+    # ECJU Query SLA exclusion criteria
+    # 1. Still open & created before cutoff time today
+    # 2. Responded to in the last working day before cutoff time today
+    return (
+        EcjuQuery.objects.filter(
+            Q(responded_at__isnull=True, created_at__lt=today(time=SLA_UPDATE_CUTOFF_TIME),)
+            | Q(responded_at__gt=yesterday(time=SLA_UPDATE_CUTOFF_TIME))
+        )
+        .values("case")
```
Mark-Spiers-kainos: if you just need the id could you add "_id" to the end?

cstobart: case is the id

cases/sla.py
```diff
@@ -72,7 +80,37 @@ def is_bank_holiday(date):
...
+def today(time=timezone.now().time()):
+    return timezone.make_aware(datetime.combine(timezone.now(), time))
+
+
+def yesterday(date=None, time=None):
+    if date:
+        day = date - timezone.timedelta(days=1)
+    else:
+        day = timezone.now() - timezone.timedelta(days=1)
+    while is_bank_holiday(day) or is_weekend(day):
```
efahy: This code calls out to the external bank holiday API many times during one call of `yesterday()`.  This will go away with the caching of bank holidays being introduced by Asmund and Jan, but maybe we should still think about getting the bank holidays once at the start of this function?

cases/sla.py
```diff
@@ -72,7 +80,37 @@ def is_bank_holiday(date):
...
+        day = day - timezone.timedelta(days=1)
+    if time:
+        day = timezone.make_aware(datetime.combine(day.date(), time))
```
efahy: Can avoid the need for `make_aware` with this slightly simpler version: `day = datetime.combine(day.date(), time, tzinfo=timezone.utc)` there are other places you could make the equivalent change

cases/sla.py
cstobart: fixed, forgot to pass the new parameter

cases/sla.py
```diff
@@ -72,7 +80,37 @@ def is_bank_holiday(date):
     return formatted_date in get_bank_holidays()
 
 
-@background(schedule=make_aware(datetime.combine(now(), SLA_UPDATE_TASK_TIME)))
+def today(time=timezone.now().time()):
```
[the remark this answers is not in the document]
cstobart: the name was chosen because it reads very well when used in context, but I will add a docstring

cases/sla.py
```diff
@@ -72,7 +80,37 @@ def is_bank_holiday(date):
...
+    return timezone.make_aware(datetime.combine(timezone.now(), time))
+
+
+def yesterday(date=None, time=None):
```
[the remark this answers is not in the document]
cstobart: same as above, I will add a docstring but the name was purposefully chosen for how it reads in context

PR #329, opened 2020-03-05
chore: Seed organisations with exporter users

andytaylorkainos:
Move the logic for seeding organisations into the exporter seeding job so that it can be a standalone script.

To add a new organisation with seeding, place the name, type and reference number in the relevant ORGANISATIONS list


local.env
```diff
@@ -4,12 +4,17 @@ PORT=8100
 ALLOWED_HOSTS=["*"]
 
 # SEED USERS
-INTERNAL_ADMIN_TEAM_USERS=[{"email": "", "role": "Super User"}]
+INTERNAL_ADMIN_TEAM_USERS=[{"email": ""}]
+#INTERNAL_ADMIN_TEAM_USERS=[{"email": "", "role": "Super User"}] # role can be specified
+
 # INTERNAL_ADMIN_TEAM_USERS are also seeded as exporter users so they do not need to be added
 # to the EXPORTER_USERS variable
 # Although, if you wish to specify what organisation and role they have as an exporter,
```
efahy: Perhaps, for convenience you tell us what the default role and org are?

static/management/commands/seedexporterusers.py
```diff
@@ -1,123 +1,154 @@
...
-            exporter_user, created = ExporterUser.objects.get_or_create(
-                email__iexact=default_data["email"], defaults=default_data
-            )
+    def _add_user_to_organisations(cls, exporter_user: ExporterUser, exporter_user_data):
+        role_name = exporter_user_data.get("role", Roles.EXPORTER_SUPER_USER_ROLE_NAME)
+        organisation_name = exporter_user_data.get("organisation")
 
-            if created:
-                cls.print_created_or_updated(ExporterUser, default_data, is_created=True)
+        # If organisation was specified, only seed the user to that chosen commercial organisation
```
efahy: Maybe clarify that commercial = not HMRC?

PR #330, opened 2020-03-05
feat: LT-1233 - WMD end use concerns section

andytaylorkainos:
https://uktrade.atlassian.net/browse/LT-1233

- Affects open and standard licences
- New fields to support questions about military and/or WMD end use concerns


applications/tests/tests_adding_goods.py
```diff
@@ -103,7 +103,8 @@ def test_add_a_good_to_open_application_failure(self):
         Then a 400 BAD REQUEST is returned
         And no goods have been added
         """
-        draft = self.create_open_application(self.organisation)
+        good_name = "A good"
```
cstobart: What is this used for?

iichr: good spot, doesn't seem to be indeed

applications/models.py
```diff
@@ -138,6 +143,24 @@ class BaseApplication(ApplicationPartyMixin, Case):
     usage = models.TextField(default=None, blank=True, null=True)
     licence_duration = models.IntegerField(default=None, null=True, help_text="Set when application finalised")
     clearance_level = models.CharField(choices=PvGrading.choices, max_length=30, null=True)
+    is_military_end_use_controls = models.CharField(
+        choices=YesNoChoiceType.yes_no_choices, default=None, max_length=50, blank=True, null=True
+    )
+    military_end_use_controls_ref = models.TextField(default=None, blank=True, null=True, max_length=2000)
+
+    is_informed_wmd = models.CharField(
```
Mark-Spiers-kainos: could you not just use a BooleanField here?

andytaylorkainos: Yes we could - but to be consistent with all of our other Yes/No fields we went with this type

applications/migrations/0013_baseapplication_wmd_end_use_details.py
```diff
@@ -0,0 +1,81 @@
...
+        ),
+        migrations.AddField(
+            model_name="baseapplication",
+            name="military_end_use_controls_ref",
+            field=models.TextField(blank=True, default=None, max_length=255, null=True),
+        ),
+        migrations.AddField(
+            model_name="baseapplication",
+            name="suspected_wmd_ref",
+            field=models.TextField(blank=True, default=None, max_length=2000, null=True),
```
Mark-Spiers-kainos: 2000 chars seems a bit big for a ref, can you check this with product?

applications/migrations/0013_baseapplication_wmd_end_use_details.py
```diff
@@ -0,0 +1,81 @@
...
+
+    dependencies = [
+        ("applications", "0012_auto_20200309_1335"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="baseapplication",
+            name="compliant_limitations_eu_ref",
+            field=models.TextField(blank=True, default=None, max_length=2000, null=True),
```
Mark-Spiers-kainos: also check the length of this

applications/models.py
```diff
@@ -139,10 +143,23 @@ class BaseApplication(ApplicationPartyMixin, Case):
     licence_duration = models.IntegerField(default=None, null=True, help_text="Set when application finalised")
     clearance_level = models.CharField(choices=PvGrading.choices, max_length=30, null=True)
 
+    is_military_end_use_controls = models.BooleanField(blank=True, default=None, null=True)
+    military_end_use_controls_ref = models.TextField(default=None, blank=True, null=True, max_length=255)
```
Mark-Spiers-kainos: a charfield, makes more sense here since it's 255 length. Textfield doesn't enforce the length

andytaylorkainos: Good point!

PR #332, opened 2020-03-05
feat: LT-1234 - hide cases with open ecju queries in work queues

Mark-Spiers-kainos:


cases/libraries/get_destination.py
```diff
@@ -48,43 +49,54 @@ def get_destination_flags(case):
...
         if isinstance(application, StandardApplication):
             flags += get_standard_application_destination_flags(application)
 
     return flags
 
 
-def get_ordered_flags(case: Case, team: Team):
-    case_flags = case.flags.all()
-    org_flags = case.organisation.flags.all()
+def get_goods_flags(case):
```
Mark-Spiers-kainos: goods flags weren't returning goods flags for all case types, have updated it to work with all case types

cases/libraries/get_destination.py
```diff
@@ -48,43 +49,54 @@ def get_destination_flags(case):
...
+    org_flags = case.organisation.flags.all()
+    goods_flags = get_goods_flags(case)
+    destination_flags = get_destination_flags(case)
 
     flag_data = (
-        sort_flags_by_team_and_name(FlagSerializer(set(goods_flags), many=True).data, team)
-        + sort_flags_by_team_and_name(FlagSerializer(set(destination_flags), many=True).data, team)
-        + sort_flags_by_team_and_name(FlagSerializer(case_flags, many=True).data, team)
-        + sort_flags_by_team_and_name(FlagSerializer(org_flags, many=True).data, team)
+        sort_flags_by_team_and_name(CaseListFlagSerializer(set(goods_flags), many=True).data, team)
```
Mark-Spiers-kainos: made a smaller serializer that doesn't run SQL to get the team attached to the flag each time, this cut down my db calls from 135 to 34 locally

cases/libraries/get_destination.py
```diff
@@ -98,6 +110,6 @@ def sort_flags_by_team_and_name(flag_data, team):
     # Group flags by user's team.
     team_flags, non_team_flags = [], []
     for flag in flag_data:
-        team_flags.append(flag) if flag["team"]["id"] == str(team.id) else non_team_flags.append(flag)
+        team_flags.append(flag) if flag["team"] == str(team.id) else non_team_flags.append(flag)
```
Mark-Spiers-kainos: serializer just returns the team reference now (id) which means we don't have to look further into it

cases/managers.py
```diff
@@ -106,24 +123,25 @@ def search(
...
         date_order=None,
+        include_hidden=None,
     ):
         """
         Search for a user's available cases given a set of search parameters.
         """
         case_qs = self.submitted().prefetch_related("queues", "status", "organisation__flags",)
+        team_id = user.team.id
 
-        if queue_id == MY_TEAMS_QUEUES_CASES_ID:
```
Mark-Spiers-kainos: moved this code out due to complexity warning from linter

cases/views/search/queue.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, id, name, team, case_count):
     @classmethod
     def from_queue(cls, queue, case_qs=None) -> "SearchQueue":
         if not case_qs:
-            case_qs = Case.objects.all()
+            case_qs = Case.objects.submitted()
```
Mark-Spiers-kainos: queues should not get draft applications, same for next change. These fixed the queue count being wrong.

cases/libraries/get_destination.py
```diff
@@ -48,43 +49,54 @@ def get_destination_flags(case):
...
             for good_on_application in goods_on_application:
                 goods_flags += good_on_application.good.flags.all()
-        else:
-            goods_types = GoodsType.objects.filter(application=case)
-            for goods_type in goods_types:
-                goods_flags += goods_type.flags.all()
+    elif case_type in [
+        CaseTypeSubTypeEnum.OPEN,
+        CaseTypeSubTypeEnum.HMRC,
+    ]:
```
a-bek: Could you use `prefetch_related`

That will load all m2m relations in one query, otherwise everytime you loop through goods_type and access flags it does an additional query at each iteration!

https://docs.djangoproject.com/en/3.0/ref/models/querysets/#prefetch-related

cases/libraries/get_destination.py
```diff
@@ -48,43 +49,54 @@ def get_destination_flags(case):
...
+        goods_types = GoodsType.objects.filter(application=case)
+        for goods_type in goods_types:
+            goods_flags += goods_type.flags.all()
+    elif case_type == CaseTypeSubTypeEnum.GOODS:
+        goods_flags += GoodsQuery.objects.get(id=case.id).good.flags.all()
```
a-bek: Would  `GoodsQuery.objects.get(id=case.id).prefetch_related('good__flags').values('good').good.flags` work?

Mark-Spiers-kainos: That seems to throw an error when I tried it.

cases/libraries/get_destination.py
Mark-Spiers-kainos: I tried to determine how many additional queries are being made as a result of the current flag stuff. From what I can see it isn't making any, and django must be handling it.

I also looked through what queries are left and 26 of the 34 are status related. so going to see if I can lower that down to 1.

Mark-Spiers-kainos: After our talk I was able to use select_related a bit, but not prefetch_related

PR #336, opened 2020-03-10
feat: LT-1656 - Intangible goods

rorysproule1:


applications/serializers/good.py
```diff
@@ -87,3 +87,14 @@ def __init__(self, **kwargs):
         else:
             self.fields["item_type"].required = False
             self.fields["other_item_type"].required = False
+
+            if self.initial_data.get("unit") == "ITG":
```
jfaracik: Could you reference the unit value instead? 

eg `if self.initial_data.get("unit") == Units.ITG:`

rorysproule1: Now referencing Units.ITG :)

PR #337, opened 2020-03-10
feat: LT-1595 - F680 questions

a-bek:


applications/models.py
```diff
@@ -196,6 +197,15 @@ class GiftingClearanceApplication(BaseApplication):
 class F680ClearanceApplication(BaseApplication):
     types = models.ManyToManyField(F680ClearanceType, related_name="f680_clearance_application")
 
+    questions = JSONField(null=True)
```
efahy: As discussed, please revise to use individual model fields as per the rest of the code and I will review more fully at that point

applications/constants.py
```diff
@@ -1 +1,19 @@
 TRANSHIPMENT_BANNED_COUNTRIES = ["GB"]
+
+
+class F680:
+    ADDITIONAL_INFORMATION_FIELDS = (
+        "expedited",
+        "expedited_date",
+        "expedited_description",
```
efahy: The scenarios in the story don't mention an additional details field for the expedited question, just the future date.  Can we do away with this field?

applications/creators.py
```diff
@@ -168,6 +170,40 @@ def _validate_end_use_details(draft, errors, application_type):
...
+        "mtcr_type",
+        "electronic_warfare_requirement",
+        "uk_service_equipment",
+        "prospect_value",
+    ]
+    required_secondary_fields = {
+        "foreign_technology": "foreign_technology_description",
+        "expedited": "expedited_date",
+        "locally_manufactured": "locally_manufactured_description",
+    }
```
efahy: The story says if they select yes to `electronic_warfare_requirement`, they "must be prompted to attach part A of the EW data release capture form to the application" but I can't see that form referred to in this PR, unless I've missed it

applications/models.py
```diff
@@ -198,6 +199,28 @@ class GiftingClearanceApplication(BaseApplication):
 class F680ClearanceApplication(BaseApplication):
     types = models.ManyToManyField(F680ClearanceType, related_name="f680_clearance_application")
 
+    expedited = models.NullBooleanField(blank=True, default=None)
+    expedited_date = models.DateField(blank=True, null=True, default=None)
```
efahy: Something I've pulled others up on is whether we should allow both null and blank.  For me, if its an optional field (and same goes for other optional fields) then we store it as null but it shouldn't ever really be blank

applications/serializers/f680_clearance.py
```diff
@@ -78,24 +109,97 @@ class F680ClearanceUpdateSerializer(GenericApplicationUpdateSerializer):
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
         if "types" in self.initial_data:
             self.initial_data["types"] = [
                 F680ClearanceTypeEnum.ids.get(clearance_type) for clearance_type in self.initial_data.get("types", [])
             ]
 
     def validate(self, data):
+        required_fields = [
```
efahy: Would be good if possible to avoid duplicating these definitions of required and secondary fields between here and the application creator since they're the same in both places as far as I can see

applications/creators.py
a-bek: The content for that is displayed in the frontend form.

applications/serializers/f680_clearance.py
a-bek: done

PR #338, opened 2020-03-11
Lt 1295 assign user and work queue

stephanieboyd:


users/tests/test_users_team_queues.py
```diff
@@ -0,0 +1,22 @@
...
+    def setUp(self):
+        super().setUp()
+        self.second_queue = self.create_queue("second queue", self.team)
+        self.url = reverse("users:team_queues", kwargs={"pk": self.gov_user.id})
+
+    def test_get_all_users_team_queues(self):
+        response = self.client.get(self.url, **self.gov_headers)
+        queues = response.json()["queues"]
+
+        self.assertEqual(len(queues), 2)
```
andytaylorkainos: would it not be better to count the queues before you start the test,
Then add a queue and assert that the returned size is the same as the counted size +1?

Makes the test future proof

PR #341, opened 2020-03-12
bug: LT-2211 - SLA number continues to increase for reopened cases

pkonopacki:
- added storing the last_closed_at (was missing)
-fixed endpoint returning errors
- changed api tests


applications/tests/tests_finalise_application.py
```diff
@@ -188,7 +188,7 @@ def test_grant_application_success(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response_data["name"], self.standard_application.name)
         self.assertEqual(response_data["licence_duration"], get_default_duration(self.standard_application))
-        self.assertEqual(response_data["status"], str(self.finalised_status.id))
+        self.assertEqual(response_data["status"]["key"], str(self.finalised_status.status))
```
jfaracik: Would you mind using the new test helper function `generate_key_value_pair`?

It would work like so:

```
self.assertEqual(response_data["status"] = generate_key_value_pair(self.finalised_status.status, CaseStatusEnum.choices)
```

applications/serializers/generic_application.py
```diff
@@ -1,4 +1,5 @@
 import abc
+from datetime import datetime
```
jfaracik: @a-bek  should this be using Django date time?

pkonopacki: It is now, I did timezone.now() form django.utils. Cheers, resolving

applications/tests/tests_finalise_application.py
pkonopacki: Done, cheers!

PR #343, opened 2020-03-16
Chore "users/me" performance tuning

Mark-Spiers-kainos:
while implementing the tests we looked to improve performance, this following image was posted to confluence in relation to the power of select_related/prefetch_related.
image
I then looked at implementing a wider variety of tests (please note the number displayed for users is wrong, it should be 10, 100, 1000 same as other tests)
Before
and added some improvements to the endpoint by only asking for what was required
after


.circleci/config.yml
```diff
@@ -72,13 +72,13 @@ jobs:
 
       - run:
           name: Run seeding tests
-          command: pipenv run ./manage.py test --tag=seeding
+          command: pipenv run ./manage.py test --tag=seeding --exclude-tag=performance
```
andytaylorkainos: i dont think you need put an `--exclude-tag` option when you have specified `--tag`

users/tests/tests_performance.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class UsersPerformanceTests(PerformanceTestClient):
+
+    url = reverse("users:me")
+
+    def make_users_me_request(self):
+        """
+        Need to wrap the get in a class method to get 'self' context into timeit
+        """
```
andytaylorkainos: you could just make it a static method with "self" as a parameter

PR #345, opened 2020-03-17
feat: LT-2274 and LT-2275

andytaylorkainos:


applications/serializers/open_application.py
```diff
@@ -58,74 +58,6 @@ class Meta:
 
 
 class OpenApplicationUpdateSerializer(GenericApplicationUpdateSerializer):
```
Mark-Spiers-kainos: do we need this class anymore if it gets the 'GenericApplicationUpdateSerializer' fields?

applications/creators.py
```diff
@@ -128,7 +128,7 @@ def _validate_ultimate_end_users(draft, errors, is_mandatory):
 
     if is_mandatory:
         ultimate_end_user_required = False
-        if next(filter(lambda x: x.is_good_incorporated, GoodOnApplication.objects.filter(application=draft),), None,):
+        if next(filter(lambda x: x.is_good_incorporated, GoodOnApplication.objects.filter(application=draft)), None):
```
jfaracik: Could this be done like: GoodOnApplication.objects.filter(application=draft, is_good_incorporated=True).count() ? That way you're not filtering in memory

test_helpers/clients.py
```diff
@@ -684,6 +687,7 @@ def create_draft_open_application(self, organisation: Organisation, reference_na
             is_military_end_use_controls=False,
             is_informed_wmd=False,
             is_suspected_wmd=False,
+            intended_end_use="intended end use is none of your business",
```
jfaracik: Sass

applications/serializers/open_application.py
andytaylorkainos: I think the reason this class existed in the first place was to allow for the "helper" method `get_application_update_serializer` in `application_helpers.py`

andytaylorkainos: Originally, before we introduced first end_use_details fields, this is what the serializer looked like

applications/creators.py
andytaylorkainos: To be honest we didnt even look at this line of code when we removed the unnecessary `,` in the statement.

I'll have a quick look at it to see if it can be optimized the way you suggested

applications/tests/tests_copy_application.py
```diff
@@ -360,13 +362,17 @@ def _validate_reset_data(self):
...
-        self.assertIsNone(self.copied_application.is_informed_wmd)
-        self.assertIsNone(self.copied_application.is_suspected_wmd)
-        self.assertIsNone(self.copied_application.is_military_end_use_controls)
-        self.assertIsNone(self.copied_application.is_eu_military)
-        self.assertIsNone(self.copied_application.is_compliant_limitations_eu)
-        self.assertIsNone(self.copied_application.compliant_limitations_eu_ref)
+    def _validate_end_use_details(self, application_type=None):
+        if application_type == CaseTypeSubTypeEnum.F680:
+            self.assertIsNone(self.copied_application.intended_end_use)
+        else:
```
andytaylorkainos: might be worth putting an if statement in here to split out standard and open

applications/creators.py
andytaylorkainos: done!

PR #347, opened 2020-03-17
feat: LT-1264 Completed Queue's Workflow

BenAndrew-Kainos:


workflow/user_queue_assignment.py
```diff
@@ -0,0 +1,31 @@
+from cases.models import Case, CaseAssignment
+from queues.models import Queue
+from static.statuses.models import CaseStatus
+
+
+def get_queues_with_case_assignments(case: Case):
+    return {assignment.queue for assignment in CaseAssignment.objects.filter(case=case).distinct("queue")}
+
+
+def get_next_non_terminal_status(status: CaseStatus):
```
stephanieboyd: minor - a better name might be `get_next_status_in_workflow_sequence`. Terminal may or may not come into it at some point

workflow/user_queue_assignment.py
```diff
@@ -0,0 +1,31 @@
...
+    return None
+
+
+def user_queue_assignment_workflow(queues: [Queue], case: Case):
+    # Remove case from queues where all gov users are done with the case
+    queues_without_case_assignments = set(queues) - get_queues_with_case_assignments(case)
+    case.queues.remove(*queues_without_case_assignments)
+
+    # Move case to next non-terminal state if unassigned from all queues
+    if case.queues.count() == 0:
```
stephanieboyd: So in the scenario where it's removed from all users on a queue and it has no status to go to next. It sits at the last status and is only visible on the `all cases` and `open cases` queues?

workflow/user_queue_assignment.py
```diff
@@ -0,0 +1,31 @@
+from cases.models import Case, CaseAssignment
+from queues.models import Queue
+from static.statuses.models import CaseStatus
+
+
+def get_queues_with_case_assignments(case: Case):
+    return {assignment.queue for assignment in CaseAssignment.objects.filter(case=case).distinct("queue")}
```
stephanieboyd: minor - You could also do `set(CaseAssignment.objects.filter(case=case).distinct("queue").values_list('queue', flat=True))`

cases/views/views.py
```diff
@@ -622,3 +624,45 @@ def put(self, request, pk):
...
+
+
+class AssignedQueues(APIView):
+    authentication_classes = (GovAuthentication,)
+    serializer_class = TinyQueueSerializer
+
+    def get(self, request, pk):
+        # Get all queues where this user is assigned to this case
+        assignments = CaseAssignment.objects.filter(user=self.request.user, case__id=pk)
+        # TODO figure out how to do this the django way
```
a-bek: 2 options:

Reverse lookup on the Queue model to retrieve the queues

Use select related and values queryset so when you retrieve through `CaseAssignment` you don't do an additional query for each queue.

cases/views/views.py
```diff
@@ -622,3 +624,45 @@ def put(self, request, pk):
...
+        queues = [assignment.queue for assignment in assignments]
+        serializer = TinyQueueSerializer(queues, many=True)
+        return JsonResponse(data={"queues": serializer.data}, status=status.HTTP_200_OK)
+
+    @transaction.atomic
+    def put(self, request, pk):
+        queues = request.data.get("queues")
+        if queues:
+            queue_names = []
+            assignments = CaseAssignment.objects.filter(user=request.user, case__id=pk, queue__id__in=queues)
```
a-bek: Same as above to avoid the adverse queries.

cases/views/views.py
BenAndrew-Kainos: Done. Thanks for your help

cases/views/views.py
BenAndrew-Kainos: As above

workflow/user_queue_assignment.py
BenAndrew-Kainos: Done

workflow/user_queue_assignment.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def get_next_non_terminal_status(case):
+    if case.case_type.reference == CaseTypeEnum.GOODS.reference:
+        return get_next_goods_query_status(case)
+    else:
+        status = case.status
+        if status.workflow_sequence:
+            next_status_id = status.workflow_sequence + 1
+            try:
```
andytaylorkainos: return CaseStatus.objects.filter(workflow_sequence=next_status_id, is_terminal=False).first()

cases/views/views.py
```diff
@@ -625,6 +628,48 @@ def put(self, request, pk):
...
+        queues = Queue.objects.filter(case_assignments__user=request.user.pk, case_assignments__case=pk)
+        serializer = TinyQueueSerializer(queues, many=True)
+        return JsonResponse(data={"queues": serializer.data}, status=status.HTTP_200_OK)
+
+    @transaction.atomic
+    def put(self, request, pk):
+        queues = request.data.get("queues")
+        if queues:
+            queue_names = []
+            assignments = CaseAssignment.objects.select_related("queue").filter(
```
andytaylorkainos: you could order the assignments by queue_name to save you having to sort it in memory later

workflow/user_queue_assignment.py
BenAndrew-Kainos: Found a better way of getting this data

cases/models.py
```diff
@@ -155,8 +157,8 @@ class CaseAssignment(TimestampableModel):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     case = models.ForeignKey(Case, on_delete=models.CASCADE, related_name="case_assignments")
-    users = models.ManyToManyField(GovUser, related_name="case_assignments")
-    queue = models.ForeignKey(Queue, on_delete=models.CASCADE)
+    user = models.ManyToManyField(GovUser, related_name="case_assignments")
```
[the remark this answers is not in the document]
BenAndrew-Kainos: I was midway through an interesting migration. Have a look now

audit_trail/payload.py
```diff
@@ -54,6 +54,8 @@ class AuditType(Enum):
     REMOVE_CASE_OFFICER_FROM_CASE = "removed {case_officer} as the Case Officer"
     GRANTED_APPLICATION = "granted licence for {licence_duration} months starting from {start_date}"
     FINALISED_APPLICATION = "finalised the application"
+    UNASSIGNED_QUEUES = "marked themselves as done for this case on the following queues: {queues}"
```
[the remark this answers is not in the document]
BenAndrew-Kainos: Sure

workflow/user_queue_assignment.py
BenAndrew-Kainos: I need to ensure there is only one!

audit_trail/payload.py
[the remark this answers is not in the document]
BenAndrew-Kainos: Anthony says it's all good

cases/views/views.py
BenAndrew-Kainos: Done 👍

audit_trail/payload.py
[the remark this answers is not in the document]
stephanieboyd: Yeah, I checked with Anthony on my audit related strings and he said he hadn't gotten to any of them yet - so there isn't a format or approach

workflow/user_queue_assignment.py
stephanieboyd: @andytaylorkainos It needed to be a set for the set operation to be performed later in the code.

workflow/user_queue_assignment.py
andytaylorkainos: but in this scenario, if there was more than one - it'd do nothing (is that the correct approach? - should there be an accompanying error message to say why it did nothing)

cases/models.py
```diff
@@ -155,8 +155,8 @@ class CaseAssignment(TimestampableModel):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     case = models.ForeignKey(Case, on_delete=models.CASCADE, related_name="case_assignments")
-    users = models.ManyToManyField(GovUser, related_name="case_assignments")
-    queue = models.ForeignKey(Queue, on_delete=models.CASCADE)
+    user = models.ForeignKey(GovUser, on_delete=models.CASCADE, related_name="case_assignments")
```
andytaylorkainos: 👍 
nicely migrated

BenAndrew-Kainos: Danke.

PR #348, opened 2020-03-17
feat: LT-1253 - Additional contacts

jfaracik:
* Added endpoint for listing and creating contacts on a case
* Set default error handler to `lite_error_handler` which automatically wraps errors in the required `errors` dict
* Added `generate_country_dict` for tests - give it a country and it'll output the json it should be compared against


users/views.py
```diff
@@ -208,7 +208,7 @@ def put(self, request, *args, **kwargs):
 
         # If (after the PUT) the user isn't assigned to any sites, raise an error
         if not combined_sites:
-            raise serializers.ValidationError({"errors": {"sites": ["Select at least one site to assign the user to"]}})
+            raise serializers.ValidationError({"sites": ["Select at least one site to assign the user to"]})
```
andytaylorkainos: can you put this error on content please

applications/tests/tests_consignee.py
```diff
@@ -297,7 +297,7 @@ def test_consignee_validate_only_failure(self):
         response = self.client.post(self.url, end_user, **self.exporter_headers)
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
-        self.assertEqual(response.json(), {"errors": {"name": [Parties.NULL_NAME]}})
+        self.assertEqual(response.json(), {"errors": {"name": [PartyErrors.NAME["null"]]}})
```
a-bek: How come you're using dicts?

jfaracik: So we don't need to specify each key in the serialiser

users/views.py
jfaracik: Done!

PR #349, opened 2020-03-19
feat: LT-2115 - automate flagging rules

Mark-Spiers-kainos:


workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+        flagging_rules = active_flagging_rules_for_level(FlagLevels.GOOD)
+    flags = list(flagging_rules.filter(matching_value__iexact=good.control_code).values_list("flag_id", flat=True))
+
+    if flags:
+        good.flags.add(*flags)
+
+
+def apply_flagging_rule_to_all_open_cases(flagging_rule):
+    if flagging_rule.status == FlagStatuses.ACTIVE and flagging_rule.flag.status == FlagStatuses.ACTIVE:
+        open_cases = Case.objects.submitted().is_open()
```
andytaylorkainos: does `is_open()` not only return submitted applications? o.O

workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+        flagging_rule_queryset = FlaggingRule.objects.filter(id=flagging_rule.id)
+
+        if flagging_rule.level == FlagLevels.CASE:
+            open_cases.filter(case_type__reference=flagging_rule.matching_value).all()
+            for case in open_cases:
```
andytaylorkainos: im fairly certain these for-loops are going to query for every single open-case in the DB one at a time

workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+
+    for good in goods:
+        apply_goods_rules_for_good(good, flagging_rules)
+
+
+def apply_goods_rules_for_good(good, flagging_rules=None):
+    if not flagging_rules:
+        flagging_rules = active_flagging_rules_for_level(FlagLevels.GOOD)
+    flags = list(flagging_rules.filter(matching_value__iexact=good.control_code).values_list("flag_id", flat=True))
```
andytaylorkainos: (duplicate comment) i dont think you have to convert this to a list to be able to do `good.flags.add(*flags)`

Infact, if you dont have to, i'd prefer you didnt -> its risky to hold all of this in memory since we're not sure how many flagging rules may exist after a few years of being live

workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+
+def apply_flagging_rules_to_case(case):
+    if not (case.status.status == CaseStatusEnum.DRAFT or CaseStatusEnum.is_terminal(case.status.status)):
+        apply_case_rules(case)
+        apply_destination_rules_for_case(case)
+        apply_good_flagging_rules_for_case(case)
+
+
+def apply_case_rules(case):
+    flags = list(
```
andytaylorkainos: (duplicate comment) i dont think you have to convert this to a list to be able to do `good.flags.add(*flags)`

Infact, if you dont have to, i'd prefer you didnt -> its risky to hold all of this in memory since we're not sure how many flagging rules may exist after a few years of being live

workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+
+    for party in parties:
+        apply_destination_rule_on_party(party, flagging_rules)
+
+
+def apply_destination_rule_on_party(party: Party, flagging_rules=None):
+    if not flagging_rules:
+        flagging_rules = active_flagging_rules_for_level(FlagLevels.DESTINATION)
+    flags = list(flagging_rules.filter(matching_value=party.country.id).values_list("flag_id", flat=True))
```
andytaylorkainos: same as other duplicate comments (unless you're certain this is going to only ever be a small list)

workflow/flagging_rules_automation.py
Mark-Spiers-kainos: it just returns non-terminal cases

workflow/flagging_rules_automation.py
Mark-Spiers-kainos: All this returns is a list of flags id's which are to be attached

workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def apply_flagging_rules_to_case(case):
+    if not (case.status.status == CaseStatusEnum.DRAFT or CaseStatusEnum.is_terminal(case.status.status)):
+        apply_case_rules(case)
+        apply_destination_rules_for_case(case)
+        apply_good_flagging_rules_for_case(case)
+
+
+def apply_case_rules(case):
```
efahy: Not enough (any!) comments in here. Please add docstring comments to all methods that are in any way complicated

workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+        active_flagging_rules_for_level(FlagLevels.CASE)
+        .filter(matching_value=case.case_type.reference)
+        .values_list("flag_id", flat=True)
+    )
+
+    if flags:
+        case.flags.add(*flags)
+
+
+def apply_destination_rules_for_case(case, flagging_rule=None):
```
efahy: Not a huge deal but you might consider specifying a type of `QuerySet` for the `flagging_rule` parameter in all applicable methods

workflow/flagging_rules_automation.py
```diff
@@ -0,0 +1,110 @@
...
+    flagging_rules = active_flagging_rules_for_level(FlagLevels.DESTINATION) if not flagging_rule else flagging_rule
+
+    if case.case_type.id == CaseTypeEnum.EUA.id:
+        if isinstance(case, Case):
```
efahy: Can it not be assumed that its always an instance of `Case` and lose the `if`?  If you want to keep the `if` then you should handle the situation where its not an instance of `Case`, probably by throwing an error

applications/views/applications.py
```diff
@@ -343,14 +346,18 @@ def put(self, request, pk):
...
         serializer = get_application_update_serializer(application)
         serializer = serializer(application, data=data, partial=True)
 
         if not serializer.is_valid():
             return JsonResponse(data={"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
 
-        serializer.save()
+        application = serializer.save()
+
+        if CaseStatusEnum.is_terminal(old_status.status) and not CaseStatusEnum.is_terminal(application.status.status):
```
efahy: Perhaps add a comment (based on the scenario in the story)

workflow/flagging_rules_automation.py
Mark-Spiers-kainos: realised this should have been if not EndUserAdvisory type, then get it.

PR #350, opened 2020-03-19
feat: LT-1493 - ECJU Query Types

cstobart:


cases/models.py
```diff
@@ -306,6 +307,9 @@ class EcjuQuery(TimestampableModel):
     responded_by_user = models.ForeignKey(
         ExporterUser, related_name="exportuser_ecju_query", on_delete=models.CASCADE, default=None, null=True,
     )
+    query_type = models.CharField(
```
jfaracik: Could we just call this `type`?

cstobart: I don't like using type as its a built in name

PR #351, opened 2020-03-20
feat: LT-1060 - Non-UK organisations and sites

jfaracik:
**Story changes**
* Adds `ForeignAddress` model - just an address field and a country field
* Updated `Site` to take both `Address` and `ForeignAddress`

**Performance and misc improvements**
* Moved and cleaned up adding sites to an application
* `PrimaryKeyRelatedSerializerField` now no longer queries for the object if its already in memory
* `disable_pagination` is now in the pagination class as opposed to having to add it to every view that needs it
* Now possible to filter an organisations user by permission (eg get me all users that don't have permission x - used for filtering users that have the manage site permission)
* BIG query reductions for the `organisations` app - eg loading `/sites/` - 105 queries down to 18 queries
* BIG 10x query reductions for /users/me/ and profile views


organisations/serializers.py
```diff
@@ -1,72 +1,126 @@
...
+        users_with_permission = set(
+            [
+                x.user
+                for x in UserOrganisationRelationship.objects.filter(
+                    organisation=instance.organisation, role__permissions__id=permission.id
+                )
+            ]
+        )
+        users_union = users.union(users_with_permission)
+        users_union = sorted(users_union, key=lambda x: x.first_name)
```
andytaylorkainos: you can do this in one query without loading into memory (i think):

```
site_users = UserOrganisationRelationship.objects.filter(sites__id__exact=instance.id).values("user")

permission_users = UserOrganisationRelationship.objects.filter(
                    organisation=instance.organisation, role__permissions__id=ExporterPermissions.ADMINISTER_SITES.name
                ).values("user")

users = site_users.union(permission_users).order_by("user__first_name")
```

andytaylorkainos: It may take a little tweaking, but essentially you can union two querysets if they both contain the same fields (in this case they both have "user")

applications/tests/tests_view_application.py
```diff
@@ -11,6 +11,7 @@
 )
 from cases.enums import CaseTypeEnum
 from goodstype.models import GoodsType
+from organisations.factories import SiteFactory
```
andytaylorkainos: Im not a fan of adding factories into the main directory of the module when its only used in context of testing.
Can you move these factories into either their `/tests/` directory or somewhere more suitable

organisations/providers.py
```diff
@@ -0,0 +1,17 @@
+import random
+
+from faker.providers import BaseProvider
+
+
+class OrganisationProvider(BaseProvider):
```
andytaylorkainos: again, if this is only used in the context of testing - can you move the classes into the `/tests/` directory

applications/tests/tests_view_application.py
andytaylorkainos: Or add them to `test_helpers`

organisations/views/users.py
```diff
@@ -23,35 +29,34 @@
...
-            p.user.status = p.status
-            p.user.role = p.role
-
-        serializer = OrganisationUserListView([p.user for p in page], many=True)
-
-        return self.get_paginated_response({"users": serializer.data})
+    serializer_class = OrganisationUserListView
+
+    def get_queryset(self):
+        _status = self.request.GET.get("status")
```
andytaylorkainos: why is status prepended with `_`? (if its a known pattern with a reason, i wouldnt mind knowing it 🤓 )

organisations/views/users.py
```diff
@@ -23,35 +29,34 @@
...
+        return (
+            ExporterUser.objects.filter(reduce(operator.and_, query))
+            .exclude(relationship__role__permissions__in=[exclude_permission])
+            .select_related("relationship__role")
+            .values(
+                "id",
+                "first_name",
+                "last_name",
+                "email",
+                status=F("relationship__status"),
```
andytaylorkainos: haven't used `F` in a while - can you explain why you're using it here?

users/views.py
```diff
@@ -135,11 +141,41 @@ class UserMeDetail(APIView):
...
+            data.update(
+                {
+                    "role": {
+                        "id": relationship.role.id,
+                        "permissions": convert_queryset_to_str(
+                            relationship.role.permissions.values_list("id", flat=True)
+                        ),
+                    }
+                }
+            )
```
andytaylorkainos: can you make a serializer that doesn't require a model and serialize the data in there please?

Pipfile
```diff
@@ -11,6 +11,7 @@ parameterized = "~=0.7"
 prospector = "==1.1.6.2" #  Our project only works with version 1.1.6.2
 unittest-xml-reporting = "~=2.5"
 faker = "~=4.0"
+factory-boy = "~=2.12.0"
```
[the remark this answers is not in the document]
jfaracik: Factories!

addresses/models.py
```diff
@@ -6,14 +6,22 @@
...
+    address_line_1 = models.CharField(default=None, blank=False, max_length=50)
+    address_line_2 = models.CharField(default=None, blank=True, null=True, max_length=50)
+    region = models.CharField(default=None, blank=False, max_length=50)
     postcode = models.CharField(max_length=10)
-    city = models.TextField(default=None, blank=False)
+    city = models.CharField(default=None, blank=False, max_length=50)
     country = models.ForeignKey(Country, blank=False, null=False, on_delete=models.CASCADE)
+
+    class Meta:
+        db_table = "address"
```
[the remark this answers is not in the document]
jfaracik: It makes a migration for you! Magical, we should really be doing this for all models as Django's default names are trash, and say in the future you move one model to another app, the name is going to make zero sense.

conf/middleware.py
```diff
@@ -34,12 +38,20 @@ def __init__(self, get_response=None):
...
         final_queries = connection.queries
 
-        elapsed_time = time.time() - start
+        elapsed_time = datetime.now() - start
+
+        _type = request.method.upper()
+        duration = round(elapsed_time.microseconds / 100000, 2)  # noqa
+        queries = len(final_queries)
+
+        print(f"\n{_type} {bold(request.path)} - ⏱  {duration}s  🗂  {queries} queries\n")
```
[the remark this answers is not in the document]
jfaracik: Not yet! I'm looking to add that, would an environment variable be good for you?

organisations/views/users.py
jfaracik: Not a pattern, it's just that "status" shadows the "status" import `from rest_framework import status`

organisations/views/users.py
jfaracik: Essentially selecting info about the user (first, last, email) and adding the information from the relationship object to it (status and role)

applications/tests/tests_view_application.py
jfaracik: Done!

organisations/providers.py
jfaracik: Done!

conf/middleware.py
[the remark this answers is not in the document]
jfaracik: `SHOW_VIEW_QUERIES` toggles between the print and the logs (I was getting fed up with them appearing in my terminal)

addresses/models.py
```diff
@@ -6,14 +6,22 @@
...
+
+
+class ForeignAddress(models.Model):
```
a-bek: Could this not just be a a row on the `Address` model, with nullable uk specific fields - postcode, etc.

The `Address` model supports country already, and to get and create foreign addresses one can just make a custom queryset and serailizer, ie, `Address.objects.exclude(country='uk')` rather than add a new model?

I believe you wouldn't really have to change much, just the model reference in the serializers and factories, etc.

applications/libraries/sites_helpers.py
```diff
@@ -0,0 +1,115 @@
...
+    if application.status.status in get_case_statuses(read_only=True):
+        raise ValidationError({"sites": [f"Application status {application.status.status} is read-only"]})
+
+    # Transhipment applications can't have sites based in certain countries
+    if application.case_type.reference == CaseTypeReferenceEnum.SITL:
+        is_site_in_banned_location = (
+            new_sites.annotate(country=Coalesce("address__country_id", "foreign_address__country_id"))
+            .values_list("country")
+            .filter(country__in=TRANSHIPMENT_BANNED_COUNTRIES)
+            .count()
```
a-bek: `.exists()` is more efficient than `count()` if all you want top know is if there is at least 1 record.

applications/views/external_locations.py
```diff
@@ -114,7 +114,7 @@ def _set_locations_and_sites(
                 actor=user,
                 verb=AuditType.REMOVED_SITES_FROM_APPLICATION,
                 target=application.get_case(),
-                payload={"sites": [site.site.name + " " + site.site.address.country.name for site in removed_sites]},
+                payload={"sites": [site.site.name for site in removed_sites]},
```
a-bek: This executed multiple queries, perhaps use a `values_list` on removed_sites above to only fetch the name?

conf/pagination.py
```diff
@@ -3,6 +3,12 @@
 
 
 class MaxPageNumberPagination(pagination.PageNumberPagination):
+    def paginate_queryset(self, queryset, request, view=None):
+        if request.GET.get("disable_pagination", False):
+            return
```
a-bek: Does the queryset not need to be returned here?

organisations/models.py
```diff
@@ -72,17 +87,26 @@ def get_by_user_organisation_relationship(self, exporter_user_organisation_relat
...
+    address = models.ForeignKey(Address, related_name="site", on_delete=models.DO_NOTHING, null=True)
+    foreign_address = models.ForeignKey(ForeignAddress, related_name="site", on_delete=models.DO_NOTHING, null=True)
+
     objects = SiteManager()
 
     class Meta:
+        db_table = "site"
         ordering = ["name"]
 
+    def get_country(self):
```
a-bek: This wouldn't be necessary with one Address table.

organisations/serializers.py
```diff
@@ -1,85 +1,135 @@
...
+
+
+class SiteViewSerializer(SiteListSerializer):
+    users = serializers.SerializerMethodField()
+
+    def get_users(self, instance):
+        users = UserOrganisationRelationship.objects.filter(
+            Q(sites__id__exact=instance.id)
+            | Q(organisation=instance.organisation, role__permissions__id=ExporterPermissions.ADMINISTER_SITES.name)
+        ).select_related("user")
```
a-bek: What about `values_list('user', flat=True)`?

addresses/models.py
jfaracik: I've made the change!

applications/libraries/sites_helpers.py
jfaracik: Done.

conf/pagination.py
jfaracik: Nope

organisations/models.py
jfaracik: Done

organisations/serializers.py
jfaracik: Doesn't work sadly

applications/views/external_locations.py
jfaracik: Done

addresses/serializers.py
```diff
@@ -2,27 +2,63 @@
...
+            "address",
             "address_line_1",
             "address_line_2",
             "city",
             "region",
             "postcode",
             "country",
         )
+
+    # def validate_country(self, value):
```
andytaylorkainos: you going to remove this code?

addresses/serializers.py
```diff
@@ -2,27 +2,63 @@
...
-    city = serializers.CharField(error_messages={"blank": "Enter a real city"})
-    region = serializers.CharField(error_messages={"blank": "Enter a real region"})
-    country = CountrySerializerField()
+    address = serializers.CharField(max_length=256, required=False, error_messages={"blank": "Enter a valid address"})
+    address_line_1 = serializers.CharField(
+        max_length=50, required=False, error_messages={"blank": "Enter a real building and street name"}
+    )
+    postcode = serializers.CharField(max_length=10, required=False, error_messages={"blank": "Enter a real postcode"})
+    city = serializers.CharField(max_length=50, required=False, error_messages={"blank": "Enter a real city"})
+    region = serializers.CharField(max_length=50, required=False, error_messages={"blank": "Enter a real region"})
```
andytaylorkainos: content errors im afraid

applications/libraries/sites_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+    Add sites to an application, handle validation and audit
+    """
+    sites_on_application = SiteOnApplication.objects.filter(application=application)
+
+    if not new_sites:
+        raise ValidationError({"sites": ["Select at least one site"]})
+
+    # Users don't specify their goods locations if they've already departed
+    if getattr(application, "have_goods_departed", False):
+        raise ValidationError({"sites": ["Your goods have already departed"]})
```
andytaylorkainos: more content :(

conf/middleware.py
```diff
@@ -15,16 +20,18 @@ def __call__(self, request):
...
-                "message": "liteolog api",
-                "corrID": request.correlation,
-                "type": "http response",
-                "method": request.method,
-                "url": request.path,
-                "elapsed_time": time.time() - start,
-            }
-        )
+
+        if not env("SHOW_VIEW_QUERIES"):
```
andytaylorkainos: i think this should always be enabled - can you revert this please

conf/middleware.py
```diff
@@ -34,23 +41,32 @@ def __init__(self, get_response=None):
...
-            }
-        )
+        elapsed_time = datetime.now() - start
+
+        _type = request.method.upper()
+        duration = round(elapsed_time.microseconds / 100000, 2)  # noqa
+        queries = len(final_queries)
+
+        if env("SHOW_VIEW_QUERIES"):
+            print(f"\n{_type} {bold(request.path)} - ⏱  {duration}s  🗂  {queries} queries\n")
```
andytaylorkainos: I do like this data showing up but i'm beginning to think it'd be better if you could move this functionality into a tool where we could just switch between what logs show instead of potentially having a lot of environment variables for the sole purpose of local development.

Maybe worth a discussion?

conf/middleware.py
jfaracik: Why should what always be enabled?

users/views.py
```diff
@@ -135,11 +141,41 @@ class UserMeDetail(APIView):
 
     def get(self, request):
         org_pk = request.headers["Organisation-Id"]
+        user = request.user
+        relationships = UserOrganisationRelationship.objects.filter(user=user)
+
+        data = {
```
andytaylorkainos: can you leave an explanation as to why we arent using a seralizer (performance implications)

addresses/serializers.py
jfaracik: Good spot!

conf/middleware.py
jfaracik: Absolutely! I'd say out of scope for this PR though :)

conf/middleware.py
andytaylorkainos: the logging info below (remove the if statement) :D

conf/middleware.py
andytaylorkainos: Agreed, so can you get rid of the environment variable for now and we can come back to it

addresses/serializers.py
jfaracik: Done

users/views.py
```diff
@@ -135,11 +141,43 @@ class UserMeDetail(APIView):
 
     def get(self, request):
         org_pk = request.headers["Organisation-Id"]
+        user = request.user
+        relationships = UserOrganisationRelationship.objects.filter(user=user)
```
a-bek: Can you do a select_related on `organisation`, as you'll be doing `n+1` queries below.

jfaracik: Done!

organisations/serializers.py
andytaylorkainos: `.values("user")`

saves you having to iterate through each `UserOrganisationRelationship`

PR #353, opened 2020-03-23
feat: LT-1447 Template visible to exporter field

BenAndrew-Kainos:
- Added visible_to_exporter field to templates (which may affect the visible_to_exporter field when generating documents)
- Updated existing tests with the create_letter_template function


letter_templates/models.py
```diff
@@ -17,6 +17,7 @@ class LetterTemplate(TimestampableModel):
     letter_paragraphs = SortedManyToManyField(PicklistItem)
     case_types = models.ManyToManyField(CaseType, related_name="letter_templates")
     decisions = models.ManyToManyField(Decision, related_name="letter_templates")
+    visible_to_exporter = models.BooleanField(blank=False, null=False)
```
cstobart: `blank=False` and `null=False` are defaults for `BooleanField`

BenAndrew-Kainos: It's always good to be explicit is it not?

cstobart: 👍

BenAndrew-Kainos: 👍

PR #354, opened 2020-03-24
feat: LT-1276 Admin System Flags

andytaylorkainos:


applications/libraries/edit_applications.py
```diff
@@ -1,7 +1,13 @@
+from django.db.models import Case
```
jfaracik: Wrong import? :P

andytaylorkainos: how the heck did that work :o

andytaylorkainos: Fixed!

applications/views/applications.py
```diff
@@ -520,8 +524,11 @@ def remove_data_from_application_copy(self):
             "submitted_at",
             "licence_duration",
             "is_informed_wmd",
+            "informed_wmd_ref",
```
stephanieboyd: Is this change unrelated to the ticket?

applications/libraries/edit_applications.py
```diff
@@ -108,3 +105,41 @@ def save_and_audit_have_you_been_informed_ref(request, application, serializer):
                     target=application.get_case(),
                     payload={"new_ref_number": new_ref_number},
                 )
+
+
+def set_case_flags_on_submitted_application(application: BaseApplication):
+    if application.case_type.sub_type in [CaseTypeSubTypeEnum.STANDARD, CaseTypeSubTypeEnum.OPEN]:
```
stephanieboyd: The check on application type should probably happen before the call to this function. It leaves the function logic to just setting case flags.

applications/libraries/edit_applications.py
```diff
@@ -108,3 +105,41 @@ def save_and_audit_have_you_been_informed_ref(request, application, serializer):
...
+                case=case, flag_id=SystemFlags.FIREARMS_ID, is_adding=GoodsCategory.FIREARMS in goods_categories,
+            )
+
+
+def _add_or_remove_flag(case: Case, flag_id: str, is_adding: bool):
+    if is_adding:
+        if not case.flags.filter(id=flag_id).exists():
+            case.flags.add(flag_id)
+    else:
+        case.flags.remove(flag_id)
```
stephanieboyd: Could you ever try to remove a flag that doesn't exist? What would happen?

applications/tests/tests_standard_application_submit.py
```diff
@@ -232,3 +234,51 @@ def test_submit_standard_application_without_end_use_details_failure(self):
         self.assertContains(
             response, text=strings.Applications.Generic.NO_END_USE_DETAILS, status_code=status.HTTP_400_BAD_REQUEST,
         )
+
+    def test_submit_standard_application_adds_system_case_flags_success(self):
```
stephanieboyd: Got any tests that show the logic isn't performed on non open/standard cases?

applications/views/applications.py
andytaylorkainos: Yes, i noticed a bug which hasnt been raised

applications/libraries/edit_applications.py
andytaylorkainos: Partially agree considering this is the only use of the function - however i've written it this way so that other case flags for other application types can happen in here (hence the generic name)

applications/libraries/edit_applications.py
andytaylorkainos: flip good catch! - i did have logic to handle that but i removed it mistakenly

applications/tests/tests_standard_application_submit.py
andytaylorkainos: sure thing!

applications/libraries/edit_applications.py
andytaylorkainos: Done!

stephanieboyd: Further discussion had, midway met :)

PR #362, opened 2020-03-25
fix: LT-2333 - Typo in Compliance Actions pick list enum value

cstobart:


picklists/enums.py
```diff
@@ -16,7 +16,7 @@ class PicklistType:
         (STANDARD_ADVICE, "Standard Advice"),
         (FOOTNOTES, "Footnotes"),
         (PRE_VISIT_QUESTIONNAIRE, "Pre-Visit Questionnaire questions (ECJU Query)"),
-        (COMPLIANCE_ACTIONS, "Compliance Actions (ECJU Query"),
+        (COMPLIANCE_ACTIONS, "Compliance Actions (ECJU Query)"),
```
Mark-Spiers-kainos: would it be possible to move these values to strings?

cstobart: They are used by the DB so no

PR #363, opened 2020-03-25
fix: Performance optimisation on flagging rules

andytaylorkainos:


workflow/flagging_rules_automation.py
```diff
@@ -118,20 +118,40 @@ def apply_flagging_rule_to_all_open_cases(flagging_rule: FlaggingRule):
...
+
+            goods = GoodOnApplication.objects.filter(application_id__in=open_cases).values_list("good_id", flat=True)
+            flagging_rule.flag.goods.add(*goods)
 
         elif flagging_rule.level == FlagLevels.DESTINATION:
-            for case in open_cases:
-                apply_destination_flagging_rules_for_case(case, flagging_rule_queryset)
+            end_users = (
+                EndUserAdvisoryQuery.objects.filter(end_user__country_id=flagging_rule.matching_value)
+                .exclude(status__status__in=CaseStatusEnum.terminal_statuses() + [CaseStatusEnum.DRAFT])
```
BenAndrew-Kainos: Could you make CaseStatusEnum.terminal_statuses() + [CaseStatusEnum.DRAFT] a constant

BenAndrew-Kainos:
Are the unit tests for this OK?

workflow/flagging_rules_automation.py
```diff
@@ -118,20 +118,40 @@ def apply_flagging_rule_to_all_open_cases(flagging_rule: FlaggingRule):
...
+            open_cases = open_cases.filter(case_type__reference=flagging_rule.matching_value).values_list(
+                "id", flat=True
+            )
+
+            flagging_rule.flag.cases.add(*open_cases)
 
         elif flagging_rule.level == FlagLevels.GOOD:
-            for case in open_cases:
-                apply_good_flagging_rules_for_case(case, flagging_rule_queryset)
+            good_in_query = GoodsQuery.objects.exclude(
```
efahy: Please add comments throughout explaining the reason why you have to flag a bunch of different things, i.e. the different case types

Also please add a comment to the lines adding the flags to make it clear that you're not really modifying the flag, you're adding the flag to the good/case/destination

workflow/flagging_rules_automation.py
andytaylorkainos: done

workflow/flagging_rules_automation.py
efahy: Please add comments throughout explaining the reason why you have to flag a bunch of different things, i.e. the different case types

Also please add a comment to the lines adding the flags to make it clear that you're not really modifying the flag, you're adding the flag to the good/case/destination

PR #369, opened 2020-03-30
Chore - Performance tests non-destructive

Mark-Spiers-kainos:


phlptt:
good stuff!

* any files which are named tests.py can you name as something as more meaningful?
* in terms of the structure I would prefer if these were all within a seperate tests/performance sub folder and together, but i am happy to go with what others prefer on this
* does this map with any documentation in terms of coverage of endpoints ?

applications/tests/tests_endpoints_response_applications.py
```diff
@@ -0,0 +1,69 @@
...
+        self.call_endpoint(
+            self.get_exporter(),
+            self.url + self.get_standard_application()["id"] + "/parties/" + self.get_party_on_application_id(),
+        )
+
+    def test_application_parties_document(self):
+        self.call_endpoint(
+            self.get_exporter(),
+            self.url
+            + self.get_standard_application()["id"]
```
andytaylorkainos: does reverse_url with kwargs not work here?

test_helpers/test_endpoints/client.py
```diff
@@ -0,0 +1,40 @@
+import requests
+
+from conf.settings import env
+
+
+def get(request, appended_address, is_gov):
+    """
+    :param request: user headers/meta for the api to look at
```
andytaylorkainos: if its headers, why not name the parameter `headers`

also, why pass a `is_gov` bool around everywhere when these headers can be built correctly in their respective `get_user` method

test_helpers/test_endpoints/client.py
```diff
@@ -0,0 +1,40 @@
...
+                "EXPORTER-USER-TOKEN": str(request["EXPORTER-USER-TOKEN"]),
+                "ORGANISATION-ID": request.get("organisation-id", "None"),
+            }
+
+        return requests.get(env("PERFORMANCE_TEST_HOST") + appended_address, headers=headers,)
+
+    return requests.get(env("PERFORMANCE_TEST_HOST") + appended_address)
+
+
+def post(request, appended_address, json):
```
andytaylorkainos: is a `post` non-destructive?

test_helpers/test_endpoints/test_endpoint_response_time.py
```diff
@@ -0,0 +1,330 @@
...
+    flag_id = None
+    flagging_rule_id = None
+    case_id = None
+    case_document = None
+    case_ecju_query_id = None
+
+    def call_endpoint(self, user, appended_address, is_gov=False, save=True):
+        response = get(user, appended_address, is_gov)
+
+        if save:
```
andytaylorkainos: `record_results` or `save_results` instead of `save`?

test_helpers/test_endpoints/test_endpoint_response_time.py
```diff
@@ -0,0 +1,330 @@
...
+            self.appended_address = appended_address
+            self.time = response.elapsed.total_seconds()
+
+        return response
+
+    @classmethod
+    def setUpClass(cls):
+        if len(times) <= 2:
+            # set up headers data in csv if not already done
+            times.append(["env", env("PERFORMANCE_TEST_HOST")])
```
andytaylorkainos: `results` instead of `times`?

test_helpers/test_endpoints/test_endpoint_response_time.py
```diff
@@ -0,0 +1,330 @@
...
+                    response = self.call_endpoint(
+                        self.get_exporter(), "/applications/page=" + str(page + 1), save=False
+                    ).json()
+                    continue
+
+                break
+
+        return self.standard_application_id
+
+    def get_open_application(self):
```
andytaylorkainos: looks like this function is identical to the one above, only difference is `CaseTypeEnum.X` (maybe make it a generic function that takes `case_type` as a parameter?)

applications/tests/tests_endpoints_response_applications.py
Mark-Spiers-kainos: reverse builds against the deployment, but because we are using simpletestcase there is no deployment of the api for it.

test_helpers/test_endpoints/user_setup.py
```diff
@@ -0,0 +1,61 @@
...
+        "EXPORTER-USER-TOKEN": response["token"],
+        "user_id": response["lite_api_user_id"],
+    }
+
+    response = get(request=exporter_user, appended_address=f"/users/me/", is_gov=False)
+    organisation_name = env("PERFORMANCE_EXPORTER_ORGANISATION")
+    if organisation_name:
+        for organisation in response.json()["organisations"]:
+            if organisation["name"] == organisation_name:
+                exporter_user["organisation-id"] = organisation["id"]
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: following the set headers up at first login implementation, it is now ORGANISATION-ID which is the header the api looks at

test_helpers/test_endpoints/client.py
Mark-Spiers-kainos: done

test_helpers/test_endpoints/client.py
Mark-Spiers-kainos: we need the post to log in the users, it is not used anywhere other than login

test_helpers/test_endpoints/test_endpoint_response_time.py
```diff
@@ -0,0 +1,330 @@
...
+    flagging_rule_id = None
+    case_id = None
+    case_document = None
+    case_ecju_query_id = None
+
+    def call_endpoint(self, user, appended_address, is_gov=False, save=True):
+        response = get(user, appended_address, is_gov)
+
+        if save:
+            # if save is set this is the final endpoint we which to hit, and we want to store the time against
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: fixed

test_helpers/test_endpoints/client.py
Mark-Spiers-kainos: also left headers as request, as that is what it is more commonly known as throughout the systems

test_helpers/test_endpoints/test_endpoint_response_time.py
Mark-Spiers-kainos: Have made a function which is called with the case_type, and attribute we which to save it against

test_helpers/test_endpoints/test_endpoint_response_time.py
Mark-Spiers-kainos: renamed to csv_data since it includes env and datetime

test_helpers/test_endpoints/test_endpoint_response_time.py
Mark-Spiers-kainos: updated to save_results

flags/tests/tests_endpoints_response_flags.py
```diff
@@ -0,0 +1,18 @@
+from test_helpers.test_endpoints.test_endpoint_response_time import EndPointTests
+
+
+class LetterTemplatesResponseTests(EndPointTests):
+
```
andytaylorkainos, line 5: small, but can you remove all the blank lines between the class name and the first variable in all files

test_helpers/test_endpoints/test_endpoint_response_time.py
```diff
@@ -0,0 +1,331 @@
...
+        If id is not already set, will request against the "/applications/"
+            endpoint one page at a time until it gets one.
+        :return: open_application_id
+        """
+        if not self.open_application_id:
+            self._get_application_by_case_type(CaseTypeEnum.OIEL.sub_type, "open_application_id")
+
+        return self.open_application_id
+
+    def _get_application_by_case_type(self, case_type, attribute):
```
andytaylorkainos, line 135: very small, should the function be marked as private? seems like a very helpful method

flags/tests/tests_endpoints_response_flags.py
Mark-Spiers-kainos: done

test_helpers/test_endpoints/test_endpoint_response_time.py
Mark-Spiers-kainos: I imagine the function should only really be used by get or fetch functions as it is at the moment, should the attribute not exist, so will leave as private for now, but if it needs opened up it can be.

PR #370, opened 2020-03-30
feat: LT-2334 - Queues

jfaracik:


queues/views/queues.py
```diff
@@ -42,9 +43,12 @@ def get(self, request, pk):
         """
         Retrieve a queue instance
         """
-        queue = get_queue(pk=pk)
+        queue = next(
```
Mark-Spiers-kainos: could you write a comment as to what this is doing?

jfaracik: Updated the docstring :)

PR #374, opened 2020-04-01
chore: Standalone seedorganisations command

andytaylorkainos:


static/management/commands/seedorganisation.py
```diff
@@ -0,0 +1,67 @@
...
+
+    help = "Seeds an organisation with a number of sites and users"
+    info = "Seeding organisation"
+    success = "Successfully seeded organisation"
+    seed_command = "seedorganisation"
+
+    def add_arguments(self, parser):
+        # Named (optional) arguments
+        parser.add_argument("--name", help="Name of organisation to seed", type=str)
+        parser.add_argument("--type", help="Type of organisation to seed", type=str)
```
efahy, line 31: The help could be more helpful if it listed the valid types, assuming that list isn't too long

static/management/commands/seedorganisation.py
```diff
@@ -0,0 +1,67 @@
...
+        org_name = options.get("name") or faker.company()
+        org_type = options.get("type") or OrganisationType.COMMERCIAL
+        no_of_sites = options.get("sites") or 1
+        no_of_users = options.get("users") or 1
+
+        self.seed_organisation(org_name, org_type, no_of_sites, no_of_users)
+
+    @classmethod
+    def seed_organisation(cls, org_name, org_type, no_of_sites, no_of_users):
+        if not Organisation.objects.filter(name__iexact=org_name).exists():
```
efahy, line 46: If the org does exist, could we have a console message to say so?

PR #375, opened 2020-04-01
feat: LT-1254 View Licences

BenAndrew-Kainos:


test_helpers/clients.py
```diff
@@ -840,13 +849,18 @@ def create_ecju_query(self, case, question="ECJU question", gov_user=None):
         return ecju_query
 
     @staticmethod
-    def create_licence(application: BaseApplication, is_complete: bool):
-        return Licence.objects.create(
+    def create_licence(application: BaseApplication, is_complete: bool, decisions=None):
```
jfaracik, line 852: Could you make this a factory instead? So LicenceFactory

jfaracik:
Some good changes! (especially around related_names) Good job!

test_helpers/clients.py
BenAndrew-Kainos: Is this a thing we do now? I haven't done it before, could you do a tech share

PR #376, opened 2020-04-01
feat: LT-1097 - HMRC SLA

andytaylorkainos:


cases/tests/tests_sla.py
```diff
@@ -196,3 +198,28 @@ def test_bank_holiday_weekend_ignored(self, weekend_func, bank_holiday_func):
 
         # Expecting update_cases_sla to be ran, but no cases found
         self.assertEqual(result, 0)
+
+
+class SlaHmrcCaseTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        self.hmrc_query = self.create_hmrc_query(self.organisation)
```
a-bek, line 206: Worth testing that another case doesn't have the SLA populated?

common/dates.py
```diff
@@ -87,3 +91,22 @@ def number_of_days_since(date, num_working_days):
         date = date - timedelta(days=1)
 
     return days
+
+
+def number_of_hours_since(start_date, end_date):
```
a-bek, line 96: also `num_working_hours_in_range` might be the one

common/dates.py
```diff
@@ -20,6 +24,29 @@ def working_days_in_range(start_date, end_date):
     return len([date for date in dates_in_range if not is_bank_holiday(date) or not is_weekend(date)])
 
 
+def working_hours_in_range(start_date, end_date):
+    hours_count = 0
+
+    # If start_date is a working day, add the remaining hours left on that day
+    if is_working_day(start_date):
+        hours_count += 24 - start_date.hour
```
efahy, line 32: It would be worth asking the business if they care whether these SLA hours are rounded up or down.  At the moment, by my reckoning you're rounding up here, e.g. if the case was submitted at 15:45, this line will give 9 hours, even though more accurately its 8 hours 15 mins

common/dates.py
```diff
@@ -20,6 +24,29 @@ def working_days_in_range(start_date, end_date):
...
+
+    # If end_date is a working day, add the hours on that day
+    if is_working_day(end_date):
+        hours_count += end_date.hour
```
efahy, line 36: And actually, this line rounds down, so for 15:45, this line gives 15.  Again, double check with the business

common/dates.py
```diff
@@ -20,6 +24,29 @@ def working_days_in_range(start_date, end_date):
...
+
+    elapsed_days = end_date.day - start_date.day
+
+    # Add 24 hours for every day that elapsed between (but not including) end_date and start_date
```
efahy, line 40: Make it clear that its only working days in the comment

common/tests/test_dates.py
```diff
@@ -45,3 +45,42 @@ def test_num_working_days_since(self, test_date, num_working_days, expected_resu
         result = number_of_days_since(test_date, num_working_days)
 
         self.assertEqual(result, expected_result)
+
+    @parameterized.expand(
+        [
+            (datetime(2020, 4, 1), datetime(2020, 4, 3), 48),
+            (datetime(2020, 4, 1, 14), datetime(2020, 4, 3, 14), 48),
+            (datetime(2020, 4, 1, 14), datetime(2020, 4, 3), 34),
+            (datetime(2020, 4, 1), datetime(2020, 4, 3, 14), 62),
```
efahy, line 54: Going back to the rounding up/down I would include some datetimes with minutes and even seconds

common/tests/test_dates.py
```diff
@@ -45,3 +45,42 @@ def test_num_working_days_since(self, test_date, num_working_days, expected_resu
...
+
+    @parameterized.expand(
+        [
+            (datetime(2020, 4, 10), datetime(2020, 4, 14), 0),
+            (datetime(2020, 4, 9), datetime(2020, 4, 14), 24),
+            (datetime(2020, 4, 9, 12), datetime(2020, 4, 14), 12),
+            (datetime(2020, 4, 9, 12), datetime(2020, 4, 14, 15), 27),
+        ]
+    )
+    def test_num_working_hours_over_bank_holidays(self, start_date, end_date, expected_result):
```
efahy, line 83: Might also be worth making a test over several weeks that has all the complexities and, for example, multiple weekends?

common/dates.py
andytaylorkainos: Good call. we're to floor the hour

common/dates.py
```diff
@@ -20,6 +32,29 @@ def working_days_in_range(start_date, end_date):
     return len([date for date in dates_in_range if not is_bank_holiday(date) or not is_weekend(date)])
 
 
+def working_hours_in_range(start_date, end_date):
```
efahy, line 35: Please comment explaining why you're working in seconds

common/dates.py
```diff
@@ -20,6 +32,29 @@ def working_days_in_range(start_date, end_date):
...
+    seconds_count = (end_date - start_date).total_seconds()
+
+    # If start_date is a non-working day, subtract the total seconds that were remaining on that day
+    if not is_working_day(start_date):
+        seconds_count -= SECONDS_IN_DAY - get_total_seconds_from_date_time(start_date)
```
efahy, line 40: `get_time_in_seconds_from_datetime`

PR #377, opened 2020-04-02
chore: Cases optimisation

jfaracik:
The `/cases/` page is awful. Let's fix that.

∞ queries down to < 30 queries (excluding flags - flag optimisation will come in the flags colours story).

### DIFF
--- a/cases/fields.py
+++ b/cases/fields.py
@@ -7,6 +7,7 @@ class CaseAssignmentRelatedSerializerField(PrimaryKeyRelatedSerializerField):
     """
     Custom field serializer required for optional query filtering based on current queue
     """
+
     def __init__(self, **kwargs):
         from cases.serializers import QueueCaseAssignmentSerializer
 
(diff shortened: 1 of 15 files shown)


cases/fields.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            return queryset.all()
+
+
+class HasOpenECJUQueriesRelatedField(PrimaryKeyRelatedField):
+    """
+    Custom field serializer required to determine if a case has open ECJU queries
+    """
+
+    def use_pk_only_optimization(self):
```
Mark-Spiers-kainos, line 34: I had to look this up to understand it, may be worthwhile to show to people?

jfaracik: Sure!

conf/middleware.py
```diff
@@ -1,8 +1,20 @@
...
+
 from django.db import connection
 
+from conf.settings import env
+from test_helpers.colours import bold
+
+
+#
+#
+# Reset to master copy when done
```
Mark-Spiers-kainos, line 14: remember to set back

cases/fields.py
efahy: A comment wouldn't hurt either

PR #378, opened 2020-04-02
LT-2354: Audit trail MVP

a-bek:
MVP activity stream

Creates a paginated json response of activities caught by the audit trail.

Current case activities that are handled for MVP:

- create
- status_update
- case_officers
- countries

The feed is accessed on API - `/audit-trail/streams/0`

Sample outputs:

Screenshot 2020-04-09 at 09 00 13

Specs:
https://uktrade.atlassian.net/wiki/spaces/ILT/pages/879951988/Design+-+Data+Reporting


common/models.py
```diff
@@ -15,3 +17,71 @@ class TimestampableModel(CreatedAt):
...
+            if related_content_type not in data.keys():
+                data[related_content_type] = []
+            data[related_content_type].append(related_object_id)
+
+    for content_type, object_ids in data.items():
+        model_class = content_type.model_class()
+        models = prefetch_generic_relations(model_class.objects.filter(pk__in=object_ids))
+
+        for model in models:
+            for weak_model in qs:
```
andytaylorkainos, line 67: seems like this entire nested for-loop is duplicated -> any chance for a function?

andytaylorkainos, line 67: seems like this is a duplicated for-loop. Any chance it can be moved to a function?

audit_trail/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+
+from audit_trail.streams import views
+
+app_name = "audit_trail"
+
+urlpatterns = [
+    path("streams/<int:n>", views.streams, name="streams"),
```
andytaylorkainos, line 8: are paginated URLs not usually handled with queryparams `?page=0`?

a-bek: Yeah but the requirement is to have it embedded in the url, not in the query params.

queries/end_user_advisories/views.py
```diff
@@ -90,15 +90,16 @@ def put(self, request, pk):
 
         request.data["status"] = get_case_status_by_status(data.get("status"))
 
+        old_status = end_user_advisory.status.status
         serializer = EndUserAdvisoryViewSerializer(end_user_advisory, data=request.data, partial=True)
 
         if serializer.is_valid():
```
iichr, line 96: I reckon `serializer.is_valid(raise_exception=True)` might make it tidier

common/models.py
a-bek: yeah I know, but given that there are exit conditions to end the loop iteration (within the duplicated segments), and it won't be used anywhere else, I would prefer just leaving it as is.

audit_trail/streams/views.py
```diff
@@ -0,0 +1,24 @@
...
+from audit_trail.streams.service import get_stream
+
+
+def get_next_page_url(request, n):
+    return request.build_absolute_uri("/").strip("/") + "/audit-trail/streams/{n}".format(n=n)
+
+
+@api_view(["GET"])
+def streams(request, n):
+    data = get_stream(n)
```
andytaylorkainos, line 13: really picky.

`stream = get_stream`

 makes more sense when you extract the data out of the dictionary on line 21:
```stream["data"]```

File: audit_trail/streams/service.py [hunk 5]
```diff
@@ -0,0 +1,164 @@
...
+    return {
+        "id": "dit:lite:case:change:{data_type}:{id}:{audit_id}:{verb}".format(
+            data_type=data_type, id=case.id, audit_id=audit.id, verb=verb
+        ),
+        "published": "{ts}".format(ts=audit.created_at),
+        "object": object_data,
+    }
+
+
+def get_stream(n):
```
andytaylorkainos, line 109: can you use a more meaningful parameter name please

audit_trail/streams/service.py
```diff
@@ -0,0 +1,164 @@
...
+    qs = prefetch_generic_relations(audit_qs)
+
+    case_ids = [value["target_object_id"] for value in qs.values("target_object_id")]
+
+    # Prefetch relevant information in bulk to be used for streams
+    countries_on_applications = CountryOnApplication.objects.filter(application__id__in=case_ids).values(
+        "application_id", "country__name"
+    )
+
+    case_types = Case.objects.filter(id__in=case_ids).select_related("case_type").values("id", "case_type__sub_type")
```
andytaylorkainos, line 136: not sure if an oversight, but this would return a sub_type of `application` for an `HMRC` query (HMRC queries are of type application in our database)

Is this a set-in-stone requirement? if not, you could just use `case_type__reference` (which would show ```SIEL/OGEL/CRE etc```

PR #380, opened 2020-04-06
feat: LT-1063 - define routing rules

Mark-Spiers-kainos:


workflow/routing_rules/models.py
```diff
@@ -0,0 +1,38 @@
...
+from teams.models import Team
+from users.models import GovUser
+from workflow.routing_rules.enum import RoutingRulesAdditionalFields
+
+
+class RoutingRule(TimestampableModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    team = models.ForeignKey(Team, on_delete=models.CASCADE)
+    queue = models.ForeignKey(Queue, related_name="routing_queue", on_delete=models.DO_NOTHING,)
+    status = models.ForeignKey(CaseStatus, related_name="routing_status", on_delete=models.DO_NOTHING,)
```
andytaylorkainos, line 21: pretty sure that the `related_name` is what is used for a reverse-lookup e.g.
`status.routing_rule`

So can you changed the related name to be `routing_rule`?

workflow/routing_rules/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+                self.initial_data["country"] = None
+
+            if RoutingRulesAdditionalFields.FLAGS in additional_rules:
+                self.fields["flags"].required = True
+                self.fields["flags"].allow_null = False
+                self.fields["flags"].allow_empty = False
+            else:
+                self.initial_data["flags"] = []
+
+    def update(self, instance, validated_data):
```
andytaylorkainos, line 104: do we need to override the update?

workflow/routing_rules/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.urls import path
+
+from workflow.routing_rules import views
+
+app_name = "routing_rules"
+
+urlpatterns = [
+    path("", views.RoutingRulesList.as_view(), name="list"),
+    path("<uuid:pk>/", views.RoutingRulesDetail.as_view(), name="detail"),
+    path("<uuid:pk>/status/<str:status>/", views.RoutingRulesActiveStatus.as_view(), name="active_status"),
```
andytaylorkainos, line 10: if `<str:status>` is the value you intend to use to update your `routing_rule` can you please change this so that the status is sent as part of the body, not the URL

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,94 @@
...
+
+class RoutingRulesActiveStatus(APIView):
+    def put(self, request, pk, status):
+
+        if status != "deactivate" and status != "reactivate":
+            raise Http404
+
+        active_status = True if status == "reactivate" else False
+
+        routing_rule = RoutingRule.objects.get(id=pk)
```
andytaylorkainos, line 84: gonna need to handle this so that if the routing rule doesnt exst, the API should still return a 404

workflow/routing_rules/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+            "case_types",
+            "country",
+            "flags",
+            "case_types",
+            "additional_rules",
+            "active",
+        )
+
+    def __init__(self, *args, **kwargs):
+        # set fields to required or not depending on array passed forward
```
andytaylorkainos, line 68: can you be explicit on what array you're referring to

workflow/routing_rules/models.py
```diff
@@ -0,0 +1,38 @@
...
+from static.countries.models import Country
+from static.statuses.models import CaseStatus
+from teams.models import Team
+from users.models import GovUser
+from workflow.routing_rules.enum import RoutingRulesAdditionalFields
+
+
+class RoutingRule(TimestampableModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    team = models.ForeignKey(Team, on_delete=models.CASCADE)
```
BenAndrew-Kainos, line 19: Could you add a related name in case we want to do a reverse lookup on this in the future

workflow/routing_rules/serializers.py
BenAndrew-Kainos: I agree, seems odd we override this

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,94 @@
...
+            filtered_qs = filtered_qs.filter(active=filter_data.get("only_active"))
+
+        return filtered_qs
+
+    def get(self, request, *args, **kwargs):
+        assert_user_has_permission(request.user, GovPermissions.MANAGE_TEAM_ROUTING_RULES)
+        return super().get(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        assert_user_has_permission(request.user, GovPermissions.MANAGE_TEAM_ROUTING_RULES)
```
BenAndrew-Kainos, line 45: This can just be in the dispatch (then you also won't need to define the GET and POST)

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,94 @@
...
+        return super().post(request, *args, **kwargs)
+
+
+class RoutingRulesDetail(RetrieveUpdateAPIView):
+    authentication_classes = (GovAuthentication,)
+
+    def get(self, request, *args, **kwargs):
+        assert_user_has_permission(request.user, GovPermissions.MANAGE_TEAM_ROUTING_RULES)
```
BenAndrew-Kainos, line 53: Same as above

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,94 @@
...
+            return EditRoutingRuleSerializer
+        else:
+            # heavy serializer for validating, saving and getting list of objects
+            return RoutingRuleSerializer
+
+    def get_queryset(self):
+        return RoutingRule.objects.filter(id=self.kwargs["pk"])
+
+    def perform_update(self, serializer):
+        if not self.request.data.get("validate_only", False):
```
BenAndrew-Kainos, line 72: Where is this used?

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,94 @@
...
+            serializer.save()
+
+
+class RoutingRulesActiveStatus(APIView):
+    def put(self, request, pk, status):
+
+        if status != "deactivate" and status != "reactivate":
+            raise Http404
+
+        active_status = True if status == "reactivate" else False
```
BenAndrew-Kainos, line 82: This can just be `status == "reactivate"`

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,94 @@
...
+
+        routing_rule = RoutingRule.objects.get(id=pk)
+
+        if routing_rule.active == active_status:
+            return JsonResponse(status=HTTP_400_BAD_REQUEST, data={"error": "status already set"})
```
BenAndrew-Kainos, line 87: content

workflow/routing_rules/serializers.py
Mark-Spiers-kainos: It wasn't saving the changes when I didn't.

workflow/routing_rules/views.py
Mark-Spiers-kainos: made a helper function for future

workflow/routing_rules/serializers.py
Mark-Spiers-kainos: updated comment

workflow/routing_rules/views.py
Mark-Spiers-kainos: the `assert_user_has_permission` function doesn't work in dispatch, the user is of class AnonymousUser which doesn't have a has_permission() attribute

workflow/routing_rules/views.py
Mark-Spiers-kainos: This is used as part of the UpdateModelMixin, after validating the serializer, it runs the function above, so I just introduced a step to only save changes if validate_only is not given.

workflow/routing_rules/models.py
Mark-Spiers-kainos: done

workflow/routing_rules/helpers.py
```diff
@@ -0,0 +1,9 @@
+from conf.exceptions import NotFoundError
+from workflow.routing_rules.models import RoutingRule
+
+
+def get_routing_rule(id):
+    try:
+        return RoutingRule.objects.get(id=id)
+    except RoutingRule.DoesNotExist:
+        raise NotFoundError({"RoutingRule": "routing rule not found"})
```
andytaylorkainos, line 9: helloo content :)

workflow/routing_rules/models.py
```diff
@@ -0,0 +1,38 @@
...
+    team = models.ForeignKey(Team, related_name="routing_rules", on_delete=models.CASCADE)
+    queue = models.ForeignKey(Queue, related_name="routing_rules", on_delete=models.DO_NOTHING,)
+    status = models.ForeignKey(CaseStatus, related_name="routing_rules", on_delete=models.DO_NOTHING,)
+    tier = models.PositiveSmallIntegerField()  # positive whole number, that decides order routing rules are applied
+    additional_rules = SeparatedValuesField(
+        choices=RoutingRulesAdditionalFields.choices, max_length=100, blank=True, null=True, default=None
+    )
+    active = models.BooleanField(default=True)
+
+    # optional fields that entry is dependent on additional fields
```
andytaylorkainos, line 28: i dont really understand the comment - can you re-phrase it?

workflow/routing_rules/serializers.py
andytaylorkainos: 🤔  thats odd

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,96 @@
...
+        if not self.request.data.get("validate_only", False):
+            serializer.save()
+
+
+class RoutingRulesActiveStatus(APIView):
+    def put(self, request, pk):
+        status = request.data.get("status")
+
+        if status != "deactivate" and status != "reactivate":
+            raise Http404
```
andytaylorkainos, line 82: might be worth explaining exactly what isnt found

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class RoutingRulesList(ListCreateAPIView):
+    authentication_classes = (GovAuthentication,)
+    serializer_class = RoutingRuleSerializer
+    queryset = RoutingRule.objects.all()
+
+    def filter_queryset(self, queryset):
+        filtered_qs = queryset
+        # get each filter
```
andytaylorkainos, line 21: comment not really needed - code reads easily

flags/views.py
```diff
@@ -45,6 +46,7 @@ def get_queryset(self):
         priority = self.request.GET.get("priority")
         team = self.request.GET.get("team")
         only_show_deactivated = optional_str_to_bool(self.request.GET.get("only_show_deactivated"))
+        include_system_flags = self.request.GET.get("include_system_flags")  # True, False
```
andytaylorkainos, line 49: friendly_boolean?

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,118 @@
...
+    def perform_update(self, serializer):
+        if not self.request.data.get("validate_only", False):
+            serializer.save()
+
+
+class RoutingRulesActiveStatus(APIView):
+    def put(self, request, pk):
+        status = request.data.get("status")
+
+        if status != "deactivate" and status != "reactivate":
```
andytaylorkainos, line 103: do we not have enums/constants for these?

workflow/routing_rules/serializers.py
BenAndrew-Kainos: I would investigate this further before doing this hacky fix

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,119 @@
...
+    def get(self, request, *args, **kwargs):
+        if request.user.has_permission(GovPermissions.MANAGE_TEAM_ROUTING_RULES) or request.user.has_permission(
+            GovPermissions.MANAGE_ALL_ROUTING_RULES
+        ):
+            return super(RoutingRulesList, self).get(request, *args, **kwargs)
+        else:
+            raise PermissionDeniedError()
+
+    def post(self, request, *args, **kwargs):
+        if request.user.has_permission(GovPermissions.MANAGE_TEAM_ROUTING_RULES) or request.user.has_permission(
```
BenAndrew-Kainos, line 54: This can be in a dispatch

workflow/routing_rules/views.py
```diff
@@ -0,0 +1,119 @@
...
+    def get(self, request, *args, **kwargs):
+        if request.user.has_permission(GovPermissions.MANAGE_TEAM_ROUTING_RULES) or request.user.has_permission(
+            GovPermissions.MANAGE_ALL_ROUTING_RULES
+        ):
+            return super(RoutingRulesDetail, self).get(request, *args, **kwargs)
+        else:
+            raise PermissionDeniedError()
+
+    def put(self, request, *args, **kwargs):
+        if request.user.has_permission(GovPermissions.MANAGE_TEAM_ROUTING_RULES) or request.user.has_permission(
```
BenAndrew-Kainos, line 74: This can be in a dispatch

workflow/routing_rules/views.py
BenAndrew-Kainos: Could you leave a comment please

workflow/routing_rules/serializers.py
efahy: I would also like us to understand why this is needed if it's just default behaviour and ideally get rid of it if at all possible

workflow/routing_rules/views.py
Mark-Spiers-kainos: It can't be in a dispatch cause the user class is unknown. Part of the dispatch is authenticating the user which gives us GovUser/ExporterUser

workflow/routing_rules/serializers.py
Mark-Spiers-kainos: I think I had fixed the issue later without thinking about it in the init. One of the problems I had is some data is optional and you may not get to it's forms, so it wasn't overriding the data, but the init now sets empty initial data for the values, if they are not required. Have removed.

workflow/routing_rules/views.py
Mark-Spiers-kainos: done

PR #381, opened 2020-04-06
feat: LT-2270 - Optional document upload

iichr:


applications/creators.py
```diff
@@ -71,12 +71,18 @@ def check_party_error(party, object_not_found_error, is_mandatory, is_document_m
 
 def _validate_end_user(draft, errors, is_mandatory):
```
andytaylorkainos, line 72: ```is_mandatory = not (draft.case_type.sub_type == CaseTypeSubTypeEnum.STANDARD and draft.export_type == ApplicationExportType.TEMPORARY)```

iichr: had to add one more condition as the ticket changed, see updated

PR #383, opened 2020-04-08
feat: LT-893 - updated exporter dashboard

rorysproule1:


applications/views/applications.py
```diff
@@ -148,6 +144,20 @@ def post(self, request, **kwargs):
         return JsonResponse(data={"id": application.id}, status=status.HTTP_201_CREATED)
 
 
+class ApplicationExisting(APIView):
+    authentication_classes = (ExporterAuthentication,)
+
+    def get(self, request):
+        organisation = self.request.user.organisation
+        if organisation.type != "hmrc":
```
cstobart, line 152: given you have the else, more conventional to if is HMRC -> get queries, else -> get applications/licences

rorysproule1: Makes sense, added :)

applications/views/applications.py
```diff
@@ -148,6 +144,20 @@ def post(self, request, **kwargs):
         return JsonResponse(data={"id": application.id}, status=status.HTTP_201_CREATED)
 
 
+class ApplicationExisting(APIView):
```
efahy, line 147: Please add a docstring comment explaining what this view returns, because its not immediately obvious from the name

rorysproule1: Added docstring

PR #385, opened 2020-04-08
bugfix: LT-2278 - Advice entities

andytaylorkainos:


cases/helpers.py
```diff
@@ -132,13 +132,48 @@ def create_grouped_advice(case, request, advice, level):
...
+
+    final_advice = FinalAdvice.objects.distinct(*FinalAdvice.ENTITIES).filter(case=case)
+
+    if advice_type:
+        final_advice = final_advice.filter(type=advice_type)
+
+    final_advice_entities = defaultdict(list)
+
+    for advice in final_advice:
+        for entity_name in FinalAdvice.ENTITIES:
```
a-bek, line 169: Why not use the entity property on the Advice model we created below?

andytaylorkainos: Because then i lose the ability to map the entity to different serializers :/

a-bek: Could map it by object type returned by entity?

cases/helpers.py
```diff
@@ -132,13 +132,45 @@ def create_grouped_advice(case, request, advice, level):
...
+        final_advice = final_advice.filter(type=advice_type)
+
+    final_advice_entities = defaultdict(list)
+
+    for advice in final_advice:
+        serializer = entity_field_serializer_map[advice.entity_field]
+        data = serializer(advice.entity).data
+        final_advice_entities[advice.entity_field].append(data)
+
+    return final_advice_entities
```
BenAndrew-Kainos, line 173: Does this function return items in lists only for objects which there may be multiple (i.e. goods)

cases/managers.py
```diff
@@ -245,3 +246,21 @@ def create(self):
...
         case_reference_code.save()
         return case_reference_code
+
+
+class AdviceManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().prefetch_related(*self.model.ENTITY_FIELDS)
+
+    def get(self, *args, **kwargs):
+        entity_id = kwargs.pop("entity_id", None)
```
BenAndrew-Kainos, line 256: Could you add a docstring for this function & its usage

cases/helpers.py
andytaylorkainos: returns a dictionary containing lists as the values

cases/managers.py
andytaylorkainos: done!

PR #387, opened 2020-04-08
bug: Site Members count & list

BenAndrew-Kainos:
Fix the members list & count for sites


organisations/serializers.py
```diff
@@ -21,25 +21,50 @@
 
 class SiteListSerializer(serializers.ModelSerializer):
     address = AddressSerializer()
+    assigned_users_count = serializers.SerializerMethodField()
+
+    def get_assigned_users_count(self, instance):
+        return (
+            UserOrganisationRelationship.objects.filter(
```
efahy, line 28: Can you add a wee comment explaining that you're getting the total of members of the site plus admin users, with duplicates removed?

organisations/serializers.py
```diff
@@ -21,25 +21,50 @@
...
+    admin_users = serializers.SerializerMethodField()
 
     def get_users(self, instance):
-        users = UserOrganisationRelationship.objects.filter(
-            Q(sites__id__exact=instance.id)
-            | Q(organisation=instance.organisation, role__permissions__id=ExporterPermissions.ADMINISTER_SITES.name)
-        ).select_related("user")
+        users = (
+            UserOrganisationRelationship.objects.filter(sites__id=instance.id)
+            .distinct()
```
efahy, line 48: Do we need all these distincts?

organisations/serializers.py
```diff
@@ -21,25 +21,50 @@
...
+    assigned_users_count = serializers.SerializerMethodField()
+
+    def get_assigned_users_count(self, instance):
+        return (
+            UserOrganisationRelationship.objects.filter(
+                Q(sites__id__exact=instance.id)
+                | Q(organisation=instance.organisation, role__permissions__id=ExporterPermissions.ADMINISTER_SITES.name)
+            )
+            .distinct()
+            .count()
```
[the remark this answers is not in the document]
BenAndrew-Kainos: It is not and unfortunately there are usages in the code that depend on it not being paginated. I am raising a bug ticket for this now

organisations/serializers.py
BenAndrew-Kainos: Done

organisations/serializers.py
BenAndrew-Kainos: Done

organisations/serializers.py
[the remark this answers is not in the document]
BenAndrew-Kainos: https://uktrade.atlassian.net/browse/LT-2412

organisations/serializers.py
```diff
@@ -21,25 +21,49 @@
 
 class SiteListSerializer(serializers.ModelSerializer):
     address = AddressSerializer()
+    assigned_users_count = serializers.SerializerMethodField()
+
+    def get_assigned_users_count(self, instance):
+        # Total number of distinct standard & admin users for the site
+        return (
```
a-bek, line 28: This will run `n` additional queries where `n` is the number of sites.

Populating `assigned_users_count` needs to be done in batch with a list of the sites, rather than per site.

One way of handling it is how sla hours are added to a list of cases: https://github.com/uktrade/lite-api/blob/master/cases/views/search/views.py#L53

BenAndrew-Kainos: That code is still running `n` queries: 

`[for id in HmrcQuery.objects.filter(id__in=hmrc_cases, have_goods_departed=False).values_list("id", flat=True)]`

a-bek: Thats 1 query

organisations/views/sites.py
```diff
@@ -40,7 +40,19 @@ def get(self, request, org_pk):
...
 
-        service.populate_assigned_users_count(org_pk, serializer_data)
+        if str_to_bool(request.GET.get("get_total_users")):
+            admin_users = UserOrganisationRelationship.objects.filter(
+                organisation=organisation, role__permissions__id=ExporterPermissions.ADMINISTER_SITES.name
+            ).values_list("user__id", flat=True)
+            total_admin_users = len(admin_users)
+
+            for site in serializer_data:
+                site["assigned_users_count"] = (
```
andytaylorkainos, line 50: it would be better if you annotated the user count into query on line 33/37 to avoid n+1 queries

andytaylorkainos, line 50: it would be better if you annotated the user count into the queries on lines 33+37 to avoid n+1 queries.

you would have to change the serializer to expect being passed `user_count` as part of the data and the `total_admin_users` as an argument so that you could do `user_count+=total_admin_users`

i.e. (pseudo)
```
site = Site.objects.filter(organisation=organisation).annotate(count= .... )

serializer(site, total_admin_count)
```

organisations/serializers.py
BenAndrew-Kainos: Done. Thanks for your help :)

organisations/views/sites.py
BenAndrew-Kainos: Already done. Thank you

organisations/views/sites.py
```diff
@@ -40,7 +41,24 @@ def get(self, request, org_pk):
...
         serializer_data = SiteListSerializer(sites, many=True).data
 
-        service.populate_assigned_users_count(org_pk, serializer_data)
+        if str_to_bool(request.GET.get("get_total_users")):
+            admin_users = UserOrganisationRelationship.objects.filter(
+                organisation=organisation, role__permissions__id=ExporterPermissions.ADMINISTER_SITES.name
+            ).values_list("user__id", flat=True)
+            total_admin_users = len(admin_users)
+
+            relationships = (
```
a-bek, line 50: 💪

PR #388, opened 2020-04-09
fix: LT-2240 - Filter by case officer

jfaracik:


a-bek:
r_1368694_YwK2T

jfaracik:
Ahem we have test

PR #390, opened 2020-04-14
feat: LT-???? - Case redesign

jfaracik:
Still in progress

### DIFF
--- a/cases/models.py
+++ b/cases/models.py
@@ -1,4 +1,5 @@
 import uuid
+from collections import defaultdict
 
 from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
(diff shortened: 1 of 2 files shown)


cases/tests/factories.py
```diff
@@ -0,0 +1,34 @@
+import factory
+
+from cases.enums import AdviceLevel, AdviceType
+from cases.models import Advice
+
+
+class UserAdviceFactory(factory.django.DjangoModelFactory):
```
iichr, line 7: 👏

cases/tests/factories.py
```diff
@@ -0,0 +1,34 @@
...
+    text = factory.Faker("word")
+    note = factory.Faker("word")
+    type = AdviceType.APPROVE
+    level = AdviceLevel.TEAM
+
+    class Meta:
+        model = Advice
+
+
+class FinalAdviceFactory(factory.django.DjangoModelFactory):
```
iichr, line 27: Very much overkill for this but am I right in thinking there could be a common abstract class, say AdviceFactory, which all of these inherit the concrete implementation of common elements such as notes/text from and override the rest?

conf/middleware.py
```diff
@@ -15,17 +15,17 @@ def __call__(self, request):
...
-                "user": request.user.id,
-                "message": "liteolog api",
-                "corrID": request.correlation,
-                "type": "http response",
-                "method": request.method,
-                "url": request.path,
-                "elapsed_time": time.time() - start,
-            }
-        )
+        # logging.info(
```
iichr, line 18: uncomment at the end

PR #402, opened 2020-04-21
feat: LT-2340 - HMRC Queries Ordering

andytaylorkainos:


cases/tests/tests_search_cases.py
```diff
@@ -446,10 +445,81 @@ def test_all_cases_queue_returns_cases_in_expected_order(self):
...
+
+    def test_work_queue_returns_cases_in_expected_order(self):
+        """Test that a work queue returns cases in expected order (hmrc queries with goods not departed first)."""
+        clc_query_1 = self.create_clc_query("Example CLC Query", self.organisation)
+        standard_app = self.create_standard_application_case(self.organisation, "Example Application")
+        hmrc_query_1 = self.submit_application(self.create_hmrc_query(self.organisation))
+        clc_query_2 = self.create_clc_query("Example CLC Query 2", self.organisation)
+        hmrc_query_2 = self.submit_application(self.create_hmrc_query(self.organisation, have_goods_departed=True))
+
+        self.queue.cases.add(clc_query_1)
```
jfaracik, line 462: `self.queue.cases.set([clc_query_1, standard_app, ...])`

jfaracik: You don't need to do queue.save after either

jfaracik: I'd give you a brownie point if you instead made a QueueFactory with the ability to set cases directly

andytaylorkainos: Done!

queues/tests/factories.py
```diff
@@ -0,0 +1,25 @@
+import factory
+from queues.models import Queue
+from teams.models import Team
+
+
+class QueueFactory(factory.django.DjangoModelFactory):
+    name = factory.Faker("word")
+    team = factory.SubFactory("queues.tests.factories.TeamFactory")
```
jfaracik, line 8: Great job!

cases/managers.py
```diff
@@ -77,7 +77,7 @@ def order_by_status(self, order=""):
 
         return self.order_by(f"{order}status__priority")
 
-    def order_by_date(self, order=""):
+    def order_by_date(self, order="-"):
```
BenAndrew-Kainos, line 80: Does this not change the default for all order by dates?

cases/managers.py
```diff
@@ -174,8 +175,18 @@ def search(
                 case_officer = None
             case_qs = case_qs.assigned_as_case_officer(user=case_officer)
 
-        if isinstance(date_order, str):
-            case_qs = case_qs.order_by_date(date_order)
+        if is_work_queue:
+            case_qs = case_qs.annotate(
+                case_order=Case(
+                    When(baseapplication__hmrcquery__have_goods_departed=False, then=1),
```
BenAndrew-Kainos, line 181: When shouldn't be needed here I don't think. Maybe try; `.annotate(case_order=Value(baseapplication__hmrcquery__have_goods_departed, BooleanField()))` (might need quotes around baseapplication__hmrcquery__have_goods_departed)

Even if that doesn't work you can change the output_field to boolean

queues/tests/tests_performance.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    def test_queue_case_ordering_performance(
+        self, std_cases, open_cases, hmrc_queries_goods_gone, hmrc_queries_goods_in_uk
+    ):
+        """
+        Test various combinations of case/application types to ensure acceptable performance on non-system
+        queue page or highlight areas for concern
+        """
+        self.create_assorted_cases(std_cases, open_cases, hmrc_queries_goods_gone, hmrc_queries_goods_in_uk)
+        self.timeit(self._make_queue_request)
```
BenAndrew-Kainos, line 31: This is a different format to our existing performance tests but is still a useful addition :)

cases/managers.py
andytaylorkainos: Yes, but we only use this sort if its a non-work queue (open + all cases) which means the default should be `"-"`

cases/managers.py
andytaylorkainos: This doesnt work because it doesn't know how to evaluate "Null" (not all cases have that field)

````
ValidationError: ["'baseapplication__hmrcquery__have_goods_departed' value must be either True or False."]
```

andytaylorkainos: we got there 👍

PR #403, opened 2020-04-21
bug: Picklists performance

BenAndrew-Kainos:
Load time for 5000 Picklists reduced from 7.56s to 0.03s (when paginated) or 0.27s (when not paginated for input lists)


picklists/serializers.py
```diff
@@ -1,15 +1,52 @@
...
 
 from conf.serializers import KeyValueChoiceField
 from picklists.enums import PicklistType, PickListStatus
 from picklists.models import PicklistItem
 from teams.models import Team
+from teams.serializers import TeamReadOnlySerializer
 
 
-class PicklistSerializer(ModelSerializer):
+class PicklistCondensedListSerializer(ModelSerializer):
```
a-bek, line 13: TinyPicklistSerializer?

picklists/serializers.py
```diff
@@ -1,15 +1,52 @@
...
+        model = PicklistItem
+        fields = (
+            "id",
+            "name",
+            "text",
+            "updated_at",
+        )
+
+
+class PicklistListSerializer(ModelSerializer):
```
a-bek, line 28: Can this just inherit from TinyPicklistSerializer

conf/custom_views.py
```diff
@@ -0,0 +1,18 @@
+from django.http import JsonResponse
+from rest_framework.generics import ListAPIView
+
+from conf.helpers import str_to_bool
+
+
+class OptionallyPaginatedEndpoint(ListAPIView):
```
a-bek, line 7: Can this be PaginatedAPIView? I don't think its necessary to embed the the param name behaviour in the class name?

picklists/serializers.py
BenAndrew-Kainos: ok

picklists/serializers.py
BenAndrew-Kainos: ok

conf/custom_views.py
BenAndrew-Kainos: Well by default ListAPIView is just a PaginatedAPIView so it would be misleading to name it so. This is new functionality we're introducing to allow to to be optionally paginated

PR #405, opened 2020-04-22
bug: LT-2465 staging case bug

BenAndrew-Kainos:


BenAndrew-Kainos:
Closing as bug was not fixed by attempted changes

PR #406, opened 2020-04-22
bug: Queues performance

BenAndrew-Kainos:


conf/custom_views.py
```diff
@@ -0,0 +1,18 @@
+from django.http import JsonResponse
+from rest_framework.generics import ListAPIView
+
+from conf.helpers import str_to_bool
+
+
+class OptionalPaginationView(ListAPIView):
```
jfaracik, line 7: Whats the purpose of this? By default our pagination already supports `disable_pagination`

BenAndrew-Kainos: removed this class

PR #408, opened 2020-04-23
bug: LT-2465  - Populate null actor columns in audit

a-bek:


audit_trail/migrations/0005_null_actor_fix.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def fill_in_missing_actor(apps, schema_editor):
+    if schema_editor.connection.alias != "default":
+        return
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    Audit = apps.get_model("audit_trail", "Audit")
+    ExporterUser = apps.get_model("users", "ExporterUser")
+    for audit in Audit.objects.filter(
+        actor_content_type__isnull=True, verb__in=[AuditType.UPDATED_STATUS.value, AuditType.CREATED.value]
```
BenAndrew-Kainos, line 13: Why do you filter by verb (is it not all Audits?)

a-bek:
This was caused by a backfill issue of audits that were not being created before a certain point in time. There is no permanent fix - all audits are created by a user, the backfill didn't add a user as an actor.

audit_trail/migrations/0005_null_actor_fix.py
a-bek: These are the only audits affected, an artificial audit was added / backfilled for some of these cases to mark the submission of an application, but left out the actor :(

PR #410, opened 2020-04-23
bug: LT-2459 Goods form validation

BenAndrew-Kainos:


a-bek:
The only thing I would comment on, but the file isn't part of the review, is that the document is there, id passed to `prepare_document(document.id)` which then gets the document again, rather than just using the document already in memory.

BenAndrew-Kainos:
> already

Yeah that seems like a good idea.  There's lots of document improvements we're hoping to do

PR #411, opened 2020-04-23
performance: Cases page

BenAndrew-Kainos:
Greatly reduce overhead for fetching queues list. 
Improvement to flags fetching/serialization. 
(10%+ improvement depending on number of queues)


queues/service.py
```diff
@@ -1,53 +1,60 @@
...
-        ),
-        SearchQueue(
-            id=queues.MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_ID,
-            name=queues.MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_NAME,
-            cases=case_qs.assigned_as_case_officer(user=user).not_terminal(),
-        ),
-        SearchQueue(
-            id=queues.UPDATED_CASES_QUEUE_ID, name=queues.UPDATED_CASES_QUEUE_NAME, cases=case_qs.is_updated(user=user)
-        ),
+        {"id": queues.ALL_CASES_QUEUE_ID, "name": queues.ALL_CASES_QUEUE_NAME, "case_count": case_qs.count()},
```
jfaracik, line 16: Is there much advantage to using a dict instead of SearchQueue? At least with SearchQueue fields are enforced, you don't get that with a dict

queues/service.py
```diff
@@ -1,53 +1,60 @@
...
     ]
 
 
 def get_all_queues(user):
-    return get_system_queues(user=user) + get_user_created_queues()
+    """
+    Returns all queues with the values id, name & case_count
+    """
+    return get_system_queues(user=user) + list(
+        Queue.objects.annotate(case_count=Count("cases")).values("id", "name", "case_count")
```
jfaracik, line 50: Could `list(Queue.objects.annotate(case_count=Count("cases")).values("id", "name", "case_count")` this be split back out into its own function? Thanks

queues/service.py
BenAndrew-Kainos: SearchQueue was using the wrong field with queues instead of queues count. Also having it in a dict format means we don't then have to convert it to a dict after (which we were doing before). This technique is just more straightforward & improves performance

queues/service.py
BenAndrew-Kainos: sure

PR #416, opened 2020-04-27
fix: LT-2359 Multiple CLCs

iichr:


licences/serializers/view_licence.py
```diff
@@ -47,11 +48,13 @@ class Meta:
 
 
 class GoodsTypeOnLicenceSerializer(serializers.ModelSerializer):
+    control_list_entries = ControlListEntryViewSerializer(many=True)
+
```
jfaracik, line 52: Absolutely breathtaking change. I'm in tears

PR #417, opened 2020-04-27
enhance: Factories for Tests

cstobart:


BenAndrew-Kainos:
Shouldn't we be doing high priority bugs instead of this?

PR #418, opened 2020-04-27
feat: LT-2310 - manage routing rules for all teams

Mark-Spiers-kainos:


teams/views.py
```diff
@@ -99,3 +101,15 @@ def get(self, request, pk):
 
         serializer = GovUserListSerializer(users, many=True)
         return JsonResponse(data={"users": serializer.data})
+
+
+class TeamQueuesList(APIView):
```
jfaracik, line 106: Can you use a ListAPIView instead and adjust the code below accordingly :)

workflow/routing_rules/tests/test_routing_rules_permissions.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_manage_team_permission_fails_for_other_team_rules(self):
+        role = Role.objects.create(name="test")
+        role.permissions.set([constants.GovPermissions.MANAGE_TEAM_ROUTING_RULES.name])
+        self.gov_user.role = role
+        self.gov_user.save()
+
+        other_team = self.create_team("other")
+
+        routing_rule = self.create_routing_rule(
```
jfaracik, line 99: If you have the time a RoutingRuleFactory would be a nice improvement :)

workflow/routing_rules/tests/test_routing_rules_permissions.py
```diff
@@ -0,0 +1,130 @@
...
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_manage_team_permission_works_for_own_team_rules(self):
+        role = Role.objects.create(name="test")
+        role.permissions.set([constants.GovPermissions.MANAGE_TEAM_ROUTING_RULES.name])
+        self.gov_user.role = role
+        self.gov_user.save()
+
+        routing_rule = self.create_routing_rule(
```
andytaylorkainos, line 55: a lot of these tests have similar setups. can you put them into a SetUp function please?

Mark-Spiers-kainos: we have 1 test with all permissions, 2 with manage all routing rules, 2 with manage team routing rules, and 1 without any routing rules. Only 2 things I can move is "create role" (but not set permissions) and create routing rule.

andytaylorkainos: if you could move those two that'd be a few less lines of clutter :) (the URL can be moved to 😄)

Mark-Spiers-kainos: Infact I can't even move the routing rule creation, since it changes which team it is depending on rule, and I can't move self.url since it depends on routing rule id.

teams/views.py
Mark-Spiers-kainos: Since I don't use this as a paginated list, have made it `OptionalPaginationView` which inherits from ListAPIView, and has the option to disable pagination built in based on GET parameters

workflow/routing_rules/tests/test_routing_rules_permissions.py
Mark-Spiers-kainos: Have wrote the tests in a different manner, setup is included now.

teams/views.py
```diff
@@ -99,3 +102,15 @@ def get(self, request, pk):
 
         serializer = GovUserListSerializer(users, many=True)
         return JsonResponse(data={"users": serializer.data})
+
+
+class TeamQueuesList(OptionalPaginationView):
```
jfaracik, line 107: AFAIK OptionalPaginationView is being removed in one of @BenAndrew-Kainos PRs

PR #420, opened 2020-04-27
bug: LT-2486 - Audit + Enum enhancement refactor

a-bek:
- New `LiteEnum` class which is based off `str` and `Enum`
- `Audit.verb` is now a choices field with values auto-generated from `AuditType` enum class.
- Mapping of `AuditType` enum class to formatted string used in frontend, which provides non-breaking flexibility in changing values
- Migrating old `Audit.verb` values to new `AuditType` enum class values


audit_trail/enums.py
```diff
@@ -0,0 +1,83 @@
...
+    REMOVE_FLAGS = autostr()
+    GOOD_REVIEWED = autostr()
+    GOOD_ADD_FLAGS = autostr()
+    GOOD_REMOVE_FLAGS = autostr()
+    GOOD_ADD_REMOVE_FLAGS = autostr()
+    DESTINATION_ADD_FLAGS = autostr()
+    DESTINATION_REMOVE_FLAGS = autostr()
+    ADD_GOOD_TO_APPLICATION = autostr()
+    REMOVE_GOOD_FROM_APPLICATION = autostr()
+    ADD_GOOD_TYPE_TO_APPLICATION = autostr()
```
andytaylorkainos, line 16: why do you need all of these?

surely you only need a 'created', 'updated', 'removed', 'added' etc. which can be formatted with any string?

a-bek: It's probably better for the audit trail to not make the AuditType generic, there are several reasons.

1)  how do we then know which object we are dealing with when we have a generic action, as we don't story the action object, and it can't be done in some cases as it doesn't necessarily act on an object.

2) We have audits for actions on multiple items, ie, ADDED_FLAGS. So we can't use `action_object` to  point to multiple model rows. We are also not storing reference to the object in `action_object`, which we currently only do for the `Case` model.

3) Probably most importantly, if an object is deleted, we don't cascade delete the audit because we never want to delete audits, but the content type will be set to NULL in the model, so we lose information about what was acted on. AuditType provides that information

PR #421, opened 2020-04-28
feat: Approval blocking flags

BenAndrew-Kainos:


flags/views.py
```diff
@@ -302,3 +304,18 @@ def put(self, request, pk):
             return JsonResponse(data={"flagging_rule": serializer.data})
 
         return JsonResponse(data={"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+
+class CaseFlags(APIView):
```
jfaracik, line 309: Could this not just be a query param on the existing flags endpoint? eg `?case_id=...`

flags/tests/tests_case_flags.py
```diff
@@ -0,0 +1,33 @@
...
+
+from test_helpers.clients import DataTestClient
+
+
+class CaseFlagsTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        self.case = self.create_standard_application_case(self.organisation)
+        self.url = reverse("flags:case_flags", kwargs={"case_pk": self.case.pk})
+        self.flag_1 = self.create_flag("Case Flag 1", "Case", self.team)
```
jfaracik, line 12: Better to just use a FlagFactory, create_flag is deprecated :) You won't need to specify a flag name either.

cases/libraries/get_destination.py
```diff
@@ -82,6 +83,20 @@ def annotate_my_team_flags(flags, priority, team):
     ).annotate(type=models.Value(priority, models.IntegerField()))
 
 
+def get_flags(case: Case) -> QuerySet:
```
jfaracik, line 86: I don't think `get_destination.py` is the right place for this function, perhaps a `flags.py`?

applications/tests/tests_finalise_application.py
```diff
@@ -82,6 +82,23 @@ def test_no_permissions_finalise_failure(self):
 
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
+    def test_approve_application_blocking_flags_failure(self):
+        flag = self.create_flag("Block", "Case", self.team)
```
jfaracik, line 86: FlagFactory would be better as discussed below. If you add `blocks_approval` to the factory you can just do `flag = FlagFactory(blocks_approval=True, level=FlagLevels.Case, team=self.team)`. No need to save.

flags/views.py
BenAndrew-Kainos: sure

flags/tests/tests_case_flags.py
BenAndrew-Kainos: How could it be deprecated? Isn't it a function we created???

I will update anyway

flags/views.py
BenAndrew-Kainos: Unfortunately the existing endpoint uses a different serializer, gets the data a different way and accesses it from a different context so I started adding to it but I couldn't actually reuse any of it, so it was more straightforward to make a different endpoint.

cases/libraries/get_destination.py
BenAndrew-Kainos: Yeah good point, will move both get_flags functions

BenAndrew-Kainos: done

applications/tests/tests_finalise_application.py
BenAndrew-Kainos: done

flags/tests/tests_case_flags.py
BenAndrew-Kainos: done :)

flags/views.py
BenAndrew-Kainos: I'm gonna have another go at it today actually

BenAndrew-Kainos: done

applications/views/applications.py
```diff
@@ -492,6 +494,23 @@ def put(self, request, pk):
...
 
+            # Check if any blocking flags are on the case
+            blocking_flags = (
+                get_flags(application.get_case())
+                .filter(status=FlagStatuses.ACTIVE, blocks_approval=True)
+                .order_by("name")
+                .values_list("name", flat=True)
+            )
+            if blocking_flags:
+                return JsonResponse(
```
jfaracik, line 505: I think this should raise a forbidden exception instead, like so `raise exceptions.PermissionDenied([f"{strings.Applications.Finalise.Error.BLOCKING_FLAGS}{','.join(list(blocking_flags))}"])`

PR #425, opened 2020-04-29
feat: LT-1413 Default Queue

andytaylorkainos:


gov_users/views/gov_user_views.py
```diff
@@ -36,10 +30,7 @@ def post(self, request, *args, **kwargs):
         :param email, first_name, last_name:
         :return token:
         """
-        try:
-            data = JSONParser().parse(request)
-        except ParseError:
-            return JsonResponse(data={"errors": "Invalid Json"}, status=status.HTTP_400_BAD_REQUEST)
+        data = request.data
```
jfaracik, line 33: Is there any point in this line seeing as you use `request.data` later on anyway?

queues/views/queues.py
```diff
@@ -43,4 +61,4 @@ def put(self, request, pk):
 
         if serializer.is_valid(raise_exception=True):
             serializer.save()
-            return JsonResponse(data={"queue": serializer.data})
+            return JsonResponse(data={"queue": serializer.data}, status=status.HTTP_200_OK)
```
jfaracik, line 64: I still don't see the point in setting the status to JsonResponse's default value :/

queues/views/queues.py
```diff
@@ -1,18 +1,36 @@
...
+class QueuesList(OptionalPaginationView):
     authentication_classes = (GovAuthentication,)
     queryset = Queue.objects.all()
     serializer_class = QueueListSerializer
 
+    def get(self, request, *args, **kwargs):
+        include_system = request.GET.get("include_system", False)
+
+        if str_to_bool(include_system):
+            system_queues = [
```
jfaracik, line 23: There is a function to already do this `get_all_queues`

andytaylorkainos: That function also returns case count etc; i don't need that - this endpoint literally returns queue-level data

gov_users/views/gov_user_views.py
andytaylorkainos: that other line you're talking about is in a different method

queues/views/queues.py
andytaylorkainos: I honestly dont mind anymore; im just making this one consistent with the rest (theres a lot more with the status specified than without)

queues/views/queues.py
andytaylorkainos: I had to a bit of a refactor to be able to re-use some of the service - if you could have a read over i'd appreciate the review: https://github.com/uktrade/lite-api/pull/425/commits/385146ac9364b662764bdb2b7ff02c9a33a99940

queues/constants.py
```diff
@@ -15,12 +15,15 @@
 MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_NAME = strings.MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_NAME
 
 SYSTEM_QUEUES = {
-    ALL_CASES_QUEUE_ID,
-    OPEN_CASES_QUEUE_ID,
-    MY_TEAMS_QUEUES_CASES_ID,
-    UPDATED_CASES_QUEUE_ID,
-    MY_ASSIGNED_CASES_QUEUE_ID,
-    MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_ID,
+    ALL_CASES_QUEUE_ID: ALL_CASES_QUEUE_NAME,
```
stephanieboyd, line 18: ID or name?

gov_users/serializers.py
```diff
@@ -118,8 +136,25 @@ class Meta:
             "status",
             "team",
             "role",
+            "default_queue",
         )
 
+    def validate(self, attrs):
+        validated_data = super().validate(attrs)
+        default_queue = str(validated_data.get("default_queue") or self.instance.default_queue)
```
BenAndrew-Kainos, line 144: You shouldn't need to do this or, just use the validated_data (same below)

gov_users/serializers.py
```diff
@@ -118,8 +136,25 @@ class Meta:
...
+        team = validated_data.get("team") or self.instance.team
+
+        is_system_queue = default_queue in SYSTEM_QUEUES.keys()
+        is_work_queue = Queue.objects.filter(id=default_queue).exists()
+
+        if not is_system_queue and not is_work_queue:
+            raise serializers.ValidationError({"default_queue": [strings.Users.NULL_DEFAULT_QUEUE]})
+
+        if is_work_queue and not Queue.objects.values_list("team_id", flat=True).get(id=default_queue) == team.id:
```
BenAndrew-Kainos, line 153: you don't need values_list if your using get. Please change to `Queue.objects.get(id=default_queue).team_id`

gov_users/tests/tests_edit.py
```diff
@@ -23,6 +23,18 @@ def test_edit_a_gov_user(self):
         self.assertNotEqual(response_data["gov_user"]["email"], "test@mail.com")
         self.assertNotEqual(response_data["gov_user"]["team"], self.team)
 
+    def test_edit_gov_user_default_queue(self):
```
BenAndrew-Kainos, line 26: Please add failure tests;
1. editing default queue and setting to an invalid UUID
2. editing default queue and setting to non team queue

queues/service.py
```diff
@@ -1,62 +1,77 @@
...
+    queue_case_count_qs_map = {
+        ALL_CASES_QUEUE_ID: case_qs,
+        OPEN_CASES_QUEUE_ID: case_qs.is_open(),
+        MY_TEAMS_QUEUES_CASES_ID: case_qs.in_team(team_id=user.team.id),
+        MY_ASSIGNED_CASES_QUEUE_ID: case_qs.assigned_to_user(user=user).not_terminal(),
+        MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_ID: case_qs.assigned_as_case_officer(user=user).not_terminal(),
+        UPDATED_CASES_QUEUE_ID: case_qs.is_updated(user=user),
+    }
+
+    return queue_case_count_qs_map[queue_id].count()
```
BenAndrew-Kainos, line 47: Can you not just do .count() inline for each

queues/service.py
```diff
@@ -1,62 +1,77 @@
...
+    MY_ASSIGNED_CASES_QUEUE_ID,
+    MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_ID,
+    UPDATED_CASES_QUEUE_ID,
+    SYSTEM_QUEUES,
+)
 from queues.models import Queue
 
-import queues.constants as queues
 
+def get_system_queues(include_team=True, include_case_count=False, user=None) -> List[Dict]:
```
BenAndrew-Kainos, line 19: Please include a docstring

queues/views/queues.py
```diff
@@ -1,18 +1,30 @@
...
+from conf.custom_views import OptionalPaginationView
+from conf.helpers import str_to_bool
 from queues.models import Queue
 from queues.serializers import QueueCreateSerializer, QueueViewSerializer, QueueListSerializer
-from queues.service import get_queue
+from queues.service import get_queue, get_work_queues_qs, get_system_queues
 
 
-class QueuesList(generics.ListAPIView):
+class QueuesList(OptionalPaginationView):
```
BenAndrew-Kainos, line 13: We don't need `OptionalPaginationView`. The `disable_pagination` argument works without it and it will be removed soon

gov_users/serializers.py
andytaylorkainos: But then you return the entire entity from the DB, when all i need is the team_id

queues/service.py
andytaylorkainos: No, because then everything will be evaluated everytime it iterates through the loop

queues/views/queues.py
andytaylorkainos: Ah didnt realise we were getting rid of it, will do!

gov_users/serializers.py
andytaylorkainos: I need the or.
If the validated data doesnt contain `default_queue` (they may only be updating the team), then i need to ensure that the instance.default_queue is valid with the team

queues/service.py
andytaylorkainos: I re-wrote it a little so that I could do this

gov_users/serializers.py
BenAndrew-Kainos: I think you might want to check that. I don't believe it does return the whole record

PR #427, opened 2020-04-30
feat: LT-1299 - countersigning queue

Mark-Spiers-kainos:


workflow/tests/tests_user_queue_assignment.py
```diff
@@ -84,6 +84,21 @@ def test_multiple_case_assignments(self):
         self.assertEqual(self.case.queues.first(), self.queue)
         self.assertNotEqual(self.case.status, self.new_status)
 
+    def test_countersigning_queue(self):
+        """
+        Tests that countersigning queues are assigned when work queue removed, and status is not change
+        """
+        old_status = self.case.status
+        countersigning_queue = self.create_queue("other", self.team)
+        self.queue.countersigning_queue = countersigning_queue
```
jfaracik, line 93: Could you update the QueueFactory to handle countersigning queues? Give me a ping if you need a hand :)

queues/serializers.py
```diff
@@ -17,6 +17,13 @@ class QueueViewSerializer(serializers.Serializer):
     id = serializers.UUIDField()
     name = serializers.CharField()
     is_system_queue = serializers.SerializerMethodField()
+    countersigning_queue = serializers.SerializerMethodField()
+
+    def get_countersigning_queue(self, instance):
+        if isinstance(instance, Queue):
+            return instance.countersigning_queue_id
+        else:
+            return instance.get("countersigning_queue", None)
```
BenAndrew-Kainos, line 26: Is this for dicts?

queues/tests/tests_create.py
```diff
@@ -20,3 +20,19 @@ def test_create_queue(self):
...
+            "countersigning_queue": countersigning_queue.id,
+        }
+
+        response = self.client.post(self.url, data, **self.gov_headers)
+        response_data = response.json()["queue"]
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response_data["name"], data["name"])
+        self.assertEqual(response_data["team"], str(self.team.id))
+        self.assertEqual(response_data["countersigning_queue"], str(countersigning_queue.id))
```
BenAndrew-Kainos, line 39: Could you add a test to cover giving a countersigning queue that is itself

queues/views/queues.py
```diff
@@ -13,6 +15,14 @@ class QueuesList(generics.ListAPIView):
     queryset = Queue.objects.all()
     serializer_class = QueueListSerializer
 
+    def filter_queryset(self, queryset):
+        if str_to_bool(self.request.GET.get("users_team_first", "False")):
+            return queryset.annotate(
+                users_team=(Case(When(team=self.request.user.team, then=1), default=0, output_field=BinaryField()))
```
BenAndrew-Kainos, line 21: nice

workflow/user_queue_assignment.py
```diff
@@ -34,13 +41,25 @@ def get_next_status_in_workflow_sequence(case):
         return None
 
 
-def user_queue_assignment_workflow(queues: [Queue], case: Case):
+def user_queue_assignment_workflow(queues: List[Queue], case: Case):
```
BenAndrew-Kainos, line 44: Do you need List here?

workflow/user_queue_assignment.py
```diff
@@ -34,13 +41,25 @@ def get_next_status_in_workflow_sequence(case):
...
 
     # Remove case from queues where all gov users are done with the case
     queues_without_case_assignments = set(queues) - get_queues_with_case_assignments(case)
     case.queues.remove(*queues_without_case_assignments)
 
+    system_user = GovUser.objects.get(id=SystemUser.LITE_SYSTEM_ID)
+
+    for queue in queues_without_case_assignments:
+        if queue.countersigning_queue_id:
+            case.queues.add(queue.countersigning_queue_id)
```
BenAndrew-Kainos, line 55: Is this the only place it's needed. I was under the impression it would also be needed when a queue was manually unassigned

queues/serializers.py
Mark-Spiers-kainos: looks like it. The cases list page breaks without this line.

workflow/user_queue_assignment.py
Mark-Spiers-kainos: I would prefer if we are going to type, that we fully type it. there are a number of other object types that fall under [] than just list and set

queues/tests/tests_create.py
Mark-Spiers-kainos: added

workflow/user_queue_assignment.py
BenAndrew-Kainos: is there a reason this function cannot support other listed types (like sets?). If not i would leave it so that it's more reusable in the future

workflow/user_queue_assignment.py
BenAndrew-Kainos: Awesome (I have a feeling it will be needed)

workflow/user_queue_assignment.py
```diff
@@ -34,13 +41,25 @@ def get_next_status_in_workflow_sequence(case):
...
+def user_queue_assignment_workflow(queues: List[Queue], case: Case):
     from workflow.automation import run_routing_rules
 
     # Remove case from queues where all gov users are done with the case
     queues_without_case_assignments = set(queues) - get_queues_with_case_assignments(case)
     case.queues.remove(*queues_without_case_assignments)
 
+    system_user = GovUser.objects.get(id=SystemUser.LITE_SYSTEM_ID)
+
+    for queue in queues_without_case_assignments:
```
efahy, line 51: I really like how easy this story was to achieve here!  Neatly done.  Perhaps a comment on this block giving a bit of business context?  I don't think  its immediately obvious to an initiated person.

Mark-Spiers-kainos: added

PR #430, opened 2020-05-01
bugfix: Seeding - System and Exporter Users

andytaylorkainos:


cases/models.py
```diff
@@ -147,7 +147,7 @@ def remove_all_case_assignments(self):
         from audit_trail import service as audit_trail_service
 
         assigned_cases = CaseAssignment.objects.filter(case_id=self.id)
-        system_user = GovUser.objects.get(id=SystemUser.LITE_SYSTEM_ID)
+        system_user = BaseUser.objects.get(id=SystemUser.id)
```
efahy, line 150: I think an `audit_trail_service.create_system` or `audit_trail_service.create_system_audit` method would be really good.  Means nobody has to figure out how to fetch the system_user themselves

local.env
```diff
@@ -4,15 +4,18 @@ PORT=8100
 ALLOWED_HOSTS=["*"]
 
 # SEED USERS
-INTERNAL_ADMIN_TEAM_USERS=[{"email": "", "role": "Super User"}]
+SYSTEM_USER={"id": "", "first_name": "", "last_name": ""}
+INTERNAL_ADMIN_TEAM_USERS=[{"email": "", role": "Super User"}]
```
efahy, line 8: Missing quote here?

cases/models.py
andytaylorkainos: Done!

PR #431, opened 2020-05-01
feat: Review organisations

BenAndrew-Kainos:
Key changes;
- Add an endpoint for getting notification the the menu widget (currently only for organisations but  scalable)
- Allow organisation status to be updated through the review screen (set to active or rejected if currently in review)
- Add an endpoint for getting properties that exist in other organisations to warn internal users what fields are duplicated from other orgs


organisations/views/organisations.py
```diff
@@ -120,3 +125,61 @@ def reopen_closed_cases_for_organisation(organisation):
...
+                primary_site__address__address_line_1=organisation.primary_site.address.address_line_1,
+            )
+            | Q(
+                primary_site__address__address__isnull=False,
+                primary_site__address__address=organisation.primary_site.address.address,
+            )
+        )
+
+        if organisations_with_matching_details.count() > 1:
+            if self._property_has_multiple_occurances(organisations_with_matching_details, organisation.name, "name"):
```
andytaylorkainos, line 153: can you space out all the `if` statements please - makes it easier to separate the logic visually

conf/urls.py
```diff
@@ -37,6 +37,7 @@
     path("queries/", include("queries.urls")),
     path("routing-rules/", include("workflow.routing_rules.urls")),
     path("licences/", include("licences.urls")),
+    path("menu/", include("menu.urls")),
```
andytaylorkainos, line 40: Would it not make more sense to put the URL under `organisations` because its related specifically to querying data from that model?

`organisations/?status=OrganisationStatus.IN_REVIEW`

menu/views.py
```diff
@@ -0,0 +1,21 @@
+from django.http import JsonResponse
+from rest_framework.views import APIView
+
+from conf.authentication import GovAuthentication
+from conf.constants import GovPermissions
+from organisations.enums import OrganisationStatus
+from organisations.models import Organisation
+
+
+class MenuNotifications(APIView):
```
andytaylorkainos, line 10: Can you please remove the endpoint and make it a query param on organisations

organisations/views/organisations.py
```diff
@@ -120,3 +125,61 @@ def reopen_closed_cases_for_organisation(organisation):
...
+
+class OrganisationsMatchingDetail(APIView):
+    @staticmethod
+    def _property_has_multiple_occurances(queryset, property, property_name):
+        return property and list(queryset.values_list(property_name, flat=True)).count(property) > 1
+
+    def get(self, request, pk):
+        matching_properties = []
+        organisation = get_organisation_by_pk(pk)
+        organisations_with_matching_details = Organisation.objects.filter(
```
andytaylorkainos, line 138: can you exclude the organisation you're using from the queryset so that you can do `if organisations_with_matching_details.exists()` on line 152?

organisations/views/organisations.py
```diff
@@ -120,3 +125,61 @@ def reopen_closed_cases_for_organisation(organisation):
...
+        )
+
+        if organisations_with_matching_details.count() > 1:
+            if self._property_has_multiple_occurances(organisations_with_matching_details, organisation.name, "name"):
+                matching_properties.append(Organisations.MatchingProperties.NAME)
+            if self._property_has_multiple_occurances(
+                organisations_with_matching_details, organisation.eori_number, "eori_number"
+            ):
+                matching_properties.append(Organisations.MatchingProperties.EORI)
+            if self._property_has_multiple_occurances(
```
andytaylorkainos, line 161:
```
matching_properties = self._get_matching_properties(organisation, queryset, 
										 [
											("eori_number", Organisations.MatchingProperties.EORI), 
											("registration_number", Organisations.MatchingProperties.REGISTRATION), 
											...
										 ])


def _get_matching_properties(organisation, queryset, properties):
   matching_properties = []

   for property_name, property_enum in properties:
       property = getattr(organisation, property_name)
       
      if property and list(queryset.values_list(property_name, flat=True)).count(property) > 1:
          matching_properties.append(property_enum)
       
     return matching_properties
```

conf/urls.py
BenAndrew-Kainos: No because this one is to do with getting notifications in the menu, nothing specific to organisations

menu/views.py
BenAndrew-Kainos: As above

organisations/views/organisations.py
BenAndrew-Kainos: good idea

conf/urls.py
andytaylorkainos: Then rebrand it to a notifications endpoint to take query params for the type of notifications

organisations/views/organisations.py
BenAndrew-Kainos: I would but unfortunately the address checks 2 fields which breaks the pattern :(

organisations/views/organisations.py
BenAndrew-Kainos: done

organisations/views/organisations.py
andytaylorkainos: damnit :(

conf/urls.py
BenAndrew-Kainos: Screenshot 2020-05-04 at 16 40 59

There is no type of notification. It's for the menu widget that is visible on all pages

organisations/views/organisations.py
BenAndrew-Kainos: Does this not start making it very hard to read? If this were a more comprehensive list I would implement but I think for 4 fields the readability of having it laid it is better. What do you think?

andytaylorkainos: Honestly not too fussed about this. it was more a suggestion - happy for you to resolve

conf/urls.py
andytaylorkainos: I'm confused, currently all it does it get organisation "notifications" (as you've put it).
And I find it a little strange to see an endpoint called "menu", when it returns a dictionary with "notifications" as the key.

I know you intend this to be an endpoint specifically for "menu" logic, but as far as I'm aware the only use-case for it right now is for organisations and for showing a form of "notification" which is why I suggested `/organisations/notifications/`.

If there's any reason that we need more of these notification-type endpoints specifically for the frontend menu then I'd rather it was under a generic name (such as '/notifications/') to indicate that it's specifically for returning a count against the entities (cases, organisations, documents).

organisations/views/organisations.py
BenAndrew-Kainos: Thanks. I honestly did want to implement something like these but just ran into this problem.

conf/urls.py
BenAndrew-Kainos: Done changes as discussed

PR #437, opened 2020-05-05
Performance: EUA list performance

BenAndrew-Kainos:
Also includes some minor fixes to the fetching of organisation after the authentication changes


users/tests/tests_exporter_user_notifications.py
```diff
@@ -193,7 +193,7 @@ def test_get_end_user_advisory_queries_with_notifications_success(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         eua_query_response_data = response.json()["results"][0]
         self.assertIn("exporter_user_notification_count", eua_query_response_data)
-        self.assertEqual(eua_query_response_data["exporter_user_notification_count"]["total"], 3)
+        self.assertEqual(eua_query_response_data["exporter_user_notification_count"], 3)
```
andytaylorkainos, line 196: got a frontend PR too?

queries/end_user_advisories/views.py
```diff
@@ -27,7 +29,26 @@ class EndUserAdvisoriesList(ListAPIView):
     serializer_class = EndUserAdvisoryListSerializer
 
     def get_queryset(self):
-        return EndUserAdvisoryQuery.objects.filter(organisation_id=get_request_user_organisation_id(self.request))
+        return EndUserAdvisoryQuery.objects.filter(
+            organisation_id=get_request_user_organisation_id(self.request)
+        ).select_related("end_user", "end_user__country")
+
+    def get_paginated_response(self, data):
+        ids = [item["id"] for item in data]
```
andytaylorkainos, line 37: just ditching the serializer to improve performance?

queries/end_user_advisories/views.py
```diff
@@ -27,7 +29,26 @@ class EndUserAdvisoriesList(ListAPIView):
...
+        notifications = (
+            ExporterNotification.objects.filter(
+                user=self.request.user, organisation_id=get_request_user_organisation_id(self.request), case__id__in=ids
```
andytaylorkainos, line 40: not sure if there is any performance benefit from this change but you can drop the dunder to the case id:

`case_id__in=ids`

queries/end_user_advisories/serializers.py
```diff
@@ -3,45 +3,34 @@
...
-        read_only_fields = fields
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.exporter_user = kwargs["context"]["request"].user
-        self.organisation_id = get_request_user_organisation_id(kwargs["context"]["request"])
+class EndUserAdvisoryListPartySerializer(serializers.Serializer):
+    name = serializers.CharField()
+    address = serializers.CharField()
+    country = EndUserAdvisoryListCountrySerializer()
```
Mark-Spiers-kainos, line 25: Might be worth looking at the source arguement for serializer Fields, rather than having a serializer that returns one field.

queries/end_user_advisories/views.py
```diff
@@ -27,7 +29,26 @@ class EndUserAdvisoriesList(ListAPIView):
...
+            )
+            .values("case")
+            .annotate(count=Count("case"))
+        )
+        cases_with_notifications = {str(notification["case"]): notification["count"] for notification in notifications}
+
+        for item in data:
+            if item["id"] in cases_with_notifications:
```
andytaylorkainos, line 48: even if the case doesnt have a notification, would still be good to have the dictionary present with {"exporter_user_notification_count": 0} against the case

users/tests/tests_exporter_user_notifications.py
BenAndrew-Kainos: https://github.com/uktrade/lite-exporter-frontend/pull/429

queries/end_user_advisories/views.py
BenAndrew-Kainos: yep. Before it was executing one query per list item. Now it executes one query for all

queries/end_user_advisories/serializers.py
BenAndrew-Kainos: Going to leave it for now as this formatting is used a lot on the frontend

queries/end_user_advisories/views.py
BenAndrew-Kainos: done

queries/end_user_advisories/views.py
BenAndrew-Kainos: done

BenAndrew-Kainos:
> this would be a good perf improvement across the board for all notifications (i think application list does this too) - not asking you to do it, but you could make a function out of your code so it can be re-used on those other api views :)

Absolutely! That will be done as part of these performance tickets

PR #438, opened 2020-05-05
bugfix: Seedall

andytaylorkainos:
Essentially the biggest changes are:
1. individual commands can be told to not system exit if they fail (if a command now throws an error, it will not prevent the following seeding operations from running in `seedall`)
2. `transaction.atomic` was removed from the `seedall` operation (individuall commands will remain atomic i.e. if `seedexporterusers` fails during `seedall`, only the exporter users will be reverted)
3. separation of the nested data from commands (the admin team is now seeded separately from users)
4. any failures are logged to console at the end of `seedall` which can now be seen in the new jenkins job

Unrelated changes:
1. Removed logging from audit migration unless something has changed
2. Removed system user environment variable as the user can no longer be authenticated

Screen Shot 2020-05-06 at 13 02 38


andytaylorkainos:
> All looks fine, but I didn't spot what fixed the issue

Essentially the biggest changes are:
1. The commands can be told to not system exit if they fail (if an individual command throws an error, it will not prevent the following seeding operations from running)
2. `transaction.atomic` was removed from the `seedall` operation (individuall commands will remain atomic i.e. if `seedexporterusers` fails during `seedall`, only the exporter users will be reverted)
3. separation of the nested data from commands (the admin team is now seeded separately from users)

PR #440, opened 2020-05-06
Performance: Authentication

BenAndrew-Kainos:
Reduce the queries and performance impact of the authentication classes (particularly ExporterAuthentication)

Organisation is now only fetched when necessary

Shaves significant load time off each Exporter endpoint

### DIFF
--- a/cases/views/views.py
+++ b/cases/views/views.py
@@ -51,6 +51,7 @@ from documents.models import Document
 from goodstype.helpers import get_goods_type
 from gov_users.serializers import GovUserSimpleSerializer
 from lite_content.lite_api.strings import Documents, Cases
+from organisations.libraries.get_organisation import get_request_user_organisation_id
 from queues.serializers import TinyQueueSerializer
 from parties.models import Party
 from parties.serializers import PartySerializer, AdditionalContactSerializer
@@ -183,7 +184,7 @@ class ExporterCaseDocumentDownload(APIView):
 
     def get(self, request, case_pk, document_pk):
         case = get_case(case_pk)
-        if case.organisation != request.user.organisation:
+        if case.organisation.id != get_request_user_organisation_id(request):
             return HttpResponse(status.HTTP_401_UNAUTHORIZED)
         try:
             document = CaseDocument.objects.get(id=document_pk, case=case)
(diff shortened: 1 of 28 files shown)


cases/views/views.py
```diff
@@ -183,7 +184,7 @@ class ExporterCaseDocumentDownload(APIView):
 
     def get(self, request, case_pk, document_pk):
         case = get_case(case_pk)
-        if case.organisation != request.user.organisation:
+        if case.organisation.id != get_request_user_organisation_id(request):
```
andytaylorkainos, line 187: i know this is out of your scope so i wont block yu with it, but can you try to add the annotation `@authorised_users(ExporterUser)` above the 'get' method and remove this `if` statement and 401_unauthorized response?

conf/authentication.py
```diff
@@ -3,21 +3,22 @@
...
+from users.models import UserOrganisationRelationship, ExporterUser, GovUser
 
 GOV_USER_TOKEN_HEADER = "HTTP_GOV_USER_TOKEN"  # nosec
 
 EXPORTER_USER_TOKEN_HEADER = "HTTP_EXPORTER_USER_TOKEN"  # nosec
 ORGANISATION_ID = "HTTP_ORGANISATION_ID"
 
-USER_DEACTIVATED_ERROR = "User has been deactivated"
-
+MISSING_TOKEN_ERROR = "You must supply the correct token in your headers"  # nosec
```
andytaylorkainos, line 18: can you change the message to "tokens" (that way we arent advertising there's only 1 token)

conf/authentication.py
```diff
@@ -62,30 +58,27 @@ def authenticate(self, request):
         """
         if request.META.get(EXPORTER_USER_TOKEN_HEADER):
             exporter_user_token = request.META.get(EXPORTER_USER_TOKEN_HEADER)
+            user_id = token_to_user_pk(exporter_user_token)
+            organisation_id = request.META.get(ORGANISATION_ID)
```
andytaylorkainos, line 62: get_request_user_organisation_id

conf/authentication.py
```diff
@@ -62,30 +58,27 @@ def authenticate(self, request):
...
+        try:
+            user = ExporterUser.objects.get(id=user_id)
+        except ExporterUser.DoesNotExist:
+            raise PermissionDeniedError(USER_NOT_FOUND_ERROR)
 
-        raise PermissionDeniedError("You don't belong to that organisation")
+        return user, None
 
 
 class ExporterOnlyAuthentication(authentication.BaseAuthentication):
```
andytaylorkainos, line 84: are we able to remove this? its only used in 1 class

BenAndrew-Kainos: We cannot (used for when an org is not selected) but I have tidied to make it like the other classes

cases/views/views.py
BenAndrew-Kainos: Unfortunately that annotation tries to use pk to get an application which isn't applicable for this endpoint

PR #445, opened 2020-05-07
Performance: Applications performance

BenAndrew-Kainos:
Improve applications list endpoints and reduce decorator processing


BenAndrew-Kainos:
> Nice job on getting the view parameters back to normal!
> 
> I think a lot of views you have edited don't require `application = get_application(pk)`, they simply just require the `pk`?

Yeah I plan on going through and removing unnecessary fetches at a later point. For now I've just decoupled the fetching of application from the authentication which will enable us to do that

PR #446, opened 2020-05-07
feat: LT-2414 - Hawk Authentication

andytaylorkainos:


conf/settings.py
```diff
@@ -127,6 +128,17 @@
 
 AUTHENTICATION_BACKENDS = ["django.contrib.auth.backends.ModelBackend"]
 
+HAWK_CREDENTIALS = {
+    "exporter-frontend": {"id": "exporter-frontend", "key": env("LITE_EXPORTER_HAWK_KEY"), "algorithm": "sha256"},
```
stephanieboyd, line 132: minor, constant for the hash algo?

static/denial_reasons/tests.py
```diff
@@ -10,7 +10,7 @@ class DenialReasonsTests(DataTestClient):
     url = reverse("static:denial-reasons:denial-reasons")
 
     def test_get_denial_reasons(self):
-        response = self.client.get(self.url)
+        response = self.client.get(self.url, **self.exporter_headers)
```
stephanieboyd, line 13: Are there gov equivalent tests? The endpoint looks to use shared authentication

andytaylorkainos: I do agree that there should have always been a "can gov user access" or "is user not allowed access" test for endpoints - I'm gonna say it's out of scope for this PR.

The reason the authentication was added to the static endpoints was because it was required for HAWK to work - all the existing tests still pass :)

PR #449, opened 2020-05-12
Bug: LT-2464 - reference integrity error when submitting multiple cases at same time

Mark-Spiers-kainos:


cases/migrations/0015_auto_20200512_1133.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("cases", "0014_auto_20200420_1444"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="casereferencecode", name="year", field=models.IntegerField(editable=False, unique=True),
```
stephanieboyd, line 14: Yaaas

PR #460, opened 2020-05-15
feat: LT-2536-audit-organisation

rorysproule1:


organisations/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+from audit_trail.enums import AuditType
+from organisations.enums import OrganisationType, OrganisationStatus
+
+
+def add_edited_audit_entry(user, organisation, organisation_field, previous_value, new_value):
+    audit_trail_service.create(
+        actor=user,
+        verb=AuditType.UPDATED_ORGANISATION,
+        target=organisation,
+        payload={"organisation_field": organisation_field, "previous_value": previous_value, "new_value": new_value,},
```
a-bek, line 11: For the status updates we have the format like:

`{"status": {"new": "", "old": ""}}`

What do you think of following the same format here for consistency:

`{organisation_field: {"new": new_value, "old": old_value}}`

Or we define a new format? Happy to discuss

audit_trail/serializers.py
```diff
@@ -23,15 +23,28 @@ class Meta:
...
     def get_user(self, instance):
-        return {
-            "first_name": instance.actor.first_name,
-            "last_name": instance.actor.last_name,
-        }
+
+        if AuditType(instance.verb) == AuditType.REGISTER_ORGANISATION:
+            # When an anonymous user is registering for an org,
+            # we pass their email in the payload to use it as the actor later
+            payload = deepcopy(instance.payload)
```
a-bek, line 30: no need to do this copy, just do `"first_name": instance.payload["email"],` below

rorysproule1: Amended :)

PR #466, opened 2020-05-18
feat: LT-1268 - case advanced search

a-bek:


cases/libraries/dates.py
```diff
@@ -0,0 +1,17 @@
+import datetime
+
+
+def make_date_from_params(prefix, data):
```
andytaylorkainos, line 4: nice!

cases/managers.py
```diff
@@ -78,6 +78,61 @@ def order_by_status(self, order=""):
...
+        return self.filter(reference_code__icontains=case_reference)
+
+    def with_organisation(self, organisation_name):
+        return self.filter(organisation__name=organisation_name)
+
+    def with_exporter_site_name(self, exporter_site_name):
+        from applications.models import SiteOnApplication
+
+        case_ids = SiteOnApplication.objects.filter(site__name=exporter_site_name).values_list(
+            "application__id", flat=True
```
andytaylorkainos, line 91: `application__id` only need one `_`

cases/managers.py
```diff
@@ -78,6 +78,61 @@ def order_by_status(self, order=""):
...
+        )
+        return self.filter(pk__in=case_ids)
```
andytaylorkainos, line 93: i think you can do ```self.filter(baseapplication__application_sites__site__name=exporter_site_name)```

cases/managers.py
```diff
@@ -78,6 +78,61 @@ def order_by_status(self, order=""):
...
+        return qs
+
+    def with_party_name(self, party_name):
+        return self.filter(baseapplication__parties__party__name__icontains=party_name)
+
+    def with_party_address(self, party_address):
+        return self.filter(baseapplication__parties__party__address__icontains=party_address)
+
+    def with_goods_related_description(self, goods_related_description):
+        return self.filter(
```
andytaylorkainos, line 130: is this really the requirement (all these description matches on goods)? o.o

cases/managers.py
```diff
@@ -149,24 +221,23 @@ def search(
...
 
-        team_id = user.team.id
-
-        if not include_hidden:
+        if not include_hidden and user is not None:
             EcjuQuery = get_model("cases", "ecjuquery")
 
             case_qs = case_qs.exclude(
-                id__in=EcjuQuery.objects.filter(raised_by_user__team_id=team_id, responded_at__isnull=True)
+                id__in=EcjuQuery.objects.filter(raised_by_user__team_id=user.team.id, responded_at__isnull=True)
```
andytaylorkainos, line 228: i think it'd be better to pass in the team_id from the view via a parameter (it gets used in a few more places along the way)

cases/managers.py
```diff
@@ -180,6 +251,45 @@ def search(
                 case_officer = None
             case_qs = case_qs.assigned_as_case_officer(user=case_officer)
 
+        if case_reference is not None:
```
andytaylorkainos, line 254: what's with all the "not None"s

cases/tests/tests_case_search.py
```diff
@@ -0,0 +1,293 @@
...
+
+        qs_1 = Case.objects.search(flags=[flag_1.name])
+        qs_2 = Case.objects.search(flags=[flag_2.name])
+
+        self.assertEqual(qs_1.count(), 1)
+        self.assertEqual(qs_2.count(), 2)
+
+    def test_filter_by_country(self):
+        """
+        What qualifies as a country on a case?
```
andytaylorkainos, line 115: destination i think (parties too)

cases/managers.py
a-bek: yes 😨

cases/tests/factories.py
```diff
@@ -1,7 +1,156 @@
...
+                self.statuses.add(status)
+
+
+class UserOrganisationRelationshipFactory(factory.django.DjangoModelFactory):
+    organisation = factory.SubFactory(OrganisationFactory)
+    role = factory.SubFactory(RoleFactory)
+    status = UserStatuses.ACTIVE
+
+
+class SiteFactory(factory.django.DjangoModelFactory):
```
jfaracik, line 100: We already have site and role factories in organisations app afaik :)

cases/tests/tests_case_search.py
```diff
@@ -0,0 +1,329 @@
+from difflib import SequenceMatcher
```
andytaylorkainos, line 1: can you move these unit tests into the already existing file for case searching (`cases/tests/tests_search_cases.py`)?
Just use a different class name :-)

cases/libraries/dates.py
andytaylorkainos: maybe rename `data` to `params`

cases/views/search/views.py
```diff
@@ -31,17 +31,22 @@ def get(self, request, *args, **kwargs):
         # we include hidden cases in non work queues (all cases, all open cases)
         # and if the flag to include hidden is added
         include_hidden = not is_work_queue or str_to_bool(request.GET.get("hidden"))
+        filters = {
+            key.replace("[]", ""): value if "[]" not in key else request.GET.getlist(key)
```
andytaylorkainos, line 35: but why?

cases/models.py
```diff
@@ -52,7 +52,7 @@ class Case(TimestampableModel):
     Base model for applications and queries
     """
 
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False, db_index=True)
```
andytaylorkainos, line 55: are UUIDs not indexable by default? :O

cases/managers.py
andytaylorkainos: still think this would save a couple of queries before the response

cases/managers.py
```diff
@@ -78,6 +78,81 @@ def order_by_status(self, order=""):
...
+            | Q(baseapplication__application_sites__site__address__region__icontains=exporter_site_address)
+            | Q(baseapplication__application_sites__site__address__postcode__icontains=exporter_site_address)
+            | Q(baseapplication__application_sites__site__address__address__icontains=exporter_site_address)
+        )
+
+    def with_control_list_entry(self, control_list_entry):
+        return self.filter(baseapplication__goods__good__control_list_entries__rating__in=[control_list_entry])
+
+    def with_flags(self, flags):
+        return self.filter(
```
andytaylorkainos, line 110: 😭

cases/managers.py
```diff
@@ -78,6 +78,81 @@ def order_by_status(self, order=""):
...
+        return self.filter(organisation__name=organisation_name)
+
+    def with_exporter_site_name(self, exporter_site_name):
+        return self.filter(baseapplication__application_sites__site__name=exporter_site_name)
+
+    def with_sla_days_elapsed(self, sla_days_elapsed):
+        return self.filter(sla_days=sla_days_elapsed)
+
+    def with_exporter_site_address(self, exporter_site_address):
+        return self.filter(
```
andytaylorkainos, line 97: 😭

cases/views/search/views.py
a-bek: Some of the filters are passed with `[]` in the key, I remove it from the key so it matches the parameter in the search function.

cases/managers.py
a-bek: It's used in conjunction with the user object and the `id` is cached so its only queried once on the first call `user.team.id`

cases/tests/tests_case_search.py
a-bek: If you don't mind I have renamed them to:

`tests_case_search_advanced`
`tests_case_search`

It would be nice to keep them separate because one uses factories and the other doesn't.

cases/views/search/views.py
andytaylorkainos: as discussed, can you shift the logic to the frontend so that the API gets the query param in the correct format :)

PR #468, opened 2020-05-19
bugfix: LT-2529 - Audit automation case status changes

andytaylorkainos:
Update to `submit_application` in `test_helpers/clients.py` so that an audit is created (to behave like the real system).

Added an audit to `workflow/automation.py` on line `71` for when the routing rules changes the case status.

Added relevant test to `cases/tests/tests_rerun_routing_rules.py`


static/statuses/enums.py
```diff
@@ -115,13 +115,13 @@ class CaseStatusEnum:
 
     @classmethod
     def get_text(cls, status):
-        for k, v in cls.choices:
+        for k, v in [*cls.choices, (cls.DRAFT, "Draft")]:
```
stephanieboyd, line 118: Stick a comment in that it's a system status. (I know it's above but just to be clear here)

test_helpers/clients.py
```diff
@@ -299,7 +302,10 @@ def create_team(name: str):
         return team
 
     @staticmethod
-    def submit_application(application: BaseApplication):
+    def submit_application(application: BaseApplication, user: ExporterUser = None):
+        if not user:
+            user = UserOrganisationRelationship.objects.filter(organisation_id=application.organisation_id).first().user
```
stephanieboyd, line 307: Are you sure you're not relying on the DB returning it in a particular order here?

applications/tests/tests_open_application_submit.py
```diff
@@ -88,7 +90,20 @@ def test_standard_application_declaration_submit_success(self):
...
         self.assertTrue(UUID(SystemFlags.ENFORCEMENT_CHECK_REQUIRED) in case.flags.values_list("id", flat=True))
 
-    def test_standard_application_declaration_submit_tcs_false_failure(self):
+        case_status_audits = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).values_list(
+            "payload", flat=True
+        )
+        self.assertIn(
+            {
+                "status": {
+                    "new": CaseStatusEnum.get_text(CaseStatusEnum.SUBMITTED),
```
stephanieboyd, line 99: Are there any tests for resubmitting? They'd have a status change from applicant editing that should need checked

applications/tests/tests_open_application_submit.py
```diff
@@ -88,7 +90,20 @@ def test_standard_application_declaration_submit_success(self):
         self.assertEqual(case.baseapplication.agreed_to_foi, True)
         self.assertTrue(UUID(SystemFlags.ENFORCEMENT_CHECK_REQUIRED) in case.flags.values_list("id", flat=True))
 
-    def test_standard_application_declaration_submit_tcs_false_failure(self):
+        case_status_audits = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).values_list(
```
a-bek, line 93: Would this also pass if you get the oldest audit and compare payload? The draft -> submitted audit is always the first for a case right?

test_helpers/clients.py
andytaylorkainos: Nope! just need to pick a user to submit the case (audit with)

applications/tests/tests_open_application_submit.py
andytaylorkainos: There is now!

applications/tests/tests_open_application_submit.py
andytaylorkainos: The reason we don't get the latest audit is because automation rules `may` set it to a different status immediately after it's submitted

test_helpers/clients.py
andytaylorkainos: I'll add another bit of logic to try and get the current exporter_user (the one that the tests mostly use everywhere) before defaulting to this

andytaylorkainos: On second thought, this allows us to create a draft as user 1 and submit it as user 2.
(in addition to that, I cant get scope to the expoter user from a static method - this method is used in other places statically)

PR #481, opened 2020-05-22
bug: LT-2651 - Cases flags performance

a-bek:


BenAndrew-Kainos:
Have you tested that this reduces the number of queries?

a-bek:
@BenAndrew-Kainos 
It increases the number of queries by the number of models that store flags, far less expensive than joining all the tables in the query.

PR #483, opened 2020-05-22
bug: Record case submitter

BenAndrew-Kainos:


letter_templates/context_generator.py
```diff
@@ -51,7 +50,7 @@ def get_document_context(case):
         "current_date": date,
         "current_time": time,
         "details": _get_details_context(case),
-        "applicant": _get_applicant_context(applicant_audit.actor) if applicant_audit else None,
+        "applicant": _get_applicant_context(case.submitted_by) if case.submitted_by else None,
```
stephanieboyd, line 53: When would there be no submitted by?

BenAndrew-Kainos: good point

PR #486, opened 2020-05-25
feat: SIEL layout

BenAndrew-Kainos:


letter_templates/layouts/siel.html
```diff
@@ -211,37 +211,39 @@ <h1>Standard Individual Export Licence</h1>
...
-          {{ applicant.primary_site.address.address_line_1 }}<br>
-          {{ applicant.primary_site.address.address_line_2 }}<br>
-          {{ applicant.primary_site.address.postcode }}<br>
-          {{ applicant.primary_site.address.city }}<br>
-          {{ applicant.primary_site.address.region }}<br>
-          {{ applicant.primary_site.address.country.name }}</span>
+          <span class="cell__uppercase">
+              {{ organisation.name }}<br>
+              {{ organisation.primary_site.address_line_1 }}<br>
+              {{ organisation.primary_site.address_line_2 }}<br>
```
rorysproule1, line 217: Just checking, is it possible for an orgs primary site to be non UK? as they wouldn't have address_line_2.

BenAndrew-Kainos: in that case the line is just blank, so we are ok :)

PR #490, opened 2020-05-27
feat: LT-874 - Search open general licences

jfaracik:
* Fixed filtering for control list entries (no more duplicates) 
* Exporters can now access the list and detail endpoints for open general licences


open_general_licences/views.py
```diff
@@ -49,9 +45,11 @@ def filter_queryset(self, queryset):
 
         filtered_qs = filtered_qs.filter(status=filter_data.get("status", "active"))
 
-        return filtered_qs
+        return filtered_qs.distinct()
```
efahy, line 48: Why do you need to distinct the qs?  Better not to do it if you don't need it

PR #491, opened 2020-05-28
Added fix for when document scan not successful and no task found - d…

efahy:
…efault to zero attempts

### DIFF
--- a/documents/tasks.py
+++ b/documents/tasks.py
@@ -35,12 +35,14 @@ def scan_document_for_viruses(document_id):
                 f"An unexpected error occurred when scanning document '{document_id}' -> {type(exc).__name__}: {exc}"
             )
 
-        # Get the task's current attempt number by retrieving the previous attempt number and adding 1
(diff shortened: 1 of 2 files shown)


documents/tasks.py
```diff
@@ -35,12 +35,14 @@ def scan_document_for_viruses(document_id):
...
 
-        # Get the task's current attempt number by retrieving the previous attempt number and adding 1
+        # Document scan not completed. Get the task's current attempt number by retrieving the previous attempt
+        # number and adding 1
         previous_attempt = (
             Task.objects.filter(queue=TASK_QUEUE, task_params__contains=document_id)
             .values_list("attempts", flat=True)
             .first()
-        )
+        ) or 0  # If no results found, default to zero
```
a-bek, line 44: Should the condition below: `if current_attempt >= settings.MAX_ATTEMPTS`

be `if current_attempt > settings.MAX_ATTEMPTS`?

If `MAX_ATTEMPTS` is 1, it will never work!

PR #492, opened 2020-05-28
feat: Picklists filtering

jfaracik:


picklists/tests/tests_views.py
```diff
@@ -33,6 +33,13 @@ def test_gov_user_can_see_all_their_teams_picklist_items_excluding_deactivated(s
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(len(response_data["results"]), 3)
 
+    def test_gov_user_can_see_all_their_teams_picklist_items_filter_by_name(self,):
```
a-bek, line 36: Maybe add one that doesn't match as well, so the test breaks if the filter by name is removed

PR #493, opened 2020-05-29
feat: LT-2448 - API/HMRC Integration

andytaylorkainos:
```
{
  "licence": {
      "id": "<String>",
      "reference": "<String>",  # case_reference_code
      "type": "<String>",  # case_type_reference
      "status": "<String>",  # case_status_status
      "start_date": "<String>",  # YYYY-MM-DD
      "end_date": "<String>",  # sart_date + duration, YYYY-MM-DD
      "organisation": {
        "name": "<String>",
        "address": {
          "line_1": "<String>",  # primary_site_name
          "line_2": "<String>",  # primary_site_address_line_1
          "line_3": "<String>", # primary_site_address_line_2
          "line_4": "<String>",  # primary_site_address_city
          "line_5": "<String>",  # primary_site_address_region
          "postcode": "<String>",  # primary_site_address_postcode
          "country": {  # primary_site_address_country
            "id": "<String>",
            "name": "<String>"
          }  
        }
      },
      "end_user": {  # only for standard applications
        "name": "<String>",
        "address": {
          "line_1": "<String>",  
          "country": {
            "id": "<String>",
            "name": "<String>"
          }
        }
      },
      "countries": [  # only for open applications
        {
          "id": "<String>",
          "name": "<String>"
        }
      ],
      "goods": [
        {
          "id": "<String>", 
          "description": "<String>",
          "usage": "<Int>", 
          "unit": "<String>",  # only for standard applications
          "licenced_quantity": "<Int>",  # only for standard applications
          "licenced_value": "<Int>",  # only for standard applications
        }
      ]
    }
}
```


documents/libraries/process_document.py
```diff
@@ -11,7 +11,7 @@ def process_document(document):
         scan_document_for_viruses(str(document.id))
     else:
         try:
-            scan_document_for_viruses.now(str(document.id))
+            scan_document_for_viruses.now(str(document.id), is_background_task=False)
```
BenAndrew-Kainos, line 14: Why have we disabled background task?

documents/tasks.py
```diff
@@ -4,51 +4,49 @@
...
 
 @background(schedule=0, queue=TASK_QUEUE)
-def scan_document_for_viruses(document_id):
-    """
-    Executed by background worker process or synchronous depending on BACKGROUND_TASK_RUN_ASYNC.
-    """
-
-    from documents.models import Document
-
+def scan_document_for_viruses(document_id, is_background_task=True):
```
efahy, line 15: Add comment somewhere explaining what `is_background_task` does, since it doesn't quite do what it says on the tin

licences/serializers/hmrc_integration.py
```diff
@@ -0,0 +1,209 @@
...
+            "country",
+        )
+        read_only_fields = fields
+
+    def __init__(self, *args, **kwargs):
+        self._address_name = kwargs.pop("address_name")
+        super().__init__(*args, **kwargs)
+
+    def get_line_1(self, _):
+        return self._address_name
```
BenAndrew-Kainos, line 134: you can replace all of these using the `source` argument like 

`line_2 = serializers.CharField(source="address_line_1")`

licences/serializers/hmrc_integration.py
```diff
@@ -0,0 +1,209 @@
...
+            self.fields.pop("end_user")
+
+        if not (
+            hasattr(self.instance.application, "openapplication")
+            and self.instance.application.openapplication.application_countries.exists()
+        ):
+            self.fields.pop("countries")
+
+    def get_reference(self, instance):
+        return instance.application.reference_code
```
BenAndrew-Kainos, line 54: you can replace all of these using the source argument like

`reference = serializers.CharField(source="application.reference_code")`

documents/tasks.py
```diff
@@ -4,51 +4,49 @@
...
-            Task.objects.filter(queue=TASK_QUEUE, task_params__contains=document_id)
-            .values_list("attempts", flat=True)
-            .first()
-        ) or 0
-
-        current_attempt = previous_attempt + 1
-
-        if current_attempt >= settings.MAX_ATTEMPTS:
-            logging.warning(f"MAX_ATTEMPTS {settings.MAX_ATTEMPTS} for document '{document_id}' has been reached")
+    if is_background_task:
```
efahy, line 34: Move to private function for exception handling

documents/tasks.py
```diff
@@ -4,51 +4,49 @@
...
+            current_attempt = task.attempts + 1
+
+            # Delete the document's file from S3 if the task has been attempted MAX_ATTEMPTS times
+            if current_attempt >= MAX_ATTEMPTS:
+                logging.warning(f"Maximum attempts of {MAX_ATTEMPTS} for document '{document_id}' has been reached")
+                doc.delete_s3()
+    else:
+        doc.delete_s3()
+
+    # Raise an exception (this will result in a serializer error or cause the task (if any) to be marked as 'Failed')
```
efahy, line 51: ... and be retried if there are retry attempts left

licences/serializers/hmrc_integration.py
```diff
@@ -0,0 +1,209 @@
...
+from goodstype.models import GoodsType
+from licences.helpers import get_approved_goods_types, get_approved_goods_on_application
+from licences.models import Licence
+from organisations.models import Organisation
+from parties.models import Party
+from static.countries.models import Country
+from static.statuses.enums import CaseStatusEnum
+
+
+class HMRCIntegrationLicenceSerializer(serializers.ModelSerializer):
```
BenAndrew-Kainos, line 15: For performance & the way used this serializer I would suggest not using a serializer and instead creating a dict.

At the very least please do not use a ModelSerializer

licences/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            # Get the task's current attempt number by retrieving the previous attempts and adding 1
+            current_attempt = task.attempts + 1
+
+            # Schedule a new task if the current task has been attempted MAX_ATTEMPTS times;
+            # HMRC Integration tasks need to be resilient and keep retrying post-failure indefinitely.
+            # This logic will make MAX_ATTEMPTS attempts to send licence changes according to the Django Background Task
+            # Runner scheduling, then wait TASK_BACK_OFF seconds before starting the process again.
+            if current_attempt >= MAX_ATTEMPTS:
+                schedule_max_tried_task_as_new_task(licence_id)
```
BenAndrew-Kainos, line 49: Do we have to do this. I would prefer if we could use a django background process method rather than writing our own code to manage MAX_ATTEMPTS and scheduling

licences/libraries/hmrc_integration_operations.py
```diff
@@ -0,0 +1,26 @@
+from rest_framework import status
+
+from conf.requests import post
+from conf.settings import LITE_HMRC_INTEGRATION_URL
+from licences.models import Licence
+from licences.serializers.hmrc_integration import HMRCIntegrationLicenceSerializer
+
+HAWK_CREDENTIALS = "lite-api"
+REQUEST_TIMEOUT = 5  # Maximum time, in seconds, to wait for a request to return a byte
```
efahy, line 9: Should these be in an .env? Everything else like this is in an env.  Maybe not intuitive that only these settings are not in an env.

licences/tests/test_get_licence.py
```diff
@@ -0,0 +1,32 @@
...
+            self.f680_application,
+            self.gifting_application,
+            self.exhibition_application,
+            self.open_application,
+        ]
+        self.licences = {
+            application: self.create_licence(application, is_complete=True) for application in self.applications
+        }
+
+    def test_get_licence(self):
```
BenAndrew-Kainos, line 27: There should already be a test class for getting a LIcence??

licences/tests/test_hmrc_integration.py
```diff
@@ -0,0 +1,234 @@
...
+        self.template = self.create_letter_template(
+            name="Template",
+            case_types=[CaseTypeEnum.SIEL.id],
+            decisions=[Decision.objects.get(name=AdviceType.APPROVE)],
+        )
+        self.create_generated_case_document(self.standard_application, self.template, advice_type=AdviceType.APPROVE)
+
+    def test_approve_standard_application_licence_success(self):
+        pass
+        # TODO: assert request sent to HMRC
```
BenAndrew-Kainos, line 220: Looks like you still have more to do

licences/models.py
```diff
@@ -16,3 +17,16 @@ class Licence(TimestampableModel):
...
+
+    def save(self, *args, **kwargs):
+        super(Licence, self).save(*args, **kwargs)
+        if self.is_complete:
+            self.send_to_hmrc_integration()
+
+    def send_to_hmrc_integration(self):
+        from licences.tasks import send_licence_to_hmrc_integration
+
+        if BACKGROUND_TASK_ENABLED:
```
efahy, line 29: I propose that you make public function where the tasks are that takes an `is_background_task` parameter and does this if in it

licences/serializers/hmrc_integration.py
```diff
@@ -0,0 +1,209 @@
...
+    reference = serializers.SerializerMethodField()
+    type = serializers.SerializerMethodField()
+    status = serializers.SerializerMethodField()
+    end_date = serializers.SerializerMethodField()
+    organisation = serializers.SerializerMethodField()
+    end_user = serializers.SerializerMethodField()
+    countries = serializers.SerializerMethodField()
+    goods = serializers.SerializerMethodField()
```
efahy, line 23: Consider pointing to other serializers instead of using `SerializerMethodField`s

licences/serializers/hmrc_integration.py
```diff
@@ -0,0 +1,209 @@
...
+            ),
+            many=True,
+        ).data
+
+    def get_goods(self, instance):
+        if instance.application.goods.exists():
+            approved_goods = get_approved_goods_on_application(instance.application)
+            return HMRCIntegrationGoodsOnApplicationSerializer(approved_goods, many=True).data
+
+        if instance.application.goods_type.exists():
```
efahy, line 84: My personal preference would be to write `if ... elif ... else` here.  I think it makes it more readable.  But happy for you to call it

licences/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+        logging.warning(str(exc))
+    except Exception as exc:  # noqa
+        logging.warning(
+            f"An unexpected error occurred when sending licence '{licence_id}' changes to HMRC Integration -> "
+            f"{type(exc).__name__}: {exc}"
+        )
+
+    error_message = f"Failed to send licence '{licence_id}' changes to HMRC Integration"
+
+    if is_background_task:
```
efahy, line 35: Move exception handling to private method

licences/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+            if current_attempt >= MAX_ATTEMPTS:
+                schedule_max_tried_task_as_new_task(licence_id)
+
+        # Raise an exception (this will cause the task to be marked as 'Failed')
+        raise Exception(error_message)
+    else:
+        logging.error(error_message)
+
+
+def schedule_max_tried_task_as_new_task(licence_id):
```
efahy, line 57: Make a comment to say you've added this to aid unit testing

documents/libraries/process_document.py
andytaylorkainos: this runs when background tasks `are` disabled - theres a condition above (`.now()`should only be used with `is_background_task=False`)

This flag is used within the task to determine how to handle exceptions (theyre handled differently depending on background tasks being enabled)

licences/tasks.py
andytaylorkainos: Sadly this is the only workaround available for ensuring that the task is never destroyed.
If LITE or HMRC Integration goes down for an extended period of time, we do not want the task to stop being attempted

licences/tests/test_get_licence.py
andytaylorkainos: this test is the only use-case of `reverse("licences:licence"` :(

licences/tests/test_hmrc_integration.py
andytaylorkainos: Yeah, this is the integration test :) -> dont worry ill be asking for indepth testing reviews soon haha

licences/serializers/hmrc_integration.py
andytaylorkainos: As discussed:
I used a serializer as we only ever opt-out from using them if we have insane performance implications.

The performance requirements for this serializer aren't as demanding as our other endpoints as it will only ever be executed upon finalising a licence (1 entity at a time).

However, I have updated them from inheriting from `ModelSerializer` to `Serializer` which will have improved the performance a little, but it has definitely cleaned the code up a lot which is a huge plus!

licences/tests/test_hmrc_integration.py
andytaylorkainos: Done now!

documents/libraries/process_document.py
BenAndrew-Kainos: Does this variable still allow background task running is they have it set in their .env

andytaylorkainos: Yes :) it doesnt affect the task itself

licences/tasks.py
BenAndrew-Kainos: ok

licences/serializers/hmrc_integration.py
BenAndrew-Kainos: 👍

licences/tasks.py
BenAndrew-Kainos: ok

licences/tests/test_get_licence.py
BenAndrew-Kainos: damn ok

documents/libraries/process_document.py
BenAndrew-Kainos: Could you rename to something like `allows_background_task` then

andytaylorkainos: how about `called_as_background_task`?
The `.now()` is what really controls whether it is being run as a task or not and the boolean is only used to handle the exceptions if the function fails

licences/libraries/hmrc_integration_operations.py
andytaylorkainos: moved to settings and increased it to 60 to match the others

conf/requests.py
```diff
@@ -0,0 +1,104 @@
...
+
+def put(url, data, headers=None, hawk_credentials=None, timeout=None):
+    return make_request("PUT", url, data=data, headers=headers, hawk_credentials=hawk_credentials, timeout=timeout)
+
+
+def make_request(method, url, data=None, headers=None, hawk_credentials=None, timeout=None):
+    headers = headers or {}  # If no headers are supplied, default to an empty dictionary
+    headers["content-type"] = "application/json"
+
+    if HAWK_AUTHENTICATION_ENABLED:
```
efahy, line 41: I see hawk_credentials parameter is defaulted to `None` - would it be wise to raise a helpful error if `HAWK_AUTHENTICATION_ENABLED` is `True` and hawk credentials are not supplied?  How does the Sender behave if it isn't given credentials?

conf/tests/test_requests.py
```diff
@@ -0,0 +1,86 @@
...
+    @mock.patch("conf.requests.send_request")
+    @mock.patch("conf.requests.get_hawk_sender")
+    def test_make_request_calls_send_request_with_hawk(self, get_hawk_sender, send_request, verify_api_response):
+        mocked_hawk_sender = MockHawkSender("authentication-header")
+        mocked_response = None
+        get_hawk_sender.return_value = mocked_hawk_sender
+        send_request.return_value = mocked_response
+        verify_api_response.return_value = None
+
+        make_request("POST", "url", data={"data": "stuff"}, headers=None, hawk_credentials="fake-id", timeout=5)
```
efahy, line 66: Can you parameterise the `make_request` tests to do all the HTTP methods? Maybe that doesn't add any value?

documents/tasks.py
```diff
@@ -4,51 +4,62 @@
...
+        if document.virus_scanned_at:
+            logging.info(f"Document '{document_id}' has already been scanned; safe={document.safe}")
+            return
 
         try:
-            return doc.scan_for_viruses()
+            document.scan_for_viruses()
         except VirusScanException as exc:
-            logging.warning(str(exc))
+            _handle_exception(str(exc), document, scheduled_as_background_task)
```
efahy, line 33: Far better 👍

documents/tasks.py
```diff
@@ -4,51 +4,62 @@
...
+                document.delete_s3()
+    else:
+        document.delete_s3()
 
-    # Raise an exception (this will cause the task to be marked as 'Failed' and the attempt number to be updated)
-    raise Exception(f"Failed to scan document '{document_id}'")
+    # Raise an exception.
+    # This will result in a serializer error or
+    # cause the task to be marked as 'Failed' and retried if there are retry attempts left
+    raise Exception(f"Failed to scan document '{document.id}'")
```
efahy, line 65: Might be worth including the retry number in the log message when we have it

conf/requests.py
andytaylorkainos: done!

conf/tests/test_requests.py
andytaylorkainos: I have tested all of the requests as separate tests :)

documents/tasks.py
andytaylorkainos: done!

licences/tasks.py
```diff
@@ -0,0 +1,92 @@
...
+def send_licence_to_hmrc_integration(licence_id, licence_reference, scheduled_as_background_task=True):
+    """
+    Sends licence details to HMRC Integration
+    :param licence_id:
+    :param licence_reference:
+    :param scheduled_as_background_task: Has this function has been scheduled as a task (used for error handling)
+    """
+
+    with transaction.atomic():
+        licence = Licence.objects.select_for_update(nowait=True).get(id=licence_id)
```
BenAndrew-Kainos, line 53: Could you comment why nowait is True

andytaylorkainos: ```Usually, if another transaction has already acquired a lock on one of the selected rows, the query will block until the lock is released. If this is not the behavior you want, call select_for_update(nowait=True). This will make the call non-blocking. If a conflicting lock is already acquired by another transaction, DatabaseError will be raised when the queryset is evaluated.``` -> essentially we want the error to be thrown so it will get logged

PR #494, opened 2020-06-01
feat: LT-2520 - Specify location where records are held when creating a site

stephanieboyd:


applications/libraries/sites_helpers.py
```diff
@@ -70,8 +71,29 @@ def add_sites_to_application(user: ExporterUser, new_sites: Union[QuerySet, List
...
+        if removed_site_ids:
+            sites_still_on_other_applications = (
+                SiteOnApplication.objects.exclude(application=application)
+                .exclude(application__status__status="draft")
+                .filter(site__id__in=removed_site_ids)
+            )
+            removed_sites_no_longer_on_other_applications = removed_sites.exclude(
+                site__id__in=sites_still_on_other_applications.values_list("site_id", flat=True)
+            )
+            for sites_on_application in removed_sites_no_longer_on_other_applications:
```
a-bek, line 87: You should be able to do this in one update using the queryset `update` method

https://docs.djangoproject.com/en/3.0/ref/models/querysets/#django.db.models.query.QuerySet.update

applications/libraries/sites_helpers.py
```diff
@@ -70,8 +71,29 @@ def add_sites_to_application(user: ExporterUser, new_sites: Union[QuerySet, List
...
+                site = Site.objects.get(id=sites_on_application.site_id)
+                site.is_used_on_application = False
+                site.save()
+
     # Save the new sites
     for site in added_sites:
+        if application.status.status != CaseStatusEnum.DRAFT:
```
a-bek, line 94: Same here!

applications/views/applications.py
```diff
@@ -364,6 +364,12 @@ def put(self, request, pk):
                 create_submitted_audit(request, application, old_status)
                 run_routing_rules(application)
 
+                # Mark the sites on this application as used so their details are no longer editable
+                for site_on_application in SiteOnApplication.objects.filter(application=application):
```
a-bek, line 368: bulk update here too?

organisations/views/sites.py
```diff
@@ -92,3 +113,59 @@ def get_serializer_class(self):
             return SiteViewSerializer
         else:
             return SiteCreateUpdateSerializer
+
+    def patch(self, request, *args, **kwargs):
```
a-bek, line 117: Can `pk` not be explicit in the method signature?

organisations/views/sites.py
```diff
@@ -92,3 +113,59 @@ def get_serializer_class(self):
...
+            return JsonResponse(
+                data={"errors": {"site_records_stored_here": [strings.Site.NO_RECORDS_LOCATED_AT]}},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+    def perform_update(self, serializer):
+        original_instance = self.get_object()
+        updated_instance = serializer.save()
+
+        if getattr(updated_instance, "site_records_located_at"):
```
a-bek, line 143: Can these fields not be accessed directly, rather than using `get_attr`

applications/libraries/sites_helpers.py
iichr: Done 👌

applications/libraries/sites_helpers.py
iichr: Sorted

applications/views/applications.py
iichr: Bueno

organisations/views/sites.py
iichr: Yes ⚔️  👀

organisations/views/sites.py
iichr: extracted as a variable instead, only used twice

applications/libraries/sites_helpers.py
```diff
@@ -58,23 +59,46 @@ def add_sites_to_application(user: ExporterUser, new_sites: Union[QuerySet, List
...
         added_sites=added_sites,
     )
 
+    # Check if site has been removed and if it is no longer used on other applications set "is_used_on_application" to False
+    # Only do this for minor/major edits
+    if application.status.status != CaseStatusEnum.DRAFT:
+        removed_site_ids = removed_sites_on_application.values_list("site_id", flat=True)
+        if removed_site_ids:
+            sites_still_on_other_applications = (
+                SiteOnApplication.objects.exclude(application=application)
```
Mark-Spiers-kainos, line 80: can you select related `application__status` and can does single dunder work for `site__id...`

a-bek: I don't think the select_related is necessary because the data isn't being used, its just for filtering.

organisations/models.py
```diff
@@ -88,6 +88,8 @@ class Site(TimestampableModel):
     )
     users = models.ManyToManyField(UserOrganisationRelationship, related_name="sites")
     address = models.ForeignKey(Address, related_name="site", on_delete=models.DO_NOTHING)
+    site_records_located_at = models.ForeignKey("self", related_name="records", on_delete=models.DO_NOTHING, null=True)
+    is_used_on_application = models.BooleanField(default=None, null=True)
```
Mark-Spiers-kainos, line 92: Can this be null? shouldn't it only be True/False?

organisations/urls.py
```diff
@@ -26,4 +26,5 @@
     path("<uuid:org_pk>/roles/<uuid:pk>/", roles.RoleDetail.as_view(), name="role"),
     path("permissions/", roles.PermissionsView.as_view(), name="permissions"),
     path("<uuid:pk>/activity/", activities.OrganisationActivityView.as_view(), name="activities"),
```
efahy, line 28: For consistency, perhaps you should change this URL to `.../organisations-activity/`?

efahy: Actually ignore me

organisations/models.py
stephanieboyd: After chatting, we've done a manual migration step to set this field accurately for existing data.

static/management/commands/seedorganisation.py
```diff
@@ -62,6 +62,10 @@ def seed_organisation(cls, org_name: str, org_type: str, no_of_sites: int, no_of
         # Since OrganisationFactory has already created a Site, subtract 1 from total number of sites to seed
         sites += [SiteFactory(organisation=organisation) for _ in range(no_of_sites - 1)]
 
+        for site in sites:
```
andytaylorkainos, line 65: this could be set inside `SiteFactory`

stephanieboyd: As discussed, will leave as is

PR #497, opened 2020-06-02
feat: 1252 Set document addressee

BenAndrew-Kainos:


cases/generated_documents/helpers.py
```diff
@@ -22,18 +25,25 @@ def html_to_pdf(request, html: str, template_name: str):
...
     if not template_id:
-        raise AttributeError(strings.Cases.MISSING_TEMPLATE)
+        raise AttributeError(strings.Cases.GeneratedDocuments.MISSING_TEMPLATE)
 
     text = request_params.get("text")
     if not text:
-        raise AttributeError(strings.Cases.MISSING_TEXT)
+        raise AttributeError(strings.Cases.GeneratedDocuments.MISSING_TEXT)
+
+    addressee = request_params.get("addressee")
```
a-bek, line 34: Since the API doesn't have the notion of an "addressee" and the title is more for UI output, what do you think of it being passed around as "party" in the backend, as it is a party object?

Other than that 👍

BenAndrew-Kainos: I think perhaps additional_contact might be better as it's technically only that party type

BenAndrew-Kainos: Renamed to additional_contact

cases/generated_documents/helpers.py
```diff
@@ -22,18 +25,27 @@ def html_to_pdf(request, html: str, template_name: str):
...
     text = request_params.get("text")
     if not text:
-        raise AttributeError(strings.Cases.MISSING_TEXT)
+        raise AttributeError(strings.Cases.GeneratedDocuments.MISSING_TEXT)
+
+    additional_contact = request_params.get("addressee")
+    if additional_contact:
+        try:
+            additional_contact = Party.objects.get(type=PartyType.ADDITIONAL_CONTACT, id=additional_contact)
+        except ObjectDoesNotExist:
```
a-bek, line 38: `Party.DoesNotExist` is better practice than a generic exception

BenAndrew-Kainos: Done

PR #502, opened 2020-06-04
bug: LT-2712 - Surrendered status error

a-bek:


applications/views/applications.py
```diff
@@ -423,12 +423,7 @@ def put(self, request, pk):
                 )
 
         if data["status"] == CaseStatusEnum.SURRENDERED:
-            if not Licence.objects.filter(application=application, is_complete=True).exists():
-                return JsonResponse(
-                    data={"errors": [strings.Applications.Generic.Finalise.Error.SURRENDER]},
-                    status=status.HTTP_400_BAD_REQUEST,
-                )
-            Licence.objects.get(application=application, is_complete=True).delete()
+            Licence.objects.filter(application=application, is_complete=True).delete()
```
BenAndrew-Kainos, line 426: Could you leave a TODO above this to update to handle multiple licences in the future (or we will have a big problem)

Also you should leave that if check in because they shouldn't be able to surrender if they don't have a licence

PR #509, opened 2020-06-09
feat: Annual Licence returns

BenAndrew-Kainos:


compliance/helpers.py
```diff
@@ -0,0 +1,40 @@
...
+    return references, cleaned_text
+
+
+def fetch_and_validate_licences(references, organisation_id):
+    licence_ids = list(
+        Licence.objects.filter(reference_code__in=references, application__organisation_id=organisation_id).values_list(
+            "id", flat=True
+        )
+    )
+    if len(licence_ids) != len(references) or len(references) == 0:
```
a-bek, line 37: If you have `if len(references) == 0` check at the top you don't have to hit the db when `references == 0`

BenAndrew-Kainos:
> How come the csv processing is split across frontend and api? Shouldn't just one component handle it (frontend)?

What do you mean. The processing is API only

compliance/helpers.py
BenAndrew-Kainos: Done

compliance/models.py
```diff
@@ -0,0 +1,16 @@
...
+
+from common.models import CreatedAt
+from licences.models import Licence
+from organisations.models import Organisation
+
+
+class OpenLicenceReturns(CreatedAt):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    organisation = models.ForeignKey(Organisation, on_delete=deletion.CASCADE)
+    file = models.TextField(blank=False, null=False)
```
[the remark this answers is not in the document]
BenAndrew-Kainos: Rename to returns_data

a-bek:
> > How come the csv processing is split across frontend and api? Shouldn't just one component handle it (frontend)?
> 
> What do you mean. The processing is API only

The API receives a csv file and reads from it. The frontend receives data from the API and creates a csv from it. Why not do all the csv work on the frontend, and just pass the data to the API?

BenAndrew-Kainos:
> > > How come the csv processing is split across frontend and api? Shouldn't just one component handle it (frontend)?
> > 
> > 
> > What do you mean. The processing is API only
> 
> The API receives a csv file and reads from it. The frontend receives data from the API and creates a csv from it. Why not do all the csv work on the frontend, and just pass the data to the API?

This is consistent with how we do other text file processing.  The frontend is responsible for converting the CSV -> text on upload and text -> CSV on download. However all the processing & validation is done on the API as we need the database & to save results

applications/views/applications.py
```diff
@@ -614,6 +614,7 @@ def put(self, request, pk):
                     return JsonResponse(data={"errors": errors}, status=status.HTTP_400_BAD_REQUEST)
 
             data["application"] = application
+            data["reference_code"] = application.reference_code
```
efahy, line 617: Its existing code, I know, but how would you feel about renaming `data` to `licence_data` or `new_licence_data`?  This is a long method so its not immediately obvious what `data` is

compliance/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+
+from rest_framework.exceptions import ValidationError
+
+from licences.models import Licence
+from lite_content.lite_api.strings import Compliance
+
+TOTAL_COLUMNS = 5
+
+
+def read_and_validate_csv(text):
```
efahy, line 11: Some comments in here please explaining the context and perhaps summarising the expectations as regards format of data

compliance/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+    references = set()
+    cleaned_text = ""
+
+    try:
+        csv_reader = csv.reader(text.split("\n"), delimiter=",")
```
efahy, line 16: We should probably sanitise the data before storing

compliance/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+    if len(references) == 0:
+        raise ValidationError({"file": [Compliance.OpenLicenceReturns.INVALID_LICENCES]})
+
+    licence_ids = list(
+        Licence.objects.filter(reference_code__in=references, application__organisation_id=organisation_id).values_list(
+            "id", flat=True
+        )
+    )
+    if len(licence_ids) != len(references):
+        raise ValidationError({"file": [Compliance.OpenLicenceReturns.INVALID_LICENCES]})
```
efahy, line 41: It would be good if we could highlight to the user which licence references are not valid

compliance/models.py
```diff
@@ -0,0 +1,16 @@
...
+from common.models import CreatedAt
+from licences.models import Licence
+from organisations.models import Organisation
+
+
+class OpenLicenceReturns(CreatedAt):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    organisation = models.ForeignKey(Organisation, on_delete=deletion.CASCADE)
+    returns_data = models.TextField(blank=False, null=False)
+    year = models.PositiveSmallIntegerField(blank=False, null=False)
```
efahy, line 15: Still think you can leave out the `blank` and `null` on these lines, since False is default

compliance/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+    created_at = serializers.DateTimeField()
+
+
+class OpenLicenceReturnsViewSerializer(OpenLicenceReturnsListSerializer):
+    returns_data = serializers.CharField()
+
+
+class OpenLicenceReturnsCreateSerializer(serializers.ModelSerializer):
+    returns_data = serializers.CharField(required=True, allow_blank=False)
+    year = serializers.IntegerField(
```
efahy, line 19: I feel like the validation should be more rigorous than just that it's a valid integer.  Validation should be that its either current or previous calendar year, right?

compliance/tests/test_add_open_licence_returns.py
```diff
@@ -0,0 +1,105 @@
...
+            "file": self._create_open_licence_returns_csv([self.licence.reference_code]),
+        }
+
+        response = self.client.post(self.url, data, **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json()["errors"], {"year": [Compliance.OpenLicenceReturns.YEAR_ERROR]})
+        self.assertFalse(OpenLicenceReturns.objects.exists())
+
+    def test_upload_licence_returns_no_file_failure(self):
```
efahy, line 56: file -> returns_data

compliance/tests/test_add_open_licence_returns.py
```diff
@@ -0,0 +1,105 @@
...
+            "year": datetime.now().year,
+        }
+
+        response = self.client.post(self.url, data, **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json()["errors"], {"file": [Compliance.OpenLicenceReturns.FILE_ERROR]})
+        self.assertFalse(OpenLicenceReturns.objects.exists())
+
+    def test_upload_licence_returns_invalid_file_format_failure(self):
```
efahy, line 67: file -> returns_data

compliance/tests/test_add_open_licence_returns.py
```diff
@@ -0,0 +1,105 @@
...
+            "year": datetime.now().year,
+        }
+
+        response = self.client.post(self.url, data, **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.json()["errors"], {"file": [Compliance.OpenLicenceReturns.INVALID_FILE_FORMAT]})
+        self.assertFalse(OpenLicenceReturns.objects.exists())
+
+    def test_upload_licence_returns_invalid_licences_failure(self):
```
efahy, line 79: If you put in logic to include the bad licence references in the error, it would be good to assert here that the correct ones are in the error

compliance/helpers.py
BenAndrew-Kainos: This was discussed in refinement and was elected against.

compliance/helpers.py
BenAndrew-Kainos: How do you mean?

licences/models.py
```diff
@@ -9,6 +9,7 @@
 
 class Licence(TimestampableModel):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    reference_code = models.CharField(max_length=30, unique=True, null=False, blank=False, editable=False)
```
efahy, line 12: Don't need `null` and `blank`

applications/views/applications.py
BenAndrew-Kainos: Done

compliance/helpers.py
BenAndrew-Kainos: Done

compliance/models.py
BenAndrew-Kainos: Done

compliance/helpers.py
efahy: I was thinking about the possibility of SQL being injected via this CSV data, but thinking more about it, this shouldn't be an issue since we never execute these strings, just store them.  However, google threw this up to me: https://owasp.org/www-community/attacks/CSV_Injection.  Maybe we should implement the recommendations here?

compliance/helpers.py
efahy: I don't recall that, but ok

compliance/tests/test_add_open_licence_returns.py
BenAndrew-Kainos: As mentioned above, we aren't including licence numbers at the moment

licences/models.py
BenAndrew-Kainos: Done

compliance/tests/test_add_open_licence_returns.py
BenAndrew-Kainos: The request data includes "file". Returns data is what we save after cleaning it. They are not the same.

compliance/tests/test_add_open_licence_returns.py
BenAndrew-Kainos: The request data includes "file". Returns data is what we save after cleaning it. They are not the same.

compliance/serializers.py
BenAndrew-Kainos: Done

compliance/helpers.py
BenAndrew-Kainos: Implemented recommendations

applications/views/applications.py
```diff
@@ -549,15 +549,15 @@ def put(self, request, pk):
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
-        data = deepcopy(request.data)
-        action = data.get("action")
+        licence_data = deepcopy(request.data)
```
andytaylorkainos, line 552: why do we have to deepcopy?

compliance/helpers.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        csv_reader = csv.reader(text.split("\n"), delimiter=",")
+        # skip headers
+        next(csv_reader, None)
+        for row in csv_reader:
+            if row:
+                if len(row) != TOTAL_COLUMNS:
+                    raise ValidationError({"file": [Compliance.OpenLicenceReturns.INVALID_FILE_FORMAT]})
+                references.add(row[0])
+                # https://owasp.org/www-community/attacks/CSV_Injection
```
andytaylorkainos, line 34: 👍

applications/views/applications.py
BenAndrew-Kainos: No clue. Will try without

efahy: This deepcopy is existing code.  i think we should leave it alone at this late stage

PR #511, opened 2020-06-10
feat: LT-2692 - exporter NLRs visible on licence page

Mark-Spiers-kainos:


licences/tests/test_get_nlrs.py
```diff
@@ -0,0 +1,53 @@
...
+        ]
+
+    def test_get_all_nlrs(self):
+        response = self.client.get(self.url, **self.exporter_headers)
+        response_data = response.json()["results"]
+        response_data.reverse()
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response_data), len(self.applications))
+        for i in range(len(self.applications)):
```
a-bek, line 42: It's nicer to read if you did:

```
for application in self.applications:
    for nlr in response_data:
        if nlr["case_id"] == str(application.id):
        ......
```

licences/serializers/view_licence.py
```diff
@@ -187,3 +189,33 @@ class Meta:
...
+                good__advice__type=AdviceType.NO_LICENCE_REQUIRED,
+                good__advice__level=AdviceLevel.FINAL,
+            ).values_list("good_id", flat=True)
+        )
+        goods = Good.objects.prefetch_related("control_list_entries").filter(id__in=goods_id)
+        return GoodLicenceListSerializer(goods, many=True).data
+
+    def get_destinations(self, instance):
+        return (
+            Party.objects.prefetch_related("parties_on_application", "country")
```
a-bek, line 216: The `prefetch_related` isn't necessary. You only need to use prefetch related if you loop through the queryset later on to access foreign members. `.annotate` and retrieving `values` combines it all into one query for you and returns the data you specify.

licences/serializers/view_licence.py
```diff
@@ -187,3 +189,33 @@ class Meta:
...
+    def get_goods(self, instance):
+        goods_id = set(
+            GoodOnApplication.objects.filter(
+                good__advice__case_id=instance.case_id,
+                good__advice__type=AdviceType.NO_LICENCE_REQUIRED,
+                good__advice__level=AdviceLevel.FINAL,
+            ).values_list("good_id", flat=True)
+        )
+        goods = Good.objects.prefetch_related("control_list_entries").filter(id__in=goods_id)
+        return GoodLicenceListSerializer(goods, many=True).data
```
a-bek, line 212: Have you checked the number of queries this makes, I think its `n+1`.

licences/serializers/view_licence.py
Mark-Spiers-kainos: I was using this with a serializer before, but got rid of serializer since it was creating stupid ammount of queries, just forgot to remove, will do so now.

licences/serializers/view_licence.py
Mark-Spiers-kainos: I haven't checked this exact function, but I know the endpoint in general is 20 queries for 3 results, and 38 queries for 10 results. From what I seen the number of goods doesn't affect it.

licences/tests/test_get_nlrs.py
Mark-Spiers-kainos: done

licences/serializers/view_licence.py
Mark-Spiers-kainos: done

licences/serializers/view_licence.py
a-bek: The query count shouldn't be growing with the number of results, that means there is an inefficiency somewhere!

PR #512, opened 2020-06-10
performance: LT-2716 - OIEL destinations

a-bek:
- Remove destinations from OpenApplicationSerializer
- Add a service level that has the function to retrieve an application's destinations
- Inject destinations into open application data objects where relevant
- Querying complexity reduced from O(1+n+m) to O(1+1+1)


cases/service.py
```diff
@@ -0,0 +1,50 @@
...
+
+from applications.models import CountryOnApplication
+from flags.enums import FlagStatuses
+
+
+def get_destinations(application_id, user_type=None):
+    """
+    Get destinations for an open application. For gov users they are ordered based on flag priority and alphabetized by name.
+    """
+    if user_type == "exporter":
```
jfaracik, line 11: `if user_type == UserType.EXPORTER`

PR #513, opened 2020-06-10
feat: LT-2723-create-compliance-case

rorysproule1:
- Introduced new model: ComplianceSiteCase
- Introduced new data on all case type enums for compliance.
- Improved performance of getting ordered flags ( 1 query per case assuming case_type is preselected)
- Added additional contacts to be on the Case as a M2M relationship, this fixes the bug that only applications can have additional contacts
- Included function generate_compliance which will generate a compliance case if needed, once the conditions are met. Can be plugged into anywhere.
- New case status Open


cases/libraries/get_flags.py
```diff
@@ -1,91 +1,90 @@
...
-    all_flags = sorted(all_flags, key=lambda x: (x.my_team, x.order, x.priority))
+def get_ordered_flags(case: Case, team: Team, limit: int = None):
+    """
+    This function will get the flags for cases looking at good, destination, case, and organisation flags. The flags
+        will be ordered with your teams flags first, in order of category (same order as above), and priority
+        (lowest first).
+
+    :param case: case object the flags relate to
+    :param team: The team for user making the request
+    :param limit: If assigned will return no more than given ammount
```
efahy, line 71: Typo "amount"

cases/libraries/get_flags.py
```diff
@@ -1,91 +1,90 @@
...
-        flag.my_team = flag.team.id != team.id
-        flag.order = order
-
-    return flags
+    return Flag.objects.none()
 
 
 def get_flags(case: Case) -> QuerySet:
     """
     Get all case flags in no particular order (order will be specified by calling function)
```
efahy, line 50: Just leave out the bit in brackets

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+from compliance.models import ComplianceSiteCase
+from goods.models import Good
+from organisations.models import Site
+from static.statuses.enums import CaseStatusEnum
+from static.statuses.libraries.get_case_status import get_case_status_by_status
+from users.enums import SystemUser
+from users.models import BaseUser
+
+
+ComplianceGoodControlCodeRegex = "(^[0-9][DE].*$)|(^ML21.*$)|(^ML22.*$)"
```
efahy, line 17: Uppercase for constant.  Also maybe add a comment explaining what it matches, with examples

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    else:
+        return False
+
+
+def get_record_holding_sites_for_case(case_id):
+    return set(
+        Site.objects.filter(sites_on_application__application_id=case_id)
+        .annotate(
+            record_site=db_case(
+                When(site_records_located_at__isnull=False, then=F("site_records_located_at")), default=F("id")
```
efahy, line 43: Where records are held for all sites is mandatory, i.e. can't be null so you can refine this a little bit I think

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+        .values_list("record_site", flat=True)
+    )
+
+
+def generate_compliance(case: Case):
+    # check if case meets conditions required
+    if not case_meets_conditions_for_compliance(case):
+        return
+
+    # Get record holding sites the case
```
efahy, line 55: Typo

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    record_holding_sites_id = get_record_holding_sites_for_case(case.id)
+
+    # Get list of record holding sites that do not relate to compliance case
```
efahy, line 58: Can you say "... that don't yet have a compliance case" instead?  I think it would be clearer

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    new_compliance_sites = set(Site.objects.filter(id__in=record_holding_sites_id, compliance__isnull=True).distinct())
```
efahy, line 59: Is the `distinct()` needed? Better to leave out if not

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # get a list compliance cases that already exist
+    existing_compliance_cases = set(
+        Case.objects.filter(compliancesitecase__site_id__in=record_holding_sites_id).distinct()
```
efahy, line 63: `distinct` needed?

compliance/views.py
```diff
@@ -0,0 +1,98 @@
...
+        cases = cases.filter(case_type__id__in=[CaseTypeEnum.OICL.id, CaseTypeEnum.OIEL.id]) | cases.filter(
+            baseapplication__goods__good__control_list_entries__rating__regex=ComplianceGoodControlCodeRegex
+        )
+
+        if reference_code:
+            cases = cases.filter(reference_code__icontains=reference_code)
+
+        return cases.filter(baseapplication__application_sites__site__compliance__id=self.kwargs["pk"],) | cases.filter(
+            baseapplication__application_sites__site__site_records_located_at__compliance__id=self.kwargs["pk"],
+        )
```
efahy, line 44: I think you only need to worry about the site_records_located_at

efahy: I think this could do with a comment - there's a fair bit of dundering going on

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+            Audit(
+                actor=system_user,
+                verb=AuditType.COMPLIANCE_SITE_CASE_NEW_LICENCE,
+                target=comp_case,
+                payload={"case_reference": case.reference_code},
+            )
+        )
+
+    if not new_compliance_sites:
+        Audit.objects.bulk_create(audits)
```
efahy, line 81: Can restructure so that we only do the audit bulk creation once at the end of the function
As you note yourself, the `if` may not be needed at all

compliance/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+        return None
+
+    # Create new compliance cases for each record holding site without a case, and audit creation
+    for site in new_compliance_sites:
+        comp_case = ComplianceSiteCase(
+            site=site,
+            status=get_case_status_by_status(CaseStatusEnum.OPEN),
+            organisation_id=case.organisation_id,
+            case_type_id=CaseTypeEnum.COMPLIANCE.id,
+            submitted_at=timezone.now(),
```
efahy, line 91: Comment explaining why you need to set this pls

compliance/tests/tests_compliance_creation.py
```diff
@@ -0,0 +1,76 @@
...
+        case = self.create_standard_application_case(self.organisation)
+
+        generate_compliance(case)
+
+        self.assertFalse(ComplianceSiteCase.objects.exists())
+
+    def tests_siel_good_control_code(self):
+        case = self.create_standard_application_case(self.organisation)
+
+        ControlListEntriesFactory(rating="ML21a")
```
efahy, line 52: Parameterise test and test all the control codes that are matched

compliance/tests/tests_compliance_creation.py
```diff
@@ -0,0 +1,76 @@
...
+    def test_multiple_cases_creates_one_compliance_case(self):
+        # Both cases uses the organisation primary site by default on creation
+        case = self.create_open_application_case(self.organisation)
+        case_2 = self.create_open_application_case(self.organisation)
+        generate_compliance(case)
+        generate_compliance(case_2)
+
+        self.assertEqual(ComplianceSiteCase.objects.count(), 1)
+
+    def test_bad_case_type(self):
```
efahy, line 71: I would like a test where an application is submitted that has a site whose records are kept at a different site (which is not on the application itself)
Also a test where there are multiple sites on an app all of whom keep their records at one record holding site
Also an application linked to two records keeping sites, one of which already has a compliance case

compliance/views.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class ComplianceManageStatus(APIView):
```
efahy, line 47: API should validate that only valid statuses are being set for compliance cases

compliance/views.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+
+    authentication_classes = (GovAuthentication,)
+
+    def get(self, request, pk, *args, **kwargs):
+        # Get record holding sites the case
+        record_holding_sites_id = get_record_holding_sites_for_case(pk)
+
+        # Get list of record holding sites that do not relate to compliance case
+
```
efahy, line 91: Remove newline

compliance/views.py
```diff
@@ -15,6 +35,86 @@
...
+        cases = Case.objects.select_related("case_type").filter(
+            baseapplication__licence__is_complete=True,
+            baseapplication__application_sites__site__site_records_located_at__compliance__id=self.kwargs["pk"],
+        )
+        cases = cases.filter(case_type__id__in=[CaseTypeEnum.OICL.id, CaseTypeEnum.OIEL.id]) | cases.filter(
+            baseapplication__goods__good__control_list_entries__rating__regex=COMPLIANCE_CASE_ACCEPTABLE_GOOD_CONTROL_CODES
+        )
+
+        if reference_code:
```
andytaylorkainos, line 55: can you .upper() the reference_code so we can drop `icontains`

PR #514, opened 2020-06-12
bug: Value/Qty length

BenAndrew-Kainos:


applications/migrations/0032_auto_20200612_1338.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("applications", "0031_auto_20200525_1120"),
+    ]
+
+    operations = [
+        migrations.RunPython(reduce_decimal_places),
```
Mark-Spiers-kainos, line 23: can you introduce a reverse migration step, even if it is just a function containing pass?

efahy: Backwards step please, like the man said

applications/migrations/0032_auto_20200612_1338.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.13 on 2020-06-12 13:38
+
+from django.db import migrations, models
+
+
+def reduce_decimal_places(apps, schema_editor):
```
efahy, line 6: Since this is a manual migration, can you please add a comment indicating what its doing?

Also, "auto" in the name is misleading - can you change it pls?

applications/migrations/0032_auto_20200612_1338.py
BenAndrew-Kainos: There is no backwards step. You reduce the resolution of the decimal places. There is no reverse

efahy: As per the original comment: "even if it is just a function containing pass"

applications/migrations/0032_auto_20200612_1338.py
BenAndrew-Kainos: Done

applications/migrations/0032_auto_20200612_1338.py
BenAndrew-Kainos: Done

PR #518, opened 2020-06-15
feat: Open licence returns upload

BenAndrew-Kainos:


licences/helpers.py
```diff
@@ -0,0 +1,11 @@
...
+from string import ascii_uppercase
+
+
+def get_reference_code(application_reference):
+    total_reference_codes = Licence.objects.filter(reference_code__icontains=application_reference).count()
+    return (
+        f"{application_reference}/{ascii_uppercase[total_reference_codes-1]}"
+        if total_reference_codes != 0
+        else application_reference
+    )
```
efahy, line 11: You need to get a database lock while doing this to avoid the possibility of two people getting the same reference code at the same time.  This was a bug we had to fix with the main application reference number function

andytaylorkainos: `cases/managers.py` has a good example

BenAndrew-Kainos: Done

licences/helpers.py
```diff
@@ -0,0 +1,17 @@
+from django.db import transaction
+
+from licences.models import Licence
+from string import ascii_uppercase
+
+
+@transaction.atomic
+def get_reference_code(application_reference):
+    # Needs to lock so that 2 Licences don't get the same reference code
+    total_reference_codes = (
```
andytaylorkainos, line 10: needs to be wrapped inside `transaction.atomic()` i think - at least thats what Mr Spiers told me

BenAndrew-Kainos: See line 7

PR #522, opened 2020-06-18
Hotfix: compliance cases and sites

Mark-Spiers-kainos:


organisations/migrations/0009_site_records_located_at.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 2.2.13 on 2020-06-12 14:21
+
+from django.db import migrations
+from django.db.models import F
+
+from compliance.helpers import generate_compliance_site_case
+
+
```
efahy, line 8: "Manual migration to ensure that all current sites have their "records held at" property set and all sites that are linked to licences have a compliance case" <- comment pls!

PR #527, opened 2020-06-22
feat: 2789 Non team queues

BenAndrew-Kainos:


cases/views/search/views.py
```diff
@@ -59,7 +61,10 @@ def get(self, request, *args, **kwargs):
         service.populate_is_recently_updated(cases)
         service.get_hmrc_sla_hours(cases)
 
-        queue = next((q for q in queues if str(q["id"]) == str(queue_id)))
+        queue = (
+            next((q for q in queues if str(q["id"]) == str(queue_id)), None)
+            or list(Queue.objects.filter(id=queue_id).annotate(case_count=Count("cases")).values())[0]
```
andytaylorkainos, line 66: list(Queue.objects.filter(id=queue_id).annotate(case_count=Count("cases")).first())

cases/views/search/views.py
```diff
@@ -59,7 +61,10 @@ def get(self, request, *args, **kwargs):
         service.populate_is_recently_updated(cases)
         service.get_hmrc_sla_hours(cases)
 
-        queue = next((q for q in queues if str(q["id"]) == str(queue_id)))
+        queue = (
```
efahy, line 64: Comment please, indicating that that the user may be on a queue that is neither a system queue or one of their team's queues, in which case we need to go to the DB again to get it

efahy: Also, should we consider an API test for this scenario?

BenAndrew-Kainos: Done

PR #530, opened 2020-06-23
feat: LT-2704 - TAU Category 3 Products

stephanieboyd:
- Add one new question to the add goods form, both from the product list and on a standard application:
  - Describe the purpose of the software/technology
    - only if the item category in Group 3 i.e. software or technology
  - validation
    - custom error message based on the item's category as above
  - tests

- Changes to ensure that the component question is not applicable for the products in group 3


goods/tests/tests_create.py
```diff
@@ -336,6 +338,77 @@ def test_add_good_information_security_details_provided_success(self):
...
+            "is_component": Component.NO,
+            "is_information_security_step": True,
+            "uses_information_security": True,
+            "information_security_details": "details about security",
+        }
+
+        response = self.client.post(URL, data, **self.exporter_headers)
+        good = response.json()["good"]
+
+        self.assertEquals(response.status_code, status.HTTP_201_CREATED)
```
jfaracik, line 362: Could this not just be a dict comparison? eg

```
self.assertEquals(good, {
   "description": data["description"],
   ....
})
```

Would be easier to read and would cut down on the number of assertions you make

stephanieboyd: Gonna push back on this :) The individual asserts allow us to see what field fails, the dict comparison hides that by giving an error like:

```AssertionError: {'id': 'a14155b1-ec89-4afa-bdcf-2ecbeae50af[752 chars]ils'} != {'description': 'coffee','is_good_controll[326 chars]ity'} Diff is 1119 characters long. Set self.maxDiff to None to see it.```

The structure of the good object and the data also differ so you'd be creating a dict with "key" keys, just to do the dict comparison

File: goods/serializers.py [hunk 2]
```diff
@@ -197,21 +199,38 @@ def update(self, instance, validated_data):
...
         uses_information_security = validated_data.get("uses_information_security")
         # if information security has changed, then set the new value and the details field
         if uses_information_security is not None and uses_information_security != instance.uses_information_security:
             instance.uses_information_security = uses_information_security
-            instance.information_security_details = validated_data.get(
-                "information_security_details", instance.information_security_details
-            )
-        instance.information_security_details = validated_data.get("information_security_details", "")
+            if uses_information_security is False:
+                instance.information_security_details = ""
```
BenAndrew-Kainos, line 219: Please add a comment

goods/serializers.py
```diff
@@ -197,21 +199,38 @@ def update(self, instance, validated_data):
...
+            else:
+                instance.information_security_details = validated_data.get(
+                    "information_security_details", instance.information_security_details
+                )
+        # If the information security details have changed
+        instance.information_security_details = validated_data.get(
+            "information_security_details", instance.information_security_details
+        )
+
+        if instance.item_category in [ItemCategory.GROUP3_TECHNOLOGY, ItemCategory.GROUP3_SOFTWARE]:
```
BenAndrew-Kainos, line 230: Do you need these if's? Can you not just set the property if it's present as validation has happened prior to this to ensure it meets requirements

goods/tests/tests_create.py
```diff
@@ -336,6 +338,77 @@ def test_add_good_information_security_details_provided_success(self):
...
+        self.assertTrue(good["uses_information_security"])
+        self.assertEquals(good["information_security_details"], data["information_security_details"])
+
+    @parameterized.expand(
+        [
+            [ItemCategory.GROUP3_SOFTWARE, "", strings.Goods.FORM_NO_SOFTWARE_DETAILS],
+            [ItemCategory.GROUP3_TECHNOLOGY, "", strings.Goods.FORM_NO_TECHNOLOGY_DETAILS],
+        ]
+    )
+    def test_add_category_three_good_failure(self, category, details, error):
```
BenAndrew-Kainos, line 378: Could you specify what is making this fail

goods/tests/tests_create.py
```diff
@@ -336,6 +338,77 @@ def test_add_good_information_security_details_provided_success(self):
...
+
+        response = self.client.post(URL, data, **self.exporter_headers)
+        errors = response.json()["errors"]
+
+        self.assertEquals(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(
+            errors["software_or_technology_details"], ["Ensure this field has no more than 2000 characters."]
+        )
+
```
BenAndrew-Kainos, line 411: Could you add a test to validate is_military_use is given if GROUP2_FIREARMS

goods/tests/tests_edit.py
```diff
@@ -231,3 +235,76 @@ def test_edit_information_security_no_selection_failure(self):
...
+        # 2 due to creating a new good for this test
+        self.assertEquals(Good.objects.all().count(), 2)
+
+    @parameterized.expand(
+        [
+            [ItemCategory.GROUP3_SOFTWARE, strings.Goods.FORM_NO_SOFTWARE_DETAILS],
+            [ItemCategory.GROUP3_TECHNOLOGY, strings.Goods.FORM_NO_TECHNOLOGY_DETAILS],
+        ]
+    )
+    def test_edit_software_or_technology_details_success(self, category, error):
```
BenAndrew-Kainos, line 266: This test has the same name as the one above (I believe this one should say failure)

goods/tests/tests_edit.py
```diff
@@ -231,3 +235,76 @@ def test_edit_information_security_no_selection_failure(self):
...
+        request_data = {
+            "is_component_step": True,
+            "is_component": Component.YES_GENERAL_PURPOSE,
+            "general_details": "some details",
+        }
+
+        response = self.client.put(url, request_data, **self.exporter_headers)
+        good = response.json()["good"]
+
+        self.assertEquals(response.status_code, status.HTTP_200_OK)
```
BenAndrew-Kainos, line 297: Why do these 'cannot edit' tests expect a 200? Surely if it's not allowed we should have a 400

goods/tests/tests_edit.py
```diff
@@ -231,3 +235,76 @@ def test_edit_information_security_no_selection_failure(self):
...
+    def test_cannot_edit_software_technology_details_non_category_three_good_success(self):
+        good = self.create_good("a good", self.organisation, item_category=ItemCategory.GROUP1_PLATFORM)
+        url = reverse("goods:good_details", kwargs={"pk": str(good.id)})
+        request_data = {"software_or_technology_details": "some details"}
+
+        response = self.client.put(url, request_data, **self.exporter_headers)
+        good = response.json()["good"]
+
+        self.assertEquals(response.status_code, status.HTTP_200_OK)
+        self.assertIsNone(good["software_or_technology_details"])
```
BenAndrew-Kainos, line 310: Seem to be missing a test for edit is_military_use on FIREARMS

goods/views.py
```diff
@@ -187,17 +192,18 @@ def post(self, request, *args, **kwargs):
...
                 return JsonResponse(
-                    data={"errors": {"item_category": ["Not implemented yet, please select an option in category 1"]}},
+                    data={
+                        "errors": {"item_category": ["Not implemented yet, please select an option in category 1 or 3"]}
+                    },
                     status=status.HTTP_400_BAD_REQUEST,
                 )
-
         serializer = GoodCreateSerializer(data=data)
+        validate_software_or_technology_details(data)
```
BenAndrew-Kainos, line 203: I like how you've put the different forms of validation into different functions, but could you make this part of the serializer

goods/views.py
```diff
@@ -263,6 +269,20 @@ def put(self, request, pk):
 
         data = request.data.copy()
 
+        if "software_or_technology_details" in data and not data.get("software_or_technology_details"):
+            raise ValidationError(
+                {
+                    "software_or_technology_details": [
```
BenAndrew-Kainos, line 275: This seems duplicated (same code in helpers)

goods/views.py
iichr: This was attempted, and it does work for the majority of cases - update and create. _However_ if the user were to click the back link on the next question and come back to this one to change their answer, we lose all validation 🙁 Reason being `validate()` on the serialiser is not called in that case at all, so we are forced to do it outside the serialiser

goods/views.py
iichr: Good spot, replaced with the function in helpers 👍

goods/tests/tests_edit.py
iichr: We've opted for catching the mistake silently, with the invalid input not persisted. This only happens if the user manually goes to that url on a potentially unsupported good

iichr: Update: Now changed, we throw a bad request error (400) upon submission

goods/serializers.py
iichr: Added 👍

goods/tests/tests_create.py
stephanieboyd: Group 2 is not yet implemented, that's the next story :)

goods/tests/tests_edit.py
stephanieboyd: This is also group 2, and will come in the next story

goods/tests/tests_create.py
stephanieboyd: Added to function name

goods/tests/tests_edit.py
stephanieboyd: Yup, thanks, good spot

goods/serializers.py
stephanieboyd: Removed

goods/tests/tests_create.py
jfaracik: That's fine with me :)

goods/tests/tests_create.py
```diff
@@ -336,6 +338,77 @@ def test_add_good_information_security_details_provided_success(self):
...
+        self.assertTrue(good["uses_information_security"])
+        self.assertEquals(good["information_security_details"], data["information_security_details"])
+
+    @parameterized.expand(
+        [
+            [ItemCategory.GROUP3_SOFTWARE, "", strings.Goods.FORM_NO_SOFTWARE_DETAILS],
+            [ItemCategory.GROUP3_TECHNOLOGY, "", strings.Goods.FORM_NO_TECHNOLOGY_DETAILS],
+        ]
+    )
+    def test_add_category_three_good_no_detailsfailure(self, category, details, error):
```
BenAndrew-Kainos, line 378: missing an underscore

goods/tests/tests_create.py
BenAndrew-Kainos: Why is there validation for group 2 on lines 74-80 in helpers?

goods/tests/tests_edit.py
BenAndrew-Kainos: Same as above. If you add the code we need the test

goods/views.py
BenAndrew-Kainos: I don't follow why validate wouldn't be called. Surely it's called every time the serializer is used

iichr: Nope.

goods/tests/tests_edit.py
iichr: Removed.

goods/tests/tests_create.py
iichr: Removed.

goods/views.py
iichr: As detailed above, multiple breakpoints in the serialiser, it just doesn't go inside `validate()` after clicking the back link, we believe a likely reason is the field not being required on the serialiser.

PR #531, opened 2020-06-23
feat: LT-1409 - OGL sites

jfaracik:
* Increased amount of information returned when getting a site's records

### DIFF
--- a/organisations/serializers.py
+++ b/organisations/serializers.py
@@ -27,13 +27,26 @@ class SiteListSerializer(serializers.Serializer):
     id = serializers.UUIDField()
     name = serializers.CharField()
     address = AddressSerializer()
-    site_records_located_at_name = serializers.SerializerMethodField()
+    records_located_at = serializers.SerializerMethodField()
 
-    def get_site_records_located_at_name(self, instance):
+    def get_records_located_at(self, instance):
         if instance.site_records_located_at:
-            site = Site.objects.filter(id=instance.site_records_located_at.id).values_list("name", flat=True).first()
+            site = Site.objects.filter(id=instance.site_records_located_at.id).first()
             if site:
-                return site
+                return {
+                    "id": site.id,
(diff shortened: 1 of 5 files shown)


organisations/serializers.py
```diff
@@ -27,13 +27,26 @@ class SiteListSerializer(serializers.Serializer):
...
     name = serializers.CharField()
     address = AddressSerializer()
-    site_records_located_at_name = serializers.SerializerMethodField()
+    records_located_at = serializers.SerializerMethodField()
 
-    def get_site_records_located_at_name(self, instance):
+    def get_records_located_at(self, instance):
         if instance.site_records_located_at:
-            site = Site.objects.filter(id=instance.site_records_located_at.id).values_list("name", flat=True).first()
+            site = Site.objects.filter(id=instance.site_records_located_at.id).first()
```
Mark-Spiers-kainos, line 34: I believe there can only be one, so can use a "get" here instead of filter and first. Also all sites require a site records located at, so should not fall over for having none.

jfaracik: Perfect if that's the case :)

organisations/tests/tests_sites.py
```diff
@@ -89,12 +89,27 @@ def test_add_uk_site(self):
...
         response = self.client.post(url, data, **self.gov_headers)
-        site = response.json()["site"]
+        response_data = response.json()["site"]
 
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         self.assertEqual(Site.objects.filter(organisation=self.organisation).count(), 2)
-        # Assert that 'site_records_located_at_name' is set to the site being created
-        self.assertEqual(site["site_records_located_at_name"], data["name"])
+        site = Site.objects.get(name=data["name"])
+        # Assert that "records_located_at" is set to the site being created
```
iichr, line 97: 👍

PR #532, opened 2020-06-23
feat: 1585 Document signing

BenAndrew-Kainos:
Screenshot 2020-06-23 at 16 29 19


cases/generated_documents/signing.py
```diff
@@ -0,0 +1,106 @@
...
+    BASE_DIR,
+    CERTIFICATE_PATH,
+    CERTIFICATE_PASSWORD,
+    SIGNING_REASON,
+    SIGNING_LOCATION,
+    SIGNING_EMAIL,
+)
+
+SIGNATURE_TITLE = "Digital Signature"
+FONT = os.path.join(BASE_DIR, "assets", "fonts", "Helvetica.ttf")
```
jfaracik, line 21: Are these documents going to be visible to exporters? If so could we use the GDS Transport font?

BenAndrew-Kainos: Sure

BenAndrew-Kainos: Screenshot 2020-06-25 at 14 10 42

.circleci/config.yml
```diff
@@ -91,6 +95,10 @@ jobs:
       - restore_cache:
           key: dependencies-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
 
+      - run:
+          name: Install OS libraries
+          command: sudo apt-get install swig;
```
Mark-Spiers-kainos, line 100: don't think I have seen a semicolon in a command before

PR #533, opened 2020-06-24
hotfix: Compliance seeding fix

BenAndrew-Kainos:


organisations/migrations/0009_site_records_located_at.py
```diff
@@ -19,6 +25,18 @@ def forward_migration(apps, schema_editor):
         baseapplication__licence__is_complete=True,
     ).distinct()
 
+    # Get or create case type & status because seeding may not have run yet
+    case_type, _ = CaseType.objects.get_or_create(
```
Mark-Spiers-kainos, line 29: I think I would prefer if you checked if the id exists, and if not reran seeding for case_type/ case_status.

PR #535, opened 2020-06-25
feat: LT-2632 - I want to see which OGLs I have already registered for when searching for new ones

jfaracik:
* Enhancement to return registrations when registering for an OGL


organisations/models.py
```diff
@@ -64,23 +64,26 @@ def register_open_general_licence(self, open_general_licence, user):
...
         # Only register open general licences for sites in the UK which don't already have that licence registered
+        registrations = []
         with transaction.atomic():
             for site in (
                 Site.objects.get_by_user_and_organisation(user, self)
                 .filter(address__country_id="GB")
                 .exclude(open_general_licence_cases__open_general_licence=open_general_licence)
             ):
-                OpenGeneralLicenceCase(
+                case = OpenGeneralLicenceCase(
```
BenAndrew-Kainos, line 74: You could just do `OpenGeneralLicenceCase.objects.create` and then you wont need a `.save()` after

PR #538, opened 2020-06-26
bug: Contact field length

BenAndrew-Kainos:


parties/models.py
```diff
@@ -33,8 +33,8 @@ def copy_detail(self, pk):
 
 class Party(TimestampableModel):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    name = models.TextField(default=None, blank=True)
-    address = models.TextField(default=None, blank=True)
+    name = models.CharField(default=None, blank=True, max_length=100)
+    address = models.TextField(default=None, blank=True, max_length=256)
```
Mark-Spiers-kainos, line 37: if this was 255 it could have been a charfield. If it's possible to remove 1 more character I would recommend it since the database doesn't enforce max length for text fields, only the serializers.

PR #539, opened 2020-06-26
feat: LT-849 - Compliance cases for open general licence registrations 

jfaracik:
* Generates a compliance case upon OGL registration


compliance/tests/tests_compliance_creation.py
```diff
@@ -108,10 +112,22 @@ def test_multiple_sites_one_already_contains_compliance_case(self):
...
+        open_general_licence = OpenGeneralLicenceFactory(case_type=CaseType.objects.get(id=CaseTypeEnum.OGTCL.id))
+        case = OpenGeneralLicenceCaseFactory(
+            open_general_licence=open_general_licence,
+            site=self.organisation.primary_site,
+            organisation=self.organisation,
+        )
+
+        generate_compliance_site_case(case)
+
+        self.assertTrue(ComplianceSiteCase.objects.exists())
```
BenAndrew-Kainos, line 133: Could you check a few properties of the ComplianceSiteCase (i.e. `ComplianceSiteCase.objects.filter(site=self.organisation.primary_site).exists()`)

compliance/helpers.py
```diff
@@ -50,16 +51,21 @@ def case_meets_conditions_for_compliance(case: Case):
...
-    return set(
-        Site.objects.filter(sites_on_application__application_id=case_id).values_list(
-            "site_records_located_at_id", flat=True
+def get_record_holding_sites_for_case(case):
+    if case.case_type.id in CaseTypeEnum.OGL_ID_LIST:
+        return {case.site.site_records_located_at.id}
+    else:
+        return set(
+            Site.objects.filter(
+                Q(sites_on_application__application_id=case.id) | Q(open_general_licence_cases__in=[case.id])
```
Mark-Spiers-kainos, line 66: Does this need a Q if you have an if else?

compliance/tests/tests_compliance_creation.py
jfaracik: Replaced it with `self.assertEqual(compliance_case.site.id, case.site)`, is that all?

compliance/helpers.py
jfaracik: I've changed it back, not too sure why I changed it in the first place 👍

PR #544, opened 2020-07-02
chore: seeding tool

mh-hmcts:


static/management/commands/seeddata.py
```diff
@@ -169,6 +196,34 @@ def action(self, options):
...
+        org_eua_data = [
+            org["organisation_id"]
+            for org in EndUserAdvisoryQuery.objects.order_by().values("organisation_id").distinct()
+        ]
+        orgs = [org for org in organisations if org.id not in org_eua_data]
+
+        jobs = [(self.add_end_user_advisory, organisation) for organisation in orgs]
+        results = [application for application in self.get_mapper(mt)(ActionBase.do_work, jobs)]
+
+        print(f"Added an end user advisory for {org_count} organisations" f", added {len(results)} EUA's in total")
```
Mark-Spiers-kainos, line 218: could just be 1 f string, rather than attaching two together

mh-hmcts: thanks

PR #545, opened 2020-07-02
Feat: Lt 2251. - Add category 2 goods

stephanieboyd:


goods/serializers.py
```diff
@@ -60,6 +76,126 @@ def validate(self, data):
...
+            "no_identification_markings_details",
+        )
+
+    def validate(self, data):
+        validated_data = super(FirearmDetailsSerializer, self).validate(data)
+
+        # Year of manufacture should be in the past and a valid year
+        year_of_manufacture = validated_data.get("year_of_manufacture")
+        if year_of_manufacture:
+            if year_of_manufacture >= timezone.now().date().year:
```
Mark-Spiers-kainos, line 123: Could it be manufactured this year? For example if it's december surely it could of been manufactured in July. Would still be in the past, but this does not allow for that.

stephanieboyd: Yeah, double checked with Shabbir and this has been changed

PR #547, opened 2020-07-06
Feat: LT-865 - exporter view compliance

Mark-Spiers-kainos:
Added new exporter views for getting compliance site case, and compliance visit case (and lists views of both).

reduced queries on getting ecju_queries to a flat 5 queries, which was scaling a lot before.
reduced queries on getting generated_documents to 5 from (6 + 1 for each object for notification deletes).
reduced queries on getting case_notes, it is not flat yet, and still grows in number of case_notes.

notifications delete is 1 query now, rather than 1 per object given.

created new function to get the notifications for a compliance case and all it's visit cases.
updated the hub notification endpoint to include compliance (both site and visit cases).


cases/libraries/delete_notifications.py
```diff
@@ -2,10 +2,10 @@
 
 
 def delete_exporter_notifications(user: ExporterUser, organisation_id, objects: list):
-    for obj in objects:
-        ExporterNotification.objects.filter(user=user, organisation_id=organisation_id, object_id=obj.id).delete()
+    id_list = [obj.id for obj in objects]
+    ExporterNotification.objects.filter(user=user, organisation_id=organisation_id, object_id__in=id_list).delete()
```
Mark-Spiers-kainos, line 6: this reduced queries for deleting notifications to 1 flat.

cases/serializers.py
```diff
@@ -413,6 +408,10 @@ class Meta:
             "responded_at",
         )
 
+    def get_responded_by_user(self, instance):
```
Mark-Spiers-kainos, line 411: we only require the users name and id, on the front-end. This reduced queries significantly, since it is not trying to get a load of other details on other tables.

users/views.py
```diff
@@ -187,18 +187,46 @@ def get(self, request):
         """
         Count the number of application, eua_query and goods_query exporter user notifications
         """
-        notifications = (
-            self.queryset.filter(user=request.user, organisation_id=get_request_user_organisation_id(request))
-            .prefetch_related("case__case_type")
-            .values("case__case_type__sub_type", "case__case_type__type")
+        organisation = get_request_user_organisation(request)
+        notifications_list = list(
```
andytaylorkainos, line 191: as discussed, theres an open task to fix this from being done in memory - ill take a look at updating the code within your PR to address this whilst we both still have a grasp on how compliance notifications work.
if this becomes a blocker, ill pull it out into a separate PR

users/libraries/notifications.py
```diff
@@ -44,3 +85,15 @@ def get_exporter_user_notification_individual_count(exporter_user: ExporterUser,
...
     return {content_type["content_type__model"]: content_type["count"] for content_type in queryset}
+
+
+def get_exporter_user_notification_individual_count_with_compliance_visit(
+    exporter_user: ExporterUser, organisation_id, case: Case
+) -> dict:
+    # Group by content_type (casenote, ecjuquery, generatedcasedocument, visitreport).
+    # Get the total number of notifications for each type
+    data = get_exporter_user_notification_individual_count(exporter_user, organisation_id, case)
+    data["visitreport"] = ExporterNotification.objects.filter(
```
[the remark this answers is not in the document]
Mark-Spiers-kainos: I kept it consistent with the others which were also lower case single word

users/views.py
```diff
@@ -187,18 +187,46 @@ def get(self, request):
...
-        case_types = [notification["case__case_type__type"] for notification in notifications]
-        case_sub_types = [notification["case__case_type__sub_type"] for notification in notifications]
+        case_types = [notification["case__case_type__type"] for notification in notifications_list]
+        case_sub_types = [notification["case__case_type__sub_type"] for notification in notifications_list]
         notifications = {
             CaseTypeTypeEnum.APPLICATION: case_types.count(CaseTypeTypeEnum.APPLICATION),
             CaseTypeSubTypeEnum.EUA: case_sub_types.count(CaseTypeSubTypeEnum.EUA),
             CaseTypeSubTypeEnum.GOODS: case_sub_types.count(CaseTypeSubTypeEnum.GOODS),
         }
+        if check_user_has_permission(self.request.user, ExporterPermissions.ADMINISTER_SITES, organisation):
```
andytaylorkainos, line 208: Can you re-write the code to match below?

```
            notifications[CaseTypeTypeEnum.COMPLIANCE] = len(
                [
                    notification["case__compliancesitecase__site_id"]
                    for notification in notifications_list
                    if <FUNCTION THAT RETURNS IF USER HAS PERMISSION AND OTHER CONDITIONS ARE MET>
                ]
            )
```

Mark-Spiers-kainos: I don't want to get the sites list if I don't have to, it's an additional query we can cut out by check if the user has permission beforehand.

andytaylorkainos: but you dont check beforehand. you query the sites on lines 196 and 197

PR #548, opened 2020-07-06
pre-compile script

r4vi:


bin/pre_compile
```diff
@@ -0,0 +1,7 @@
+echo "-----> Exporting SWIG path"
+export SWIG_LIB=/home/vcap/deps/0/apt/usr/share/swig3.0
+
+# permanently make the environment variable available in the runtime container
```
BenAndrew-Kainos, line 4: Will it be permanent? I thought all of the variables were reset when built through Jenkins

r4vi: this will be loaded by the .profile when the container starts up, as long as jenkins isn't unsetting variables then it should stick

bin/pre_compile
```diff
@@ -0,0 +1,7 @@
...
+PROFILE_PATH="$HOME/profile.d/swig.sh"
```
efahy, line 5: Does `profile.d` not need to be in `/etc`?

bin/pre_compile
lgarvey: The problem here is that env vars aren't available at the point that the python buildpack does a `pip install -r requirements.txt` - setting it in pre_compile, .profile and manifest.yml did not work

bin/pre_compile
r4vi: that's the standard place but I don't think you can write to /etc/, I think the version of debian cloud foundry uses also looks in $HOME

bin/pre_compile
r4vi: ah so you've already tried `pre_compile` that's strange as I think we had the exact same issue on `staff-sso` with xmlsec1 and we did something similar there: https://github.com/uktrade/staff-sso/blob/master/bin/pre_compile - some of those needed to be set before `pip install -r`

bin/pre_compile
efahy: ok

bin/pre_compile
lgarvey: Ahh xmlsec is slightly different - as the binary is only needed at runtime.  

Anything set in .profile / profile.d/some-custom-script.js should be available at runtime. Just not at the point that the buildpack installs deps.  

Frustratingly, if we remove endesive & pyk11cs from Pipfile.lock, and set the path using any method, then ssh into the running app and do a `pip3 install endesive` it installs without any errors.  But consistently fails using when installed via the buildpack.

r4vi: we could make a binary wheel and push it to s3 and pull it back down similar to how we handle `xmlsec` that would remove the need for swig all together - but for now let's just go the feature flag route and I can look at it at some point in the future

lgarvey: This sounds like a solution.  The paas team also have an open ticket to see if they have a solution, and I've raised an issue on the python-buildpack repo.

PR #549, opened 2020-07-07
feat: LT-1446-outstanding-annual-return

rorysproule1:


compliance/tests/factories.py
```diff
@@ -12,7 +12,6 @@
 
 class OpenLicenceReturnsFactory(factory.django.DjangoModelFactory):
     returns_data = "\na,b,c,d,e"
-    year = datetime.now().year
```
Mark-Spiers-kainos, line 15: Is there a reason this has been removed?

rorysproule1: Unaware of that, you're right.. i'll revert it :)

PR #550, opened 2020-07-07
feat: LT-1188 - I want to have the option to add a note whenever I progress a case

jfaracik:
* Added optional notes for various case progression mechanisms 
* Updated filtering for audit entries by case note to include audit entries with notes
* Removed compliance status endpoint, now using the standard case patch endpoint 
* Tests


cases/helpers.py
```diff
@@ -51,3 +53,25 @@ def remove_next_review_date(case, request, pk):
...
+    reference_type = case.case_type.reference
+
+    if reference_type == CaseTypeReferenceEnum.COMP_SITE and status not in CaseStatusEnum.compliance_site_statuses:
+        return False
+    elif reference_type == CaseTypeReferenceEnum.COMP_VISIT and status not in CaseStatusEnum.compliance_visit_statuses:
+        return False
+
+    if case.case_type.reference == CaseTypeReferenceEnum.COMP_VISIT and CaseStatusEnum.is_terminal(status):
+        comp_case = ComplianceVisitCase.objects.get(id=case.id)
+        if not compliance_visit_case_complete(comp_case):
```
andytaylorkainos, line 74: can you turn this into `if not ComplianceVisitCase.objects.filter(id=case.id).exists()` please

PR #555, opened 2020-07-09
Add staff SSO for django admin login + use django.conf.settings

richtier:
# access granted
image

# access denied
image

# it's alive
admin

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -41,6 +41,7 @@ notifications-python-client = "~=5.6.0"
 dataclasses = "*"
 sentry-sdk = "~=0.16" 
 elastic-apm = "~=5.8"
+django-staff-sso-client = "~=1.0.0"
 
 [requires]
 python_version = "3.7"
--- a/users/models.py
+++ b/users/models.py
@@ -1,6 +1,7 @@
 import uuid
 from abc import abstractmethod
 
+from django.conf import settings
 from django.contrib.auth.base_user import BaseUserManager
 from django.contrib.auth.models import AbstractUser
 from django.contrib.contenttypes.fields import GenericForeignKey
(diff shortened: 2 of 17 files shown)


users/models.py
```diff
@@ -86,10 +86,10 @@ class BaseUser(AbstractUser, TimestampableModel):
     username = None
     email = models.EmailField(default=None, blank=True)
     password = None
-    is_superuser = None
+    is_superuser = True
     last_login = None
-    is_staff = None
-    is_active = None
+    is_staff = True
+    is_active = True
```
richtier, line 92: only users that are staff and superuser can access django admin

impact of this is all GovUsers will be able to access admin.

given this will only be turned on in local/dev/staging I think this is ok

users/models.py
```diff
@@ -132,6 +132,10 @@ class GovNotification(BaseNotification):
 
 
 class ExporterUser(BaseUser):
+    is_superuser = None
```
r4vi, line 135: slightly scared of the defaults being `True` in the base class, I'd prefer the `@property` implementation but it's your call

richtier: will do property sure :+1:

Pipfile
```diff
@@ -41,6 +41,7 @@ notifications-python-client = "~=5.6.0"
 dataclasses = "*"
 sentry-sdk = "~=0.16" 
 elastic-apm = "~=5.8"
+django-staff-sso-client = "~=1.0.0"
```
r4vi, line 44: maybe this can go in devPackages so it'll be impossible to deploy in production? or do you like the idea of having it available in non-prod deployed envs too?

users/models.py
```diff
@@ -86,12 +87,17 @@ class BaseUser(AbstractUser, TimestampableModel):
...
+    def has_django_admin_permission(self):
+        return self.email in settings.ALLOWED_ADMIN_EMAILS
+
+    is_superuser = has_django_admin_permission
+    is_staff = has_django_admin_permission
+    is_active = True
+
     # Set this to use id as email cannot be unique in the base user model
     # (and we couldn't think of anything else to use instead)
     USERNAME_FIELD = "id"
```
r4vi, line 103: I've done this before too!

Pipfile
richtier: I like the idea of having it on non-prod envs too :+1: 

makes debugging easier

PR #560, opened 2020-07-10
bug: LT-2753-set-hmrc-closed

rorysproule1:


applications/views/applications.py
```diff
@@ -409,6 +410,9 @@ def put(self, request, pk):
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
+        if not can_set_status(application, data["status"]):
```
jfaracik, line 413: I'd raise a ValidationError here instead of returning a JsonResponse :)

rorysproule1: Added! :)

PR #561, opened 2020-07-10
Hotfix: background tasks

BenAndrew-Kainos:
Fix for background task collection as suggested in https://github.com/arteria/django-background-tasks/issues/130


cases/__init__.py
```diff
@@ -0,0 +1 @@
+default_app_config = "cases.apps.CasesConfig"
```
andytaylorkainos, line 1: ❤️

licences/__init__.py
```diff
@@ -0,0 +1 @@
+default_app_config = "licences.apps.LicencesConfig"
```
andytaylorkainos, line 1: ❤️

PR #565, opened 2020-07-13
feat: LT-1917 and LT-2691 - Surrendering and Re-issuing Licences

andytaylorkainos:


licences/apps.py
```diff
@@ -23,7 +23,9 @@ def schedule_not_sent_licences():
         )
 
         for licence in licences_not_sent:
-            schedule_licence_for_hmrc_integration(str(licence.id), licence.application.reference_code)
+            schedule_licence_for_hmrc_integration(
+                str(licence.id), LicenceStatus.lite_to_hmrc_intergration.get(licence.status)
```
efahy, line 27: Spelling mistake in `lite_to_hmrc_intergration`

licences/models.py
```diff
@@ -68,7 +66,7 @@ def save(self, *args, **kwargs):
     def send_to_hmrc_integration(self):
         from licences.tasks import schedule_licence_for_hmrc_integration
 
-        schedule_licence_for_hmrc_integration(str(self.id), self.reference_code)
+        schedule_licence_for_hmrc_integration(str(self.id), LicenceStatus.lite_to_hmrc_intergration.get(self.status))
```
efahy, line 69: I notice in the save method above that we send a licence to HMRC every time its saved as long as its not in draft status.  Is that right?  What if we save a licence object to update its usage?  It doesn't seem to me that should trigger an update to HMRC

efahy: Oh usage is on licence_on_application, which means we're probably okay putting this on the save method?

licences/apps.py
andytaylorkainos: Thanks! rejigged this slightly to `LicenceStatus.hmrc_intergration_action.get(...)`

licences/models.py
andytaylorkainos: exactly :)

We only ever send the licence to HMRC if it's cancelled/surrendered/issued/reissued

PR #568, opened 2020-07-15
bug: 2677, 2758, 2792, 2694

BenAndrew-Kainos:
Various validation & document processing bugs

- 2677: Failed AV scanned documents stuck on "processing" & cannot re-upload if first document is detected as a virus
- 2758: No validation on giving advice and then you get API error
- 2792: API error on phone number if you enter a string on additional contacts
- 2694: Case error due to audit serialization


applications/tests/tests_standard_application_submit.py
```diff
@@ -188,6 +210,28 @@ def test_status_code_post_with_infected_document_failure(self):
...
+            response.json()["errors"],
+            {"supporting-documents": [strings.Applications.Standard.ADDITIONAL_DOCUMENTS_PROCESSING]},
+        )
+
+    def test_submit_standard_application_with_infected_additional_documents_failure(self):
+        self.create_application_document(self.draft, safe=False)
+
+        response = self.client.put(self.url, **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
rorysproule1, line 229: for all these tests, are we able to assert if the application document hasn't been added to the draft due to it failing?

applications/creators.py
```diff
@@ -29,6 +32,18 @@ def _validate_locations(application, errors):
...
+    document_statuses = documents.values_list("safe", flat=True)
+
+    # If safe is None, then the document hasn't been virus scanned yet
+    if not all([safe is not None for safe in document_statuses]):
+        return processing_error
+
+    # If safe is False, the file contains a virus
+    if not all(document_statuses):
+        return virus_error
+
```
andytaylorkainos, line 45: what if the background task fails? (never gets scanned)

applications/tests/tests_standard_application_submit.py
BenAndrew-Kainos: The document is added to the application, but it won't allow the user to submit until they either wait for it to process, or remove it because it has a virus

PR #569, opened 2020-07-15
feat: LT-2360 - Exhaust Licences (HMRC Integration)

andytaylorkainos:


licences/libraries/hmrc_integration_operations.py
```diff
@@ -134,14 +135,49 @@ def _validate_good_on_licence(licence_id: UUID, data: dict) -> dict:
...
+    change_status = None
+
+    if action == HMRCIntegrationActionEnum.EXHAUST:
+        change_status = licence.exhaust
+    elif action == HMRCIntegrationActionEnum.CANCEL:
+        change_status = licence.cancel
+    elif action == HMRCIntegrationActionEnum.SURRENDER:
+        change_status = licence.surrender
+    elif action == HMRCIntegrationActionEnum.EXPIRE:
+        change_status = licence.expire
```
BenAndrew-Kainos, line 153: Do you need to add a catch state for it it's not one of these 4 (i.e. NotImplementedError)

licences/libraries/hmrc_integration_operations.py
```diff
@@ -134,14 +135,49 @@ def _validate_good_on_licence(licence_id: UUID, data: dict) -> dict:
...
+
+    if gols and data["action"] != HMRCIntegrationActionEnum.EXHAUST:
+        exhausted_goods_count = 0
+        for gol in gols:
+            if gol["status"] == HMRCIntegrationActionEnum.EXHAUST:
+                exhausted_goods_count += 1
+
+        # If all Goods have been Exhausted; Exhaust the Licence
+        if exhausted_goods_count >= licence.goods.count():
```
BenAndrew-Kainos, line 162: You could make this logic simpler by changing to something like;

```
all_goods_exhausted = all([gol["status"] == HMRCIntegrationActionEnum.EXHAUST for gol in gols])
if all_goods_exhausted:
  action = HMRCIntegrationActionEnum.EXHAUST
  change_status = licence.exhaust
```

licences/tasks.py
```diff
@@ -1,50 +1,47 @@
...
-def schedule_max_tried_task_as_new_task(licence_id, action):
-    """
-    Used to schedule a max-tried task as a new task (starting from attempts=0)
-    This function was abstracted from 'send_licence_to_hmrc_integration' to enable unit testing of a recursive operation
-    """
-    logging.warning(
-        f"Maximum attempts of {MAX_ATTEMPTS} for licence '{licence_id}', action '{action}' has been reached"
-    )
+            # transaction.atomic + select_for_update + nowait=True will throw an error if row has already been locked
+            Licence.objects.filter(id__in=expired_licence_ids).select_for_update(nowait=True).update(
```
BenAndrew-Kainos, line 32: Why have Licences been queried twice? Also I don't think you need nowait here. Can you change to;

```
Licence.objects.filter(
                status__in=LicenceStatus.open_statuses, end_date__lt=timezone.now()
).select_for_update().update(status=LicenceStatus.EXPIRED)
```

licences/tasks.py
```diff
@@ -1,50 +1,47 @@
...
+                status=LicenceStatus.EXPIRED
+            )
 
-    schedule_datetime = timezone.now() + timedelta(seconds=TASK_BACK_OFF)
-    logging.info(
-        f"Scheduling new task for licence '{licence_id}', action '{action}' to commence at {schedule_datetime}"
-    )
-    send_licence_to_hmrc_integration(licence_id, action, schedule=TASK_BACK_OFF)  # noqa
```
BenAndrew-Kainos, line 44: Why has the `schedule=TASK_BACK_OFF` been removed?

licences/tasks.py
```diff
@@ -70,13 +67,43 @@ def send_licence_to_hmrc_integration(licence_id, action, scheduled_as_background
         )
 
 
+def schedule_licence_for_hmrc_integration(licence_id, action):
+    if BACKGROUND_TASK_ENABLED:
+        logging.info(f"Scheduling licence '{licence_id}', action '{action}' for HMRC Integration")
+        task = Task.objects.filter(queue=HMRC_INTEGRATION_QUEUE, task_params=f'[["{licence_id}", "{action}"], {{}}]')
```
BenAndrew-Kainos, line 73: Could you make the task_params bit of the filter more flexible so that is just checks the licence_id & action is in the task params. Otherwise this will break if the params format changes

licences/libraries/hmrc_integration_operations.py
andytaylorkainos: Already caught by the serializer
```
class HMRCIntegrationUsageUpdateLicenceSerializer(serializers.Serializer):
    id = serializers.UUIDField(required=True, allow_null=False)
    action = serializers.CharField(required=True, allow_null=False)
    goods = serializers.ListField(required=True, allow_null=False, allow_empty=True)

    def validate_action(self, value):
        valid_actions = HMRCIntegrationActionEnum.from_hmrc

        if value not in valid_actions:
            raise serializers.ValidationError([f"Must be one of {valid_actions}"])

        return value
```

licences/tasks.py
andytaylorkainos: Needs to be queried twice so we can collect the licence IDs (an update operation doesnt return the ids of the rows created sadly)

licences/tasks.py
andytaylorkainos: It hasnt, the code has just been moved

licences/tasks.py
andytaylorkainos: Thats the idea of this test -> task params will be `exactly` this which is a more efficient query.
If the params change, the test needs to update to match

licences/libraries/hmrc_integration_operations.py
andytaylorkainos: Only partially catches everything. 
Every good that HMRC has sent us may have been exhausted, but we need to ensure that every good on the licence is exhausted (not just the ones HMRC have sent)

licences/tasks.py
andytaylorkainos: Ah. You mean if the library ever changes its structure?
I'll write a test to ensure that using this filter will return a task from the DB

licences/libraries/hmrc_integration_operations.py
BenAndrew-Kainos: OK, could you just leave a comment mentioning that

andytaylorkainos: I renamed the data object to "validated_data" :

```
action = validated_data["action"]
```

licences/libraries/hmrc_integration_operations.py
```diff
@@ -134,11 +137,43 @@ def _validate_good_on_licence(licence_id: UUID, data: dict) -> dict:
...
+        audit_trail_service.create_system_user_audit(
+            verb=AuditType.LICENCE_UPDATED_STATUS,
+            target=licence.application.get_case(),
+            payload={
+                "licence": licence.reference_code,
+                "status": hmrc_integration_action_to_licence_status.get(action),
+            },
+        )
+
+    return str(licence.id)
```
jfaracik, line 176: Why return a string instead of the UUID?

andytaylorkainos: To be honest, just personal preference. I am happy to return the UUID instead if you'd prefer

licences/libraries/hmrc_integration_operations.py
BenAndrew-Kainos: Ok just extend the code like so;

```
all_goods_exhausted = all([gol["status"] == HMRCIntegrationActionEnum.EXHAUST for gol in gols])
if all_goods_exhausted and len(gols) >= licence.goods.count():
  action = HMRCIntegrationActionEnum.EXHAUST
  change_status = licence.exhaust
```

andytaylorkainos: Can i ask whats the issue with:

``` 
if not licence.goods.filter(usage__lt=F("quantity")).exists():
    action = HMRCIntegrationActionEnum.EXHAUST
    change_status = licence.exhaust
```

licences/libraries/hmrc_integration_operations.py
andytaylorkainos: Decided to return the UUID in the end!

licences/libraries/hmrc_integration_operations.py
BenAndrew-Kainos: Nothing wrong with that :)

PR #577, opened 2020-07-20
bug: LT-2794 - Disallow UK end users on SITLs

iichr:
- United Kingdom shouldn't be an option to select for end users on standard transhipments


applications/views/existing_parties.py
```diff
@@ -36,6 +40,12 @@ def get_queryset(self):
         uncopied_parties = self.get_uncopied_parties(application.organisation, params)
         newest_copied_parties = self.get_newest_copied_parties(application.organisation, params)
 
+        # Exclude the UK if end user on standard transhipment
+        if application.case_type.id == CaseTypeEnum.SITL.id:
+            if "party_type" in request_data and request_data["party_type"][0] == "end_user":
```
BenAndrew-Kainos, line 45: This seems to rely of end user being the first party type. Shouldn't it be `if "end_user" in request_data["party_type"]`

PR #581, opened 2020-07-21
fixes: LT-2575, LT-2253

jfaracik:
* LT-2575 - Notification is no longer sent to internal users when an application is submitted
* LT-2253 - Selecting other on select the role of the third party question should force user to give a description
* Small code cleanup


audit_trail/managers.py
```diff
@@ -23,23 +23,22 @@ def create(self, *args, **kwargs):
         from cases.models import Case
 
         target = kwargs.get("target")
+        actor = kwargs.get("actor")
+        send_notification = kwargs.pop("send_notification") if "send_notification" in kwargs else True
+        ignore_case_status = kwargs.pop("ignore_case_status") if "ignore_case_status" in kwargs else False
```
andytaylorkainos, line 28: you can do kwargs.pop("ignore_case_status", False)

applications/libraries/application_helpers.py
```diff
@@ -68,4 +68,6 @@ def create_submitted_audit(request, application, old_status):
                 "old": CaseStatusEnum.get_text(old_status),
             }
         },
+        ignore_case_status=True,
```
BenAndrew-Kainos, line 71: What is ignore case status for (surely we shouldn't ignore it)

audit_trail/managers.py
jfaracik: Thanks! The more you know 🌟

applications/libraries/application_helpers.py
jfaracik: AuditManager doesn't create audit entries for cases with a status of `draft`, so there's an optional parameter `ignore_case_status` to ignore this check.

PR #583, opened 2020-07-21
bugfix: LT-2245 and LT-2401

andytaylorkainos:


users/models.py
```diff
@@ -50,35 +50,21 @@ class Meta:
...
-        if extra_fields.get("is_staff") is not True:
-            raise ValueError("Superuser must have is_staff=True.")
-        if extra_fields.get("is_superuser") is not True:
-            raise ValueError("Superuser must have is_superuser=True.")
-
-        return self._create_user(email, password, **extra_fields)
+    def create(self, **kwargs):
+        email = kwargs.get("email")
+        if email:
+            kwargs["email"] = email.lower()
```
BenAndrew-Kainos, line 66: Can we not just put the .lower() on the save?

andytaylorkainos: yep - probably overkill on both. however this class cannot be deleted which is why i was trying to repurpose it

andytaylorkainos: done!

PR #589, opened 2020-07-22
bug: Approve UK Continental Shelf application

BenAndrew-Kainos:
- Changed Advice to use BaseUser so that the system user can produce approve advice (GovUser still validated on serializer)
- Fixed Country reverse lookup


applications/serializers/advice.py
```diff
@@ -100,6 +101,12 @@ def validate_proviso(self, value):
 
         return value
 
+    def validate_user(self, value):
+        if value.type != UserType.INTERNAL:
+            raise ValidationError("Only GovUsers can give advice")
```
rorysproule1, line 106: can we put this in content?

BenAndrew-Kainos: It's a dev only error and wouldn't show for users

PR #595, opened 2020-07-31
Create search index for Cases

saruniitr:


search/management/commands/esindex_migrate.py
```diff
@@ -0,0 +1,64 @@
...
+
+        data = []
+        for case in cases:
+            case_document = documents.case_model_to_document(case=case, index_name=self.new_index_name)
+            data.append(case_document.to_dict(True))
+        bulk(self.client, data)
+
+    def update_aliases(self):
+        current_indices = [
+            index for index, _ in self.client.indices.get(PATTERN).items() if index != self.new_index_name
```
richtier, line 42: could this be
```
index for index in self.client.indices.get(PATTERN).keys() if index != self.new_index_name
```?

Comment (saruniitr) [reply search/management/commands/esindex_migrate.py:42]: yep, no need of the unused variable, will update this

File: conf/settings.py [hunk 2]
```diff
@@ -249,6 +251,23 @@
 
 USE_TZ = True
 
+# Elasticsearch configuration
+LITE_API_ENABLE_ES = env.bool("LITE_API_ENABLE_ES", False)
+if LITE_API_ENABLE_ES:
+    ELASTICSEARCH_DSL = {
+        'default': {
+            'hosts': 'localhost:9200'
```
kerin, line 259: Could you make that an env var?

saruniitr: yes forgot about this, will fix it.

PR #598, opened 2020-08-10
Settings: Allow DEBUG=True if no django_extensions installed

r4vi:
The dev environments are run with DEBUG=True but we don't install
django_extensions in those environments so the app fails to start.

This only activates django_extensions if it is installed

### DIFF
--- a/conf/settings.py
+++ b/conf/settings.py
@@ -336,7 +336,7 @@ SIGNING_LOCATION = env("SIGNING_LOCATION")
 SIGNING_REASON = env("SIGNING_REASON")
 
 # Django Extensions
-if DEBUG:
+if DEBUG and "django_extensions" in sys.modules:
     INSTALLED_APPS.append("django_extensions")
 
     GRAPH_MODELS = {


conf/settings.py
```diff
@@ -336,7 +336,7 @@
 SIGNING_REASON = env("SIGNING_REASON")
 
 # Django Extensions
-if DEBUG:
+if DEBUG and "django_extensions" in sys.modules:
```
richtier, line 339: perhaps a feature flag `FEATURE_DJANGO_EXTENSIONS_ON`

r4vi: the way I was doing it before you added it was using a local.py that looked like this:

```
from .settings import *
INSTALLED_APPS.insert(0, "django_extensions")
LOGGING = {
    "version": 1,
    "disable_existing_loggers": False,
    "handlers": {"console": {"class": "logging.StreamHandler",},},
    "root": {"handlers": ["console"], "level": "WARNING",},
}
```
and setting that as my django_settings_module

r4vi: I did consider FEATURE_DJANGO_EXTENSIONS_ON but it's kind of weird:

* Why wouldn't you want it on if you have it installed? 
* If you flip that to FEATURE_DJANGO_EXTENSIONS_ON=True then deployed envs will still break

Happy to move to that instead if you prefer, it is more explicit...

richtier: makes sense :+1:

r4vi:
This is one of the reasons UAT build is red right now

PR #605, opened 2020-08-20
LTD-180: Read demo_flags, demo_queues and demo_teams from Vault

saruniitr:


static/management/commands/seedinternaldemodata.py
```diff
@@ -10,6 +15,11 @@
 TEAMS_FILE = "lite_content/lite_api/demo_teams.csv"
```
r4vi, line 15: can be deleted now?

conf/settings.py
```diff
@@ -327,6 +327,11 @@
     env("EXPORTER_BASE_URL") if env("EXPORTER_BASE_URL") else f"https://exporter.lite.service.{ENV}.uktrade.digital"
 )
 
+# Demo flags
+LITE_API_DEMO_FLAGS_CSV = env.bytes("LITE_API_DEMO_FLAGS_CSV")
+LITE_API_DEMO_QUEUES_CSV = env.bytes("LITE_API_DEMO_QUEUES_CSV")
+LITE_API_DEMO_TEAMS_CSV = env.bytes("LITE_API_DEMO_TEAMS_CSV")
```
r4vi, line 333: `env.str(..., multiline=True)` might work around the issue of eating newlines

static/management/commands/seedinternaldemodata.py
saruniitr: yes

conf/settings.py
saruniitr: yes. for some reason it didn't work before. I might've made a mistake.

r4vi (approved):

PR #607, opened 2020-08-26
LTD-126 Make Docker-compose work

r4vi:
- Now allows communication with frontend using docker shared network
- Dependencies required for document signing now work in docker (by removing
- Procfile hack)
- Updated Readme for docker-compose setup

related to: https://github.com/uktrade/lite-frontend/pull/24


r4vi:
Note (to self): removing the `pip3 install endesive` hack from the Procfile doesn't work - will need to figure out why or roll back

Procfile
```diff
@@ -1,2 +1,2 @@
-web: pip3 install endesive && python manage.py migrate && gunicorn --worker-class gevent -c api/conf/gconfig.py -b 0.0.0.0:$PORT api.conf.wsgi
+web: python manage.py migrate && gunicorn --worker-class gevent -c api/conf/gconfig.py -b 0.0.0.0:$PORT conf.wsgi
```
saruniitr, line 1: should it be `api.conf.wsgi` at the end?

r4vi: good spot

local.env
```diff
@@ -1,7 +1,11 @@
 DEBUG=True
+
 DATABASE_URL=postgres://postgres:password@localhost:5432/postgres
+# Comment above and uncomment below for running in docker
+#DATABASE_URL=postgres://postgres:password@db:5432/postgres
```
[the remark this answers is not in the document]
r4vi: yes you're right, I changed it to match local config but forgot to update the example

r4vi:
@saruniitr updated as per your comments and backed out the change in `Procfile` to remove the `pip3 install endsive` - fixing that can be part of another ticket.

saruniitr (approved):
looks good to me. I guess we can remove DON'T MERGE label now @r4vi ?

r4vi:
@saruniitr let me do one final check to see if it deploys to cloudfoundry and yes we can remove the label

r4vi:
deploy is clean - merging

PR #609, opened 2020-08-28
LTD-152 Search by Product

saruniitr:
## Change description

Add index based on `BaseApplication` to be able search by application attributes (case_type, status etc), associated product attributes (part_number, description etc), associated parties (name, address etc).

Further required in terms of testing searches by all attributes with different keywords, combinations but this PR represents the base required to carry additional work.


richtier:
can we be more explicit (but yes more verbose) with the name, go `good_on_application` instead of goa

api/search/goa/documents.py
```diff
@@ -0,0 +1,74 @@
...
+
+@registry.register_document
+class GoodOnApplicationDocumentType(Document):
+    id = TextField()
+    quantity = FloatField()
+    value = FloatField()
+    unit = TextField()
+    item_type = TextField()
+    incorporated = BooleanField(attr="is_good_incorporated")
+    good = ObjectField(
```
richtier, line 17: would `InnerDoc` work here? https://elasticsearch-dsl.readthedocs.io/en/latest/persistence.html#document i.e. 

```
class Good(InnerDoc):
   id = TextField()
   ...

class GoodOnApplicationDocumentType(Document):
    ...
    good = Object(Good)
```

saruniitr: yes definitely, it wasn't working when I first tried so flattened it to to unblock. I will investigate further and replace it with InnerDoc

saruniitr:
> can we be more explicit (but yes more verbose) with the name, go `good_on_application` instead of goa

Agree, couldn't sensibly shorten it so went with that.

richtier (approved):

api/search/application/documents.py
```diff
@@ -0,0 +1,77 @@
...
+    products = NestedField(
+        attr="goods",
+        properties={
+            "quantity": FloatField(),
+            "value": FloatField(),
+            "unit": TextField(),
+            "item_type": TextField(),
+            "incorporated": BooleanField(attr="is_good_incorporated"),
+            "good": ObjectField(
+                properties={
```
richtier, line 40: we can flatten. may result in results that are easier to query and render on the frontend

```
 "description": TextField(attr="good.description"),
 "part_number": TextField(attr="good.part_number"),
...
```

local.env
```diff
@@ -78,3 +78,5 @@ PERMISSIONS_FINDER_URL='' # required for accessing exporter lite content, can be
 LITE_API_DEMO_FLAGS_CSV="field\nteam1\nteam2"
 LITE_API_DEMO_QUEUES_CSV="field\nteam1\nteam2"
 LITE_API_DEMO_TEAMS_CSV="field\nteam1\nteam2"
+
+ELASTICSEARCH_HOST=http://localost:9200
```
r4vi, line 82: missing new line at end of file here

r4vi (approved):
LGTM :+1:

local.env
saruniitr: fixed

PR #612, opened 2020-09-07
Expose queue id and incase filters

richtier:


saruniitr (approved):
lgtm.
Autocomplete should help with search so it makes senses to remove some of the fields from search, we can always add them later if required.

PR #615, opened 2020-09-08
LTD-173: Single auth model

r4vi:
They now have a conventional relationship (onetoone) with BaseUser which is the
`AUTH_USER_MODEL`, this paves the way for more work to allow setting
`request.user` as BaseUser everywhere.

I also added some type annotations on the way to help me (and my IDE) understand what was going on during the refactoring


kerin (approved):
:shipit:

api/audit_trail/service.py
```diff
@@ -101,7 +111,7 @@ def filter_object_activity(
     if team:
         gov_content_type = ContentType.objects.get_for_model(GovUser)
         user_ids = audit_qs.filter(actor_content_type=gov_content_type).values_list("actor_object_id", flat=True)
-        team_user_ids = GovUser.objects.filter(id__in=list(user_ids), team=team).values_list("id", flat=True)
+        team_user_ids = GovUser.objects.filter(pk__in=list(user_ids), team=team).values_list("pk", flat=True)
         audit_qs = audit_qs.filter(actor_object_id__in=list(team_user_ids))
```
richtier, line 115: not new code I know, but interesting use of application-side filtering. I wonder why not done at db level

PR #620, opened 2020-09-10
Fix hawk auth on search

richtier:


api/core/authentication.py
```diff
@@ -36,7 +36,7 @@ def authenticate(self, request):
 
         from api.organisations.libraries.get_organisation import get_request_user_organisation_id
 
-        _, hawk_receiver = HawkOnlyAuthentication().authenticate(request)
+        hawk_receiver = _authenticate(request, _lookup_credentials)
```
richtier, line 39: doing the same thing but less complex

api/core/authentication.py
```diff
@@ -125,13 +126,7 @@ def authenticate(self, request):
...
         """
 
-        try:
-            hawk_receiver = _authenticate(request, _lookup_credentials)
-        except HawkFail as e:
-            logging.error(f"Failed HAWK authentication {e}")
-            raise e
-
-        return AnonymousUser, hawk_receiver
+        return AnonymousUser, _authenticate(request, _lookup_credentials)
```
richtier, line 129: doing the same thing but less complex

api/search/application/views.py
```diff
@@ -21,6 +22,7 @@ class MatchBoolPrefix(Query):
 class ApplicationDocumentView(DocumentViewSet):
     document = ApplicationDocumentType
     serializer_class = ApplicationDocumentSerializer
+    authentication_classes = (GovAuthentication,)
```
richtier, line 25: it turns out the frontend requires all endpoints to have hawk auth enabled

r4vi (approved):

api/search/application/views.py
r4vi: that would explain it

PR #621, opened 2020-09-11
Fix signals to update ES index

saruniitr:
## Change description

Previous changes to update the ES index on data updates are not working.
The registry expects model corresponding to the index which is BaseApplication
in our case so the updates for related models were not happening. Update
signals to get the relevant BaseApplication instance.


api/search/signals.py
```diff
@@ -1,41 +1,59 @@
...
-            registry.update(instance)
+    """
+    To keep the index fresh we need to update them whenever the underlying data is updated.
+    The immediate attributes of the index are handled by django_elasticsearch_dsl but
+    additional changes are required to update the related models.
 
+    The registry update code requires model corresponding to the index which is BaseApplication
+    in our case. We intercept the signals and update the index.
+    """
+    if app_label == "applications":
```
richtier, line 23: would this do the same? `if issubclass(instance, BaseApplication) or issubclass(instance, BaseApplication):`

api/search/signals.py
```diff
@@ -1,41 +1,59 @@
...
+        if model_name in [
+            "standardapplication",
+            "baseapplication",
+            "openapplication",
+            "exhibitionclearanceapplication",
+        ]:
+            application = BaseApplication.objects.get(id=instance.id)
```
richtier, line 30: `application = instance.baseapplication`

api/search/signals.py
```diff
@@ -1,41 +1,59 @@
...
+                registry.update(poa.application)
 
     if app_label == "organisations":
         if model_name == "organisation":
-            instances = instance.cases.all()
-            for _instance in instances:
-                registry.update(_instance)
+            for case in instance.cases.all():
+                try:
+                    base_application_id = BaseApplication.objects.get(id=case.id)
```
richtier, line 56: `base_application_id` implies the variable is a primary key, but it's really and instance of BaseApplication

saruniitr: indeed, will update this.

api/search/signals.py
saruniitr: ok

api/search/signals.py
saruniitr: yes that is better

api/search/signals.py
```diff
@@ -1,41 +1,52 @@
...
+    To keep the index fresh we need to update them whenever the underlying data is updated.
+    The immediate attributes of the index are handled by django_elasticsearch_dsl but
+    additional changes are required to update the related models.
 
-        if model_name == "goodonapplication":
-            registry.update(instance)
+    The registry update code requires model corresponding to the index which is BaseApplication
+    in our case. We intercept the signals and update the index.
+    """
+    if issubclass(instance.__class__, BaseApplication):
```
richtier, line 23: I think `sender` is the instance's class

api/search/signals.py
```diff
@@ -1,41 +1,52 @@
...
+                registry.update(poa.application)
 
     if app_label == "organisations":
         if model_name == "organisation":
-            instances = instance.cases.all()
-            for _instance in instances:
-                registry.update(_instance)
+            for case in instance.cases.all():
+                try:
+                    base_application = BaseApplication.objects.get(id=case.id)
```
richtier, line 49: `case.baseapplication` works here too. When a model is subclassed, django adds a pointer on the subclass to the superclass (with the superclass's app label as the attribute name) 

so it works on the `caseassignment` case too

saruniitr: fixed

api/search/signals.py
saruniitr: updated

saruniitr (commented):
+1 for #623 changes

PR #622, opened 2020-09-14
Support multi-index application search

richtier:


api/conf/settings.py
```diff
@@ -103,8 +103,6 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
-    "api.conf.middleware.LoggingMiddleware",
-    "api.conf.middleware.DBLoggingMiddleware",
```
saruniitr, line 107: Any particular reason this is being removed? I am good with this as I comment locally also, just wanted to understand the reason.

api/search/application/documents.py
```diff
@@ -90,28 +90,14 @@ class Product(InnerDoc):
     quantity = fields.FloatField()
     value = fields.KeywordField()
     unit = fields.KeywordField()
-    item_type = fields.KeywordField()
-    incorporated = fields.BooleanField(attr="is_good_incorporated")
```
saruniitr, line 94: incorporated - this is one of the acceptance criteria

api/search/application/documents.py
```diff
@@ -90,28 +90,14 @@ class Product(InnerDoc):
...
-    status = fields.KeywordField(attr="good.status")
-    comment = fields.TextField(attr="good.comment", copy_to="wildcard", analyzer=descriptive_text_analyzer)
-    grading_comment = fields.TextField(
-        attr="good.grading_comment", copy_to="wildcard", analyzer=descriptive_text_analyzer
-    )
-    report_summary = fields.TextField(
-        attr="good.report_summary", copy_to="wildcard", analyzer=descriptive_text_analyzer
-    )
-    is_military_use = fields.TextField(attr="good.is_military_use")
-    is_pv_graded = fields.TextField(attr="good.is_pv_graded")
```
saruniitr, line 112: Could you please clarify why these are removed? Is it it because they are not available in SPIRE?

api/conf/settings.py
richtier: the middleware is noisy and causes more harm than good while trying to understand the cause of errors :)

sentry + django debug toolbar are far better

api/search/application/documents.py
richtier: correct - the fields are not currently being used for searching and not displayed in the UI, so I removed them as it's extra effort to get parity with SPIRE

api/conf/settings.py
saruniitr: +1

api/search/application/documents.py
saruniitr: in that case I think we can retain `comment` and `organisation`

richtier: I'll re-add comment

will product `organisation` be any different than application `organisation`? I suspect not.

saruniitr (approved):
looks good to me

PR #623, opened 2020-09-14
Fix reindex

richtier:


api/users/models.py
```diff
@@ -142,10 +142,6 @@ class ExporterUser(models.Model, BaseUserCompatMixin):
 
     baseuser_ptr = models.OneToOneField(BaseUser, on_delete=models.CASCADE, primary_key=True,)
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.baseuser_ptr.type = UserType.EXPORTER
```
richtier, line 147: this was doing a db read during instantiation: accessing `self.baseuser_ptr` does db read. That harmed ES indexing performance

saruniitr (approved):
looks good to me

PR #628, opened 2020-09-16
Include spire index only if the feature is enabled

saruniitr:
## Change description

Spire index is only included in the search if the feature is enabled.
Some of the spire related apps are not available in all envs so include the index selectively based on this feature flag.


api/conf/settings.py
```diff
@@ -251,12 +251,17 @@
         "default": {"hosts": env.str("ELASTICSEARCH_HOST")},
     }
 
+    LITE_ENABLE_SPIRE_SEARCH = env.bool("LITE_ENABLE_SPIRE_SEARCH", False)
```
richtier, line 254: I think the `LITE_` prefix is redundant - the whole thing is LITE

richtier (approved):

api/conf/settings.py
saruniitr: ok will remove it

PR #629, opened 2020-09-18
Update request.user to Django BaseUser

saruniitr:
## Change description

Django expects request.user to be an instance of BaseUser but this is either
an ExporterUser/GovUser at the moment.

This interefers with basic Django requirement and can lead to incompatibilities,
issues going forward.


r4vi (approved):
the changes you've made look good to me, hard part is if we introduced any bugs but I guess only time will tell on that front

PR #634, opened 2020-09-24
Perform good control review at application level

richtier:


r4vi (approved):
code changes are good but I'd try to run the front end browser tests running backed by this version of the API to ensure no text changes have happened because of the change to `GoodControlled` enum

PR #637, opened 2020-10-09
LTD-215: Reporting - Add query for standard licences with goods and ratings

r4vi:
To act as a template for building reports that aggregate this raw data like
summary of SIELs

### DIFF
--- a/api/reports/queries/standard.py
+++ b/api/reports/queries/standard.py
@@ -1,0 +1,52 @@
+GOODS_AND_RATINGS = """
+WITH goods as (
+    select ll.reference_code         "licence_ref"
+         , sa.baseapplication_ptr_id "application_id"
+         , ll.id "licence_pk"
+         , g.id                      "good_id"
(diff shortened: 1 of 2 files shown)


api/reports/queries/standard.py
```diff
@@ -0,0 +1,52 @@
+GOODS_AND_RATINGS = """
+WITH goods as (
+    select ll.reference_code         "licence_ref"
+         , sa.baseapplication_ptr_id "appliaction_id"
```
kerin, line 4: typo?

saruniitr, line 4: typo in `application_id`

kerin (commented):
Is it worth thinking about adding tests at this stage, or is this proof of concept -ish?

api/reports/queries/standard.py
r4vi: thanks - good spot

kerin (approved):

PR #639, opened 2020-10-13
Add expoint to expose OGL types

saruniitr:
This will be used by Data workspace to generate reports

### DIFF
--- a/api/open_general_licences/tests/test_gets.py
+++ b/api/open_general_licences/tests/test_gets.py
@@ -123,3 +123,9 @@ class ExporterListTests(DataTestClient):
         self.assertEquals(response.status_code, status.HTTP_200_OK)
         self.assertEquals(len(response.json()["results"]), 1)
         self.assertEquals(len(response.json()["results"][0]["registrations"]), 2)
+
+
+class DataWorkspaceTests(DataTestClient):
+    def test_dw_view_ogl_types(self):
+        response = self.client.get(reverse("open_general_licences:ogl-dw-list"))
+        self.assertEquals(response.status_code, status.HTTP_200_OK)
--- a/api/open_general_licences/urls.py
+++ b/api/open_general_licences/urls.py
@@ -1,11 +1,17 @@
-from django.urls import path
+from django.urls import path, include
+from rest_framework.routers import DefaultRouter
 
 from api.open_general_licences import views
 
 app_name = "open_general_licences"
 
+router = DefaultRouter()
+router.register("", views.OpenGeneralLicenceListDW, basename="ogl-dw")
+
+
 urlpatterns = [
     path("", views.OpenGeneralLicenceList.as_view(), name="list"),
     path("<uuid:pk>/", views.OpenGeneralLicenceDetail.as_view(), name="detail"),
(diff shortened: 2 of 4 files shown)


api/open_general_licences/urls.py
```diff
@@ -1,11 +1,17 @@
...
 
+router = DefaultRouter()
+router.register("", views.OpenGeneralLicenceListDW, basename="ogl-dw")
+
+
 urlpatterns = [
     path("", views.OpenGeneralLicenceList.as_view(), name="list"),
     path("<uuid:pk>/", views.OpenGeneralLicenceDetail.as_view(), name="detail"),
     path("<uuid:pk>/activity/", views.OpenGeneralLicenceActivityView.as_view(), name="activity"),
+    path("dw", include(router.urls)),
```
r4vi, line 16: it's ok here for now but once we have a bunch of endpoints (not just for OGL definitions) for DW consumption we might want to move them all to their own namespace.

e.g 
```
/dw/v1/ogls/
/dw/v1/licences/
/dw/v1/cases/
``` which would be defined by a dw django app containing it's own urls.py

**Review by r4vi — CHANGES_REQUESTED** [review]
looks good, can we add a permission check to make sure only DW hawk key can access this (new permission class?) not sure if/how you can work out who a hawk key belongs to.

File: api/open_general_licences/views.py [hunk 2]
```diff
@@ -195,3 +195,13 @@ def get(self, request, pk):
         filters = audit_trail_service.get_objects_activity_filters(pk, content_type)
 
         return JsonResponse(data={"activity": data, "filters": filters}, status=status.HTTP_200_OK)
+
+
+class OpenGeneralLicenceListDW(viewsets.ReadOnlyModelViewSet):
+    authentication_classes = (HawkOnlyAuthentication,)
+    serializer_class = OpenGeneralLicenceSerializer
```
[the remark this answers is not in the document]
saruniitr: ok will add permissions class. Are we going to create specific users for DW or it will be AnonymousUser?

r4vi:
From Django's perspective it can just be anonymous user

api/open_general_licences/urls.py
saruniitr: should I create the app `dataworkspace` or simply `dw` as part of this?

r4vi:
Don't worry about DW app for now, let's think about it when we add the next
DW endpoint

r4vi (approved):

api/open_general_licences/tests/test_gets.py
```diff
@@ -123,3 +129,24 @@ def test_exporter_view_registered_licences_success(self):
         self.assertEquals(response.status_code, status.HTTP_200_OK)
         self.assertEquals(len(response.json()["results"]), 1)
         self.assertEquals(len(response.json()["results"][0]["registrations"]), 2)
+
+
+class DataWorkspaceTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        test_host = "http://testserver"
```
saruniitr, line 137: Receiver uses `build_absolute_uri` because Sender provides absolute url hence in the tests also we have to provide absolute url

r4vi (approved):

PR #642, opened 2020-10-20
LTD-264:  Cannot generate documents compliance cases

saruniitr:
## Change description

- Fix context preparation that is used to generate documents
- Fix template also to display the values correctly

### DIFF
--- a/api/compliance/enums.py
+++ b/api/compliance/enums.py
@@ -37,4 +37,4 @@ class ComplianceRiskValues:
 
     @classmethod
     def to_str(cls, risk_value):
-        return next(choice[1] for choice in cls.choices if choice[0] == risk_value)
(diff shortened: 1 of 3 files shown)


api/compliance/enums.py
```diff
@@ -37,4 +37,4 @@ class ComplianceRiskValues:
 
     @classmethod
     def to_str(cls, risk_value):
-        return next(choice[1] for choice in cls.choices if choice[0] == risk_value)
+        return next(choice[1] for choice in cls.choices if choice[0] == risk_value) if risk_value else ""
```
richtier, line 40: can you add a unit test checking expecting outcomes happen with ComplianceRiskValues please :)

richtier (approved):

PR #643, opened 2020-10-21
Mark user text as safe to fix underlining

saruniitr:
We are using markdown to render user content in bold, italic and underline but
markdown doesn't support underline so inline html is used in markdown.

The current code escaping html tags, marking it as safe to avoid escaping
underline html tags.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -49,6 +49,7 @@ django-elasticsearch-dsl-drf = "~=0.20"
 django-elasticsearch-dsl = "~=7.1"
 gunicorn = "~=20.0"
 gevent = "~=20.6"
+bleach = "*"
 
 [requires]
 python_version = "3.7"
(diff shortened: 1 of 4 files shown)


r4vi (changes requested):
if we allow this couldn't a user add any HTML tag they like including malicious javascript?

I would suggest running through a tool like: https://bleach.readthedocs.io/en/latest/clean.html#allowed-tags-tags with a whitelist of tags before `mark_safe`ing the string

saruniitr:
> if we allow this couldn't a user add any HTML tag they like including malicious javascript?
> 
> I would suggest running through a tool like: https://bleach.readthedocs.io/en/latest/clean.html#allowed-tags-tags with a whitelist of tags before `mark_safe`ing the string

Yes this can be issue but don't see any other way of supporting this.
I will check that tool.

r4vi (approved):

PR #644, opened 2020-10-23
Add end_use_control to good-on-application level

richtier:


README.md
```diff
@@ -39,6 +39,15 @@ Service for handling backend calls in LITE.
 - At this point you might want to seed your database with some static
   - run `docker-compose run ./manage.py seedall`
 
+## Add a single user:
+
+Run the following command to add new users:
+
+```
+INTERNAL_USERS='[{"email"=>"foo@bar.gov.uk"}]' ./manage.py seedinternalusers
```
r4vi, line 47: does this format work? I always do `'[{"email":"foo@bar.gov.uk", "role": "Super User"}]'`

api/applications/models.py
```diff
@@ -296,9 +297,13 @@ class AbstractGoodOnApplication(TimestampableModel):
     is_good_controlled = models.BooleanField(default=None, blank=True, null=True)
     comment = models.TextField(help_text="control review comment", default=None, blank=True, null=True)
     report_summary = models.TextField(default=None, blank=True, null=True)
-
     application = models.ForeignKey(BaseApplication, on_delete=models.CASCADE, null=False)
     control_list_entries = models.ManyToManyField(ControlListEntry)
+    end_use_control = ArrayField(
```
r4vi, line 302: can you have more than one end use?

r4vi (approved):
left some questions but no blockers

api/applications/models.py
richtier: Daniel thinks maybe. We can lock it down to single when we get a concrete answer :)

README.md
richtier: I tied that and it did not work, but `=>` did...odd

PR #645, opened 2020-10-27
Add healthcheck endpoint

richtier:


api/conf/settings.py
```diff
@@ -42,8 +42,11 @@
 ADMIN_ENABLED = True
 
 ALLOWED_HOSTS = env.list("ALLOWED_HOSTS", default=[])
-# Application definition
 
+# django-allow-cidr
+ALLOWED_CIDR_NETS = ["10.0.0.0/8"]
```
saruniitr, line 47: Does it help if it is an environment variable? is it going to be the same for all environments?

saruniitr (approved):
Requested a clarification, otherwise looks good to me.

api/conf/settings.py
richtier: same for all envs :)

PR #650, opened 2020-11-05
Ltd 305 attach docs to ecju query response

saruniitr:
## Change description

API changes to support document uploads when responding to an ECJU query or pre-visit questionnaire or compliance action.


api/cases/views/views.py
```diff
@@ -542,6 +560,99 @@ def put(self, request, pk, ecju_pk):
...
+        if data.get("has_document_to_upload"):
+            document_to_upload = str_to_bool(data["has_document_to_upload"])
+            if not document_to_upload:
+                ecju_query.missing_document_reason = data["missing_document_reason"]
+                serializer = EcjuQueryMissingDocumentSerializer(instance=ecju_query, data=data, partial=True)
+                if serializer.is_valid():
+                    serializer.save()
+                    ecju_query_detail = EcjuQueryCreateSerializer(ecju_query).data
+                else:
+                    return JsonResponse(data={"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
```
richtier, line 578:
```suggestion
               serializer.is_valid(raise_exception=True)
               serializer.save()
               ecju_query_detail = EcjuQueryCreateSerializer(ecju_query).data

```
`raise_exception` will make DRF return a 400 with the serializer erorrs

richtier (approved):

api/cases/views/views.py
saruniitr: yes, will update this.

PR #651, opened 2020-11-05
Allow caseworker and exporter to download document

richtier:


r4vi (approved):

api/documents/views.py
```diff
@@ -13,7 +13,7 @@ class DocumentDetail(APIView):
     Get information about a Document
     """
 
-    authentication_classes = (GovAuthentication,)
+    authentication_classes = (SharedAuthentication,)
```
r4vi, line 16: Question about this - if I figure out a random document ID do I end up with the ability to download an arbitrary document?
Do we need something in the vein of "django object permissions" to ensure that if you are an exporter user you can only download your own ones and not "any" document ?

r4vi (changes requested):
just looked at this again and had a thought about enumeration attacks / brute forcing document IDs

api/documents/views.py
richtier: I'll check :+1:

r4vi (approved):

PR #652, opened 2020-11-09
Add firearm control details to good on application level

richtier:


api/goods/models.py
```diff
@@ -45,6 +45,7 @@ class FirearmGoodDetails(models.Model):
     has_identification_markings = models.BooleanField(null=False)
     identification_markings_details = models.TextField(blank=True, max_length=2000, null=True)
     no_identification_markings_details = models.TextField(blank=True, max_length=2000, null=True)
+    serial_number = models.TextField(default="")
```
saruniitr, line 48: `CharField` may be more appropriate as S/N may not be very long?

saruniitr (approved):
requested clarification, otherwise looks good to me

api/goods/models.py
richtier: We don't know the length of the seriali number

Given there is no performance loss by using `TextField` over `CharField`, but there is a negative of using CharField: we need to set max_length. will it be long enough? if we dont know how long we need it then we could be like `max_length=100` but then that's extra noise we don't need

https://djangodoctor.medium.com/thinking-charfield-youll-think-again-after-this-8d207abbca4d

saruniitr: thank you, agree on the performance side but S/N probably cannot go more than 32/64 chars I would imagine. As it is safe to go with `TextField`, lets go with that.

api/goods/models.py
```diff
@@ -45,6 +45,12 @@ class FirearmGoodDetails(models.Model):
...
+    has_proof_mark = models.BooleanField(
+        help_text='Has been proofed (by a proof house) indicating it is safe to be used.',
+        null=True,
+    )
+    no_proof_mark_details = models.TextField(default="", blank=True)
```
r4vi, line 53: I assume this is a text explanation about why there is no proof mark. can we have help_text saying that?

PR #653, opened 2020-11-20
Add FOI reason field

eadpearce:


api/applications/models.py
```diff
@@ -166,6 +166,7 @@ class BaseApplication(ApplicationPartyMixin, Case):
 
     intended_end_use = models.CharField(default=None, blank=True, null=True, max_length=2200)
     agreed_to_foi = models.BooleanField(blank=True, default=None, null=True)
+    foi_reason = models.TextField(blank=True, null=True)
```
saruniitr, line 169: Documentation suggests avoid using `null=True` for string-based fields,
https://docs.djangoproject.com/en/3.1/ref/models/fields/#null

saruniitr (changes requested):

api/applications/models.py
eadpearce: Fixed!

r4vi (approved):

PR #657, opened 2020-11-24
LTD-398: Add additional firearms types to categorise accurately

saruniitr:
## Change description

To accurately categorise firearms products, additional types are introduced.
Added new enums for these types and some of the fields are not applicable for these types so the validation criteria and model field restrictions are modified accordingly.
Eg sections are not applicable for a software product relating to a firearm


api/goods/tests/tests_create.py
```diff
@@ -446,23 +446,6 @@ def test_add_category_two_good_no_type_selected_failure(self):
...
         self.assertEqual(errors["type"], [strings.Goods.FIREARM_GOOD_NO_TYPE])
 
-    def test_add_category_two_good_no_year_of_manufacture_or_calibre_failure(self):
-        data = {
-            "description": "coffee",
-            "is_good_controlled": False,
-            "is_pv_graded": GoodPvGraded.NO,
-            "item_category": ItemCategory.GROUP2_FIREARMS,
-            "validate_only": True,
-            "firearm_details": {"type": FirearmGoodType.AMMUNITION, "calibre": "", "year_of_manufacture": ""},
```
saruniitr, line 456: this is allowed now, eg if we are adding a firearm accessory or software relating to a firearm

api/goods/tests/tests_create.py
```diff
@@ -870,35 +853,6 @@ def test_add_category_two_good_has_markings_details_too_long_failure(
...
-        [
-            ["is_military_use", "True"],
-            ["modified_military_use_details", "some details"],
-            ["is_component", "True"],
-            ["designed_details", "some details"],
-            ["modified_details", "some details"],
-            ["general_details", "some details"],
-            ["uses_information_security", "True"],
-            ["information_security_details", "some details"],
-            ["software_or_technology_details", "some details"],
```
saruniitr, line 883: It is now possible to have these fields added/edited when adding firearm products hence removed the test

api/goods/models.py
```diff
@@ -35,9 +35,9 @@ class PvGradingDetails(models.Model):
 
 
 class FirearmGoodDetails(models.Model):
-    type = models.CharField(choices=FirearmGoodType.choices, max_length=25, blank=False)
-    year_of_manufacture = models.PositiveSmallIntegerField()
-    calibre = models.TextField(blank=False, max_length=15)
+    type = models.CharField(choices=FirearmGoodType.choices, max_length=50, blank=False)
```
richtier, line 38: just a callback to our previous talk about `CharField` vs `TextField`: is this was a `TextField` this change would not have been needed. Helps with the idea :sunglasses:

api/goods/models.py
```diff
@@ -35,9 +35,9 @@ class PvGradingDetails(models.Model):
...
+    year_of_manufacture = models.PositiveSmallIntegerField(blank=True, null=True)
+    calibre = models.TextField(blank=True, max_length=15)
```
richtier, line 40: max_length on a text field will mean the database does not validate the length but the form will. Is that expected? Seems worst of both worlds :)

api/goods/models.py
saruniitr: :+1: it did remind me our previous discussion.

api/goods/models.py
saruniitr: will remove it

richtier (approved):

PR #659, opened 2020-11-27
Support Year of Manufacture as part of add preexisting good flow

r4vi:
This extends the GoodOnApplicationCreateSerializer to support
year of manufacture

### DIFF
--- a/api/applications/serializers/good.py
+++ b/api/applications/serializers/good.py
@@ -111,10 +111,23 @@ class GoodOnApplicationCreateSerializer(serializers.ModelSerializer):
         error_messages={"required": strings.Goods.OTHER_ITEM_TYPE, "blank": strings.Goods.OTHER_ITEM_TYPE},
     )
     has_proof_mark = BooleanField(allow_null=True, required=False)
-    no_proof_mark_details = serializers.CharField(default="", allow_blank=True)
+    no_proof_mark_details = serializers.CharField(allow_blank=True, required=False)
+    year_of_manufacture = serializers.IntegerField(
+        allow_null=True,
+        required=False,
+        error_messages={
+            "required": strings.Goods.FIREARM_GOOD_NO_YEAR_OF_MANUFACTURE,
+            "invalid": strings.Goods.FIREARM_GOOD_YEAR_MUST_BE_VALID,
+        },
+    )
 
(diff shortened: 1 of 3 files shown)


api/applications/serializers/good.py
```diff
@@ -159,20 +163,30 @@ def __init__(self, **kwargs):
                     data["quantity"] = 1
                 if not data["value"]:
                     data["value"] = 0
-            if data.get("has_proof_mark") == "False" and data.get("no_proof_mark_details") == "":
```
r4vi, line 162: @richtier I hope this doesn't break your frontend branch - I think my changes achieve the same thing but not 100% sure. Let me know the corresponding frontend branch name and I'll try before merging

saruniitr (approved):
looks good to me

api/applications/views/goods.py
```diff
@@ -80,7 +80,7 @@ def post(self, request, pk):
             # note: Goods attached to applications also need documents. This is validated at a later stage.
             serializer = GoodOnApplicationCreateSerializer(data=data, partial=True)
             if serializer.is_valid():
-                return HttpResponse(status=status.HTTP_200_OK)
+                return JsonResponse(status=status.HTTP_200_OK, data=serializer.errors)
```
eadpearce, line 83: since this is under `if serializer.is_valid()` presumably serializer.errors will be empty? also may not be what the frontend is expecting but not sure

r4vi: yeah that's the intention, send an empty dict... before it was sending a blank http response and the "validate_good" method was barfing when trying to parse the json. Really just needs to be any valid json

r4vi: btw the frontend `validate_good` function wasn't being used before that's why this was never noticed

eadpearce: fair enough. seems frontend only checks for an "errors" key in the response so i think this should be fine?

eadpearce (approved):

api/applications/views/goods.py
r4vi: I think frontend only even goes into the value of the response if validation fails and you get a 400, so in the 200 case it never checks it, it just checks if it got a 200.

PR #662, opened 2020-12-02
LTD-411 Update Document Generation for FirearmDetails on Good on App

r4vi:
This changes the way Document generation works so it looks at the firearm
details on the good on application instead of the good. This will continue to
work for older cases where there is no firearms detail on the GOA by falling
back to the good.

### DIFF
--- a/api/letter_templates/context_generator.py
+++ b/api/letter_templates/context_generator.py
@@ -512,30 +512,21 @@ def _get_good_on_application_context(good_on_application, advice=None):
         if good_on_application.good.uses_information_security:
             good_context["information_security_details"] = good_on_application.good.information_security_details
     elif good_on_application.good.item_category in ItemCategory.group_two:
-        good_context["firearm_type"] = FirearmGoodType.to_str(good_on_application.good.firearm_details.type)
-        good_context["year_of_manufacture"] = good_on_application.good.firearm_details.year_of_manufacture
-        good_context["calibre"] = good_on_application.good.firearm_details.calibre
+        firearm_details = good_on_application.firearm_details or good_on_application.good.firearm_details
+        good_context["firearm_type"] = FirearmGoodType.to_str(firearm_details.type)
+        good_context["year_of_manufacture"] = firearm_details.year_of_manufacture
+        good_context["calibre"] = firearm_details.calibre
         good_context["is_covered_by_firearm_act_section_one_two_or_five"] = friendly_boolean(
-            good_on_application.good.firearm_details.is_covered_by_firearm_act_section_one_two_or_five
+            firearm_details.is_covered_by_firearm_act_section_one_two_or_five
         )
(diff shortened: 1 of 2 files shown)


api/letter_templates/context_generator.py
```diff
@@ -512,30 +512,21 @@ def _get_good_on_application_context(good_on_application, advice=None):
         if good_on_application.good.uses_information_security:
             good_context["information_security_details"] = good_on_application.good.information_security_details
     elif good_on_application.good.item_category in ItemCategory.group_two:
-        good_context["firearm_type"] = FirearmGoodType.to_str(good_on_application.good.firearm_details.type)
-        good_context["year_of_manufacture"] = good_on_application.good.firearm_details.year_of_manufacture
-        good_context["calibre"] = good_on_application.good.firearm_details.calibre
+        firearm_details = good_on_application.firearm_details or good_on_application.good.firearm_details
```
saruniitr, line 515: Should we include sporting shotgun and replica details?
Also not all the fields are applicable for all types, in exporter and internal side we show them conditionally perhaps they need to be added conditionally here as well otherwise letters include details not applicable for the product types.

r4vi: this is just setting up the context so I see no harm in adding other fields to the context, but we should maybe think about using a nested firearms_details structure in that case. right now it adds goods to the context like this: ```[{'description': 'program',
   'control_list_entries': ['ML1a'],
   'is_controlled': None,
   'part_number': '9627820545446',
   'applied_for_quantity': '100.0 Number of articles',
   'applied_for_value': '£1500.00',
   'is_incorporated': None,
   'item_category': 'Firearms',
   'is_pv_graded': 'No',
   'firearm_type': 'Ammunition',
   'year_of_manufacture': 1919,
   'calibre': '5.56x45mm',
   'is_covered_by_firearm_act_section_one_two_or_five': 'Yes',
   'section_certificate_number': 'section certificate number?',
   'section_certificate_date_of_expiry': datetime.date(2020, 12, 2),
   'has_identification_markings': 'Yes',
   'identification_markings_details': 'manage'}]```

for consistency we should maybe change it to have the firearms fields nested (and expose all of them).

Then templates can decide what they want to show based on type?

saruniitr: that sounds good :+1:

saruniitr (approved):

api/letter_templates/context_generator.py
r4vi: @saruniitr would you prefer I make those changes in this PR or merge this one and create a new ticket for that work?

richtier (approved):

api/letter_templates/context_generator.py
saruniitr: @r4vi  if its just one template may be in this PR, but if there are more best to do in a separate ticket but please decide whichever is easier.

PR #668, opened 2020-12-09
LTD-375: [API] Firearms section 1-2 certificate changes

saruniitr:
## Change description

Changes for Firearms section 1/2 question
 - Update Firearm details to include additional fields
 - Add new model `GoodOnApplicationDocument` to save Firearms certificate
 - Fix broken tests and add new ones

Bulk of the changes are adding views to support certificate create/retrieve/delete operations
Firearms details serializer validations and resetting not applicable fields as they change.


api/goods/models.py
```diff
@@ -42,7 +42,10 @@ class FirearmGoodDetails(models.Model):
     is_replica = models.BooleanField(blank=True, null=True)
     replica_description = models.TextField(blank=True, default="")
     # this refers specifically to section 1, 2 or 5 of firearms act 1968
-    is_covered_by_firearm_act_section_one_two_or_five = models.BooleanField(null=True)
+    is_covered_by_firearm_act_section_one_two_or_five = models.TextField(blank=True, default="")
```
r4vi, line 45: not entirely following why it was changed from Bool(null=True) to Text field

api/goods/tests/tests_edit.py
```diff
@@ -407,9 +407,10 @@ def test_edit_category_two_section_question_and_details_success(self):
         url = reverse("goods:good_details", kwargs={"pk": str(good.id)})
         request_data = {
             "firearm_details": {
-                "is_covered_by_firearm_act_section_one_two_or_five": "True",
+                "is_covered_by_firearm_act_section_one_two_or_five": "Yes",
+                "firearms_act_section": "firearms_act_section1",
                 "section_certificate_number": "ABC123",
-                "section_certificate_date_of_expiry": "2020-12-12",
+                "section_certificate_date_of_expiry": "2022-12-12",
```
r4vi, line 413: should we use `now() + timedelta(year=1)` so that these tests don't break when we go past 2022-12-12?

r4vi (approved):
:+1: assuming there is a good reason for changing is_covered_by_firearm_act_section_one_two_or_five to Text from bool it looks good

api/goods/models.py
saruniitr: It currently has three values, `Yes`, `No` and `I don't know`

api/goods/tests/tests_edit.py
saruniitr: ok, this may not be required now but I will double check and update.

PR #671, opened 2020-12-15
LTD-319: Digital signature text in licence template mismatch

r4vi:
The text in the licence template was hard coded and therefore wouldn't match the
"include_digital_signature" boolean set by the user when creating the template.

This change passes the value to the part where the template is generated so it
can be correctly displayed.

### DIFF
--- a/api/goods/tests/tests_edit.py
+++ b/api/goods/tests/tests_edit.py
@@ -1,3 +1,6 @@
+from datetime import timedelta
+
+from django.utils.timezone import now
 from parameterized import parameterized
 from rest_framework import status
 from rest_framework.reverse import reverse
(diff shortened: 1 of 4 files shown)


api/goods/tests/tests_edit.py
```diff
@@ -405,11 +408,12 @@ def test_edit_category_two_section_question_and_details_success(self):
         )
 
         url = reverse("goods:good_details", kwargs={"pk": str(good.id)})
+        future_expiry_date = (now() + timedelta(days=365)).date().isoformat()
         request_data = {
             "firearm_details": {
                 "is_covered_by_firearm_act_section_one_two_or_five": "True",
                 "section_certificate_number": "ABC123",
-                "section_certificate_date_of_expiry": "2020-12-12",
+                "section_certificate_date_of_expiry": future_expiry_date,
```
saruniitr, line 416: thank you. I have fixed it my PR

saruniitr (approved):

api/goods/tests/tests_edit.py
r4vi: if yours is ready to merge I can remove this commit

saruniitr: no, please keep it

PR #680, opened 2021-01-04
Use spire country name for reports

richtier:


r4vi (approved):

api/reports/queries/standard.py
```diff
@@ -92,7 +92,7 @@
          , sub_type
          , role
          , country.id
-         , country.name
+         , country.report_name
```
r4vi, line 95: can we keep both here? I want `report_name` to only be there for SPIRE legacy matching - I don't want it to persist to be the lite "country_name" even after SPIRE is dead

r4vi (approved):
approved, despite my comment - I have some small tweaks to names to do in another PR so better to get this in and I'll fix various small things with column names

PR #684, opened 2021-01-06
Denial ingest, matching denials on application

richtier:
## Change description

Model, views for uploading denials, creating denials matched as part of application and deleting them.


api/applications/views/denials.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Denial matches belonging to a standard application
+    """
+
+    authentication_classes = (GovAuthentication,)
+
+    def get(self, request, pk):
+        denial_matches = DenialMatchOnApplication.objects.filter(application_id=pk, denial__is_revoked=False)
+        denial_matches_data = denial.DenialMatchOnApplicationViewSerializer(denial_matches, many=True).data
+        return JsonResponse(data={"denial_matches": denial_matches_data})
```
r4vi, line 23: do we expect many denials coming back from this endpoint? do we need to worry about paging?

r4vi (approved):

api/applications/views/denials.py
saruniitr: I think initially we are not expecting any matches, even if they match not a lot but eventually we may need paging.

r4vi: ok, it's fine for now given that most of the time it will return 0 matches

PR #689, opened 2021-01-12
LTD-478 MI Report: Cases by SLA

r4vi:
Allows reporting on cases/applications/licences based on their status and how
many of the 20 sla days have been used to process them

### DIFF
--- a/api/reports/queries/standard.py
+++ b/api/reports/queries/standard.py
@@ -339,3 +339,33 @@ from changes where
                    or audit_created_at between %(start_date)s::date and %(end_date)s::date
 order by audit_created_at desc;
 """
+
+SLA_CASES = """
+select cases_case.reference_code
+     , cases_case.submitted_at
+     , sla_days
+     , sla_remaining_days
+     , sla_updated_at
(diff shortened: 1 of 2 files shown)


api/reports/queries/standard.py
```diff
@@ -336,3 +338,32 @@
...
+              on cases_case.status_id = statuses_casestatus.id
+         left outer join applications_baseapplication on cases_case.id =
+                                                         applications_baseapplication.case_ptr_id
+         left outer join advice on cases_case.id = advice.case_id
+         left outer join advice_denial_reasons
+                         on advice.id = advice_denial_reasons.advice_id
+         left outer join licences_licence ll on cases_case.id = ll.case_id
+where
+    statuses_casestatus.status != 'draft'
+    and cases_case.submitted_at between %(start_date)s::date and %(end_date)s::date
```
r4vi, line 364: this should probably be between advice `created_at` instead of case created at, will also expose advice `created_at` field

saruniitr (approved):
adding new query, lgtm.

PR #701, opened 2021-01-18
Add further fields to required denials

richtier:


saruniitr (approved):
adding new fields, looks good to me.

PR #705, opened 2021-01-19
Ingest sanction lists

richtier:


api/external_data/management/commands/ingest_un_consolidated_list.py
```diff
@@ -0,0 +1,140 @@
...
+            except FileNotFoundError:
+                xml = function(*args)
+                with open(filename, 'wb') as f:
+                    f.write(xml)
+                return xml
+        return inner
+    return wrapper
+
+
+@cache_to_file('consolidated_un_list.xml')
```
richtier, line 35: this is for local debugging. will remove soon

api/external_data/management/commands/ingest_un_consolidated_list.py
```diff
@@ -0,0 +1,140 @@
...
+
+class Command(BaseCommand):
+
+    def rebuild_index(self):
+        connection = connections.get_connection()
+        connection.indices.delete(index=documents.SanctionDocumentType.Index.name)
+        documents.SanctionDocumentType.init()
+
+    def handle(self, *args, **options):
+        self.rebuild_index()
```
richtier, line 84: I will add arg to not do this every time. useful when changing the mapping

PR #707, opened 2021-01-21
LTD-486: Add name of signatory to the End user party

saruniitr:
## Change description

Add a new field to Party model to save name of the signatory signing the end user undertaking.


api/parties/serializers.py
```diff
@@ -86,6 +91,15 @@ def __init__(self, *args, **kwargs):
                 else:
                     self.fields.pop("role_other")
 
+    def validate(self, data):
```
r4vi, line 94: could you add error messages to `signatory_name_euu = serializers.CharField(allow_blank=True)`
and 
```
if party_type == PartyType.END_USER:
                self.fields["signatory_name_euu"].allow_blank = False
```
then you can remove `validate`?

saruniitr: yes but it is only applicable for "End user" type. For other types "third_party", "consignee" etc this is optional hence bypassed the default validators by making it optional and handled in `validate` method for the end user case.

r4vi (approved):

PR #721, opened 2021-02-24
Refactor document context generator

eadpearce:
Also fixes bug LTD-546 missing product description/name


api/letter_templates/templates/letter_templates/application_form.html
```diff
@@ -387,9 +387,7 @@ <h2 class="govuk-heading-m"> {% lcs 'GoodsDetailSummary.HEADING' %} {{ forloop.c
 										{{ good.has_identification_markings }}
 										<span class="govuk-hint">
 										<br>
-											{% if good.has_identification_markings %}
```
saruniitr, line 390: We currently capture serial number if there are identification markings, do we need to include those here?
Also this is only applicable for core firearms types - firearms, ammunition, components for firearm, components for firearm ammunition. Other types this is not applicable.

eadpearce: Yeah I wasn't sure what to do here. We could do this in a new ticket as we'll probably need design work to figure out the best way to display them. Could tie in with some of the new document type tickets in next sprint

richtier (approved):

PR #722, opened 2021-02-26
LTD-503: Handle firearms act question in case of RFD user

saruniitr:
## Change description

In this case the RFD certificate covers section 1 and 2 so the firearms act
question is mainly for section5 and user is really not asked to select
a section. So handle the validation messages and update the selected
section in this case.


api/goods/views.py
```diff
@@ -208,9 +213,28 @@ def post(self, request, *args, **kwargs):
...
+                    if (
+                        rfd_status
+                        and data["firearm_details"].get("is_covered_by_firearm_act_section_one_two_or_five") == "Yes"
+                    ):
+                        data["firearm_details"]["firearms_act_section"] = "firearms_act_section5"
+
         serializer = GoodCreateSerializer(data=data)
 
-        return create_or_update_good(serializer, data.get("validate_only"), is_created=True)
+        return create_or_update_good(serializer, data, is_created=True)
```
r4vi, line 237: is there any other code calling `create_or_update_good` but passing in `validate_only` which doesn't appear in data?

r4vi (approved):

api/goods/views.py
saruniitr: There are few other places in this file only, modified in all places

PR #723, opened 2021-03-04
Hotfix for exporting config

r4vi:
This removes username from RoutingRule natural key as these will probably not be
consistent across environments.

Updates FlaggingRule natural key to support updated model


saruniitr (approved):
looks good to me

PR #724, opened 2021-03-08
Add permissions for removing approval-blocking flags

eadpearce:


api/organisations/models.py
```diff
@@ -40,9 +40,7 @@ def get_user_relationship(self, user):
             raise NotFoundError({"user": "User does not belong to this organisation"})
 
     def get_users(self):
-        user_organisation_relationships = UserOrganisationRelationship.objects.filter(organisation=self).order_by(
-            "user__first_name"
-        )
```
r4vi, line 45: why did we get rid of the `order_by`?

api/flags/views.py
```diff
@@ -127,13 +139,17 @@ def put(self, request):
...
-                )
+                try:
+                    self._assign_flags(
+                        flags=serializer.validated_data.get("flags"),
+                        level=level.title(),
+                        note=serializer.validated_data.get("note"),
+                        obj=obj,
+                        user=request.user.govuser,
+                    )
+                except serializers.ValidationError as e:
```
r4vi, line 150: if `serializer.is_valid()` how do we get into a situation where a `serializers.ValidationError` is raised?

r4vi (approved):
some questions, no blockers

api/organisations/models.py
eadpearce: this function is only used in seed commands and it errors. complains that the user has no first_name property

api/flags/views.py
eadpearce: I thought I wasn't able to put it in the serializer validation because I didn't have access to some data but looking again I can probably move it

PR #730, opened 2021-03-15
LTD-719: Email reports using Django SMTP Email backend

saruniitr:
## Change Description

Add changes to create a background task which runs the query, generates report and saves it in a temporary file and then emails it to the Users using SMTP Emailbackend.

Sending emails is behind a feature flag as it is temporary which is by default turned off.
Recipient list is configurable using env variables.


api/reports/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+                        "report_filename": temp_file.name.split("/")[-1],
+                        "link_to_file": prepare_upload(f, is_csv=True),
+                    }
+                    response = notify_client.send_email(recipient, settings.LITE_REPORTS_EMAIL_TEMPLATE_ID, data)
+                    logger.info(response)
+        except errors.HTTPError as err:
+            raise err
+
+
+@background(queue=EMAIL_REPORTS_QUEUE, schedule=0)
```
r4vi, line 103: won't this be constantly running instead of 7am every day?

saruniitr: Forgot to change this, it is like this only for testing

api/reports/views.py
```diff
@@ -0,0 +1,13 @@
+from django.http import HttpResponse
+from rest_framework import status
+from rest_framework.views import APIView
+
+
+from api.reports.tasks import email_reports_task
+
+
+class EmailReportView(APIView):
```
r4vi, line 9: do we want to disable/remove the view?

r4vi (approved):

api/reports/views.py
saruniitr: ah, yes will disable the route in the urls.py

PR #732, opened 2021-03-17
Applications Finalised Report

r4vi:
- Flip "FOI objection" flag as lite stores it as "agrees_to_foi"
- add case status as if there is NLR then there is no licence and therefore status is blank


api/reports/queries/standard.py
```diff
@@ -490,12 +490,13 @@
...
 applications.licensee "Licensee",
 applications.intended_end_use "End use",
-applications.agreed_to_foi "FOI Objection flag",
+not applications.agreed_to_foi "FOI Objection flag",
 applications.foi_reason "FOI Objection",
 applications.submitted_at "Submitted Datetime",
 applications.case_status "Case closed reason",
 applications.last_closed_at "Closed Datetime",
-licences.reference_code "Licence reference",
+applications.reference_code "Licence reference",
```
r4vi, line 498: use ref code from case as if the join on licence doesn't happen then this is blank

saruniitr (approved):

PR #739, opened 2021-03-23
[LTD-699] SIEL cover letter

alixedi:
## [LTD-699](https://uktrade.atlassian.net/browse/LTD-699) Gov - I want to generate a licence covering letter

There are 2 parts to this -

1. Include references in the letter 8a12422e90f9bdf09e8c06c8f3d53215b6f624a2
2. Include case officer's name in the letter 8a12422e90f9bdf09e8c06c8f3d53215b6f624a2

In order to do (2), I had to do some digging into the models.

After these changes, we should be able to generate the letter specified in the ticket using LITE.

P.S. I have changed the standard ECJU letter template. This should be ok for the references but I am not sure about the case officer's name. LMK if there are any issues with this.

### DIFF
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -227,6 +227,13 @@ class Case(TimestampableModel):
                 verb=AuditType.REMOVE_CASE_FROM_ALL_USER_ASSIGNMENTS, action_object=case,
             )
 
+    def get_case_officer_name(self):
+        """
+        Returns the name of the case officer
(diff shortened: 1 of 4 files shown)


saruniitr (approved):
lgtm

r4vi:
code climate not picking up any diff-coverage for some reason, but feel free to merge - it's not a required check

PR #741, opened 2021-03-25
LTD-718: Add ELA statistics report

saruniitr:
## Change description

- Add query to generate the ELA statistics report. Part of the query string is generated in python and added to the query string because the list of teams the case is assigned can vary and difficult to build this completely in SQL.
- Update SLA update logic to use the through table `CaseQueue` instead of `CaseAssignment` as the latter is used mainly when users are assigned to a case. `CaseQueue` is used when a case is assigned to queue(s)
- A flag added to `Team` model to determine whether a team is part of ECJU or not as it is required to calculate SLA days in the report. Updated unit tests.
- Pin openpyxl to v3.0.6 as in v3.0.7 WINDOWS_EPOCH is not timezone aware causing issue when generating reports.


api/teams/models.py
```diff
@@ -10,7 +10,8 @@ def get_by_natural_key(self, name):
 
 class Team(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    name = models.CharField(max_length=60, default=None, blank=False, null=False, unique=True)
+    name = models.TextField(default=None, blank=False, null=False, unique=True)
```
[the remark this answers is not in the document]
saruniitr: updated

api/reports/helpers.py
```diff
@@ -0,0 +1,39 @@
+from api.teams import models as team_models
+from api.teams import serializers as team_serializers
+
+
+def get_teams():
+    return team_serializers.TeamSerializer(team_models.Team.objects.all(), many=True).data
+
+
+def get_team_query_options():
```
saruniitr, line 9: This function builds part of the sql query in python. This is done here because it is not possible to build this in sql without hard coding the team names. As both the number and names of teams depend on how they are configured we cannot really hard code in the query also so part of the query is built here.

alixedi: @saruniitr Sorry if the following comment reflects my lack of understanding -

It seems like we are parametrizing the query by string manipulation. Would it not be better to use [this](https://docs.djangoproject.com/en/3.1/topics/db/sql/#passing-parameters-into-raw)?

saruniitr: As you might have noticed this is just part of a query, the complete query string is `MI_ELA_STATISTICS` and this part is substituted at line 876 below in the file `api/reports/queries/standard.py`. We are not executing the query here, it is executed via a management command or a task that generates and emails report.
Also eventually this will be moved to Dataworkspace where the queries are executed.

api/teams/serializers.py
```diff
@@ -22,7 +23,21 @@ class TeamSerializer(serializers.ModelSerializer):
...
     class Meta:
         model = Team
-        fields = ("id", "name")
+        fields = "__all__"
+
+    def validate(self, attrs):
+        validated_data = super().validate(attrs)
+
+        if "part_of_ecju" not in validated_data:
+            raise serializers.ValidationError({"part_of_ecju": "Select yes if the team is part of ECJU"})
```
alixedi, line 41: A few things here. 

1. Why does `part_of_ecju` need to be a required field? Why can't we e.g. have a default value of `False`?
2. If it needs to be required for some reason, why can't we use the [required](https://www.django-rest-framework.org/api-guide/fields/#required) attribute that is built into DRF?
3. Again, if this field needs be required at the API level for some reason, we need to say this in the `ValidationError`.

api/teams/models.py
```diff
@@ -10,7 +10,8 @@ def get_by_natural_key(self, name):
 
 class Team(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    name = models.CharField(max_length=60, default=None, blank=False, null=False, unique=True)
+    name = models.TextField(default=None, unique=True)
```
alixedi, line 13: Are we sure we want to remove `null=False, blank=False` from here? Why are we having to do this? Are there teams with no name?

api/reports/queries/standard.py
```diff
@@ -814,3 +814,174 @@
     or audit_created_at between %(start_date)s::date and %(end_date)s::date
 order by audit_created_at desc;
 """
+
+
+MI_ELA_STATISTICS = """
```
alixedi, line 819: As someone new in the team, it is impossible for me to follow what is happening here or why. This means that if the code here misbehaves in the future (like all software does), it will be very hard to fix.

I understand the argument that this is temporary etc. but we have no idea exactly when DW integration would land and we should probably discuss how can we make this tractable at the very least. 

On that, I don't notice any tests for this query. Is this the plan for all reporting stuff?

api/teams/models.py
saruniitr: I am not sure why they initially set like that but I have removed them as Team name is mandatory. As per the documentation it is also not recommended to use `null` for string based fields

api/teams/serializers.py
saruniitr: 1. The requirement for this field has come up when building this report and it is not part of any feature so didn't assume any default value. Also there are teams configured in all the environments, assuming `False` makes existing ECJU teams also show up as `False` unless they are edited. Similarly assuming `True` makes non-ECJU teams part of ECJU which can create confusion to the users so no default value is assumed.
2. Yes `required` can be used but in that case since `null=True` in the model, objects can still be created, hence it is being handled in the validation method.
3. The error message content is provided by the content team, usually they provide content for any error/info messages. I have asked when making this change and already reviewed by them.

api/reports/queries/standard.py
saruniitr: We don't have tests at the moment for the reporting queries mainly because this is iterative and not the final set of fields required. We are continuously working and finalising the fields required so this list is volatile. Also trying to match up what we have in LITE with SPIRE and as we iterate we may get refined set of fields in future. We also need to setup data for these queries to pull. Since we run the tests on CI also then we need to expose the data as well. All of this is additional effort and considering these are temporary, tests are not added.

PR #742, opened 2021-03-29
LTD-745 API errors

alixedi:
This bug resulted from a lack of exception handling and was exacerbated by the poor code that looked for `error` in a string. 

In that, if a user puts `error` in the form for customising the text, not only did the API incorrectly thought there was an error in generating the document but in trying to find that (nonexistent) error, ran into a real `KeyError`.

### DIFF
--- a/api/letter_templates/helpers.py
+++ b/api/letter_templates/helpers.py
@@ -71,6 +71,10 @@ def format_user_text(user_text):
     return markdown_to_html(mark_safe(cleaned_text))
 
 
+class DocumentPreviewError(Exception):
+    pass
+
+
 def generate_preview(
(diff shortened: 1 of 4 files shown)


api/letter_templates/helpers.py
```diff
@@ -71,6 +71,10 @@ def format_user_text(user_text):
     return markdown_to_html(mark_safe(cleaned_text))
 
 
+class PreviewError(Exception):
```
saruniitr, line 74: Does it help if we rename it as `DocumentPreviewError`? preview available only for documents at the moment may be not required.

saruniitr (approved):

PR #744, opened 2021-03-30
Allow user advice to be created when team advice exists

eadpearce:


alixedi:
Would be nice to add some tests for before/after this change.

PR #745, opened 2021-03-31
NLR template

richtier:
image


saruniitr (commented):
Recently we added company registration number to the Licence outcome document, perhaps this is required in this template as well?
Please ask the content team to check the content, otherwise it looks good to me.

alixedi:
@saruniitr If we choose to include the company registration number, it will be on the base template under the addressee details. That should be a different PR.

saruniitr (approved):

PR #746, opened 2021-04-05
LTD-593-594: Collect phone number and website when registering an organisation

saruniitr:
## Change description

For commercial organisations we collect organisation phone number and site admin phone number whereas for individual orgs only contact number is collected. Website details are collected in both cases.
Updated tests and add new tests.

Noticed the field related to site in Organisation model was not referenced consistently. The field is called `primary_site` and in some places it is referenced as `site`, added changes to make it consistent.


api/organisations/views/users.py
```diff
@@ -28,12 +33,19 @@
 
 class UsersList(generics.ListCreateAPIView):
     authentication_classes = (SharedAuthentication,)
-    serializer_class = OrganisationUserListView
+
+    def get_serializer_class(self):
+        organisation = get_organisation_by_pk(self.kwargs["org_pk"])
+        if organisation.type == OrganisationType.INDIVIDUAL and hasattr(self.request.user, "govuser"):
```
saruniitr, line 39: We don't ask for contact number in case of Individual organisations hence separate serializer is used to omit this field.

api/organisations/views/users.py
```diff
@@ -45,18 +57,27 @@ def get_queryset(self):
         if _status:
             query.append(Q(relationship__status=UserStatuses.from_string(_status)))
 
+        values = ("baseuser_ptr_id", "first_name", "last_name", "email", "status", "role_name")
+        if (
+            organisation.type == OrganisationType.COMMERCIAL
+            or organisation.type == OrganisationType.HMRC
+            or hasattr(self.request.user, "exporteruser")
```
saruniitr, line 64: Interesting in case of exporters when adding members we also ask for contact number hence we include this for exporters.

api/addresses/models.py
```diff
@@ -17,6 +17,8 @@ class Address(models.Model):
         default=None, blank=True, null=True, max_length=256, help_text="Used for addresses not in the UK"
     )
     country = models.ForeignKey(Country, blank=False, null=False, on_delete=models.CASCADE)
+    phone_number = models.TextField(default="")
+    website = models.URLField(blank=True, default="")
```
alixedi, line 21: @saruniitr curious to learn why we are choosing to put `phone_number` and `website` in the `Address` model?

saruniitr: In the UI they are collected in the same form so I have reflected the same in data model also.

alixedi: Do you see any repercussions for this choice? 

E.g. A company can have 10 addresses (because of branches aka `Site`) but usually a single website. If we put the website with the `Address`, how are we going to ensure that that we are recording the website consistently for all the addresses?

P.S. I am concerned because generally the data model should **not** be informed by UI design. Happy to learn why this should be an exception.

api/addresses/models.py
```diff
@@ -17,6 +17,8 @@ class Address(models.Model):
         default=None, blank=True, null=True, max_length=256, help_text="Used for addresses not in the UK"
     )
     country = models.ForeignKey(Country, blank=False, null=False, on_delete=models.CASCADE)
+    phone_number = models.TextField(default="")
```
alixedi, line 20: Can we consider using e.g. [this](https://pypi.org/project/django-phonenumber-field/) or something similar here to ensure that we are indeed storing valid phone numbers?

api/addresses/serializers.py
```diff
@@ -52,6 +70,19 @@ def validate(self, data):
...
             validated_data["country"] = get_country("GB")
 
+        if "phone_number" in validated_data:
+            error = (
+                "Enter an organisation phone number"
+                if self.context.get("commercial", "commercial")
+                else "Enter a phone number"
+            )
+            if validated_data["phone_number"] == "":
+                raise serializers.ValidationError({"phone_number": error})
```
alixedi, line 80: Is this a required field? If so, we can set that up on the model which should get us this validation for free.

api/addresses/models.py
r4vi: if you do use phonenumberfield, I think the DIT standard is to use `E164` as the format

api/organisations/serializers.py
```diff
@@ -179,7 +190,7 @@ class OrganisationCreateUpdateSerializer(serializers.ModelSerializer):
         },
     )
     user = ExporterUserCreateUpdateSerializer(write_only=True)
-    site = SiteCreateUpdateSerializer(write_only=True)
+    primary_site = SiteCreateUpdateSerializer(write_only=True)
```
alixedi, line 193: When we hit this serializer (create or update) are we always updating the `primary_site` (and not any of the other sites)?

saruniitr:
It is being reworked in a different PR hence closing this

PR #750, opened 2021-04-13
LTD-590/591/593/594: User registration changes

saruniitr:
## Change description

For commercial organisations we collect organisation phone number and site admin phone number whereas for individual orgs only contact number is collected. Website (optional) details are collected in both cases.
Updated tests and add new tests.

Phone number and website fields for an organisation site are added to Organisation model because an organisation can have multiple sites but usually a single website. Also multiple sites can only be created from the exporter site but not from the internal site.


api/organisations/serializers.py
```diff
@@ -238,6 +244,17 @@ def validate_vat_number(self, value):
             return stripped_vat
         return value
 
+    def validate_phone_number(self, value):
```
saruniitr, line 247: This is a mandatory field but this method is added because content changes expect custom error message depending on whether the organisation is "commercial" or "individual".
`Phonenumberfield` handles validation of different phone number formats.

api/organisations/views/organisations.py
```diff
@@ -67,13 +67,16 @@ def post(self, request):
         """ Create a new organisation. """
         data = request.data.copy()
         validate_only = request.data.get("validate_only", False)
+        print(f"===> data: {data}")
```
saruniitr, line 70: will remove this

api/organisations/views/organisations.py
```diff
@@ -117,7 +121,10 @@ def put(self, request, pk):
                     data={"errors": Organisations.NO_PERM_TO_EDIT_NAME}, status=status.HTTP_400_BAD_REQUEST,
                 )
 
-        serializer = OrganisationCreateUpdateSerializer(instance=organisation, data=request.data, partial=True)
+        if "primary_site" in data:
+            data["site"] = data.pop("primary_site")
```
saruniitr, line 125: very likely that this is not necessary, checking in few other places before removing

alixedi (approved):
Brilliant stuff @saruniitr 👏 👏 👏 

LGTM!

PR #751, opened 2021-04-14
Add TokenIntrospectionMiddleware.

alixedi:


alixedi:
Closing this because after conversations with other devs, it seems we do all the SSO-dance in FE.

PR #754, opened 2021-04-26
Squashing migrations

agalamatis:
Excludes circular dependancies of the following apps:

1. organisations
2. goodstype
3. applications
4. external_data
5. cases
6. static_data.statuses

... and licenses that threw too many integrity errors to fix now.


alixedi (approved):
I checked the following -

✅  DB schema hasn’t changed
✅  `makemigrations` does not create any migration file

CI shows no tests break so LGTM.

PR #755, opened 2021-04-26
LTD-767: Update SIEL template and css to workaround the pdf rendering issue

saruniitr:
## Change description

Weasyprint cannot break and render an element properly if it spans over a
page. Fortunately the tables are getting rendered correctly - if any part of the
table row crosses a page then the complete row is getting moved to the next page
which is fine. However if the row in some cases if the row content is too long
and emdedded under a single div (in case of SIEL licence document it is the
terms and conditions section) then the content is getting truncated and not
being rendered.

To workaround this, the content is included in multiple top level div elements.
In this case atmost the content in last div gets pushed to the next page. This
is a workaround and may need further work if it fails in any other sections.


api/letter_templates/helpers.py
```diff
@@ -98,6 +99,12 @@ def generate_preview(
         if case:
             context = {**context, **get_document_context(case, additional_contact)}
 
-        return load_css(layout) + template.render(Context(context))
+        css_string = ""
+        if include_css:
+            css_string = load_css(layout)
+            if layout == "siel":
+                css_string = load_css("siel_preview")
```
alixedi, line 106: The fact that we are making a hard-coded exception at this level in the stack for `siel_preview` bothers me.

Just an idea - Given that the views for preview and pdf are different, can we load up the CSS in the view function and maybe pass it down the stack? (instead of this + `load_css` flag).

It might require having "preview" copies of existing CSS. Like we have put in for `siel_preview`. Except for other documents, `preview_<layout>.css` will be the same as `<layout>.css`.

api/cases/generated_documents/tests/test_generate_document.py
```diff
@@ -305,6 +309,20 @@ def test_get_document_preview_when_get_html_contains_error_string(self):
...
+        self.assertTrue(set(expected).issubset(css_files))
+
+    def test_generating_siel_licence_pdf_with_css(self):
+        resp = html_to_pdf("<div>Hello World !!</div>", "siel")
+        with NamedTemporaryFile(suffix=".pdf", delete=True) as tmp_file:
+            tmp_file.write(resp)
+
+        resp = html_to_pdf("<div>Hello World !!</div>", "siel_preview")
+        with NamedTemporaryFile(suffix=".pdf", delete=True) as tmp_file:
+            tmp_file.write(resp)
```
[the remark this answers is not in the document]
saruniitr: Tests whether the required css files are present or not, if they are not available then `html_to_pdf` fails which reminds me to actually check for that exception, I will update this.

api/cases/generated_documents/helpers.py
```diff
@@ -23,7 +23,7 @@
 def html_to_pdf(html: str, template_name: str):
     html = HTML(string=html)
     css = CSS(filename=get_css_location(template_name), font_config=font_config)
-    return html.write_pdf(stylesheets=[css], font_config=font_config)
+    return html.write_pdf(stylesheets=[css], font_config=font_config, presentational_hints=True)
```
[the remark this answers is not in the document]
saruniitr: [1] I will remove if doesn't give any other issues
[2] sure I will add a comment

api/letter_templates/helpers.py
saruniitr: I will check if it can be improved and update accordingly.

alixedi: Another (hopefully better) idea -

1) Rename `load_css` to `preview`.

2) Pass it down to the template as part of the context.

3) Inside the template - {% if preview %} <load-css-from-file> {% endif %}

P.S. This way we can keep the same templates and load or not load CSS depending on whether it's a preview or not.

alixedi (approved):
Don't forget to fix the tests.

PR #756, opened 2021-04-26
Emit audit logs

alixedi:
This does 3 things.

1) It adds `django-log-formatter-ecs`.
2) It adds a signal that emits Audit logs in ECS format.
3) Fixes a bunch of tests (and application logic) that was always broken but only got exposed because we started serializing all the `Audit` instances.

(2) looks like the following:

```
{
    "asctime": "2021-04-26 14:37:40,970",
    "levelname": "INFO",
    "message": "ExporterUser object (f0aaf9d1-1bae-4681-b2e3-318da86e46c8) created_final_advice StandardApplication object (0bc97090-4152-42b2-93c6-1bf9d5eefa6f) 0 minutes ago",
    "filename": "signals.py",
    "lineno": 19,
    "threadName": "MainThread",
    "name": "api.audit_trail.signals",
    "thread": 4626759168,
    "created": 1619444260.970651,
    "process": 33461,
    "processName": "MainProcess",
    "relativeCreated": 29715.826988220215,
    "module": "signals",
    "funcName": "emit_audit_log",
    "levelno": 20,
    "msecs": 970.6509113311768,
    "pathname": "/Users/alixedi/Projects/dit/lite-api/api/audit_trail/signals.py",
    "id": "b7e46ed4-469c-4bdf-9fc8-21235f9aa4aa",
    "created_at": "2021-04-26T14:37:40.967040+01:00",
    "user": {
        "id": "f0aaf9d1-1bae-4681-b2e3-318da86e46c8",
        "first_name": "Tiffany",
        "last_name": "Carlson",
        "type": "exporter"
    },
    "text": "created final advice.",
    "additional_text": ""
}
```

I would recommend a commit-wise review.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -62,7 +62,7 @@ pyexcel-ods3 = "*"
 Pillow = ">8.1.0,<8.2"
 django-phonenumber-field = "==5.0.0"
 phonenumbers = "~=8.12"
-django-log-formatter-ecs = "*"
+django-log-formatter-ecs = "==0.0.5"
 
 [requires]
 python_version = "3.7"
--- a/Pipfile
+++ b/Pipfile
@@ -62,6 +62,7 @@ pyexcel-ods3 = "*"
 Pillow = ">8.1.0,<8.2"
 django-phonenumber-field = "==5.0.0"
 phonenumbers = "~=8.12"
+django-log-formatter-ecs = "*"
 
 [requires]
 python_version = "3.7"
--- a/api/conf/settings.py
+++ b/api/conf/settings.py
@@ -307,10 +307,17 @@ if "test" not in sys.argv:
             "json": {
                 "class": "pythonjsonlogger.jsonlogger.JsonFormatter",
                 "format": "(asctime)(levelname)(message)(filename)(lineno)(threadName)(name)(thread)(created)(process)(processName)(relativeCreated)(module)(funcName)(levelno)(msecs)(pathname)",  # noqa
-            }
+            },
+            "ecs_formatter": {"class": "django_log_formatter_ecs.ECSFormatter",},
         },
-        "handlers": {"console": {"class": "logging.StreamHandler", "formatter": "json"}},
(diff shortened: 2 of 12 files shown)


Pipfile
```diff
@@ -62,6 +62,7 @@ pyexcel-ods3 = "*"
 Pillow = ">8.1.0,<8.2"
 django-phonenumber-field = "==5.0.0"
 phonenumbers = "~=8.12"
+django-log-formatter-ecs = "*"
```
r4vi, line 65: can we pin major version?

api/conf/settings.py
```diff
@@ -307,10 +307,17 @@
             "json": {
                 "class": "pythonjsonlogger.jsonlogger.JsonFormatter",
                 "format": "(asctime)(levelname)(message)(filename)(lineno)(threadName)(name)(thread)(created)(process)(processName)(relativeCreated)(module)(funcName)(levelno)(msecs)(pathname)",  # noqa
-            }
+            },
+            "ecs_formatter": {"class": "django_log_formatter_ecs.ECSFormatter",},
+        },
+        "handlers": {
+            "console": {"class": "logging.StreamHandler", "formatter": "json"},
```
r4vi, line 314: shouldn't we drop json formatter and use ecs formatter for everything now?

Pipfile
alixedi: Yes, I thought about it but didn't because the version ATM is `0.0.5` 😶

api/conf/settings.py
alixedi: I was wondering if it would be better off to turn off the current logger after we have verified that the ECS logger is working as expected? 

I am easy with this one too though if you would prefer to switch right away.

Pipfile
r4vi: maybe just lock to `0.0.5` and assume nothing about `0.0.6` being compatible

api/conf/settings.py
r4vi: ok that sounds good, no harm in doing it incrementally

r4vi (approved):

PR #759, opened 2021-04-28
LTD-716: Fix missing logo in generated pdf

saruniitr:
## Change description
 
`base_url` is not provided to weasyprint so it was not able to refer to `img` resource correctly resulting in this bug.


r4vi (approved):

alixedi:
About -

https://github.com/uktrade/lite-api/blob/cbb137ef4fbc8f04df52a74313d9d812a276f706/api/letter_templates/helpers.py#L92-L97

Variable substitution in user content is broken so we can do the following -


```
        context = {"include_digital_signature": include_digital_signature, "user_content": text}
```

And inside the template -

```
{{ user_content }}
```

r4vi:
> About -
> 
> https://github.com/uktrade/lite-api/blob/cbb137ef4fbc8f04df52a74313d9d812a276f706/api/letter_templates/helpers.py#L92-L97
> 
> Variable substitution in user content is broken so we can do the following -
> 
> ```
>         context = {"include_digital_signature": include_digital_signature, "user_content": text}
> ```
> 
> And inside the template -
> 
> ```
> {{ user_content }}
> ```

I can't understand why it wasn't doing that in the first place.

PR #762, opened 2021-05-07
LTD: Disable Spire product search view when ENABLE_SPIRE_SEARCH is disabled

saruniitr:
## Change description

Disable SPIRE product search view when `ENABLE_SPIRE_SEARCH` is disabled, currently the route is enabled.


eadpearce (approved):

api/conf/settings.py
```diff
@@ -290,10 +290,10 @@
 
     ELASTICSEARCH_PRODUCT_INDEXES = {"LITE": ELASTICSEARCH_PRODUCT_INDEX_ALIAS}
     ELASTICSEARCH_APPLICATION_INDEXES = {"LITE": ELASTICSEARCH_APPLICATION_INDEX_ALIAS}
+    SPIRE_APPLICATION_INDEX_NAME = env.str("SPIRE_APPLICATION_INDEX_NAME", "spire-application-alias")
+    SPIRE_PRODUCT_INDEX_NAME = env.str("SPIRE_PRODUCT_INDEX_NAME", "spire-products-alias")
```
alixedi, line 294: Moving these out of the `ENABLE_SPIRE_SEARCH` condition seems weird. 

Presume this is because there is some code that tries to access these without checking if `ENABLE_SPIRE_SEARCH is True`?

saruniitr: yes, there is a view where the index name is set to this.
I have included the view under this variable so it is not available when the variable is disabled.
Setting the index doesn't cause any issues though.

PR #765, opened 2021-05-14
Fix product details in application form

eadpearce:


api/letter_templates/templates/letter_templates/application_form.html
```diff
@@ -219,25 +223,26 @@ <h2 class="govuk-heading-m">Exhibition details</h2>
...
+							{% if good_on_application.good.description %}{{ good_on_application.good.description }}{% endif %}
 						</dd>
 					</div>
 					<div class="govuk-summary-list__row">
 						<dt class="govuk-summary-list__key">
 							{% lcs 'GoodsDetailSummary.CONTROLLED' %}
 						</dt>
 						<dd class="govuk-summary-list__value">
-							{{ good.is_controlled }} - {{ good.control_list_entries|join:", "|default_na }}
+							{{ good_on_application.good.is_controlled }} - {{ good_on_application.good.control_list_entries|join:", "|default_na }}
```
saruniitr, line 245: Depending upon the product is reviewed or not the list of control list entries vary. Do we need to check for that here?
Is this document generated only when the application is submitted or it is updated whenever there is a change to the application?

saruniitr (commented):
Requested clarifications.
The code changes looks good to me.

Because the way we render the template is modified, could you generate a licence and see if the correct formatting is maintained?

api/letter_templates/templates/letter_templates/application_form.html
eadpearce: Yeah this one is only generated when submitted so I think whether reviewed or not isn't relevant yet. As far as I know it's only generated once on application submission

saruniitr: :+1:  thanks.

api/letter_templates/tests/test_helpers.py
```diff
@@ -1,7 +1,25 @@
...
+            "case_context_test",
+            "ecju_queries_and_notes",
+            "ecju",
+            "nlr",
+            "refusal",
+            "siel",
+        ]
+        for layout in layouts:
+            # check it renders with no errors
+            assert generate_preview(layout=layout, case=case, text="")
```
alixedi, line 25: 1. `assert` is probably not needed if all you are testing is that no Exception is raised.
2. Might be better to use `pytest.mark.parametrize` for the layouts.

alixedi (approved):
Couple of comments that I will leave with you 🙏 🚀

api/letter_templates/tests/test_helpers.py
eadpearce: 1. Fair enough. can remove.
2. I wasn't sure if I could do that within a test case but can give it a go as that was what I planned to do originally

api/letter_templates/helpers.py
```diff
@@ -83,21 +83,18 @@ def generate_preview(
...
-        css_string = ""
-        if include_css:
-            css_string = load_css(layout)
-            if layout == "siel":
-                css_string = load_css("siel_preview")
-
-        return css_string + template.render(Context(context))
-    except (FileNotFoundError, TemplateDoesNotExist):
-        raise DocumentPreviewError(strings.LetterTemplates.PREVIEW_ERROR)
+    template_name = f"letter_templates/{layout}.html"
```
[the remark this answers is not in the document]
eadpearce: Yes actually if I remove that I can also remove another function that was only used in here

api/letter_templates/tests/test_helpers.py
eadpearce: I keep getting `TypeError: test_document_layouts() missing 1 required positional argument: 'layout'` whether I put the parametrize at the function or class level so I'm gonna leave this for now

PR #766, opened 2021-05-17
[LTD-682] Remove address from the ES query & update tests.

alixedi:
This PR does 2 things -

1. It removes addresses from the ES query for sanctions.
2. It removes the tests that were related to addresses.

It does not -

1. Remove the machinery that ingests addresses for sanctions and generate indices etc.
2. Fix the way tests are written - using weird class inheritance instead of pytest fixtures.
3. Strictly do what the spec says e.g. ATM "Jackson, Jeremy" and "Mr. Jeremy Jackson" will match "Jeremy Jackson". I think that (1) this is how it should be and (2) we can always do a second pass.

### DIFF
--- a/api/applications/helpers.py
+++ b/api/applications/helpers.py
@@ -208,7 +208,7 @@ def auto_match_sanctions(application):
         parties.append(item.party)
 
     for party in parties:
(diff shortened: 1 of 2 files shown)


agalamatis (approved):
looks good

PR #769, opened 2021-05-25
Add additional certificate expiry validation

eadpearce:


api/goods/libraries/save_good.py
```diff
@@ -4,26 +4,34 @@
 from api.core.helpers import str_to_bool
 
 
+def flatten_errors(errors, data):
+    pv_grading_errors = errors.pop("pv_grading_details", {})
```
saruniitr, line 8: nice refactoring

api/goods/serializers.py
```diff
@@ -163,10 +168,27 @@ class Meta:
             "deactivation_standard_other",
             "number_of_items",
             "serial_numbers",
+            # non model fields below included for validation only
+            "is_registered_firearm_dealer",
+            "document_on_organisation",
         )
 
+    def create(self, validated_data):
```
saruniitr, line 176: I am not sure about this. These documents are related to Organisation and not product specific so I think this should not be here.
Also we already do this here,
https://github.com/uktrade/lite-api/blob/master/api/applications/serializers/good.py#L225

api/goods/views.py
```diff
@@ -232,7 +233,9 @@ def post(self, request, *args, **kwargs):
                     ):
                         data["firearm_details"]["firearms_act_section"] = "firearms_act_section5"
 
-        serializer = GoodCreateSerializer(data=data)
+        serializer = GoodCreateSerializer(
+            data=data, context={"validate_only": validate_only, "organisation": get_request_user_organisation(request)}
```
saruniitr, line 237: `validate_only` doesn't seem to be used in the serializer.

saruniitr (commented):
Have a query on possible duplication, otherwise looks good to me.

api/goods/libraries/save_good.py
```diff
@@ -9,11 +9,18 @@ def create_or_update_good(serializer, data, is_created):
...
-            )
+            document_on_organisation_errors = firearm_errors.pop("document_on_organisation", None)
+            flattened_errors = {
+                **errors,
+                **firearm_errors,
+            }
+            if pv_grading_errors:
+                flattened_errors = {**flattened_errors, **pv_grading_errors}
+            if document_on_organisation_errors:
+                flattened_errors = {**flattened_errors, **document_on_organisation_errors}
```
alixedi: This can generally be simplified (generally = not just your changes) -

```
pv_grading_errors = errors.pop("pv_grading_details", {})
firearm_errors = errors.pop("firearm_details", {})
document_on_organisation_errors = firearm_errors.pop("document_on_organisation", {})
if firearm_errors:
    flattened_errors = {**errors, firearm_errors, **pv_grading_errors, **document_on_organization_errors}
...
```

api/goods/views.py
eadpearce: It's used by the nested `DocumentOnOrganisationSerializer`

api/goods/serializers.py
eadpearce: The problem is the part of the form with the document fields on it is part of the product flow so if we want to keep validation in the api this is where we need to put it. The validate_only flag that the frontend sends skips creation until the very end of the form where the document gets uploaded to the organisation not as part of the product.

saruniitr: Ok this is only to validate the document fields correct? in that case I think its ok.

saruniitr (approved):
looks good to me.
Definitely fixes the original issue of Expiry date for firearms certificate and refactors some code nicely.

PR #770, opened 2021-05-27
Rename tests

alixedi:
A bunch of tests haven't been running because the modules are called `tests_*.py` instead of `test_*.py`. So -

1. I have renamed them.
2. On renaming them, I found out that ~50 were failing
3. I have fixed a bunch of them except - 74aef3d1ad988ab357cee6ba76b7c92b7c258e9d & b8521b5f3ace6a77d1a544ec012497a83f5b5815
4. I have made tickets for these in JIRA. We need to review why these tests are failing.
5. We were getting `DeprecationWarning` for all the `assertEquals` in the code. I fixed that.
6. Finally, we were getting a warning b8521b5f3ace6a77d1a544ec012497a83f5b5815. I removed this. 

Coverage for the entire repo seems to have gone up by 14% 🌞 🎉 🚀 

P.S. Please review commit-wise.


agalamatis (approved):
Cool, add the SPIKE for the tests that still fail in JIRA.

eadpearce (approved):
removing those deprecation warnings will make looking through test results SO much easier!! 🎉 🎉 🎉

PR #774, opened 2021-06-04
[LTD-888] APM

alixedi:
This removes `python-json-logger` and simplifies logging a little. Supposed to help with APM.


saruniitr (approved):
lgtm, similar to that of lite-hmrc

PR #779, opened 2021-06-22
Update readme

dgnball:
Some things I stumbled over when setting up my dev environment (from a new starters point of view). Just a few changes and corrections to the suggested environment variables and readme.


local.env
```diff
@@ -1,7 +1,7 @@
 DEBUG=True
 
 # update username/password to match your local configuration
-DATABASE_URL=postgres://postgres:password@localhost:5432/lite-api
+DATABASE_URL=postgres://postgres:password@localhost:5462/lite-api
```
saruniitr, line 4: `5432` is the default port, I think we should keep the default values in this example env

saruniitr (changes requested):
One minor change requested, otherwise looks good to me.

local.env
dgnball: Ok, but 5462 is the port exposed by docker-compose. So if it's localhost it should be 5462 (if using pipenv). I can change this back, but I would at least want to mention this in the readme as I found this tripped me up.

saruniitr: I mean the default port for postgresql is 5432. Yes please mention as a comment that it can be a different as you faced some issue with this.

dgnball: Ok, I have addressed this.

saruniitr (approved):

PR #784, opened 2021-07-14
Add explicit init for report task.

alixedi:


saruniitr (approved):
verified that it is working locally

PR #786, opened 2021-07-22
[LTD-1055] Match system flags

alixedi:
This was meant to sync up the system flags across the CSV and the `enum`.

But I found that 3 (flaky) tests were breaking in master and so I also fixed them here.


test_helpers/clients.py
```diff
@@ -742,8 +742,8 @@ def create_mod_clearance_application(
 
         if case_type == CaseTypeEnum.EXHIBITION:
             application.title = "title"
-            application.required_by_date = "2021-07-20"
-            application.first_exhibition_date = "2022-08-19"
+            application.required_by_date = "2030-07-20"
+            application.first_exhibition_date = "2030-07-20"
```
alixedi, line 746: This was broken in master. Unrelated to this PR.

api/applications/tests/test_edit_application.py
```diff
@@ -475,7 +475,7 @@ def test_edit_exhibition_first_exhibition_date_draft_success(self):
         data = {
             "title": self.application.title,
             "required_by_date": self.application.required_by_date,
-            "first_exhibition_date": "2022-05-03",
+            "first_exhibition_date": "2030-08-03",
```
alixedi, line 478: This was broken in master. Unrelated to this PR.

api/flags/enums.py
```diff
@@ -106,4 +106,7 @@ class SystemFlags:
         ENFORCEMENT_THIRD_PARTY_MATCH,
         ENFORCEMENT_SITE_MATCH,
         ENFORCEMENT_ORGANISATION_MATCH,
+        SANCTION_UN_SC_MATCH,
+        SANCTION_OFSI_MATCH,
+        SANCTION_UK_MATCH,
```
alixedi, line 111: These were present in the `Enum` but missing from the list 🤷‍♂️

alixedi: (I know this is not an Enum FML)

api/staticdata/management/commands/seedflags.py
```diff
@@ -18,4 +18,5 @@ class Command(SeedCommand):
     @transaction.atomic
     def operation(self, *args, **options):
         csv = self.read_csv(FLAGS_FILE)
+        self.delete_unused_objects(Flag, csv)
```
alixedi, line 21: This is necessary IMO to keep the DB and the CSV in sync.

alixedi:
**Update**

It seems at least some of the system flags are being referenced in the codebase to using the `get_flag_name` function and a direct query -

https://github.com/uktrade/lite-api/blob/b468b753b1c31c2e42a5a81c0c107c9261ea2469/api/applications/views/countries.py#L158-L161

This means that removing what seemed at the time like [unused system flags](https://github.com/uktrade/lite-content/pull/285) breaks the tests.

We had a chat about this in the dev-huddle and the following suggestions were made - 

1. I can add the missing system flags to `SystemFlags`. This is the simplest way to solve our immediate problems.
2. We can remove `SystemFlags` and shift exclusively to using direct queries. When we need the list of all the system flags, we can perhaps use the fact that the system flags have the same team ID.

I am voting for (1) right now and perhaps a different ticket for (2). In that ticket, I would probably put in a `BooleanField` on the `Flags` model that explicitly identifies the system flags instead of overloading what `team_id` means.

api/flags/enums.py
alixedi: This is important for [this](https://github.com/uktrade/lite-api/blob/master/api/flags/views.py#L77) query to work.

saruniitr (approved):

PR #787, opened 2021-07-27
[LTD-1028] Stop SLA for cases with terminal status.

alixedi:
If you read carefully, we are supposed to only stop the counter for `FINALISED`, `CLOSED`, `WITHDRAWN`.

I am going with `CaseStatusEnum.terminal_statuses()` because that seems like a more exhaustive list, which means it includes `FINALISED`, `CLOSED`, `WITHDRAWN` amongst others.

### DIFF
--- a/api/cases/tests/test_sla.py
+++ b/api/cases/tests/test_sla.py
@@ -540,3 +540,32 @@ class SlaHmrcCaseTests(DataTestClient):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         response_data = response.json()["results"]["cases"]
         self.assertNotIn("sla_hours_since_raised", response_data[0])
+
+
+class TerminalCaseSlaTests(DataTestClient):
+    @mock.patch("api.cases.tasks.is_weekend")
+    @mock.patch("api.cases.tasks.is_bank_holiday")
+    def test_sla_not_update_for_terminal(
+        self, mock_is_weekend, mock_is_bank_holiday,
+    ):
+        mock_is_weekend.return_value = False
+        mock_is_bank_holiday.return_value = False
+        application = self.create_draft_standard_application(self.organisation)
+        case = self.submit_application(application)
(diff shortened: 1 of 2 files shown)


api/cases/tests/test_sla.py
```diff
@@ -540,3 +543,35 @@ def test_sla_hours_does_not_appear_on_other_cases(self):
...
+        case = self.submit_application(application)
+        _set_submitted_at(case, HOUR_BEFORE_CUTOFF)
+
+        results = update_cases_sla.now()
+        case.refresh_from_db()
+        self.assertEqual(results, 1)
+        self.assertEqual(case.sla_days, 1)
+        self.assertEqual(case.sla_remaining_days, STANDARD_APPLICATION_TARGET_DAYS - 1)
+
+        for case_status in CaseStatusEnum.terminal_statuses():
```
alixedi, line 569: It's a shame that I cannot parameterise a test that has to inherit from `DataTestClient` :(

api/cases/tests/test_sla.py
```diff
@@ -2,6 +2,7 @@
 from unittest import mock
 from unittest.mock import patch
 
+import pytest
```
saruniitr, line 5: this is not used

saruniitr (approved):

wkeeling (approved):

api/cases/tests/test_sla.py
alixedi: TY, Yes, wonder why the linter didn't complain.

api/cases/tests/test_sla.py
eadpearce: I noticed this the other day! Might be worth trying https://github.com/uktrade/lite-api/blob/49291035c015178cdd5bad9343409d4f48b50cef/api/common/tests/test_dates.py#L11

alixedi: Awesome! TY @eadpearce. I tried it but it complicates some of the setup so sticking to the loop for now.

PR #785, opened 2021-07-16
[LTD-1011] System Flags

alixedi:
This PR adds a technical document and fixes a bug that was exposed while writing this document.

PR #788, opened 2021-07-29
[LTD-1055] Revert changes

alixedi:
Reverts some changes introduced in #786 and #785. There are a couple of reasons for this -

1) Most important one - we cannot `delete_unused_object` for `Flag` model. The reason is that, unlike other "seeded" models, the `Flag` model comprises seeded (system) flags as well as user-defined flags. In the same model 🤦 

(this was my bad, I should have realised this but didn't)

2) The second one is more insidious. The order in which filters are applied in `FlagsListCreateView.get_queryset` matters. As a result, this doesn't work e.g. if I have `name=UK` and `include_system_flags=True`. The results will include flags that have `UK` in the name as well as ALL the system flags.

### DIFF
--- a/api/flags/views.py
+++ b/api/flags/views.py
@@ -74,7 +74,7 @@ class FlagsListCreateView(ListCreateAPIView):
         if case:
             flags = get_flags(get_case(case))
         else:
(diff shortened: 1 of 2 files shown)


dgnball (approved):

saruniitr (approved):
looks good to me

PR #792, opened 2021-08-06
LTD-1025: Expose Application specific endpoints to data workspace

saruniitr:
## Change description

Expose Applications specific endpoints to Data workspace.


api/data_workspace/urls.py
```diff
@@ -1,11 +1,14 @@
 from django.urls import path, include
 from rest_framework.routers import DefaultRouter
 
+from api.data_workspace import application_views
 from api.data_workspace import license_views
 
 app_name = "data_workspace"
 
 router = DefaultRouter()
+router.register("v1/baseapplication", application_views.BaseApplicationListView, basename="dw-baseapplications")
```
alixedi, line 10: Do we want to use singular or plural terms in URLs? Do we want to use hyphens? E.g. here, `baseapplication` vs `baseapplications` vs `base-applications`?

api/applications/migrations/0047_auto_20210809_1557.py
```diff
@@ -0,0 +1,53 @@
...
+                    options={"db_table": "applications_goodonapplication_control_list_entries",},
+                ),
+                migrations.AlterField(
+                    model_name="goodonapplication",
+                    name="control_list_entries",
+                    field=models.ManyToManyField(
+                        through="applications.GoodOnApplicationControlListEntry",
+                        to="control_list_entries.ControlListEntry",
+                    ),
+                ),
```
alixedi, line 50: I wonder what would happen to existing data if we do this? Have you checked?

agalamatis: Locally, yes, we paired on it. We checked that the DB schema is not altered by the migrations, and counted the data before and after the migration.

agalamatis (approved):
LGTM, before merging, (1) check the test that failed, (2) deploy as a branch on a dev env and check the DB schema and data before and after migrating there too. (3) run the functional tests and (4) have a look whether the related_name string chosen create any issue. I don't expect any of these to fail, but let's be thorough before merging.

saruniitr:
@agalamatis I will be splitting non `applications` endpoints to a separate PR and add tests for the endpoints.

alixedi (approved):
LGTM! Great work Arun 👏

PR #795, opened 2021-08-11
LTD-1025: Add remaining endpoints

saruniitr:
This adds the remaining endpoints required for `LTD-1025`.

I reckon it would be best to review commit-wise.

### DIFF
--- a/api/data_workspace/license_views.py
+++ b/api/data_workspace/license_views.py
@@ -47,3 +47,10 @@ class GoodOnLicenceList(viewsets.ReadOnlyModelViewSet):
     serializer_class = serializers.GoodOnLicenceViewSerializer
     pagination_class = LimitOffsetPagination
     queryset = models.GoodOnLicence.objects.all()
+
+
+class LicencesList(viewsets.ReadOnlyModelViewSet):
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    serializer_class = serializers.LicenceListSerializer
+    pagination_class = LimitOffsetPagination
+    queryset = models.Licence.objects.all()
--- a/api/data_workspace/license_views.py
+++ b/api/data_workspace/license_views.py
@@ -6,26 +6,25 @@ from api.cases.enums import CaseTypeEnum
 from api.core.authentication import DataWorkspaceOnlyAuthentication
 from api.open_general_licences.models import OpenGeneralLicence
 from api.open_general_licences.serializers import OpenGeneralLicenceSerializer
-from api.licences.models import Licence
-from api.licences.enums import LicenceStatus
-from api.licences.serializers.view_licence import LicenceListSerializer
+from api.licences import models, enums
+from api.licences.serializers import view_licence as serializers
 from api.staticdata.statuses.enums import CaseStatusEnum
 from api.staticdata.statuses.models import CaseStatus
 
 
 class LicencesListDW(viewsets.ReadOnlyModelViewSet):
     authentication_classes = (DataWorkspaceOnlyAuthentication,)
-    serializer_class = LicenceListSerializer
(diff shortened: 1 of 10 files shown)


api/data_workspace/license_views.py
```diff
@@ -6,26 +6,25 @@
...
-    serializer_class = LicenceListSerializer
+    serializer_class = serializers.LicenceListSerializer
     pagination_class = LimitOffsetPagination
     non_active_states = CaseStatus.objects.filter(status=CaseStatusEnum.SURRENDERED)
 
     queryset = (
         # OGL's are always hidden as we don't treat them as a licence
         # and they shouldn't be viewed from this endpoint
-        Licence.objects.exclude(
+        models.Licence.objects.exclude(
```
agalamatis, line 24: Why keep the exclude? We might need those objects in the future.

agalamatis: It will also make the class following (OpenGeneralLicenceListDW) redundant. We can sort out what is necessary and what not in DWs.

api/data_workspace/urls.py
```diff
@@ -15,8 +15,19 @@
 router.register(
     "v1/party-on-applications", application_views.PartyOnApplicationListView, basename="dw-party-on-applications"
 )
+router.register(
```
agalamatis, line 18: Doesn't router take a path prefix as a param? If it does, we can use that for versioning.

alixedi: It does but we will still have to do include `prefix="v1"` in every call to `register`. Might be better to write a small custom router that takes a prefix and applies it to all routes. But probably not in this PR.

api/data_workspace/license_views.py
alixedi: This was one of the OG endpoints. I didn't modify it because of the comment -

> OGL's are always hidden as we don't treat them as a licence and they shouldn't be viewed from this endpoint

If we do what we are doing elsewhere, DW will have to take ☝️ into consideration - a detail that should probably remain internal to LITE.

api/data_workspace/urls.py
agalamatis: The DefaultRouter class init, not its register super's method, but it apparently [does not support a prefix](https://github.com/encode/django-rest-framework/blob/master/rest_framework/routers.py).

But for versioning, we can do the following...
```
router = routers.DefaultRouter()
router.register(r'old_licence', OldLicenseViewset)
router1 = routers.DefaultRouter()
router1.register(r'new_licence', NewLicenseViewset)
urlpatterns = [
    path('v0/', include(router.urls)),
    path('v1/', include(router1.urls)),
```

Slam the old viewsets in a v0 and have ours in v1. Should we need to upgrade, we can continue this pattern.

api/data_workspace/license_views.py
agalamatis: See versioning in the urls.py below on how to keep these legacy views, while bringing in all data for ours and moving the deriving/selection/filtering logic in DWs.

alixedi: See - 1f34c1fee2a8fa18a301012c4dfda7801ce8768d

api/data_workspace/urls.py
alixedi: See - 1f34c1fee2a8fa18a301012c4dfda7801ce8768d

api/data_workspace/license_views.py
alixedi: See - 674f08dc839e5344126d78bfe9e5ddd3b469bf0a

agalamatis:
Awesome stuff 💯  waiting the tests.

agalamatis (approved):

PR #796, opened 2021-08-18
LTD-1025: Add missing endpoint and extend serializer fields

saruniitr:
## Change description

Add missing endpoint to expose `GoodOnApplicationControlListEntry` and `GoodControlListEntry` models to DW
Add additional fields to serializer to support reporting pipelines

### DIFF
--- a/api/applications/serializers/good.py
+++ b/api/applications/serializers/good.py
@@ -65,7 +65,7 @@ class GoodOnStandardLicenceSerializer(serializers.ModelSerializer):
         return data
 
 
-class GoodOnApplicationControlListEntryViewSerializer(serializers.ModelSerializer):
+class GoodOnApplicationControlListEntrySerializer(serializers.ModelSerializer):
     class Meta:
         model = GoodOnApplicationControlListEntry
         fields = "__all__"
--- a/api/applications/serializers/good.py
+++ b/api/applications/serializers/good.py
@@ -74,6 +74,8 @@ class GoodOnApplicationViewSerializer(serializers.ModelSerializer):
         model = GoodOnApplication
         fields = (
             "id",
+            "created_at",
+            "updated_at",
             "good",
             "application",
             "quantity",
--- a/api/applications/serializers/good.py
+++ b/api/applications/serializers/good.py
@@ -4,7 +4,12 @@ from rest_framework.relations import PrimaryKeyRelatedField
 
 from django.forms.models import model_to_dict
 
-from api.applications.models import GoodOnApplicationDocument, BaseApplication, GoodOnApplication
+from api.applications.models import (
+    GoodOnApplicationDocument,
+    BaseApplication,
+    GoodOnApplication,
+    GoodOnApplicationControlListEntry,
+)
 from api.audit_trail import service as audit_trail_service
 from api.audit_trail.enums import AuditType
 from api.audit_trail.serializers import AuditSerializer
@@ -58,6 +63,12 @@ class GoodOnStandardLicenceSerializer(serializers.ModelSerializer):
         if data["quantity"] > self.context.get("applied_for_quantity"):
             raise serializers.ValidationError({"quantity": strings.Licence.INVALID_QUANTITY_ERROR})
         return data
+
+
+class GoodOnApplicationControlListEntryViewSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = GoodOnApplicationControlListEntry
+        fields = "__all__"
 
 
 class GoodOnApplicationViewSerializer(serializers.ModelSerializer):
--- a/api/applications/tests/test_goods_type_on_application.py
+++ b/api/applications/tests/test_goods_type_on_application.py
@@ -40,9 +40,9 @@ class GoodsTypeOnApplicationTests(DataTestClient):
         response_data = response.json()["good"]
         self.assertEqual(response_data["description"], "Widget")
         self.assertEqual(response_data["is_good_controlled"], True)
-        self.assertEqual(
-            response_data["control_list_entries"], [{"rating": "ML1a", "text": get_control_list_entry("ML1a").text}]
-        )
+        self.assertEqual(len(response_data["control_list_entries"]), 1)
+        self.assertEqual(response_data["control_list_entries"][0]["rating"], "ML1a")
+        self.assertEqual(response_data["control_list_entries"][0]["text"], get_control_list_entry("ML1a").text)
         self.assertEqual(response_data["is_good_incorporated"], True)
 
     def test_create_goodstype_multiple_clcs_on_open_application_as_exporter_user_success(self):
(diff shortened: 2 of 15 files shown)


api/applications/serializers/good.py
```diff
@@ -60,20 +65,28 @@ def validate(self, data):
         return data
 
 
+class GoodOnApplicationControlListEntryViewSerializer(serializers.ModelSerializer):
```
alixedi, line 68: Should this be `GoodOnApplicationControlListEntrySerializer`?

saruniitr: no strong preference, will update this.

api/applications/serializers/good.py
```diff
@@ -60,20 +65,28 @@ def validate(self, data):
...
+
 class GoodOnApplicationViewSerializer(serializers.ModelSerializer):
     good = GoodSerializerInternal(read_only=True)
     good_application_documents = serializers.SerializerMethodField()
     unit = KeyValueChoiceField(choices=Units.choices)
     flags = serializers.SerializerMethodField()
     control_list_entries = ControlListEntrySerializer(many=True)
     audit_trail = serializers.SerializerMethodField()
-    is_good_controlled = KeyValueChoiceField(choices=GoodControlled.choices)
+    is_good_controlled = serializers.BooleanField()
```
alixedi, line 81: @saruniitr careful with this. You are changing the schema of the API and it might break the FE.

saruniitr: Reverted this as there is a chance it may break FE and handled it using transform in data-flow

api/compliance/tests/test_exporter_compliance_views.py
```diff
@@ -11,7 +11,6 @@
 
 class ComplianceExporterViewTests(DataTestClient):
     def compare_compliance_case_in_list(self, data, case, site):
-        self.assertEqual(data["id"], str(case.id))
```
saruniitr, line 14: this test started failing for some reason and it is not related to SIEL or changes in this PR. It was failing because of difference in `id` and don't know why. We could skip this test but removing this check is enough.

api/applications/tests/test_goods_type_on_application.py
```diff
@@ -40,6 +40,8 @@ def test_create_goodstype_on_open_application_as_exporter_user_success(self):
         response_data = response.json()["good"]
         self.assertEqual(response_data["description"], "Widget")
         self.assertEqual(response_data["is_good_controlled"], True)
+        for item in response_data["control_list_entries"]:
+            item.pop("id", None)
```
alixedi, line 44: I am easy on this but it seems like instead of popping keys we could just compare the ones we want to compare - 

```
self.assertEqual(response_data["control_list_entries"]["rating"], ...
self.assertEqual(response_data["control_list_entries"]["text"], ...
```

api/applications/tests/test_goods_type_on_application.py
```diff
@@ -53,6 +55,8 @@ def test_create_goodstype_multiple_clcs_on_open_application_as_exporter_user_suc
...
```
alixedi, line 59: In addition to my last comment, if we know the number of `control_list_entries` (b/c we assigned them in the fixture), might be a good idea to e.g.

```
assertEqual(len(response_data["control_list_entries"]), 1)
control_list_entry = response_data["control_list_entries"][0]
self.assertEqual(control_list_entry["rating"], ...)
```

api/data_workspace/tests/test_good_views.py
```diff
@@ -7,8 +10,9 @@
 class DataWorkspaceTests(DataTestClient):
     def setUp(self):
         super().setUp()
-        DataTestClient.create_good(description="Test good", organisation=self.organisation)
+        self.good = DataTestClient.create_good(description="Test good", organisation=self.organisation)
 
+    @override_settings(HAWK_AUTHENTICATION_ENABLED=False)
```
alixedi, line 15: This should not be needed. IIRC, Tests tun by default with `HAWK_AUTHENTICATION_ENABLED=False`.

api/data_workspace/tests/test_good_views.py
```diff
@@ -48,5 +52,27 @@ def test_goods(self):
 
         response = self.client.options(url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        options = response.json()["actions"]["OPTIONS"]
+        options = response.json()["actions"]["GET"]
+        self.assertEqual(tuple(options.keys()), expected_fields)
+
+    @override_settings(HAWK_AUTHENTICATION_ENABLED=False)
```
alixedi, line 58: This should not be needed. IIRC, Tests tun by default with `HAWK_AUTHENTICATION_ENABLED=False`.

api/goods/tests/test_create.py
```diff
@@ -60,6 +60,8 @@ def _assert_response_data(self, response_data, request_data):
 
     if request_data["is_good_controlled"] == True:
         self.assertEqual(response_data["is_good_controlled"]["key"], "True")
+        for item in response_data["control_list_entries"]:
+            item.pop("id", None)
```
alixedi, line 64: See earlier comments.

api/goods/tests/test_create.py
```diff
@@ -1128,6 +1130,8 @@ def test_when_creating_a_good_with_multiple_clcs_then_created_response_is_return
 
         response = self.client.post(URL, self.request_data, **self.exporter_headers)
         response_data = response.json()["good"]["control_list_entries"]
+        for item in response_data:
+            item.pop("id", None)
```
alixedi, line 1134: See earlier comments.

alixedi (commented):
Some minor comments.

api/data_workspace/tests/test_good_views.py
saruniitr: yes but locally when you run this could be enabled in `.env` and pytest loads it from that file, in that case this is required.

api/data_workspace/tests/test_good_views.py
saruniitr: same as above

api/applications/tests/test_goods_type_on_application.py
saruniitr: Ah I see what you mean, I tried to make current comparison work, will fix it.

api/data_workspace/tests/test_application_views.py
```diff
@@ -102,7 +105,31 @@ def test_dw_good_on_application_views(self):
...
+        response = self.client.options(self.good_on_applications_clc_entries)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        actual_keys = response.json()["actions"]["GET"].keys()
+        expected_keys = [
+            "id",
+            "goodonapplication",
+            "controllistentry",
+        ]
+        for key in expected_keys:
+            self.assertTrue(key in actual_keys)
```
alixedi, line 135: I think the following is better - 

```
self.assertEqual(tuple(actual_keys), expected_keys)
```

What you are testing is whether `expected_keys` is a subset of `actual_keys`.

(This applies to other places in this PR)

api/data_workspace/tests/test_good_views.py
alixedi: I think we should remove it in order to be consistent with other tests. Or add it in all the tests. IMO, remove it b/c it's easier.

alixedi (approved):
Couple of comments that I will leave with you. Looks good otherwise.

PR #797, opened 2021-08-23
Add DenialMatchOnApplicationListView

dgnball:


api/data_workspace/tests/test_application_views.py
```diff
@@ -106,3 +106,24 @@ def test_dw_good_on_application_views(self):
...
+        ]
+        for key in expected_keys:
+            self.assertIn(key, actual_keys)
+
+        response = self.client.options(self.denial_on_applications)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        actual_keys = response.json()["actions"]["OPTIONS"].keys()
+        expected_keys = ["id", "application", "denial", "category"]
+        for key in expected_keys:
+            self.assertIn(key, actual_keys)
```
saruniitr, line 129: duplicate of above snippet (L112-122)

api/data_workspace/tests/test_application_views.py
```diff
@@ -106,3 +106,24 @@ def test_dw_good_on_application_views(self):
         expected_keys = ["id", "application", "party", "flags"]
         for key in expected_keys:
             self.assertTrue(key in actual_keys)
+
+    @override_settings(HAWK_AUTHENTICATION_ENABLED=False)
+    def test_dw_denial_on_application_views(self):
+        response = self.client.options(self.denial_on_applications)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        actual_keys = response.json()["actions"]["OPTIONS"].keys()
```
saruniitr, line 114: `options` method returns actions for various http methods but we need tocheck the actions of `GET` here instead of `OPTIONS`. It is not correct in few places that were added initially.

api/data_workspace/urls.py
```diff
@@ -19,6 +19,9 @@
 router_v1.register(
     "party-on-applications", application_views.PartyOnApplicationListView, basename="dw-party-on-applications"
 )
+router_v1.register(
+    "denial-on-applications", application_views.DenialMatchOnApplicationListView, basename="dw-denial-on-applications"
```
saruniitr, line 23: Please add `match` also in the name eg `denial-match-on-applications` as they are referred as denial matches.

api/data_workspace/tests/test_application_views.py
alixedi: Might want to `assertEqual(actual_keys, expected_keys)`. ATM, this tests if `actual_keys` is a subset of `expected_keys`, which is probably not what we want.

Pithikos:
I wrongly added my work on this PR but reverted now with a force-push. So none of my work should be here now and will make a separate PR for that.

api/data_workspace/tests/test_application_views.py
dgnball: Ok, done.

api/data_workspace/tests/test_application_views.py
dgnball: Ok, done.

api/data_workspace/urls.py
dgnball: Ok, done.

api/data_workspace/tests/test_application_views.py
```diff
@@ -106,3 +108,16 @@ def test_dw_good_on_application_views(self):
         expected_keys = ["id", "application", "party", "flags"]
         for key in expected_keys:
             self.assertTrue(key in actual_keys)
+
+    @override_settings(HAWK_AUTHENTICATION_ENABLED=False)
```
saruniitr, line 112: sorry forgot about this earlier, this is not required as `HAWK_AUTHENTICATION_ENABLED` is `False` by default on CI so please remove to be consistent with other tests.

api/data_workspace/application_views.py
```diff
@@ -25,3 +25,10 @@ class PartyOnApplicationListView(viewsets.ReadOnlyModelViewSet):
     serializer_class = party.PartyOnApplicationViewSerializer
     pagination_class = LimitOffsetPagination
     queryset = models.PartyOnApplication.objects.all()
+
+
+class DenialMatchOnApplicationListView(viewsets.ReadOnlyModelViewSet):
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    serializer_class = denial.DenialMatchOnApplicationViewSerializer
+    pagination_class = LimitOffsetPagination
+    queryset = models.DenialMatchOnApplication.objects.all()
```
saruniitr, line 34: Please `order_by` `id` as we just found out it can cause issues with duplicates when running the pipeline.

dgnball: Done

api/data_workspace/tests/test_application_views.py
dgnball: Done

saruniitr (approved):

PR #799, opened 2021-08-24
LTD-1026 Add endpoint data_workspace/case-assignment

Pithikos:


api/cases/models.py
```diff
@@ -244,6 +244,10 @@ class Meta:
 
 
 class CaseAssignmentSla(models.Model):
+    """
+    Similar to CaseAssignment but tracks time for deadlines inside ECJU
+    and total time (based on ravi's Slack comments)
```
saruniitr, line 249: Usually we don't mention names etc, please update the comment.

api/data_workspace/tests/test_case_views.py
```diff
@@ -12,6 +13,25 @@ def setUp(self):
         self.queue.cases.add(self.case)
         CaseAssignmentSla.objects.create(sla_days=4, queue=self.queue, case=self.case)
 
+        # Create CaseAssignment
+        user = GovUserFactory(
+            baseuser_ptr__email="manos@manos.com",
+            baseuser_ptr__first_name="Manos",
+            baseuser_ptr__last_name="Sonam",
```
saruniitr, line 20: I think it is better to use test user details for names.

api/data_workspace/tests/test_case_views.py
```diff
@@ -12,6 +13,25 @@ def setUp(self):
...
+
+    def test_case_assignment(self):
+        url = reverse("data_workspace:dw-case-assignment-list")
+        expected_fields = {"user", "case"}
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        results = response.json()["results"]
+        self.assertGreater(len(results), 0)
+        self.assertEqual(set(results[0].keys()), expected_fields)
```
saruniitr, line 33: please use `self.client.options(url)` method to the structure of the response instead of `get` and check for all fields rather than subset.
Eg https://github.com/uktrade/lite-api/blob/master/api/data_workspace/tests/test_application_views.py#L108

saruniitr (changes requested):

Pithikos:
@saruniitr I think all things should be resolved now. Let me know if I missed something. Cheers!

api/data_workspace/tests/test_case_views.py
```diff
@@ -15,22 +15,36 @@ def setUp(self):
...
 
+        # Check GET
+        expected_fields = {"user", "case"}
+        expected_user_fields = {"id", "first_name", "last_name", "email", "team"}
         response = self.client.get(url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         results = response.json()["results"]
         self.assertGreater(len(results), 0)
         self.assertEqual(set(results[0].keys()), expected_fields)
+        self.assertEqual(set(results[0]["user"].keys()), expected_user_fields)
```
saruniitr: this can be removed as the `options` check is enough

api/data_workspace/tests/test_case_views.py
```diff
@@ -15,22 +15,36 @@ def setUp(self):
...
+
+        # Check OPTIONS
+        expected_fields = {"user", "case"}
+        expected_case_fields = {"type", "required", "read_only", "label"}
+        expected_user_fields = {"type", "required", "read_only", "label", "children"}
```
saruniitr, line 41: these should be the same as the ones in lines 30 above.
Also we should check the actions for `GET`.

`options = response.json()["actions"]["OPTIONS"]`
the above should be,
`options = response.json()["actions"]["GET"]`

api/data_workspace/tests/test_case_views.py
Pithikos: You mean remove the test case for GET?

Pithikos:
@saruniitr Still a bit unclear to me why we test in that specific way (mixing OPTIONS with GET in the same test case) but I pushed the changes now

api/data_workspace/tests/test_case_views.py
saruniitr: I mean the highlighted parts where (`# Check GET` block)
The reason is the it is redundant as we are checking with `options`.
The `options` method returns structure of the data for various actions which include `GET` hence we check for `GET` action in the response of `options`. Checking again using `client.get` is redundant, hope it is clear now.

saruniitr:
@Pithikos please see my comment earlier.
In these tests we are more concerned about the structure of the data (fields in the response) rather than the response itself hence we are using `options` method

api/data_workspace/tests/test_case_views.py
```diff
@@ -12,6 +13,25 @@ def setUp(self):
...
+            baseuser_ptr__first_name="John",
+            baseuser_ptr__last_name="Conam",
+            team=self.team,
+        )
+        CaseAssignment.objects.create(queue=self.queue, case=self.case, user=user)
+
+    def test_case_assignment(self):
+        url = reverse("data_workspace:dw-case-assignment-list")
+
+        expected_fields = {"user", "case"}
```
saruniitr, line 28: Sorry we need all fields of the model, currently `id, queue` are missing.
They are missing in the serializer also so please update the serializer and update the test.

api/data_workspace/case_views.py
```diff
@@ -2,14 +2,22 @@
...
     CaseTypeSerializer,
     CaseQueueSerializer,
 )
 
 
+class CaseAssignmentList(viewsets.ReadOnlyModelViewSet):
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    serializer_class = CaseAssignmentSerializer
+    pagination_class = LimitOffsetPagination
+    queryset = CaseAssignment.objects.all()
```
saruniitr, line 18: Also please `order_by("id")` as this affects the pipeline during execution.

saruniitr (commented):
last set of comments

saruniitr (approved):

PR #809, opened 2021-09-02
Support for system user type

wkeeling:


api/users/tests/test_serializers.py
```diff
@@ -0,0 +1,28 @@
...
+from test_helpers.clients import DataTestClient
+
+
+class BaseUserViewSerializerTests(DataTestClient):
+    def test_serialize_exporter(self):
+        serializer = BaseUserViewSerializer(instance=self.exporter_user)
+
+        data = serializer.data
+
+        self.assertEqual(data["type"], "exporter")
```
saruniitr, line 12: Since we have `UserType` enums, please use that for comparison.

saruniitr (approved):
one change requested, otherwise looks good to me.

api/users/tests/test_serializers.py
wkeeling: Thanks, now using the enum.

PR #812, opened 2021-09-08
Refactor existing tests (prework for LTD-1026 case-assignment endpoint)

Pithikos:
This PR refactors existing tests in order to remove an older hack introduced in the `AddressFactory.country` where it would always be set to GB (`factory.Iterator(Country.objects.filter(id="GB"))`).

This would make it impossible to create factories which use this as a subfactory since `queryset` would be empty in some occasions.


test_helpers/clients.py
```diff
@@ -163,6 +164,15 @@ def setUp(self):
         if settings.TIME_TESTS:
             self.tick = timezone.localtime()
 
+    def setup_exporter_headers(self, exporter_user):
```
eadpearce, line 167: these headers are already set here https://github.com/uktrade/lite-api/pull/812/files#diff-9ee85736e9813e7b51f9bb83bd8ff7543cb0258c3efe69d47bd8b6186aa437ceR147-R150

what's the reason for putting this in a separate method?

Pithikos: It's mainly so I can login as a different `exporter_user` in a test case. I needed it at some point while debugging

eadpearce: Oh I see! Yeah that's useful to have

eadpearce (approved):
lgtm

Pithikos:
There seems to be some flakiness with the tests after the last update from master. However the PR https://github.com/uktrade/lite-api/pull/814 which includes these changes does seem to be fine.

So I close this for now without merge.

PR #814, opened 2021-09-08
LTD-1026 Add missing field created_at

Pithikos:


api/core/decorators.py
```diff
@@ -113,6 +113,7 @@ def authorised_to_view_application(user_type: Union[Type[GovUser], Type[Exporter
     def decorator(func):
         @wraps(func)
         def inner(request, *args, **kwargs):
+            # TODO: Wrong function signature. Should be (self, request, ..)
```
dgnball, line 116: Remove this comment please

dgnball (changes requested):

api/core/decorators.py
Pithikos: Happy to, but what is the reasoning? This needs to be resolved at some point imo since it makes the code a bit hard to read and this is a reminder on how to do so

dgnball (approved):

PR #815, opened 2021-09-09
Pin elasticsearch version

eadpearce:
Fixes this error that was happening during deployment `elasticsearch.exceptions.UnsupportedProductError: The client noticed that the server is not a supported distribution of Elasticsearch`


ejayesh (approved):
approved, builds are now successful.

PR #820, opened 2021-09-20
Create endpoint for retrieving move case audit events

wkeeling:
This endpoint exposes 'move case' audit events in the following format:
```
        {
            "created_at": "2021-06-08T17:31:15.520183+01:00",
            "user": "00000000-0000-0000-0000-000000000001",
            "case": "b878c4f8-8f7c-45f9-ab06-143de04de3b5",
            "queue": "4dfc0d0b-c0c0-4f59-897b-b7d596bed930"
        },
```

### DIFF
--- a/api/data_workspace/audit_views.py
+++ b/api/data_workspace/audit_views.py
@@ -1,0 +1,46 @@
+from django.contrib.contenttypes.models import ContentType
+from rest_framework import viewsets
+from rest_framework.pagination import LimitOffsetPagination
+
+
+from api.audit_trail.enums import AuditType
+from api.audit_trail.models import Audit
+from api.cases.models import Case
+from api.core.authentication import DataWorkspaceOnlyAuthentication
+from api.data_workspace.serializers import AuditMoveCaseSerializer
+
+
+class AuditMoveCaseListView(viewsets.ReadOnlyModelViewSet):
+    """Expose 'move case' audit events to data workspace."""
+
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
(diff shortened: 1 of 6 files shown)


api/data_workspace/audit_views.py
```diff
@@ -0,0 +1,46 @@
...
+
+        # This returns a queryset of audit records for the "move case" audit event.
+        # For each record, it exposes the nested "queues" JSON property as a top
+        # level column called "queue" and splits into multiple rows when "queues"
+        # contains multiple entries.
+        # It also deals with the fact that the value of "queues" is sometimes an
+        # array of queue names but sometimes a single string.
+        return Audit.objects.raw(
+            """
+            with audit_move_case as (
```
eadpearce, line 31: I'm guessing this is written as SQL because it's easier than trying to write the query with django?

wkeeling: Yes it was tricky without the use of postgres specific functions. I needed to be able to split a single audit row into multiple rows when the row contained multiple queues in it's JSON payload, and that required the use of a `cross join` with `jsonb_array_elements()`. I didn't see a way to do this using the regular Django API.

eadpearce (approved):

api/data_workspace/audit_views.py
alixedi: I wonder if the SQL that splits the rows should live in DW? I think we already have `queues` in there.

The reason I am saying this is because it is hard to conceive this being used for anything but the reports and therefore it might make more sense for it to live in the codebase that is specifically for the reports.

That said, I am happy to listen to arguments on why this is a better place for it.

PR #822, opened 2021-09-23
Feature flag compliance case creation for SIELs

eadpearce:


api/compliance/helpers.py
```diff
@@ -38,15 +40,14 @@ def get_compliance_site_case(pk):
...
                 goods_on_application__application_id=case.id,
                 control_list_entries__rating__regex=COMPLIANCE_CASE_ACCEPTABLE_GOOD_CONTROL_CODES,
                 goods_on_application__licence__quantity__isnull=False,
-            ).exists()
-        ):
-            return False
-        return True
+            ).exists():
+                return True
+        return False
```
Pithikos, line 50: LGTM, just thinking that the complexity of the logic (all these nested ifs) and the lack of comments will make this hard to figure out in the future.

local.env
```diff
@@ -76,3 +76,5 @@ SIGNING_REASON=SIGNING_REASON
 PERMISSIONS_FINDER_URL='' # required for accessing exporter lite content, can be empty string
 
 ELASTICSEARCH_HOST=http://localost:9200
+
+FEATURE_SIEL_COMPLIANCE_ENABLED=False
```
Pithikos, line 80: Might be worth to have a comment here so we know if we need it on/off on local?

api/compliance/tests/test_compliance_creation.py
```diff
@@ -72,6 +74,22 @@ def tests_siel_good_control_code(self, control_code, exists):
 
         self.assertEqual(ComplianceSiteCase.objects.exists(), exists)
 
+    @override_settings(FEATURE_SIEL_COMPLIANCE_ENABLED=False)
+    def tests_siel_no_compliance_feature_flag_off(self):
```
Pithikos, line 78: What are we testing here? That the flag works or some specific behaviour from the flag? I think again some comment would be helpful

Pithikos (approved):
LGTM. I just think we generally have a culture of not adding comments. Maybe it's just me but personally I like to not having to dig into the code to tell what a function is doing, or alternatively having the code read like prose so that no more than a glance is needed

api/compliance/tests/test_compliance_creation.py
eadpearce: That no compliance case is created when the flag is off even when other conditions are met ie a SIEL with the ML22 CLC. I think that's more or less clear from the test code? But I could rename it

api/compliance/helpers.py
eadpearce: Yeah it's too complex and we should refactor but we'll get to that when we deal with compliance cases which I think is a way off yet, hence turning this off.

local.env
eadpearce: I can put a comment where it's set in settings

api/compliance/tests/test_compliance_creation.py
Pithikos: OK, I think it's just because I'm still not fully confident I understand what the flag is for. Is it a temporary feature flag for the UI? When will it be removed? Is it for certain environments only?

Clearly I lack the context for this and it might not matter much, I've approved anyway :)

api/conf/settings.py
```diff
@@ -426,3 +426,6 @@
     LITE_OPS_EMAIL = env.str("LITE_OPS_EMAIL")
     LITE_REPORTS_RECIPIENTS = env.list("LITE_REPORTS_RECIPIENTS", default=[])
     LITE_REPORTS_EMAIL_TEMPLATE_ID = env.str("LITE_REPORTS_EMAIL_TEMPLATE_ID", default="reports-email-template_id")
+
+# Controls whether a ComplianceSiteCase is automatically created when a SIEL licence is issued that has goods with certain control codes
```
saruniitr, line 430: this comment gives all the required context

saruniitr (approved):

api/conf/settings.py
Pithikos: @saruniitr This describes me the technical solution with different wording, doesn't really help. 

I think the key point that should be mentioned is what @eadpearce mentioned in the other comment. "..compliance cases which I think is a way off yet, hence turning this off." <- Something along the lines of "We had to turn off compliance cases and used this flag temporarily since we won't deal with such in the near future" would be much more helpful I think, and then the dev doesn't even have to dig in the code.

I think such comments are what is missing in the code.. giving some **reason** as to why certain things are there.

eadpearce: Yeah good point. I can reference the jira ticket number in the comment here for context

PR #823, opened 2021-09-24
LTD-1025: Identify if the application is amended or not

saruniitr:
## Change description

User can do minor and major edits and there is no database field for this.
Add changes to determine if an application is amended or not by using the
audit logs created when minor/major edits are done.


api/applications/serializers/standard_application.py
```diff
@@ -99,6 +105,32 @@ def get_denial_matches(self, instance):
         denial_matches = DenialMatchOnApplication.objects.filter(application=instance, denial__is_revoked=False)
         return DenialMatchOnApplicationViewSerializer(denial_matches, many=True).data
 
+    def get_amended_status(self, instance):
```
Pithikos, line 108: Not any issue with the function but just a small note.

When looking at the function name, I expect to get some sort of status as a string (e.g. "amended", "partialy-amended"). But the function returns a boolean, so wouldn't it make more sense to either rename it to `is_amended` or just add a small docstring mentioning what to expect (e.g. "Returns a boolean denoting if application has been amended at all by applicant user")?

saruniitr: yeah good point. I will rename it to `is_amended()`.

api/applications/serializers/standard_application.py
```diff
@@ -99,6 +105,32 @@ def get_denial_matches(self, instance):
...
+            )
+        )
+        # in case of doing major edits then the status is set as "Applicant editing"
+        # Here we are detecting the transition from "Submitted" -> "Applicant editing"
+        for item in audit_qs.filter(verb=AuditType.UPDATED_STATUS):
+            status = item.payload["status"]
+            if status["old"] == CaseStatusEnum.get_text(CaseStatusEnum.SUBMITTED) and status[
+                "new"
+            ] == CaseStatusEnum.get_text(CaseStatusEnum.APPLICANT_EDITING):
+                major_edit = True
```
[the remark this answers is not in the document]
saruniitr: querysets are lazy so it would've actually evaluated on the first iteration.
We could return from here but favouring readability I am doing at the end but maybe it is not that bad.
I will update this one also.

Pithikos (approved):

PR #824, opened 2021-09-24
Introduce is_ogd flag

wkeeling:
Introducing an `is_ogd` flag on to `Team` so that we can differentiate which teams belong to OGDs - which is something we need for reporting purposes.


api/teams/models.py
```diff
@@ -12,6 +12,7 @@ class Team(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     name = models.TextField(default=None, unique=True)
     part_of_ecju = models.BooleanField(default=None, null=True)
+    is_ogd = models.BooleanField(default=None, null=True)
```
alixedi, line 15: A default value of `None` for a boolean seems off because either a team is `OGD` or not and we would (should) know in either case.

(I know `part_of_ecju` does the same but this is a different conversation)

If you

wkeeling: Yes I wondered about the "Admin" team - which I think is `null` for `part_of_ecju` - but perhaps we'd just set to `False` for `is_ogd`.

wkeeling: `is_ogd` is now a mandatory field

api/teams/models.py
```diff
@@ -12,6 +12,7 @@ class Team(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     name = models.TextField(default=None, unique=True)
     part_of_ecju = models.BooleanField(default=None, null=True)
+    is_ogd = models.BooleanField(default=False)
```
Pithikos, line 15: Minor comment for future.. I think it's not consistent with what we have but it would be helpful if we populate `help_text` in some fields that can be not as straightforward. In this case it could e.g. be helpful if we explained that a team can be both in OGD and ECJU. But not a major thing for now

Pithikos (approved):
LGTM

api/teams/models.py
wkeeling: Thanks, good point - will update.

PR #826, opened 2021-09-27
Add Departments and DepartmentsSLA endpoints

dgnball:


wkeeling (approved):

api/cases/models.py
```diff
@@ -244,12 +244,20 @@ class Meta:
...
+    Keeps track of days passed since case assigned to a team
+    """
+
     sla_days = models.IntegerField()
     queue = models.ForeignKey(Queue, related_name="slas", on_delete=models.CASCADE)
     case = models.ForeignKey(Case, related_name="slas", on_delete=models.CASCADE)
 
 
-class DepartmentSLA(models.Model):
+class DepartmentSla(models.Model):
```
alixedi, line 256: I would like to keep `DepartmentSLA` please.

api/cases/models.py
```diff
@@ -244,12 +244,20 @@ class Meta:
 
 
 class CaseAssignmentSla(models.Model):
+    """
+    Keeps track of days passed since case assigned to a team
+    """
+
```
alixedi, line 250: Nice one 👌

alixedi (approved):
One comment that I will leave with you 🙏

api/cases/models.py
Pithikos: I changed it because `CaseAssignmentSla` uses that pattern.. also I'm thinking it might make it hard to be picked up by any meta-programming that relies on naming conventions. E.g. `DepartmentSLAFactory` or `DepartmentSLAView`

But not so fuss. Though if we change it, shouldn't we also change CaseAssignmentSla for consistency?

Pithikos:
@alixedi FYI I've changed both models to *SLA. No migrations required so this shouldn't affect anything else.

PR #827, opened 2021-09-28
Django management command for editing a team

wkeeling:


api/teams/management/commands/edit_teams.py
```diff
@@ -0,0 +1,57 @@
...
+                    converters = {
+                        "id": uuid.UUID,
+                        "name": str,
+                        "department": lambda x: uuid.UUID(x) if x else None,
+                        "part_of_ecju": lambda x: x.lower() == "true",
+                        "is_ogd": lambda x: x.lower() == "true",
+                    }
+
+                    for attr_name, attr_val in record.items():
+                        setattr(team, attr_name, converters[attr_name](attr_val))
```
alixedi, line 55: Might want to consider the slightly less verbose - 

https://github.com/uktrade/lite-api/blob/master/api/staticdata/management/SeedCommand.py#L95-L97

Naturally, you may need to replace `** attributes` with e.g. 

```
{
    "name": str(record["name"]),
    "department": record["department"] or None,
    "part_of_ecju": record["part_of_ecju"].lower() == "true",
    "is_ogd": record["is_ogd"].lower() == "true",
}
```

Notice that I didn't update ID. Are we sure we want to do that?

api/teams/management/commands/edit_teams.py
```diff
@@ -0,0 +1,57 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument("input_csv", type=str, help="Path to the input CSV file")
+        parser.add_argument(
+            "--dry", action="store_true", help="Print out what action will happen without applying any changes"
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options["dry"]
+        if dry_run:
+            self.stdout.write("Dry run only, no changes will be applied")
```
alixedi, line 34: Should we use a logger for this? Has the benefit that the logs will persist beyond a dev closing the ssh terminal.

api/teams/management/commands/edit_teams.py
wkeeling: Thanks. That looks cleaner, although it would assume that all model attributes are present in the CSV? The CSV currently allows you to supply only the attributes you want to update - so `record` may only contain a subset of `Team` attributes. Good spot on the `id` too - that shouldn't be in there.

api/teams/management/commands/edit_teams.py
wkeeling: Yes, good idea. Will change to a logger - thanks

api/teams/management/commands/edit_teams.py
alixedi: Good point about the attributes being "optional" in the CSV because of the fact that we are iterating on the columns. 

Happy for you to keep the OG code b/c it is less invasive than a complete update of the records.

wkeeling: Thanks, left the functionality, but switched to using an `update(**attrs)` as slightly less verbose.

alixedi (approved):

PR #833, opened 2021-10-07
LTD-1277: Limit ARS updates to specific product on application

saruniitr:
## Change description

Annual report summary (ARS) is specific to a product on application. If the same product
is used on other application then updating ARS in one application was updating the
value in the other application also. This has been fixed but the fix was failing
if the same product is added twice in the same application.

This is because API receives list of underlying product ids (Good) and we use them to
retrieve all matching products on application (GoodOnApplication). This is the current
behaviour because if the underlying product is same for all products on applications
then reviewing one good CLC entries automatically marks all of them as verified.
This is correct but it causes issues with Report summary.

This is fixed by taking the current product on application as input and limiting the
updates for this field only to this object.

An Audit log is also created whenever ARS value is updated.


api/goods/views.py
```diff
@@ -96,7 +96,32 @@ def post(self, request, case_pk):
             serializer.is_valid(raise_exception=True)
             old_control_list_entries = list(good.control_list_entries.values_list("rating", flat=True))
             old_is_controlled = good.is_good_controlled
-            serializer.save()
+
+            old_report_summary = good.report_summary
+            obj = serializer.save()
+
+            if request.data["report_summary"] != old_report_summary:
+                if str(good.id) != request.data.get("current_object", str(good.id)):
```
wkeeling, line 104: This looks like it puts back the old report_summary for those goods that don't represent the "current good"? Rather than putting the original report_summary back again, was wondering if it might be possible to tell the serializer not to change the report_summary in the first place, e.g.

```python
for good in self.get_queryset():
    serializer = self.get_serializer(good)

    # Ensure that non-current goods retain their report_summary
    if str(good.id) != request.data.get("current_object", str(good.id)):
        serializer.initial_data['report_summary'] = good.report_summary

    old_control_list_entries = list(good.control_list_entries.values_list("rating", flat=True))
    old_is_controlled = good.is_good_controlled
    serializer.save()

```
No idea if that would work - just an idea :)

saruniitr: modifying `initial_data` was updating all instances but I have updated similar to this idea i.e., changing the value conditionally before saving the serializer.

PR #835, opened 2021-10-08
Added updated status Audit endpoint

nahmed99:


api/data_workspace/audit_views.py
```diff
@@ -44,3 +44,14 @@ def get_queryset(self):
...
+
+class AuditUpdatedStatusListView(viewsets.ReadOnlyModelViewSet):
+    """Expose 'updated status' audit events to data workspace."""
+
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    serializer_class = AuditUpdatedStatusSerializer
+    pagination_class = LimitOffsetPagination
+
+    def get_queryset(self):
+        return Audit.objects.filter(verb=AuditType.UPDATED_STATUS)
```
saruniitr: please `order_by("id")` and add new line at the end of the file here and other places below.

api/data_workspace/audit_views.py
```diff
@@ -44,3 +44,14 @@ def get_queryset(self):
             """,
             {"verb": AuditType.MOVE_CASE, "action_type": content_type.pk, "target_type": content_type.pk},
         )
+
+
+class AuditUpdatedStatusListView(viewsets.ReadOnlyModelViewSet):
```
saruniitr, line 49: `AuditUpdatedCaseStatusListView` is better I think otherwise it is not clear whose status.
The audit verb cannot be changed but we can make this more meaningful.

api/data_workspace/tests/test_audit_views.py
```diff
@@ -55,3 +55,36 @@ def test_payload_multiple_queues(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        results = response.json()["results"]
+        self.assertEqual(len(results), 1)
+        self.assertEqual(tuple(results[0].keys()), expected_fields)
+        self.assertEqual(results[0]["status"], "Submitted")
+
+        response = self.client.options(self.url)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        options = response.json()["actions"]["OPTIONS"]
```
saruniitr, line 89: Please check the actions for `GET` here instead of `OPTIONS`

api/data_workspace/audit_views.py
saruniitr, line 57: Please also `order_by("id")`

wkeeling: @saruniitr would it be better to order by `created_at` as I think `id` is a UUID.

Pithikos (approved):
LGTM

PR #840, opened 2021-10-19
Add ability to delete user advice

wkeeling:


api/cases/views/views.py
```diff
@@ -249,6 +249,14 @@ def post(self, request, pk):
         """
         return post_advice(request, self.case, AdviceLevel.USER)
 
+    def delete(self, request, pk):
+        """
+        Delete user level advice on a case for the current user.
+        """
+        self.advice.filter(user=request.user.govuser).delete()
```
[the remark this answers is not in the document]
saruniitr: :+1:

api/audit_trail/payload.py
```diff
@@ -59,6 +59,7 @@ def format_payload(audit_type, payload):
     AuditType.CREATED_TEAM_ADVICE: strings.Audit.CREATED_TEAM_ADVICE,
     AuditType.CLEARED_TEAM_ADVICE: strings.Audit.CLEARED_TEAM_ADVICE,
     AuditType.CREATED_USER_ADVICE: strings.Audit.CREATED_USER_ADVICE,
+    AuditType.CLEARED_USER_ADVICE: "cleared user advice",
```
saruniitr, line 62: sorry one last thing, can we format it with more details, like including the user and advice type cleared.

saruniitr: ok, yeah that covers it

PR #845, opened 2021-11-01
Audit licence status changes

wkeeling:
We need to expose licence status changes to Data Workspace so that reports can see them. Reports consider licence status changes as "incidents" and they're reported on individually.

This PR ensures auditing of licence status changes happens centrally via a pre_save signal - so that no licence status changes will escape audit.

### DIFF
--- a/api/audit_trail/signals.py
+++ b/api/audit_trail/signals.py
@@ -1,10 +1,13 @@
 import logging
 
-from django.db.models.signals import post_save
+from django.db.models.signals import post_save, pre_save
 from django.dispatch import receiver
 
+from api.audit_trail import service as audit_trail_service
+from api.audit_trail.enums import AuditType
 from api.audit_trail.models import Audit
 from api.audit_trail.serializers import AuditSerializer
+from api.licences.models import Licence
 
 
 logger = logging.getLogger(__name__)
@@ -17,3 +20,21 @@ def emit_audit_log(sender, instance, **kwargs):
     text = str(instance)
     extra = AuditSerializer(instance).data
     logger.info(text, extra={"audit": extra})
+
+
+@receiver(pre_save, sender=Licence)
+def audit_licence_status_change(sender, instance, **kwargs):
+    """Audit a licence status change."""
+    try:
+        Licence.objects.get(id=instance.id, status=instance.status)
+    except Licence.DoesNotExist:
+        # The `pre_save` signal is called *before* the save() method is run and
+        # the `instance` is for the object that is about to be saved. In this case,
+        # if a `License` with the given parameters does not exist, it implies
+        # a change in status.
+        audit_trail_service.create_system_user_audit(
+            verb=AuditType.LICENCE_UPDATED_STATUS,
+            action_object=instance,
+            target=instance.case.get_case(),
+            payload={"licence": instance.reference_code, "status": instance.status},
+        )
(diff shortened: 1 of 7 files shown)


api/licences/libraries/hmrc_integration_operations.py
```diff
@@ -216,19 +216,11 @@ def _update_licence(validated_data: dict) -> str:
...
+        # Changing the licence status will trigger an auditlog entry
         change_status(send_status_change_to_hmrc=send_status_change_to_hmrc)
-        audit_trail_service.create_system_user_audit(
-            verb=AuditType.LICENCE_UPDATED_STATUS,
-            target=licence.case.get_case(),
-            payload={
-                "licence": licence.reference_code,
-                "status": hmrc_integration_action_to_licence_status.get(action),
-            },
-        )
```
wkeeling, line 231: This audit call has been removed as licence status changes are now handled centrally by `Licence.save()`

api/audit_trail/signals.py
```diff
@@ -17,3 +20,17 @@ def emit_audit_log(sender, instance, **kwargs):
     text = str(instance)
     extra = AuditSerializer(instance).data
     logger.info(text, extra={"audit": extra})
+
+
+@receiver(pre_save, sender=Licence)
+def audit_licence_status_change(sender, instance, **kwargs):
+    """Audit a licence status change."""
+
+    if not Licence.objects.filter(id=instance.id or "", status=instance.status).exists():
```
Pithikos, line 29: `instance.id or ""` this feels a bit smelly to me having it in there in the query.. Not sure I'm reading this correctly but I think a better clearer pattern would be to use `Licence.objects.get` and catch `Licence.DoesNotExist` or whatever to avoid the surströmming :fish:

Pithikos (approved):
LTGM. Just a readability comment on that query

api/audit_trail/signals.py
wkeeling: Thanks - yes you're right, catching `Licence.DoesNotExist` is cleaner here (I'd moved the `filter()` code from somewhere else where it made more sense). Now updated.

api/audit_trail/signals.py
```diff
@@ -17,3 +20,18 @@ def emit_audit_log(sender, instance, **kwargs):
...
     logger.info(text, extra={"audit": extra})
+
+
+@receiver(pre_save, sender=Licence)
+def audit_licence_status_change(sender, instance, **kwargs):
+    """Audit a licence status change."""
+    try:
+        Licence.objects.get(id=instance.id, status=instance.status)
+    except Licence.DoesNotExist:
+        # Only create an audit record when the licence status has actually changed.
```
alixedi, line 31: Minor comment - it might be hard to decipher why and how this works e.g. 6 months from now. Would propose adding something along the following lines to the comment -

```
`pre_save` signal is called *before* the save() method is run and the `instance` is for the object that is about to be saved. In this case, if a `License` with the given parameters does not exist, it implies a change in status.
```

alixedi (approved):
One minor comment that I will leave with you.

api/audit_trail/signals.py
wkeeling: Thanks - yes definitely a good idea. Have updated.

Pithikos: I'd counter and say that it's unnecessary to duplicate the Django documentation. IMO we should focus on adding comments for **esoteric logic** that it's hard to find anywhere if at all and comments that can **reason** about the code.

If there would be any improvement of the comments it would be making the docstring a bit more zoomed out so it reads like a Harry Potter book. E.g. "Record any licence change in the audit table" but I think the snippet is so tiny and not that cryptic so that it's not that big of a deal

PR #846, opened 2021-11-02
LTD-1171: Add endpoint to Countersign advice

saruniitr:


api/audit_trail/payload.py
```diff
@@ -127,4 +127,5 @@ def format_payload(audit_type, payload):
     AuditType.LICENCE_UPDATED_STATUS: strings.Audit.LICENCE_UPDATED_STATUS,
     AuditType.DOCUMENT_ON_ORGANISATION_CREATE: "added {document_type} '{file_name}' to organization",
     AuditType.REPORT_SUMMARY_UPDATED: "updated ARS for {good_name} from {old_report_summary} to {report_summary}",
+    AuditType.COUNTERSIGN_ADVICE: "countersigned advice",
```
saruniitr, line 130: should we add more info here?

api/audit_trail/migrations/0004_auto_20211101_1242.py
```diff
@@ -0,0 +1,246 @@
+# Generated by Django 3.1.8 on 2021-11-01 12:42
+
+import api.audit_trail.enums
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
alixedi, line 7: Is this migration the result of adding an `AuditType`?

api/audit_trail/payload.py
saruniitr: The details of the user who countersigned is already included in the audit. It comes out as,
<user> countersigned advice

api/cases/models.py
```diff
@@ -388,6 +388,14 @@ class Advice(TimestampableModel):
     pv_grading = models.CharField(choices=PvGrading.choices, null=True, max_length=30)
     # This is to store the collated security grading(s) for display purposes
     collated_pv_grading = models.TextField(default=None, blank=True, null=True)
+    comments = models.TextField(
```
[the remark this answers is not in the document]
saruniitr: will rename it and as discussed not going to add `null=True` as documentation says it is not recommended for string based fields

api/cases/models.py
```diff
@@ -388,6 +388,14 @@ class Advice(TimestampableModel):
...
+        blank=True,
+        default="",
+        help_text="Reasons provided by the countersigner when they agree/disagree with the advice during countersigning",
+    )
+    countersigned_by = models.ForeignKey(
+        GovUser, on_delete=models.DO_NOTHING, related_name="countersigned_advice", blank=True, null=True
```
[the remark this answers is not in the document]
saruniitr: ok will change it to `countersigned_by`

api/audit_trail/migrations/0004_auto_20211101_1242.py
saruniitr: yes

api/applications/serializers/advice.py
```diff
@@ -153,6 +157,20 @@ def _footnote_fields_setup(self):
                 self.initial_data[i]["footnote_required"] = None
 
 
+class AdviceUpdateListSerializer(serializers.ListSerializer):
```
alixedi, line 160: `CountersignAdviceListSerializer` would be a better name and to be consistent, the one below can be called `CountersignAdviceSerializer`.

api/cases/views/views.py
```diff
@@ -931,3 +931,32 @@ def put(self, request, pk):
...
+        serializer = CountersignedAdviceUpdateSerializer(
+            Advice.objects.filter(id__in=advice_ids), data=data, many=True, partial=True
+        )
+        if not serializer.is_valid():
+            return JsonResponse({"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        serializer.save()
+        audit_trail_service.create(
+            actor=request.user, verb=AuditType.COUNTERSIGN_ADVICE, target=case, payload={},
+        )
```
alixedi, line 960: I think the payload could contain the `advice_ids` that were countersigned.

alixedi (approved):
Couple of comments I will leave with you. Awesome stuff 🚀

PR #847, opened 2021-11-02
LTD-1382 Add field Good.no_document_comments

Pithikos:
Add field `no_document_comments` and make it required when no document is supplied by applicant


api/goods/models.py
```diff
@@ -107,6 +99,13 @@ class Good(TimestampableModel):
     flags = models.ManyToManyField(Flag, related_name="goods")
     is_document_available = models.BooleanField(default=None, null=True)
     is_document_sensitive = models.BooleanField(default=None, null=True)
+    no_document_comments = models.TextField(
+        default=None,
+        blank=True,
+        null=True,
+        max_length=2000,
```
saruniitr, line 106: You don't need to specify max length for `TextField`, please remove, also we don't want to limit it to the user.

api/goods/tests/test_document_reasons.py
```diff
@@ -17,7 +17,7 @@ def setUp(self):
         self.document_sensitivity_url = reverse("goods:good_document_sensitivity", kwargs={"pk": self.good.id})
 
     @parameterized.expand(
-        [[{"is_document_available": "yes"}], [{"is_document_available": "no"}],]
+        [[{"is_document_available": "yes"}], [{"is_document_available": "no", "no_document_comments": "yada yada"}]]
```
saruniitr, line 20: Please add sensible test data instead of above
eg datasheet not available or similar

api/goods/tests/test_document_reasons.py
```diff
@@ -34,9 +34,22 @@ def test_document_available_to_upload_failure(self):
...
+        # Missing no_document_comments
+        data = {"is_document_available": "no"}
+        response = self.client.post(self.document_availability_url, data, **self.exporter_headers)
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        data = {"is_document_available": "no", "no_document_comments": ""}
+        response = self.client.post(self.document_availability_url, data, **self.exporter_headers)
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+        # Given no_document_comments
+        data = {"is_document_available": "no", "no_document_comments": "yada yada yada"}
```
saruniitr, line 48: same as above

saruniitr (changes requested):

api/goods/serializers.py
```diff
@@ -552,7 +543,7 @@ def update(self, instance, validated_data):
             firearm_details = validated_data.get("firearm_details")
             if firearm_details:
                 instance.firearm_details = GoodCreateSerializer._update_firearm_details(
-                    firearm_details=firearm_details, instance=instance.firearm_details,
```
saruniitr, line 555: I think we need `no_document_comments` in `GoodCreateSerializer` above at line 350

api/goods/serializers.py
```diff
@@ -47,21 +47,12 @@ class PvGradingDetailsSerializer(serializers.ModelSerializer):
     issuing_authority = serializers.CharField(allow_blank=False, allow_null=False)
     reference = serializers.CharField(allow_blank=False, allow_null=False)
     date_of_issue = serializers.DateField(
-        allow_null=False,
-        error_messages={"invalid": "Enter the products date of issue and include a day, month, year."},
+        allow_null=False, error_messages={"invalid": "Enter the products date of issue and include a day, month, year."}
```
wkeeling, line 50: Minor grammar: "products" should be "product's"

api/goods/views.py
```diff
@@ -267,6 +267,10 @@ def post(self, request, *args, **kwargs):
 
 
 class GoodDocumentAvailabilityCheck(APIView):
+    """
+    Check document is attached to application good/product
+    """
+
     authentication_classes = (ExporterAuthentication,)
 
     def get(self, request, pk):
```
wkeeling, line 276: Wondering whether this be a `post` endpoint now that it's going to be receiving a block of text for the `no_document_comments`?

api/goods/serializers.py
Pithikos: Good catch. This was just the linter, but I'll change it since I don't want to be git-blamed for poor grammar :')

api/goods/serializers.py
Pithikos: True, I think this is not caught by the test coverage since we actually don't seem to call serializer's `is_valid()` at least in the step I was working on

saruniitr (approved):
lgtm, thanks

PR #848, opened 2021-11-02
Endpoint for exposing licence status updates to Data Workspace

wkeeling:


api/data_workspace/serializers.py
```diff
@@ -79,3 +79,29 @@ def get_case(self, instance):
...
+        return None
+
+    def get_case(self, instance):
+        return instance.target_object_id or None
+
+    def get_licence(self, instance):
+        return instance.action_object_object_id or None
+
+    def get_status(self, instance):
+        status = instance.payload["status"].lower()
```
saruniitr, line 106: it is up to you, perhaps we can get rid of `status` and return the expression result as is?

saruniitr (approved):

api/data_workspace/serializers.py
```diff
@@ -79,3 +79,29 @@ def get_case(self, instance):
...
+        model = Audit
+        fields = ("created_at", "user", "case", "licence", "status")
+
+    def get_user(self, instance):
+        if instance.actor:
+            return instance.actor.pk
+        return None
+
+    def get_case(self, instance):
+        return instance.target_object_id or None
```
Pithikos, line 100: What if `object_id == 0`? It will return `None`. I think we should do all this sort of transformation at the DW tbh or else 
there's a chance that data will change during transit

api/data_workspace/serializers.py
wkeeling: Do you mean return the whole payload rather than extracting the status out of it?

api/data_workspace/serializers.py
wkeeling: The `object_id` here is a UUID so won't be 0 - but yes performing the transformation in DW is probably a better idea since the pipelines do provide for that. Have updated - thanks.

api/data_workspace/serializers.py
wkeeling: Have updated to `return instance.payload["status"].lower()`

PR #849, opened 2021-11-02
Fix test-suite seed: don't seed when we use --reuse-db

Pithikos:
- Stops seeding unit-tests when we use pytest's `--reuse-db`
- Fix timezone failing tests.

### DIFF
--- a/api/organisations/tests/test_organisations.py
+++ b/api/organisations/tests/test_organisations.py
@@ -40,20 +40,28 @@ class GetOrganisationTests(DataTestClient):
         response_data = next(data for data in response.json()["results"] if data["id"] == str(organisation.id))
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(
-            response_data,
-            {
-                "id": str(organisation.id),
-                "name": organisation.name,
-                "sic_number": organisation.sic_number,
-                "eori_number": organisation.eori_number,
-                "type": generate_key_value_pair(organisation.type, OrganisationType.choices),
-                "registration_number": organisation.registration_number,
-                "vat_number": organisation.vat_number,
-                "status": generate_key_value_pair(organisation.status, OrganisationStatus.choices),
-                "created_at": date_to_drf_date(organisation.created_at),
-            },
-        )
+
+        expected_fields = {
+            "id",
+            "name",
+            "sic_number",
+            "eori_number",
+            "type",
+            "registration_number",
+            "vat_number",
+            "status",
+            "created_at",
+        }
+        assert response_data.keys() == expected_fields
+        assert response_data["id"] == str(organisation.id)
+        assert response_data["name"] == organisation.name
+        assert response_data["sic_number"] == organisation.sic_number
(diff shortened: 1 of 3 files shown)


api/organisations/tests/test_organisations.py
```diff
@@ -38,16 +38,27 @@ def test_list_organisations(self):
...
+            "name",
+            "sic_number",
+            "eori_number",
+            "type",
+            "registration_number",
+            "vat_number",
+            "status",
+            "created_at",
+        }
+        assert response_data.keys() == expected_fields
```
wkeeling, line 52: I wonder whether it would be enough to assert the number of keys here. That would then eliminate the need for the `expected_fields` set and the duplication of the field names. So e.g.
```
assert len(response_data) == 9
assert response_data["id"] == str(organisation.id)
assert response_data["name"] == organisation.name
... etc ...
```

Pithikos: I see your point.. But I guess now the test cases are covered totally decoupled (1) check keys exist, (2) check values of those keys. Your logic makes sense but it will mangle the two things together

test_helpers/clients.py
```diff
@@ -120,6 +121,16 @@ class DataTestClient(APITestCase, URLPatternsTestCase):
     def setUpClass(cls):
         """ Run seed operations ONCE for the entire test suite. """
         if not Static.seeded:
+
+            # HACK: Don't seed if we already seeeded and use --reuse-db or similar
```
eadpearce, line 125: good job spotting this 👍

api/organisations/tests/test_organisations.py
wkeeling: I guess I'm just a bit hung up on having to have the key names specified twice. I wonder then if we could keep the expected values together with the keys, something like:

```
expected = {
    "id": str(organisation.id),
    "name": organisation.name,
    "sic_number": organisation.sic_number,
    "eori_number": organisation.eori_number,
    "type": generate_key_value_pair(organisation.type, OrganisationType.choices),
    "registration_number": organisation.registration_number,
    "vat_number": organisation.vat_number,
    "status": generate_key_value_pair(organisation.status, OrganisationStatus.choices),
    "created_at": None,
}

assert response_data.keys() == expected.keys()

for key, value in expected.items():
    if value is not None:
        assert response_data[key] == value
    else:
        assert response_data[key]

```

wkeeling (approved):

api/organisations/tests/test_organisations.py
Pithikos: Don't see why you're so hung up on it lol. We all have IDEs with multiple cursors B)

Sure, it's more verbose, but it's a test that any junior dev can read. You can think of it as two tests smashed together though:  `test_we_get_the_expected_keys` and `test_values_are_correct`. If you had those tests separate, then you would have that "duplication" still and there's no way around it. I'm pretty sure that's happening already in some tests.

In the end of the day though the only reason I kept my own version over yours was because it covered the `created_at` while if I recall correctly your fix was missing that check.

PR #850, opened 2021-11-03
Autoload all factories to use directly in pytest tests

Pithikos:
Loads all factories as `pytest.fixture` so they can be used directly without having to manually import

```
class TestTeam():
  def test_my_team(team_factory):
      my_team = team_factory()
```

NOTE: This will only work with pytest tests, you will still need to load manually the factories in existing vanilla Django tests.


wkeeling (approved):
Looks good to me

PR #851, opened 2021-11-03
WIP

alixedi:
This is massively under-cooked atm but contains a prototype of how I would want to set a system flag when the advice for a given case has covered all the destinations.

Please take a look and lmk how I am screwing this up 🙏


api/cases/libraries/post_advice.py
```diff
@@ -126,3 +127,26 @@ def case_advice_contains_refusal(case_id):
...
+    routing rules that will move the case to countersigning
+    queue when this flag is set.
+    """
+    # Get all the destinations for the case
+    case_destinations = {case.end_user.party.country}
+    for user in case.ultimate_end_users:
+        case_destinations.add(user.party.country)
+    # Get all destinations for which advice has been given
+    advice_destinations = {advice.end_user.party.country for advice in case.advice.all()}
+    for advice in case.advice.all():
```
alixedi, line 145: I should prolly be filtering for user advice here.

api/cases/libraries/post_advice.py
```diff
@@ -126,3 +127,26 @@ def case_advice_contains_refusal(case_id):
...
+    """If all the destination on a case are covered by advice,
+    we set the "Advice Completed" flag. Otherwise, we unset this
+    flag. This is useful because for FCDO, we will be writing
+    routing rules that will move the case to countersigning
+    queue when this flag is set.
+    """
+    # Get all the destinations for the case
+    case_destinations = {case.end_user.party.country}
+    for user in case.ultimate_end_users:
+        case_destinations.add(user.party.country)
```
alixedi, line 142: Does the logic for getting destinations make sense?

api/cases/tests/test_post_advice.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.standard_application = self.create_draft_standard_application(self.organisation)
+        self.standard_case = self.submit_application(self.standard_application)
+
+    def test_update_advice_creates_flag(self):
+        self.create_advice(self.gov_user, self.standard_case, "end_user", AdviceType.APPROVE, AdviceLevel.USER)
+        self.assertFalse(self._check_if_flag_exists())
+        update_advice_completed_flag(self.standard_case)
```
alixedi, line 27: I am adding tests so please ignore for the moment.

api/flags/enums.py
```diff
@@ -112,6 +112,7 @@ class SystemFlags:
     SANCTION_UN_SC_MATCH = "00000000-0000-0000-0000-000000000039"
     SANCTION_OFSI_MATCH = "00000000-0000-0000-0000-000000000040"
     SANCTION_UK_MATCH = "00000000-0000-0000-0000-000000000041"
+    ADVICE_COMPLETED_ID = "00000000-0000-0000-0000-000000000042"
```
alixedi, line 115: There is a PR to `lite-content` that adds this flag to the csv.

api/cases/libraries/post_advice.py
saruniitr: Yes, I think you need third_parties and consignee as well. `end_user` is covered below.

api/cases/libraries/post_advice.py
saruniitr: yeah, filter by advice level 'user'

api/cases/libraries/post_advice.py
```diff
@@ -126,3 +127,26 @@ def case_advice_contains_refusal(case_id):
...
+    # Get all destinations for which advice has been given
+    advice_destinations = {advice.end_user.party.country for advice in case.advice.all()}
+    for advice in case.advice.all():
+        advice_destinations = advice_destinations | {user.party.country for user in advice.ultimate_end_users}
+    # If advice covers all destinations, set the flag, unset otherwise
+    flag = Flag.objects.get(id=SystemFlags.ADVICE_COMPLETED_ID)
+    if advice_destinations - case_destinations == {}:
```
saruniitr, line 149: `advice_destinations == case_destinations` ?

saruniitr: Do we need to emit an Audit event here?

api/cases/libraries/post_advice.py
```diff
@@ -126,3 +127,26 @@ def case_advice_contains_refusal(case_id):
...
         if flag in case.flags.all():
             case.flags.remove(flag)
+
+
+def update_advice_completed_flag(case):
+    """If all the destination on a case are covered by advice,
+    we set the "Advice Completed" flag. Otherwise, we unset this
+    flag. This is useful because for FCDO, we will be writing
+    routing rules that will move the case to countersigning
+    queue when this flag is set.
```
saruniitr, line 137: Does the routing rules are triggered automatically as soon as we set this flag?
If yes then there is a potential issue that as soon as the advice is given for the last destination then the rule triggers and case gets moved to countersigning queue. User would still be at the view countersigning page looking at the options for editing etc, if he decides to edit then that won't be possible as case is moved to a different queue.
If routing rules are triggered only when he clicks "Move case forward" then it is not an issue.

alixedi: That is a very good question. I don't know if routing rules are triggered automatically. I should find out. 

If the routing rules are not triggered automatically, it means that there needs to be some event that triggers them. ATM, I think the event is when a user presses the "I am done" or "Move case forward" button.

We want to avoid that ideally because atm, the case is moved over to countersigning queue even if a single adviser hits the "I am done" button.

PR #855, opened 2021-11-09
LTD-1370

Pithikos:
Remove `Party.name` length limit


api/parties/models.py
```diff
@@ -46,14 +46,14 @@ def copy_detail(self, pk):
 
 class Party(TimestampableModel):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    name = models.CharField(default=None, blank=True, max_length=100)
+    name = models.TextField(default=None, blank=True)
```
saruniitr, line 49: `default=""` is better for text based fields

saruniitr (approved):
one minor thing otherwise it is good, nice quick one.

api/parties/models.py
agalamatis: This! Let's fix old bad practices. String-like fields must have an empty string as a default. The blank parameter affects only validation, so None is the correct choice.

PR #857, opened 2021-11-11
LTD-1368 - Page 1 (Exporter)

Pithikos:
Add `Document.is_content_english` and `Document.includes_company_letterhead`


api/applications/models.py
```diff
@@ -177,12 +177,17 @@ class Meta:
...
+    Application for any of below licences:
+        - SIEL (Standard Individual Export Licence)
+        - SITCL (Standard Individual Trade Control Licence)
+    """
+
     export_type = models.CharField(choices=ApplicationExportType.choices, default=None, max_length=50)
     reference_number_on_information_form = models.CharField(blank=True, null=True, max_length=255)
     have_you_been_informed = models.CharField(
-        choices=ApplicationExportLicenceOfficialType.choices, blank=True, null=True, default=None, max_length=50,
+        choices=ApplicationExportLicenceOfficialType.choices, blank=True, null=True, default=None, max_length=50
```
alixedi, line 190: Was `black` complaining about this @Pithikos ?

(this comment applies to a bunch of changes in this file)

api/conftest.py
```diff
@@ -62,3 +66,109 @@ def {fixture_name}(db):
 if settings.DEBUG:
     for fixture_name in sorted(fixture_names):
         print(f"pytest.fixture: {fixture_name}")
+
+
+# -------------------------- Clients ----------------------------
```
alixedi, line 71: This is not necessary I reckon.

api/conftest.py
```diff
@@ -62,3 +66,109 @@ def {fixture_name}(db):
...
+
+
+class LiteClient(APIClient):
```
alixedi, line 74: What is the use for this in the API?

Elsewhere, we are using the standard Django test client and it works well.

alixedi (changes requested):
Left some initial comments.

api/conftest.py
Pithikos: You mean the header? Yeah, just put it since it's different things in the same conftest so makes it easier to split into separate files in the future

api/applications/models.py
Pithikos: Yep, as discussed on Slack, the pre-commit hook was using a different version of black and auto-formatting the file. I've added a fix for this on this PR.

I think it's fine to keep the changes. `black` won't fix it back (from what I tested locally)

api/documents/models.py
```diff
@@ -16,6 +16,11 @@ class Document(TimestampableModel):
     virus_scanned_at = models.DateTimeField(null=True, blank=True)
     safe = models.NullBooleanField()
 
+    is_content_english = models.NullBooleanField(help_text="Is the document in English?")
```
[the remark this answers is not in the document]
eadpearce: ^ what the doc said

api/conftest.py
```diff
@@ -62,3 +66,109 @@ def {fixture_name}(db):
...
+
+        elif isinstance(user, GovUser):
+            if organisation:
+                raise Exception("No organisation required for gov user")
+            self.headers = {"HTTP_GOV_USER_TOKEN": user_to_token(user)}
+        else:
+            raise Exception(f"Unknown user type: {user}")
+
+    def get(self, *args, **kwargs):
+        return super().get(*args, **kwargs, **self.headers)
```
eadpearce, line 122: do we need all these methods in here if they're just doing the same as the superclass?

eadpearce: ah wait. is `self.headers` the different part?

api/parties/serializers.py
```diff
@@ -137,21 +137,23 @@ class PartyDocumentSerializer(serializers.ModelSerializer):
...
         document = super(PartyDocumentSerializer, self).create(validated_data)
         document.save()
         process_document(document)
         return document
 
+    def validate(self, attrs):
+        validated_data = super().validate(attrs)
+        if attrs["party"].is_end_user:
+            if "is_content_english" not in attrs:
+                raise serializers.ValidationError({"is_content_english": "Required field"})
```
eadpearce, line 152: based on the help text we have for these fields a better error message could be "Select yes if the document is in English"

api/parties/serializers.py
```diff
@@ -137,21 +137,23 @@ class PartyDocumentSerializer(serializers.ModelSerializer):
...
+            if "includes_company_letterhead" not in attrs:
+                raise serializers.ValidationError({"includes_company_letterhead": "Required field"})
```
eadpearce, line 154: "Select yes if the document includes at least one page on company letterhead"

api/conftest.py
Pithikos: Yep, essentially the whole client just sets the right headers once `login()` is called

alixedi: Did you not see this on the superclass @Pithikos ?

```
    def credentials(self, **kwargs):
        """
        Sets headers that will be used on every outgoing request.
        """
        self._credentials = kwargs
```

PR #859, opened 2021-11-12
Add new fields for locations form, make export_type not required

eadpearce:
Relies on this PR on lite-frontend https://github.com/uktrade/lite-frontend/pull/299


api/applications/models.py
```diff
@@ -179,7 +179,23 @@ class Meta:
...
+    VIA_CONSIGNEE = "via_consignee"
+    VIA_CONSIGNEE_AND_THIRD_PARTIES = "via_consignee_and_third_parties"
+
+    WHO_ARE_GOODS_GOING_TO_CHOICES = [
+        (DIRECT_TO_END_USER, "Directly to the end-user"),
+        (VIA_CONSIGNEE, "To an end-user via a consignee"),
+        (VIA_CONSIGNEE_AND_THIRD_PARTIES, "To an end-user via a consignee, with additional third parties"),
+    ]
+
+    export_type = models.CharField(choices=ApplicationExportType.choices, null=True, blank=True, default=None, max_length=50)
```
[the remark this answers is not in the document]
eadpearce: Was going to leave this as it is because the frontend may be checking if export_type is None but it doesn't appear to be. Probably safe to fix this

api/applications/serializers/standard_application.py
```diff
@@ -192,15 +194,21 @@ def create(self, validated_data):
 
 
 class StandardApplicationUpdateSerializer(GenericApplicationUpdateSerializer):
+    export_type = KeyValueChoiceField(choices=ApplicationExportType.choices, required=False)
```
saruniitr, line 197: iirc this is a mandatory field, is not the case anymore?

api/applications/models.py
```diff
@@ -179,7 +179,23 @@ class Meta:
 
 # Licence Applications
 class StandardApplication(BaseApplication):
-    export_type = models.CharField(choices=ApplicationExportType.choices, default=None, max_length=50)
+    GB = "GB"
+    NI = "NI"
+    SENT_FROM_GB_OR_NI_CHOICES = [
```
saruniitr, line 184: `SENT_FROM_CHOICES` is better I think because if we add another option then it won't be consistent.

api/applications/models.py
```diff
@@ -197,6 +213,8 @@ class StandardApplication(BaseApplication):
     trade_control_product_categories = SeparatedValuesField(
         choices=TradeControlProductCategory.choices, blank=False, null=True, max_length=50
     )
+    who_are_goods_going_to = models.CharField(choices=WHO_ARE_GOODS_GOING_TO_CHOICES, default="", max_length=100)
+    sent_from_gb_or_ni = models.CharField(choices=SENT_FROM_GB_OR_NI_CHOICES, default="", max_length=2)
```
saruniitr, line 217: Please go with `TextField`s

api/applications/serializers/standard_application.py
```diff
@@ -192,15 +194,21 @@ def create(self, validated_data):
...
     class Meta:
         model = StandardApplication
         fields = GenericApplicationUpdateSerializer.Meta.fields + (
+            "export_type",
             "have_you_been_informed",
             "reference_number_on_information_form",
             "is_shipped_waybill_or_lading",
             "non_waybill_or_lading_route_details",
+            "sent_from_gb_or_ni",
+            "who_are_goods_going_to",
```
wkeeling, line 211: Thinking about reporting again, and I guess we'll need to update `LITEStandardApplicationsPipeline` to pull these fields through into DW at some point.

saruniitr (commented):
added few comments, please add tests for this.

api/applications/serializers/standard_application.py
wkeeling: When an export_type is not set is the default now an empty string `''`? Just thinking about how this might impact reports.

api/applications/models.py
eadpearce: Since we control the length of the choices I think it's fine to use CharFields here

api/applications/serializers/standard_application.py
eadpearce: @wkeeling The export_type should only ever be an empty string on draft applications that haven't been submitted yet. Do we use that data in reports?

wkeeling: Ah ok that's probably fine then. The reports do use export_type, but I think they all look at cases that are finalised (or in some terminal state).

api/applications/models.py
saruniitr: Even so I think it is good practice to go with `TextField` where possible to avoid unnecessary bugs.
Eg There is a character limit set on End user name but it is not restricted in the UI. User tried to enter longer name but failing with server error which could've been avoided if it was `TextField` to begin with. I agree these two are not the same but idea is to prevent bugs in future.

eadpearce: Apparently there's no performance improvement in postgres for using a varchar over a text field so I will change this to text here

api/applications/models.py
```diff
@@ -197,6 +213,8 @@ class StandardApplication(BaseApplication):
     trade_control_product_categories = SeparatedValuesField(
         choices=TradeControlProductCategory.choices, blank=False, null=True, max_length=50
     )
+    goods_recipients = models.TextField(choices=GOODS_RECIPIENTS_CHOICES, default="", max_length=100)
```
saruniitr, line 216: we don't need `max_length` with `TextField`

api/applications/models.py
```diff
@@ -179,7 +179,23 @@ class Meta:
...
+    VIA_CONSIGNEE = "via_consignee"
+    VIA_CONSIGNEE_AND_THIRD_PARTIES = "via_consignee_and_third_parties"
+
+    GOODS_RECIPIENTS_CHOICES = [
+        (DIRECT_TO_END_USER, "Directly to the end-user"),
+        (VIA_CONSIGNEE, "To an end-user via a consignee"),
+        (VIA_CONSIGNEE_AND_THIRD_PARTIES, "To an end-user via a consignee, with additional third parties"),
+    ]
+
+    export_type = models.CharField(choices=ApplicationExportType.choices, blank=True, default="", max_length=50)
```
saruniitr, line 198: Because we are changing this I think it is better to change to `TextField` as well.

saruniitr (approved):
Leaving few minor comments, with that fixed it looks good to me.

api/applications/models.py
eadpearce: oops! I can remove this

PR #862, opened 2021-11-18
Sporting Shotgun (code) references removed 

nahmed99:


saruniitr (commented):
There are many unrelated formatting issues. This could be mainly because your `black` version is different. Please use the correct version to not make these changes.
Please check with @Pithikos , he had similar problem and corrected it now.

eadpearce (approved):
lgtm

api/goods/helpers.py
```diff
@@ -201,7 +190,7 @@ def check_if_unsupported_fields_edited_on_firearm_good(data):
 
 def has_valid_certificate(organisation_id, document_type):
     certificate_exists = DocumentOnOrganisation.objects.filter(
-        organisation=organisation_id, document_type=document_type,
+        organisation=organisation_id, document_type=document_type
```
saruniitr, line 193: these unrelated black formatting changes need to be reverted

saruniitr (changes requested):
Please remove unrelated black formatting changes, there are many to ignore.

api/goods/helpers.py
nahmed99: Done.

saruniitr (approved):
ty, looks good to me now

PR #863, opened 2021-11-19
LTD-1485: Use GoodOnApplication id as line item identifier to record usage data

saruniitr:
## Change description

When licence details are sent to HMRC we include details about the products
on the application in the same order as they are added to the application.
They are numbered as line item 1, 2, 3 ....
When we get the usage data we get the usage against these line item. To
identify which product this line item number is referring to, lite-hmrc
maintains a mapping of line number -> lite_id which is provided when
licence details are sent.

Currently Good id is being used as lite_id. The problem with this is that
if the same Good is used to add multiple products to the application then
we will have multiple GoodOnApplication ids pointing to the same Good id.

So as per the current code when retrieving the corresponding GoodOnLicence
to record the usage using the line number mapping then we get
MultipleObjectsReturned exception and won't be able to record the usage.

To fix this we are going to use GoodOnApplication id which is unique
and can record the usage correctly.


wkeeling (approved):
Looks good

PR #867, opened 2021-11-24
LTD-1358: Add option to specify serial number can be provided later

saruniitr:
## Change description

We are updating the identification markings/serial numbers choices question for the Exporter and adding a third option to provide serial numbers later. Currently in the API this is a `boolean` and only supports two options. To accommodate the new option a new field is being added.
A management command is also added to copy the data from existing field to new field.
In preparation for the FE change serializers are also updated with new fields so FE can be released with new option. Finally we can remove existing fields later in a separate PR.


api/goods/models.py
```diff
@@ -42,8 +51,10 @@ class FirearmGoodDetails(models.Model):
     section_certificate_number = models.CharField(blank=True, max_length=100, null=True)
     section_certificate_date_of_expiry = models.DateField(blank=True, null=True)
 
-    has_identification_markings = models.BooleanField(null=True)
-    no_identification_markings_details = models.TextField(blank=True, max_length=2000, null=True)
+    has_identification_markings = models.TextField(choices=FirearmSerialNumberChoices.choices, blank=True, default="")
+    no_identification_markings_details = models.TextField(
```
saruniitr, line 55: For some reason this is restricted in the API but there was no indication in the UI.
Discussed with design and removed as there is no such requirement

api/goods/enums.py
```diff
@@ -186,3 +186,19 @@ class FirearmGoodType:
     @classmethod
     def to_str(cls, obj):
         return next(choice[1] for choice in cls.choices if choice[0] == obj) if obj else None
+
+
+class FirearmSerialNumberChoices:
+    YES_AVAILABLE = "yes_available"
```
eadpearce, line 192: can you move these into the `FirearmGoodDetails` model? It's much easier to find them then and really clear which model they belong to

saruniitr: :+1:

eadpearce (approved):

api/goods/models.py
```diff
@@ -43,6 +53,8 @@ class FirearmGoodDetails(models.Model):
 
     has_identification_markings = models.BooleanField(null=True)
     no_identification_markings_details = models.TextField(blank=True, max_length=2000, null=True)
+    serial_numbers_available = models.TextField(choices=serial_number_choices, blank=True, default="")
```
wkeeling, line 56: Did we want to use a `CharField` for choices or are we sticking with `TextField`?

api/goods/serializers.py
```diff
@@ -116,6 +116,8 @@ class FirearmDetailsSerializer(serializers.ModelSerializer):
     no_identification_markings_details = serializers.CharField(
         required=False, allow_blank=True, allow_null=True, max_length=2000
     )
+    serial_numbers_available = serializers.CharField(allow_null=True, allow_blank=True, required=False)
```
wkeeling, line 119: We're allowing `null` to be passed here. How will the underlying model respond if `null` is passed - since the model field `serial_numbers_available` doesn't permit `null`? No idea, just asking out of curiosity.

api/goods/models.py
saruniitr: Prefer `TextField`, with `CharField` again we have to specify `max_length`. In this case it is probably ok to choose some max length to cover all choices but we might as well go with textfield to avoid with, agree @wkeeling ?

api/goods/serializers.py
saruniitr: actually `allow_null=True` is likely required, copy paste from above, I will check again and remove

api/goods/management/commands/tests/test_update_sn_choice_fields.py
```diff
@@ -0,0 +1,44 @@
...
+
+from api.goods.models import FirearmGoodDetails, FirearmGoodType
+from api.goods.tests.factories import FirearmFactory
+from test_helpers.clients import DataTestClient
+
+
+class UpdateSerialNUmberChoiceFieldsMgmtCommandTests(DataTestClient):
+    def test_update_sn_choice_fields_command_with_serial_numbers_available(self):
+        firearm = FirearmFactory()
+        self.assertEqual(firearm.has_identification_markings, True)
```
wkeeling, line 11: `self.assertTrue(firearm.has_identification_markings)` ?

api/goods/management/commands/tests/test_update_sn_choice_fields.py
```diff
@@ -0,0 +1,44 @@
...
+        firearm.refresh_from_db()
+        self.assertEqual(firearm.serial_numbers_available, FirearmGoodDetails.SN_AVAILABLE)
+
+    def test_update_sn_choice_fields_command_with_serial_numbers_not_available(self):
+        firearm = FirearmFactory(
+            type=FirearmGoodType.FIREARMS,
+            has_identification_markings=False,
+            no_identification_markings_details="Products not yet manufactured",
+        )
+        self.assertEqual(firearm.has_identification_markings, False)
```
wkeeling, line 24: `self.assertFalse(firearm.has_identification_markings)`?

api/goods/management/commands/tests/test_update_sn_choice_fields.py
```diff
@@ -0,0 +1,44 @@
...
+        self.assertEqual(firearm.serial_numbers_available, FirearmGoodDetails.SN_NOT_AVAILABLE)
+        self.assertEqual(firearm.no_serial_numbers_reason, "Products not yet manufactured")
+
+    def test_update_sn_choice_fields_command_non_core_firearms(self):
+        firearm = FirearmFactory(
+            type=FirearmGoodType.FIREARMS_ACCESSORY,
+            has_identification_markings=False,
+            no_identification_markings_details="Products not yet manufactured",
+        )
+        self.assertEqual(firearm.has_identification_markings, False)
```
wkeeling, line 38: as above

api/goods/models.py
wkeeling: OK

api/goods/serializers.py
```diff
@@ -116,6 +116,8 @@ class FirearmDetailsSerializer(serializers.ModelSerializer):
...
+    no_serial_numbers_reason = serializers.CharField(required=False, allow_blank=True, allow_null=True,)
```
wkeeling, line 120: May need to check `allow_null` here too, since `FirearmGoodDetails.no_serial_numbers_reason` doesn't permit null.

saruniitr: :+1:

api/goods/management/commands/tests/test_update_sn_choice_fields.py
saruniitr: same explanation as above

api/goods/management/commands/tests/test_update_sn_choice_fields.py
saruniitr: As we have both fields this is ensuring the current field has a value, after running the command test ensure the new fields gets relevant value for this option.
True - means user has serial numbers available and can add now
False - means user doesn't have serial numbers
later - currently it is not applicable so cannot be tested in this test

wkeeling (approved):
Look good.

PR #869, opened 2021-12-01
LTD-1358: Serial number choices API fixes

saruniitr:


api/goods/helpers.py
```diff
@@ -78,16 +78,16 @@ def validate_identification_markings(validated_data):
         raise serializers.ValidationError({"number_of_items": "Enter the number of items"})
 
     # Mandatory question for firearm goods (Group 2) with conditional details fields based on the answer
-    has_identification_markings = validated_data.get("has_identification_markings")
```
saruniitr, line 81: since FE is updated, this condition is no longer relevant

api/goods/helpers.py
```diff
@@ -78,16 +78,16 @@ def validate_identification_markings(validated_data):
...
-        raise serializers.ValidationError({"has_identification_markings": [strings.Goods.FIREARM_GOOD_NO_MARKINGS]})
-
-    if has_identification_markings is False and not validated_data.get("no_identification_markings_details"):
+    serial_numbers_available = validated_data.get("serial_numbers_available")
+    if "serial_numbers_available" in validated_data and not serial_numbers_available:
         raise serializers.ValidationError(
-            {"no_identification_markings_details": ["Enter a reason why the product has not been marked"]}
+            {
+                "serial_numbers_available": [
+                    "Select whether you can enter serial numbers now, later or the product does not have them"
```
saruniitr, line 86: the error we display now is different as we have an additional option

PR #873, opened 2021-12-01
Ltd 1368 page 1 cherry picked

Pithikos:
Not squashing to make reverting/merging easier to resolve


alixedi (changes requested):
Please undo the changes made by the wrong version of black.

api/parties/serializers.py
```diff
@@ -137,21 +137,27 @@ class PartyDocumentSerializer(serializers.ModelSerializer):
...
         document.save()
         process_document(document)
         return document
 
+    def validate(self, attrs):
+        validated_data = super().validate(attrs)
+        if attrs["party"].is_end_user:
+            if "is_content_english" not in attrs:
+                raise serializers.ValidationError({"is_content_english": "Select if the document is in English"})
+            if "includes_company_letterhead" not in attrs:
```
saruniitr, line 153: Is there no validation required if none of these options are selected

Pithikos: So the default validation is run in all cases like before `validated_data = super().validate(attrs)`

This is just extra checks I've added

PR #875, opened 2021-12-06
Exclude NLR goods from serialization

wkeeling:
Filter out any goods that are not controlled (NLR) before attempting to serialize the goods on a licence.

### DIFF
--- a/api/licences/helpers.py
+++ b/api/licences/helpers.py
@@ -51,7 +51,7 @@ def serialize_goods_on_licence(licence):
 
     if licence.goods.exists():
         # Standard Application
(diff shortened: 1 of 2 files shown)


saruniitr (approved):
this looks good to me.
We have created `GoodOnLicenceReportsViewSerializer` mainly for reporting purposes that gives the correct `id` and `GoodOnApplication.id` on the `GoodOnLicence` model, just wondering if we need to make use of this or it is not related to that.

wkeeling:
This fix actually fixes a LITE API issue (which indirectly affects reporting). I'm not sure we'd be able to use the `GoodOnLicenceReportsViewSerializer` since that's reporting specific - but that serializer should be unaffected.

PR #877, opened 2021-12-07
Don't delete unused denial reasons

wkeeling:
This command will also remove any associated advice->denial reason
mappings via the many-to-many through table which is not desired.


api/staticdata/management/commands/seeddenialreasons.py
```diff
@@ -22,4 +22,3 @@ def operation(self, *args, **options):
             {"id": row["id"], "deprecated": row["deprecated"], "description": row["description"]} for row in csv
         ]
         self.update_or_create(DenialReason, filtered_csv)
-        self.delete_unused_objects(DenialReason, filtered_csv)
```
saruniitr, line 25: not sure why this is added in the first place

saruniitr (approved):

agalamatis (approved):
For whatever reason, get rid of it now, so we can restore the data in the table.

PR #880, opened 2021-12-08
Additionally order by id so not to break pagination

wkeeling:
Ordering by name alone causes non-deterministic behaviour when multiple parties exist with the same name (which is common in production). This breaks pagination - because the same party can be emitted on different pages. That in turn breaks other things like unique constraints in Data Workspace.

### DIFF
--- a/api/data_workspace/views.py
+++ b/api/data_workspace/views.py
@@ -24,7 +24,13 @@ class PartyListView(viewsets.ReadOnlyModelViewSet):
     authentication_classes = (DataWorkspaceOnlyAuthentication,)
     serializer_class = PartyViewSerializer
     pagination_class = LimitOffsetPagination
-    queryset = Party.objects.all()
+    # We ensure that parties with the same name always come out
+    # in the same order by additionally ordering by created_at
+    # and by id. If just name is used, the order is non-deterministic
+    # when multiple parties have the same name. This can mean the
+    # same party being emitted on different pages, causing
+    # primary key violations in Data Workspace.
+    queryset = Party.objects.order_by("name", "created_at", "id")
 
 
 class QueueListView(viewsets.ReadOnlyModelViewSet):


api/parties/models.py
```diff
@@ -75,7 +75,10 @@ class Party(TimestampableModel):
     objects = PartyManager()
 
     class Meta:
-        ordering = ["name"]
+        # We order by id as well as name to ensure ordering is
+        # deterministic when multiple parties exist with the same name.
+        # In such cases the id will act as a tie breaker.
+        ordering = ["name", "id"]
```
saruniitr, line 81: Is it better to order by `name` and `created_at` since `id` is `uuid` and not a simple int? I guess it won't make much difference but just wondering if `created_at` is better than `id`.

saruniitr (approved):
just one thing to consider

api/parties/models.py
wkeeling: Ideally we want to use something that is 100% unique to act as a tie breaker. Although `created_at` _should_ be unique, there is a possibility that it might not be unique - e.g. bulk import of data from a different system (pointed out by Michal in the DI team). 

Perhaps compromise with:  
```
ordering = ["name", "created_at", "id"]
```

wkeeling: @saruniitr on a related note, is there likely to be any impact here on the licence as I remember a recent bug related to mis-matched ordering? Do we need to modify the ordering to match?

agalamatis: @wkeeling to avoid changing the ordering in the model, I think you can redefine the queryset in viewset class using it, something like
```
class PartyViewset(...):
    return super(...).order_by("name", "created_at", "id")
```

saruniitr: @wkeeling only the ordering of goods on licence that needs to match with the order on the products on the application so parties order has no impact.

PR #883, opened 2021-12-14
Reduce number of SQL queries when retrieving a Case

wkeeling:
Approximate 40% reduction in number of SQL queries when serializing a `Case`.

### DIFF
--- a/api/cases/libraries/get_case.py
+++ b/api/cases/libraries/get_case.py
@@ -3,15 +3,25 @@ from api.core.exceptions import NotFoundError
 from lite_content.lite_api import strings
 
 
-def get_case(pk, include_draft=False):
+def get_case(pk, include_draft=False, prefetch_related=None):
     """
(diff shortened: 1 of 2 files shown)


api/cases/libraries/get_case.py
```diff
@@ -3,15 +3,25 @@
 from lite_content.lite_api import strings
 
 
-def get_case(pk, include_draft=False):
+def get_case(pk, include_draft=False, prefetch_related=None):
```
alixedi, line 6: Might want to use `false` as the default here?

alixedi (approved):
One small comment that I will leave with you.

api/cases/libraries/get_case.py
wkeeling: When supplied it's passed as a list of strings, so I thought better to use `None` (as didn't want to use `[]` as the default).

PR #885, opened 2021-12-15
Backwards compatible SIEL location flow updates

eadpearce:


api/applications/models.py
```diff
@@ -179,7 +179,23 @@ class Meta:
...
+    VIA_CONSIGNEE = "via_consignee"
+    VIA_CONSIGNEE_AND_THIRD_PARTIES = "via_consignee_and_third_parties"
+
+    GOODS_RECIPIENTS_CHOICES = [
+        (DIRECT_TO_END_USER, "Directly to the end-user"),
+        (VIA_CONSIGNEE, "To an end-user via a consignee"),
+        (VIA_CONSIGNEE_AND_THIRD_PARTIES, "To an end-user via a consignee, with additional third parties"),
+    ]
+
+    export_type = models.TextField(choices=ApplicationExportType.choices, blank=True, default="")
```
wkeeling, line 198: We need to update `LITEStandardApplicationsPipeline` in the reports repo, and switch the `export_type` to be a `Text` type there too (it's currently a `String` type of maxlen 50). I'll raise a ticket.

wkeeling (approved):
Looks good to me - perhaps worth squashing the commits before merging though?

wkeeling (approved):
Looks good to me - perhaps worth squashing the commits before merging though?

alixedi (changes requested):
Can you please clean up the commits before I review this?

api/applications/views/applications.py
```diff
@@ -203,10 +203,12 @@ def put(self, request, pk):
         Update an application instance
         """
         application = get_application(pk)
-        serializer = get_application_update_serializer(application)
+        update_serializer = get_application_update_serializer(application)
```
saruniitr, line 206: `partial=True` implies update so this renaming is not necessary

eadpearce: this was only renamed because there's another variable further down called serializer and it's less confusing this way

PR #889, opened 2022-01-04
LTD-1597: Update Party type values and expose it in serializer

saruniitr:
These are displayed in the internal UI.


api/parties/serializers.py
```diff
@@ -20,6 +20,7 @@ class PartySerializer(serializers.ModelSerializer):
     website = serializers.CharField(required=False, allow_blank=True)
     signatory_name_euu = serializers.CharField(allow_blank=True)
     type = serializers.ChoiceField(choices=PartyType.choices, error_messages=PartyErrors.TYPE)
+    type_value = serializers.SerializerMethodField()
```
wkeeling, line 23: Wonder whether this may be clearer if called `type_display_value` or `type_display_name` since it holds the display name for the type?

wkeeling (commented):
Added 1 comment

api/parties/serializers.py
saruniitr: that is better, I will update this

wkeeling: Or maybe `type_display`

wkeeling (approved):
Looks good

PR #891, opened 2022-01-06
Revert "Backwards compatible SIEL location flow updates"

alixedi:
Reverts uktrade/lite-api#885


saruniitr (approved):

api/applications/migrations/0050_auto_20211210_1618.py
```diff
@@ -1,28 +0,0 @@
...
-            model_name='standardapplication',
-            name='goods_starting_point',
-            field=models.TextField(choices=[('GB', 'Great Britain'), ('NI', 'Northern Ireland')], default=''),
-        ),
-        migrations.AlterField(
-            model_name='standardapplication',
-            name='export_type',
-            field=models.TextField(blank=True, choices=[('permanent', 'Permanent'), ('temporary', 'Temporary')], default=''),
-        ),
-    ]
```
wkeeling, line 28: Think we might have to keep this migration, and add a new migration that reverts what this does? That way the database will also get rolled back.

wkeeling (approved):

PR #892, opened 2022-01-06
LTD-1603: Add display_value field to DenialReason to display in UI

saruniitr:
## Change description

Some of the refusal criteria items have a descriptive text to display instead of
an acronym used in id. Changing the id is not an option as we will endup creating
new items when the seed command is executed.

This PR adds a new field to the model which is used to display in the UI instead.
For those items where there is no descriptive text, we use the value of id itself.


api/staticdata/denial_reasons/models.py
```diff
@@ -5,6 +5,7 @@ class DenialReason(models.Model):
     id = models.TextField(primary_key=True, editable=False)
     deprecated = models.BooleanField(default=False, null=False, blank=False)
     description = models.TextField(default="")
+    display_value = models.TextField(default="")
```
wkeeling, line 8: Do we need to add this to the Data Workspace reports?

wkeeling (approved):

api/staticdata/denial_reasons/models.py
saruniitr: as discussed internally, marking it as resolved

PR #899, opened 2022-01-14
Check good exists on application

wkeeling:
This fixes a support issue currently being seen in production where an exporter is resubmitting an application. The submission falls over because a good has been removed from the application. This PR fixes that by checking that the good exists before proceeding.


saruniitr (commented):
with a unit test this is good to go

api/letter_templates/tests/test_context_generation.py
```diff
@@ -465,6 +465,35 @@ def test_generate_context_with_advice_on_goods(self):
...
+            case,
+            "good",
+            AdviceType.REFUSE,
+            AdviceLevel.FINAL,
+            advice_text="abc",
+            good=case.goods.first().good,
+        )
+        final_advice = self.create_advice(
+            self.gov_user, case, "good", AdviceType.REFUSE, AdviceLevel.FINAL, advice_text="abc", good=good.good
+        )
```
saruniitr, line 487: these two are the same correct? do we need both @wkeeling ?

wkeeling: The first advice references the existing good on the case, the second advice references the new good that was created on line 473. We then delete the existing good on the case, so that the first advice item points at a non-existent good. The fix then prevents that situation from falling over.

saruniitr: :+1:

saruniitr (approved):

PR #911, opened 2022-01-27
LTD-1676: Support default values for audit substitution params

wkeeling:
If a timeline message is changed it potentially no longer matches the payload parameters that are stored for historic audit records. This PR allows a default value to be supplied for parameters so the code has a fallback option.

Also improves error reporting.

### DIFF
--- a/api/audit_trail/payload.py
+++ b/api/audit_trail/payload.py
@@ -19,8 +19,12 @@ class DefaultValueParameterFormatter(Formatter):
                 return kwds[key]
             except KeyError:
                 try:
-                    return key.split("|")[1].strip()
-                except IndexError:
+                    key, val = key.split("|")
+                    try:
+                        return kwds[key.strip()]
+                    except KeyError:
+                        return val.strip()
+                except ValueError:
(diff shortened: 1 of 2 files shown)


api/audit_trail/payload.py
```diff
@@ -128,5 +153,5 @@ def format_payload(audit_type, payload):
     AuditType.LICENCE_UPDATED_STATUS: strings.Audit.LICENCE_UPDATED_STATUS,
     AuditType.DOCUMENT_ON_ORGANISATION_CREATE: "added {document_type} '{file_name}' to organization",
     AuditType.REPORT_SUMMARY_UPDATED: "updated ARS for {good_name} from {old_report_summary} to {report_summary}",
-    AuditType.COUNTERSIGN_ADVICE: "countersigned all {department} recommendations",
+    AuditType.COUNTERSIGN_ADVICE: "countersigned all {department|FCDO} recommendations",
```
wkeeling, line 156: The downside I suppose is that this won't work so well when non-FCDO departments start countersigning.

alixedi: Would it be safer to default to a `""` when we don't have the department?

saruniitr (approved):

PR #915, opened 2022-02-01
LTD-1327: Changes to incorporate new case/application referencing scheme

saruniitr:
## Change description
This PR implementes changes to generate Case/application reference code as per the new naming scheme.
 - removes the use of 'GB' in the reference code to cover applications from NI as well.
 - removes the export_type which is either 'permanent' or 'temporary'.
 - only includes the decade instead of the whole year
 - revisions of licences are tracked with incrementing number suffixes (-01, -02, ...)


depsiatwal (approved):
LGTM

PR #921, opened 2022-02-11
Locations changes as per LTD-1372

sekharpanja:
This is to address locations changes as per LTD-1372


api/applications/creators.py
```diff
@@ -32,6 +33,26 @@ def _validate_locations(application, errors):
     return errors
 
 
+def _validate_siel_locations(application, errors):
+    old_locations_invalid = (
+        not SiteOnApplication.objects.filter(application=application).exists()
+        and not ExternalLocationOnApplication.objects.filter(application=application).exists()
+        and not getattr(application, "have_goods_departed", False)
+        and not getattr(application, "goodstype_category", None) == GoodsTypeCategory.CRYPTOGRAPHIC
```
kevincarrogan, line 41: These `getattr`s seem really odd to me here. It's hard to know what this logic is actually trying to convey. What's the scenario where these attributes aren't on the application?

api/applications/tests/test_edit_application.py
```diff
@@ -38,6 +38,46 @@ def test_edit_unsubmitted_application_name_success(self):
         # Unsubmitted (draft) applications should not create audit entries when edited
         self.assertEqual(Audit.objects.count(), 0)
 
+    def test_edit_unsubmitted_application_export_type_success(self):
+        """ Test edit the application export_type of an unsubmitted application. An unsubmitted application
+        has the 'draft' status.
+        """
+        application = self.create_draft_standard_application(self.organisation)
+        # export_type is set to permanent in create_draft_standard_application
```
kevincarrogan, line 46: You could add an assertion for this to make sure that is always valid if your test has an expectation that it will be e.g. `self.assertEqual(application.export_type, "permanent")`

api/applications/creators.py
sekharpanja: I agree this is slightly clumsy, but this gracefully allows both old and new locations to co-exist. This can be refactored during django forms migration.

api/applications/models.py
```diff
@@ -201,6 +218,8 @@ class StandardApplication(BaseApplication):
     trade_control_product_categories = SeparatedValuesField(
         choices=TradeControlProductCategory.choices, blank=False, null=True, max_length=50
     )
+    goods_recipients = models.TextField(choices=GOODS_RECIPIENTS_CHOICES, default="")
+    goods_starting_point = models.TextField(choices=GOODS_STARTING_POINT_CHOICES, default="")
```
wkeeling, line 223: These new attributes will potentially need to be pulled into Data Workspace for the reports. That's a separate piece of work but I'll raise a ticket for it.

wkeeling (approved):

api/applications/tests/test_create_application.py
```diff
@@ -45,6 +45,25 @@ def test_create_draft_standard_individual_export_application_successful(self):
...
+        data = {
+            "name": "Test",
+            "application_type": CaseTypeReferenceEnum.SIEL,
+            "have_you_been_informed": ApplicationExportLicenceOfficialType.YES,
+            "reference_number_on_information_form": "123",
+        }
+
+        response = self.client.post(self.url, data, **self.exporter_headers)
+        response_data = response.json()
+        standard_application = StandardApplication.objects.get()
```
saruniitr, line 61: I think its better to assert on the status code first before retrieving this

saruniitr (approved):
few minor comments other lgtm

api/applications/models.py
wkeeling: Created https://uktrade.atlassian.net/browse/LTD-1885

PR #923, opened 2022-02-15
ltd-1814-pre-commit-pii-hook

depsiatwal:


README.md
```diff
@@ -42,6 +42,12 @@ Service for handling backend calls in LITE.
 - At this point you might want to seed your database with some static
   - run `docker-compose run api ./manage.py seedall`
 
+#### Git Hub pre-commit setup
+- Install pre-commit
```
wkeeling, line 46: Might be worth being explicit about the installation of the library vs installation of the hook itself, e.g.

- Install the pre-commit library with `pip install pre-commit`
- Then install the pre-commit hook with `pre-commit install`

depsiatwal: pip is more environment based and I try to keep read.me environment agnostic 
hence added but as an example

wkeeling (approved):

PR #933, opened 2022-02-28
LTD-1704-gov-sigin-login 

depsiatwal:


wkeeling (approved):

api/users/views.py
```diff
@@ -49,28 +49,38 @@ def post(self, request, *args, **kwargs):
...
+                    data={"errors": [strings.Login.Error.USER_PROFILE]},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
         try:
             user = ExporterUser.objects.get(baseuser_ptr__email__iexact=data.get("email"))
-            # Update the user's first and last names
-            user.baseuser_ptr.first_name = first_name
-            user.baseuser_ptr.last_name = last_name
-            user.baseuser_ptr.save()
+            if first_name and last_name:
```
alixedi, line 79: What if we get `first_name` but not `last_name` or vice versa? Can we do away with this check?

api/users/views.py
```diff
@@ -49,28 +49,38 @@ def post(self, request, *args, **kwargs):
         Returns a token which is just our ID for the user
         """
         data = request.data
-
-        user_profile = data.get("user_profile")
-        if not user_profile:
-            return JsonResponse(
-                data={"errors": [strings.Login.Error.USER_PROFILE]},
-                status=status.HTTP_400_BAD_REQUEST,
-            )
```
alixedi, line 58: If I understand correctly, it seems that we can keep all of this code and basically put our new logic here under this `if`?

api/users/views.py
depsiatwal: nope bad design we can come to this view with no name 
where we want to return a user with no name or we may come to this view to update the user.

The correct way would be to have two separate views one for update and one for retrieve. Probably one for when we get rid of great log-in. We can remove all the profile stuff.

api/users/views.py
```diff
@@ -49,28 +49,38 @@ def post(self, request, *args, **kwargs):
...
-
-        first_name = data.get("user_profile").get("first_name")
-        last_name = data.get("user_profile").get("last_name")
-        if not first_name or not last_name:
-            return JsonResponse(
-                data={"errors": [strings.Login.Error.USER_PROFILE]},
-                status=status.HTTP_400_BAD_REQUEST,
-            )
-
+        # We require this to be feature flagged because no profile is returned for GOV.UK
```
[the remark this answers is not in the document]
depsiatwal: removed

PR #936, opened 2022-03-03
fixing locations issues

sekharpanja:
fixing Locations issues by making the whole journey into single lite forms group. Issues as shown in LTD-1891 and more.

### DIFF
--- a/api/applications/serializers/standard_application.py
+++ b/api/applications/serializers/standard_application.py
@@ -6,6 +6,8 @@ from api.applications.enums import (
     YesNoChoiceType,
     ApplicationExportLicenceOfficialType,
     ApplicationExportType,
+    GoodsStartingPoint,
+    GoodsRecipients,
 )
 from api.applications.mixins.serializers import PartiesSerializerMixin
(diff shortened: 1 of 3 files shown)


api/applications/serializers/standard_application.py
```diff
@@ -198,8 +200,10 @@ class StandardApplicationUpdateSerializer(GenericApplicationUpdateSerializer):
     export_type = KeyValueChoiceField(
         choices=ApplicationExportType.choices, required=False, allow_blank=True, allow_null=True
     )
-    goods_starting_point = serializers.CharField()
-    goods_recipients = serializers.CharField()
+    goods_starting_point = KeyValueChoiceField(
+        choices=GoodsStartingPoint.choices, required=False, allow_blank=True, allow_null=True
```
wkeeling, line 204: I notice `allow_null=True` is set here, but the underlying model `StandardApplication.goods_starting_point` doesn't permit null. Do we need to change either the serializer or the model so that they match?

wkeeling (approved):

PR #941, opened 2022-03-08
LTD-1902-nlr-fix-finialise 

depsiatwal:
This change is to check if we have 'no license required' document and all products are not controlled and remove the approval doc. 
Before the change the user was presented with the following documents. The user is forced to generate the approval doc even through it has no products. 

With the new change this document is no longer required which enables the case worker to continue to finalise the case. 
Screen Shot 2022-02-28 at 11 49 43 AM

Screenshot 2022-03-07 at 22 07 21


api/cases/libraries/finalise.py
```diff
@@ -18,6 +19,14 @@ def get_required_decision_document_types(case):
         required_decisions.add(AdviceType.APPROVE)
         required_decisions.remove(AdviceType.PROVISO)
 
+    # Check if no controlled good on application then no approval document required.
+    controlled_goods = GoodOnApplication.objects.filter(application=case.id).values_list(
+        "is_good_controlled", flat=True
+    )
+
+    if True not in controlled_goods and AdviceType.NO_LICENCE_REQUIRED in required_decisions:
```
saruniitr, line 27: is it not enough to check if all final decisions are NLR?

kevincarrogan, line 27: Instead of checking whether one of the values is True in the evaluated queryset you should just be able to do the whole check using a queryset and a .exists e.g. `GoodOnApplication.objects.filter(application=case.id, is_good_controlled=True).exists()` and assign this to an appropriately named variable.

api/cases/tests/test_final_advice_documents.py
```diff
@@ -58,6 +59,76 @@ def test_get_final_advice_with_document(self):
...
+        )
+
+        GoodOnApplicationFactory(application=self.case, good=good, is_good_controlled=False)
+
+        response = self.client.get(self.url, **self.gov_headers)
+        response_data = response.json()["documents"]
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        self.assertEqual(response_data.get(AdviceType.APPROVE), None)
```
wkeeling, line 83: `self.assertIsNone()` may be clearer here.

api/cases/tests/test_final_advice_documents.py
```diff
@@ -58,6 +59,76 @@ def test_get_final_advice_with_document(self):
...
+
+        self.assertEqual(
+            response_data[AdviceType.NO_LICENCE_REQUIRED]["value"], AdviceType.get_text(AdviceType.NO_LICENCE_REQUIRED)
+        )
+        self.assertEqual(
+            list(response_data[AdviceType.NO_LICENCE_REQUIRED]["document"].keys()),
```
wkeeling, line 89: Minor - `keys()` not needed here as dictionary keys are iterated by default.

api/cases/tests/test_final_advice_documents.py
```diff
@@ -58,6 +59,76 @@ def test_get_final_advice_with_document(self):
...
+            self.case, self.template, advice_type=AdviceType.NO_LICENCE_REQUIRED, licence=self.licence
+        )
+
+        response = self.client.get(self.url, **self.gov_headers)
+        response_data = response.json()["documents"]
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response_data[AdviceType.APPROVE]["value"], AdviceType.get_text(AdviceType.APPROVE))
+        self.assertEqual(
+            list(response_data[AdviceType.APPROVE]["document"].keys()), ["id", "advice_type", "user", "created_at"]
```
wkeeling, line 119: As above

api/cases/tests/test_final_advice_documents.py
```diff
@@ -58,6 +59,76 @@ def test_get_final_advice_with_document(self):
...
+        )
+        self.assertEqual(response_data[AdviceType.APPROVE]["document"]["id"], str(document_one.pk))
+
+        self.assertEqual(
+            response_data[AdviceType.NO_LICENCE_REQUIRED]["value"], AdviceType.get_text(AdviceType.NO_LICENCE_REQUIRED)
+        )
+        self.assertEqual(
+            list(response_data[AdviceType.NO_LICENCE_REQUIRED]["document"].keys()),
```
wkeeling, line 127: As above

api/cases/libraries/finalise.py
depsiatwal: yeah I was thinking that as well , but was trying to following exiting pattern above. 
I will change it as I think it's cleaner.

depsiatwal: nope as sometimes you can have a mix which will require approval doc and NLR

kevincarrogan (approved):

api/cases/libraries/finalise.py
saruniitr: I meant if all decisions are NLR only, if we have a mix then that condition will be false and we generate both Approve and NLR documents

PR #945, opened 2022-03-11
LTD-1368: Update end-user document validation when submitting application

saruniitr:
## Change description

End-user document(s) are mandatory but we are also providing an option to skip this with a valid reason so there may not be any documents also.

The validation logic is updated to consider this otherwise the application cannot be submitted.
Updated unit test that validates this.

Previously an input field was incorrectly associated as `PartyDocument` description as it was below the file upload but it is a separate piece of information we are capturing from End-user. A new field in `Party` model is created for this and updated serializers.


wkeeling:
Looks fine, but should we cover this condition with a test somewhere?

saruniitr:
> Looks fine, but should we cover this condition with a test somewhere?

@wkeeling test was already available so I have updated it

wkeeling (approved):

kevincarrogan (approved):

sekharpanja (approved):

PR #946, opened 2022-03-14
LTD-1279-Adds-Destinations-to-payload

seijihg:
This PR is a support for LTD-1279 where it needs to list out the list of countries for destinations. When checking details for each cases you can see that those destinations included all consignee, end user and etc.

This is adding to the end-point payload the necessary informations needed to list them out.

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-1279
- [ ] Jira ticket has the correct status.

### Reviewing help

### Housekeeping

### Merging


api/applications/serializers/serializers.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import serializers
+
+from api.applications.models import PartyOnApplication
+from api.parties.serializers import PartySerializer
+
+
+class PartyOnApplicationSerializer(serializers.ModelSerializer):
+    party = PartySerializer()
```
kevincarrogan, line 8: Is this right? Should we just be instantiating a serializer object here? Usually this is some kind of field.

api/cases/views/search/service.py
```diff
@@ -189,3 +191,20 @@ def get_hmrc_sla_hours(cases: List[Dict]):
...
+
+
+def populate_destinations(cases: List[Dict]):
+    for case in cases:
+        id = case["id"]
+        destinations = []
+
+        if case["case_type"]["type"]["key"] == CaseTypeTypeEnum.APPLICATION:
+            for poa in PartyOnApplication.objects.filter(application=id):
+                if poa.deleted_at is None:
```
kevincarrogan, line 203: This could just be part of the above filter in the queryset.

api/cases/views/search/service.py
```diff
@@ -189,3 +191,20 @@ def get_hmrc_sla_hours(cases: List[Dict]):
...
+                    serializer = PartyOnApplicationSerializer(poa)
```
kevincarrogan, line 204: Did we actually need to add this new serializer? You could just get the party object from `poa` and then use the `PartySerializer` directly instead of having this other serializer in the middle.

seijihg: ah right so much to learn about Django :D

kevincarrogan (approved):

PR #948, opened 2022-03-15
Minio local

sekharpanja:
add minio to docker compose and conditionally use it


docker-compose.yml
```diff
@@ -13,6 +13,19 @@ services:
     networks:
       - lite
 
+  lite-minio-s3:
+    image: minio/minio:RELEASE.2020-10-28T08-16-50Z-29-g07d7dd632
+    environment:
+      MINIO_ACCESS_KEY: AKIAIEXAMPLE
+      MINIO_SECRET_KEY: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
saruniitr, line 20: Please replace it with some example password

local.env
```diff
@@ -28,6 +28,7 @@ AWS_ACCESS_KEY_ID=AWS_ACCESS_KEY_ID
 AWS_SECRET_ACCESS_KEY=AWS_SECRET_ACCESS_KEY
 AWS_STORAGE_BUCKET_NAME=AWS_STORAGE_BUCKET_NAME
 AWS_REGION=eu-west-2
+S3_LOCAL_ENDPOINT_URL=
```
saruniitr, line 31: some example value instead of blank?

docker-compose.yml
kevincarrogan: This is actually better than an example password as it's pretty clear this isn't a real password and it's fine that it's the value that is used by minio.

saruniitr: problem is if we were to type it (not sure minio interface allows copy paste)

local.env
sekharpanja: This was purposefully kept blank to give the dev the choice of using MinIO when they need to. If this was filled in with default and copied across to .env, it might come as a surprise. Happy to discuss further.

sekharpanja:
closing as agreed

PR #952, opened 2022-03-21
Ltd 1731 advice performance optimisations

kevincarrogan:
This optimises the performance of the `CaseDetail` endpoint by utilising `select_related` and `prefetch_related` where possible.

This is only the beginning of finding all the cases where we can optimise in this manner but we are stopping here for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -172,3 +172,5 @@ bank-holidays.csv
 
 # pii file
 .pii-secret-hook
+
+*.prof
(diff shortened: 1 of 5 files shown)


depsiatwal (approved):
nicely done :-) 
Optional would be useful to see evidence of optimisation i.e perhaps a query which shows optimised vs non-optimised timings.

kevincarrogan:
> nicely done :-) Optional would be useful to see evidence of optimisation i.e perhaps a query which shows optimised vs non-optimised timings.

That information is in the ticket as opposed to here in Github.

PR #955, opened 2022-03-24
Add category field to firearm good details model

kevincarrogan:


api/goods/models.py
```diff
@@ -41,6 +50,7 @@ def has_serial_numbers(cls, value):
             return value in cls.get_has_serial_numbers_values()
 
     type = models.TextField(choices=FirearmGoodType.choices, blank=False)
+    category = ArrayField(models.CharField(choices=FirearmCategory.choices, max_length=255, default=""), default=list)
```
alixedi, line 53: Wonder why do we need `default=""` for the `CharField` here? Is `""` a valid category?

kevincarrogan: Ye. It is because we need to provide backwards compatibility with models that don't have this currently set.

kevincarrogan: Oh. Actually I see what you're saying, this is from me copying and pasting poorly, the whole thing should be nullable.

wkeeling (approved):

PR #956, opened 2022-03-25
LTD-1430: Update PV grading choices

saruniitr:
## Change description

New designs introduce different choices without prefix and capture prefix
as separate field. Update enums to include new choices as separate group.

To make it backward compatible these are added as a separate group because
some of the applications may be using existing choices. The current FE
gets existing choices whereas update FE gets new choices.


api/goods/serializers.py
```diff
@@ -47,33 +46,7 @@ class PvGradingDetailsSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = PvGradingDetails
-        fields = ("grading", "custom_grading", "prefix", "suffix", "issuing_authority", "reference", "date_of_issue")
```
saruniitr, line 50: `custom_grading` is already optional in current `master` and it is completely removed in the updated design

PR #966, opened 2022-04-06
LTD-2165 fixing missing serial number api filter

sekharpanja:
LTD-2165 fixing missing serial number API filter

### DIFF
--- a/api/applications/views/applications.py
+++ b/api/applications/views/applications.py
@@ -175,8 +175,8 @@ class ApplicationsRequireSerialNumbersList(ListAPIView):
         applications = applications.filter(
             Q(
                 goods__firearm_details__serial_numbers_available__in=FirearmGoodDetails.SerialNumberAvailability.get_has_serial_numbers_values()
-            ) &
-            (
+            )
+            & (
                 Q(goods__firearm_details__serial_numbers__len__lt=F("goods__firearm_details__number_of_items"))
                 | Q(goods__firearm_details__serial_numbers__contains=[""])
             )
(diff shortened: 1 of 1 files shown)


kevincarrogan (commented):
I think it's worth adding or amending a test to verify this change (I can see someone trying to "refactor" this into the older version and breaking it again).

api/applications/views/applications.py
```diff
@@ -173,11 +173,13 @@ def get_queryset(self):
...
-        applications = applications.filter(
-            Q(goods__firearm_details__serial_numbers__len__lt=F("goods__firearm_details__number_of_items"))
-            | Q(goods__firearm_details__serial_numbers__contains=[""])
+            Q(
+                goods__firearm_details__serial_numbers_available__in=FirearmGoodDetails.SerialNumberAvailability.get_has_serial_numbers_values()
+            ) &
+            (
+                Q(goods__firearm_details__serial_numbers__len__lt=F("goods__firearm_details__number_of_items"))
+                | Q(goods__firearm_details__serial_numbers__contains=[""])
+            )
```
saruniitr, line 182: looks like it is doing the same as before? or
is it that because `filter` is lazy the `applications.filter` in line 178 is the original list of applications rather than the filtered value from line 175?

kevincarrogan: This doesn't result in the same query that you would think it does.

The two filters aren't actually doing an `&` between them, this is explicitly doing what we want now.

kevincarrogan: But I think this pull request could do with a test to show that these two queries are actually doing something subtly different.

kevincarrogan: To be more concise it's the difference between:

`False and False or True`

and 

`False and (False or True)`

saruniitr (approved):

PR #970, opened 2022-04-08
LTD-2156: Update Good serializer to be able to save document related options

saruniitr:
## Change description

Currently we don't have any option to edit product document availability and sensitivity options as in the current wizard we ask these questions after a product is created.

In the new wizard however we ask these before a product is created and user has an option to edit these from summary screen. So update serializer to be able to save changes and also add unit tests.

### DIFF
--- a/api/applications/tests/test_standard_application_submit.py
+++ b/api/applications/tests/test_standard_application_submit.py
@@ -532,9 +532,6 @@ class StandardApplicationTests(DataTestClient):
         self.assertNotIn(SystemFlags.WMD_END_USE_ID, case_flags)
         self.assertNotIn(SystemFlags.MARITIME_ANTI_PIRACY_ID, case_flags)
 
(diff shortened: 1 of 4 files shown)


api/applications/tests/test_standard_application_submit.py
```diff
@@ -532,9 +532,6 @@ def test_resubmit_edited_standard_application_removes_system_case_flags_success(
         self.assertNotIn(SystemFlags.WMD_END_USE_ID, case_flags)
         self.assertNotIn(SystemFlags.MARITIME_ANTI_PIRACY_ID, case_flags)
 
-        # only good on the draft not firearms
-        self.assertNotIn(SystemFlags.FIREARMS_ID, case_flags)
```
saruniitr, line 536: removed because the product category in the factory is updated below

depsiatwal (approved):
LGTM

PR #978, opened 2022-04-20
[LTD-2174] Precedent

alixedi:
This adds an endpoint that exposes precedents.

Precedents are defined as `GoodOnApplication` that share the same `Good` but are associated with an application that has already been submitted.

### DIFF
--- a/api/goods/serializers.py
+++ b/api/goods/serializers.py
@@ -1,6 +1,7 @@
 from rest_framework import serializers
 from rest_framework.relations import PrimaryKeyRelatedField
 
+from api.applications.models import PartyOnApplication
 from api.core.helpers import str_to_bool
 from api.core.serializers import KeyValueChoiceField, ControlListEntryField, GoodControlReviewSerializer
 from api.documents.libraries.process_document import process_document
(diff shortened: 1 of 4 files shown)


api/goods/serializers.py
```diff
@@ -728,3 +729,48 @@ def update(self, instance, validated_data):
...
+
+    def get_queue(self, obj):
+        queue = obj.application.queues.first()
+        return queue.id if queue else None
+
+    def get_control_list_entries(self, obj):
+        return [cle.rating for cle in obj.get_control_list_entries().all()]
+
+    def get_wassenaar(self, obj):
+        return "WASSENAAR" in [flag["name"] for flag in obj.good.flags.values("name")]
```
wkeeling, line 770: Could perhaps use `values_list` instead, then can do away with the list comp:
```python
return "WASSENAAR" in obj.good.flags.values_list("name", flat=True)
```

api/goods/serializers.py
```diff
@@ -728,3 +729,48 @@ def update(self, instance, validated_data):
...
+            "unit",
+            "value",
+            "control_list_entries",
+            "destinations",
+            "wassenaar",
+            "submitted_at",
+        )
+
+    def get_queue(self, obj):
+        queue = obj.application.queues.first()
```
wkeeling, line 763: If the application is assigned to more than one queue, will it matter that it's only considering `first()`?

alixedi: That is a very good point @wkeeling and worth putting in a comment. 

Basically, the only purpose of the `queue_id` in this endpoint is that the FE generates a link from it -

```
{% url cases:case queue_pk=... %}
```

It is not clear whether one link would be better than another at the point when it is generated.

wkeeling (approved):

api/goods/serializers.py
```diff
@@ -728,3 +729,48 @@ def update(self, instance, validated_data):
...
         return instance
+
+
+class GoodOnApplicationSerializer(serializers.ModelSerializer):
+
+    queue = serializers.SerializerMethodField()
+    reference = serializers.ReadOnlyField(source="application.reference_code")
+    control_list_entries = serializers.SerializerMethodField()
+    wassenaar = serializers.SerializerMethodField()
+    report_summary = serializers.ReadOnlyField(source="good.report_summary")
```
saruniitr, line 740: I think the application report summary is specific to `GoodOnApplication` but not the underlying good. This field is available in both models though.

api/goods/serializers.py
```diff
@@ -728,3 +729,48 @@ def update(self, instance, validated_data):
...
+            "wassenaar",
+            "submitted_at",
+        )
+
+    def get_queue(self, obj):
+        queue = obj.application.queues.first()
+        return queue.id if queue else None
+
+    def get_control_list_entries(self, obj):
+        return [cle.rating for cle in obj.get_control_list_entries().all()]
```
saruniitr, line 767: Can this be used here?
https://github.com/uktrade/lite-api/blob/master/api/staticdata/control_list_entries/serializers.py#L6

alixedi: We can but I reckon the current implementation is simpler because it is not nested and we only need ratings on the FE from this endpoint.

saruniitr: :+1:

PR #982, opened 2022-04-22
LTD-1437-onward-export-save

depsiatwal:


api/applications/serializers/good.py
```diff
@@ -203,6 +203,10 @@ def create(self, validated_data):
         if validated_data.get("firearm_details"):
             # copy the data from the "firearm detail on good" level to "firearm detail on good-on-application" level
             firearm_data = model_to_dict(validated_data["good"].firearm_details)
+            # since in the instance no manufacture date is set it retrieves as null. This is requires to avoid a
+            # validation error as this is an optional question when saving against a good application
+            if "year_of_manufacture" in firearm_data and firearm_data["year_of_manufacture"] is None:
+                del firearm_data["year_of_manufacture"]
```
saruniitr, line 209: My understanding was with django forms we are not doing this type of validation etc in the API but keep it in FE only.

kevincarrogan (approved):

api/applications/serializers/good.py
depsiatwal: this is a work around from a copy of firearm_details which gets copied as null, 
hence when we try to reinsert we get None validation exception.

PR #989, opened 2022-04-28
LTD-2227: Allow empty alias in csv data

wkeeling:
Permit seed data to have an empty `alias` column. Certain teams and queues don't have an alias set.

### DIFF
--- a/api/staticdata/management/commands/seedinternaldemodata.py
+++ b/api/staticdata/management/commands/seedinternaldemodata.py
@@ -37,6 +37,8 @@ class Command(SeedCommand):
         teams = {}
 
         for row in rows:
+            if not row.get("alias"):
+                row["alias"] = None
             team = Team.objects.filter(name__iexact=row["name"])
             if not team.exists():
                 team_id = Team.objects.create(**row).id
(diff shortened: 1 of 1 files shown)


api/staticdata/management/commands/seedinternaldemodata.py
```diff
@@ -37,6 +37,8 @@ def seed_teams(cls) -> dict:
         teams = {}
 
         for row in rows:
+            if not row.get("alias"):
```
saruniitr, line 40: this can be a problem correct? is it not better to update data to setup an alias?
Could you give an example of where this can happen?

wkeeling: I took a snapshot of the teams and queues we have setup in `devdata` and dumped them as csv. Many of those don't use an alias (the alias is just NULL in the database). So this allows us to import the data in the same state, e.g. for the UI tests.

wkeeling: But actually perhaps the UI tests aren't dependent on the data with NULL aliases anyway, so maybe better to strip them out of the seed data completely. I'll try that instead.

PR #990, opened 2022-04-28
LTD-2228: Add command to seed routing rules

wkeeling:
Add management command to seed routing rules for UI tests.

### DIFF
--- a/api/staticdata/management/commands/seedroutingrules.py
+++ b/api/staticdata/management/commands/seedroutingrules.py
@@ -1,0 +1,50 @@
+import uuid
+
+from django.db import transaction
+
+from api.cases.models import CaseStatus
+from api.queues.models import Queue
+from api.teams.models import Team
+from api.staticdata.management.SeedCommand import SeedCommand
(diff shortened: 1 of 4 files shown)


api/staticdata/management/commands/seedroutingrules.py
```diff
@@ -0,0 +1,50 @@
...
+from django.db import transaction
+
+from api.cases.models import CaseStatus
+from api.queues.models import Queue
+from api.teams.models import Team
+from api.staticdata.management.SeedCommand import SeedCommand
+from api.workflow.routing_rules.models import RoutingRule
+
+
+ROUTING_RULES_FILE = "lite_content/lite_api/routing_rules.csv"
```
saruniitr, line 12: we also need a way to keep it upto date in case if any rules are added/edited.

wkeeling: Yes agreed. I'd originally just dumped the csv using my SQL client, but would be better to have management command to do that.

saruniitr (approved):

PR #994, opened 2022-05-03
 LTD-2275-uk-sanctions-pyexcel-to-xml 

depsiatwal:


Pipfile
```diff
@@ -58,8 +58,6 @@ gunicorn = "~=20.0"
 gevent = "~=20.6"
 bleach = "~=3.3"
 xmltodict = "~=0.12.0"
-pyexcel = "==0.6.6"
```
wkeeling, line 61: Would it be more flexible to allow minor version updates rather than fixing at `0.6.6`, or does it have to be fixed?

depsiatwal: pyexcel is being removed in PR

api/external_data/management/commands/ingest_sanctions.py
```diff
@@ -41,20 +40,15 @@ def get_office_financial_sanctions_implementation():
...
-    for sheet_name in book.sheet_names():
-        records = iter(book[sheet_name])
-        # Top 2 lines is just meta
-        next(records)
-        next(records)
-        headers = next(records)
-        for row in records:
-            data = dict(zip(headers, row))
-            yield {**data, "sheet": sheet_name}
+    response = requests.get(settings.SANCTION_LIST_SOURCES["uk_sanctions_file"])
```
wkeeling, line 43: If the response status code is `404` or some other non OK status, I wonder whether it would be worth printing that and exiting rather than trying to proceed - since subsequent code would blow up.

api/external_data/management/commands/ingest_sanctions.py
```diff
@@ -174,30 +169,36 @@ def populate_uk_sanctions_list(self):
...
+                    filtered_name = filter(
+                        lambda n: n.get("nametype", "").lower() == "primary name",
+                        item["names"]["name"],
                     )
-                    postcode = normalize_address(item["Address Postal Code"])
-                    if postcode not in normalize_address(address):
-                        address += " " + postcode
-                    name = join_fields(item, fields=["Name 1", "Name 2", "Name 3", "Name 4", "Name 5", "Name 6"])
-                    unique_id = hash_values([item["Unique ID"], name, address, postcode, item["Regime Name"]])
+                    primary_name = list(filtered_name)[0]
```
wkeeling, line 184: Could simplify 180 - 184 with:
```python
primary_name = next(filter(
    lambda n: n.get("nametype", "").lower() == "primary name",
    item["names"]["name"],
))
```

api/external_data/management/commands/ingest_sanctions.py
depsiatwal: ahh good pint added raise.status_for_status()

api/external_data/management/commands/ingest_sanctions.py
depsiatwal: done :-)

wkeeling (approved):

PR #1000, opened 2022-05-13
 LTD-2237-upload-internal-docs 

depsiatwal:


api/applications/models.py
```diff
@@ -433,6 +433,19 @@ class GoodOnApplicationDocument(Document):
     good_on_application = models.ForeignKey(GoodOnApplication, on_delete=models.CASCADE, blank=True, null=True)
 
 
+class GoodOnApplicationInternalDocument(Document):
+
+    document_title = models.TextField(
+        max_length=200,
```
saruniitr, line 439: I think in UCD we decided to not restrict the user in any way

api/applications/serializers/good.py
```diff
@@ -301,3 +308,35 @@ class GoodOnApplicationDocumentViewSerializer(serializers.Serializer):
...
     def get_s3_key(self, instance):
         return instance.s3_key if instance.safe else "File not ready"
+
+
+class GoodOnApplicationInternalDocumentCreateSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = GoodOnApplicationInternalDocument
+
+        fields = (
+            "name",
```
saruniitr, line 318: we also need `safe` field from the `Document`, we only allow downloads if the document is safe

api/goods/views.py
```diff
@@ -576,3 +582,45 @@ def delete(self, request, pk, doc_pk):
                 good_on_application.delete()
 
         return JsonResponse({"document": "deleted success"})
+
+
+class DocumentGoodOnApplicationInternalView(APIView):
+    authentication_classes = (SharedAuthentication,)
+    serializer_class = GoodOnApplicationInternalDocumentCreateSerializer
+
+    def post(self, request, goods_on_application_pk):
```
saruniitr, line 591: Because this is internal document, lets check the user creating them is also an internal user

api/goods/views.py
```diff
@@ -576,3 +582,45 @@ def delete(self, request, pk, doc_pk):
...
+
+        return JsonResponse({"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+
+class DocumentGoodOnApplicationInternalDetailView(APIView):
+    authentication_classes = (SharedAuthentication,)
+    serializer_class = GoodOnApplicationInternalDocumentViewSerializer
+
+    def get(self, request, doc_pk):
+        document = get_object_or_404(GoodOnApplicationInternalDocument.objects.all(), pk=doc_pk)
```
saruniitr, line 607: Similarly, is Exporter allowed to retrieve/update this document? if not then we should check user and report error accordingly.

PR #1004, opened 2022-05-18
Copy over certificate number if available when creating product on ap…

kevincarrogan:
…plication

### DIFF
--- a/api/goods/helpers.py
+++ b/api/goods/helpers.py
@@ -106,7 +106,9 @@ def update_firearms_certificate_data(organisation_id, firearm_data):
         return firearm_data
 
     if not has_valid_rfd:
-        if has_valid_section5:
+        # If we have a value for the certificate number then we use it.
+        # If it's not set then we just discard the empty value.
+        if has_valid_section5 and not firearm_data.get("section_certificate_number"):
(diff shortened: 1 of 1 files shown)


api/goods/helpers.py
```diff
@@ -106,7 +106,9 @@ def update_firearms_certificate_data(organisation_id, firearm_data):
         return firearm_data
 
     if not has_valid_rfd:
-        if has_valid_section5:
+        # If we have a value for the certificate number then we use it.
+        # If it's not set then we just discard the empty value.
+        if has_valid_section5 and not firearm_data.get("section_certificate_number"):
```
saruniitr, line 111: Not sure this is really required, if there is a valid section5 then we would've have asked the question again so certificate number is empty anyway.

kevincarrogan: This is necessary because I'm fixing a bug where this wasn't being populated correctly.

The problem is that this information needs to be copied across from the firearm_details on the good over to the good_on_application and we're leaving this off during that copy.

saruniitr: ok

saruniitr (approved):

PR #1010, opened 2022-05-30
prefill ES in order for tests based on it will work during CI

sekharpanja:
prefill ES in order for tests based on it will work during CI


start-e2e.sh
```diff
@@ -8,5 +8,8 @@ EXPORTER_USERS='[{"email"=>"foo@bar.com"}]' /app/manage.py seedexporterusers
 # Rebuild ES index
 LITE_API_ENABLE_ES=true /app/manage.py search_index --rebuild -f
 
+# Fill sanctions in ES
+LITE_API_ENABLE_ES=true /app/manage.py ingest_sanctions
```
saruniitr, line 12: I think this should be before the above re-index step.
Once we ingest sanctions data we need to re-index.

saruniitr (approved):

PR #1011, opened 2022-05-31
Add alias to TAU.

alixedi:


api/teams/migrations/0007_auto_20220531_1344.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.12 on 2022-05-31 12:44
+
+from django.db import migrations
+
+
+def populate_dev_team_alias(apps, schema_editor):
+    teams = {
+        "TAU": "TAU",
```
saruniitr, line 8: In UAT the team name is `Technical Assessment Unit` so we need to define alias for that

saruniitr (approved):

PR #1012, opened 2022-06-06
 LTD-2395-store-user-external-id 

depsiatwal:


api/users/tests/test_exporter_authentication.py
```diff
@@ -12,7 +12,11 @@ def test_authentication_success(self):
         """
         Authorises user then checks the token which is sent is valid upon another request
         """
-        data = {"email": self.exporter_user.email, "user_profile": {"first_name": "Matt", "last_name": "Berninger"}}
+        data = {
+            "email": self.exporter_user.email,
+            "sub": "xyz123456",
+            "user_profile": {"first_name": "Matt", "last_name": "Berninger"},
+        }
```
kevincarrogan, line 19: It would be good to add some assertions here to check that the `sub` is being saved to the user object.

api/users/views.py
```diff
@@ -51,6 +51,7 @@ def post(self, request, *args, **kwargs):
         data = request.data
         first_name = data.get("user_profile", {}).get("first_name", "")
         last_name = data.get("user_profile", {}).get("last_name", "")
+        external_id = data.get("sub")
```
saruniitr, line 54: if the key `sub` is not available then we get `None` but `external_id` is a `CharField` so may be use a default of `""`?

api/users/tests/test_exporter_authentication.py
```diff
@@ -12,17 +13,25 @@ def test_authentication_success(self):
...
 
         response = self.client.post(self.url, data)
         response_data = response.json()
-
+        updated_user = ExporterUser.objects.all().last()
```
saruniitr, line 24: we have Exporter email so better to use that and retrieve the correct user

PR #1018, opened 2022-06-14
LTD-2430: Pre-define Agent role so exporters can use this for new members

saruniitr:
## Change description

Exporters currently can define custom roles with selected permissions but that
is a multistep process and can be complicated. To make it easier an 'Agent' role
is predefined so they can quickly add new members and assign this role.


api/core/constants.py
```diff
@@ -43,10 +47,12 @@ class Roles:
     EXPORTER_SUPER_USER_ROLE_NAME = "Super User"
     EXPORTER_DEFAULT_ROLE_ID = UUID("00000000-0000-0000-0000-000000000004")
     EXPORTER_DEFAULT_ROLE_NAME = "Default"
+    EXPORTER_AGENT_ROLE_ID = UUID("00000000-0000-0000-0000-000000000005")
```
depsiatwal, line 50: do we need to add any seed changes for this to work ?

saruniitr: yeah we need to run `seedrolepermissions` which gets run after deployment

depsiatwal (approved):
lgtm

alixedi (approved):

PR #1021, opened 2022-06-20
LTD-2412: Add audit event to show product review and report summary updated details

saruniitr:
## Change description

Currently we emit two separate audit events to show that a product is reviewed
along with its old and new control list entries and also old and new
report summary values. These are also shown in a way which is not readable.

Create new event to combine these two logs and present them in a readble way.
The existing event cannot be used as this requires a new piece of data (line item)
which would be missing from existing events and will give a 'KeyError' hence
a new event is created. This would mean only new cases will use this event and
existing cases display them using two separate events.


api/goods/tests/test_control_codes.py
```diff
@@ -326,53 +329,6 @@ def setUp(self):
         self.case = self.submit_application(self.application)
         self.url = reverse_lazy("goods:control_list_entries", kwargs={"case_pk": self.case.id})
 
-    def test_verify_single_good(self):
```
currycoder, line 329: Why have these tests been removed?  I think I am missing something as they seem unrelated to the change in question

api/goods/views.py
```diff
@@ -100,6 +105,17 @@ def post(self, request, case_pk):
 
         case = get_case(case_pk)
 
+        line_items = {}
+        good_on_application_qs = self.get_queryset()
+
+        try:
```
currycoder, line 111: I feel like this try/except could be narrower and more specific.  e.g. only around the lines that are expected to throw the exception and catching a specific exception (or list of exceptions)

currycoder: In addition, could this block be broken out to a helper?  It seems like this post function is very long already

api/workflow/tests/test_flagging_rules.py
```diff
@@ -96,7 +96,7 @@ def test_goods_flag_before_and_after_reviewing_good(self):
         )
 
         apply_flagging_rules_to_case(case)
-        self.assertEqual(good.flags.count(), 1)
+        self.assertEqual(good.flags.count(), 2)
```
currycoder, line 99: I can't quite tally up what might require this change.  Could you help me to understand why this test needs to be modified?

currycoder (commented):
Just a few minor comments and questions for my understanding

api/goods/tests/test_control_codes.py
saruniitr: Yeah I was going to add a comment about this.

The changes now required `StandardApplication` (https://github.com/uktrade/lite-api/pull/1021/files#diff-ef0780a5a4caf1a7292cda0f0bd6d9953314d7078608568b37e0d9bddf4e0507R112) and these tests were part of `Open` licence application which are not supported yet hence removed.

api/goods/views.py
saruniitr: it is sufficient for the `for` loop but included for retrieving the application also which is really not required because officer is reviewing so application should exist. I will update.

PR #1032, opened 2022-06-27
LTD-2179: Remove few system audit events and replace with summary event

saruniitr:


api/audit_trail/formatters.py
```diff
@@ -155,3 +156,27 @@ def update_product_usage_data(**payload):
...
+    decision = payload["decision"]
+
+    if decision == AdviceType.APPROVE:
+        return "added a decision of licence approved."
+    elif decision == AdviceType.REFUSE:
+        return "added a decision of licence refused."
+    elif decision == AdviceType.NO_LICENCE_REQUIRED:
+        return "added a decision of no licence needed."
+
+    return f"added a decision {decision}."
```
saruniitr, line 171: unlikely to come here but just added a default text

api/cases/generated_documents/views.py
```diff
@@ -120,21 +118,21 @@ def post(self, request, pk):
...
-                )
-
                 s3_operations.upload_bytes_file(raw_file=pdf, s3_key=s3_key)
         except Exception:  # noqa
             return JsonResponse(
                 {"errors": [strings.Cases.GeneratedDocuments.UPLOAD_ERROR]},
                 status=status.HTTP_500_INTERNAL_SERVER_ERROR,
             )
 
+        if advice_type in [AdviceType.REFUSE, AdviceType.NO_LICENCE_REQUIRED]:
```
saruniitr, line 128: In case of approve we generate licence document, asked a query about whether we need to emit event in this case also so once that is clarified then we may add it for approve case as well.

api/cases/libraries/post_advice.py
```diff
@@ -100,9 +99,10 @@ def post_advice(request, case, level, team=False):
         else:
             department = "department"
 
-        audit_trail_service.create(
-            actor=request.user, verb=audit_verbs[level], target=case, payload={"department": department}
-        )
+        if level in audit_verbs:
```
saruniitr, line 102: for 'final' advice we are logging a different event with more details (`CREATED_FINAL_RECOMMENDATION` below)

api/cases/views/views.py
```diff
@@ -830,36 +824,50 @@ def put(self, request, pk):
...
-            verb=AuditType.UPDATED_STATUS,
-            target=case,
-            payload={
-                "status": {"new": case.status.status, "old": old_status},
-                "additional_text": request.data.get("note"),
-            },
-        )
+        decisions = required_decisions.copy()
+        if AdviceType.APPROVE in decisions:
+            decisions.remove(AdviceType.APPROVE)
```
saruniitr, line 862: `approve` is removed here because we are adding that above L827

kevincarrogan (approved):

PR #1042, opened 2022-07-08
[LTD-2510] Email notification unit tests and exporter registration notification

currycoder:


api/organisations/views/organisations.py
```diff
@@ -81,20 +82,27 @@ def post(self, request):
...
-                    verb=AuditType.CREATED_ORGANISATION,
-                    target=get_organisation_by_pk(serializer.data.get("id")),
-                    payload={"organisation_name": serializer.data.get("name")},
-                )
-            else:
-                audit_trail_service.create_system_user_audit(
-                    verb=AuditType.REGISTER_ORGANISATION,
-                    target=get_organisation_by_pk(serializer.data.get("id")),
-                    payload={"email": request.data["user"]["email"], "organisation_name": serializer.data.get("name")},
-                )
```
currycoder, line 97: My understanding of `validate_only` is that with this flag off, we should save the model/fire audit events/fire notifications.  I have changed the indentation of the audit events below to respect that as I believe that there was a possibility for these to fire before save otherwise.

gov_notify/enums.py
```diff
@@ -6,6 +6,7 @@ class TemplateType(Enum):
     ECJU_COMPLIANCE_CREATED = "ecju_compliance_created"
     APPLICATION_STATUS = "application_status"
     ORGANISATION_STATUS = "organisation_status"
+    EXPORTER_REGISTERED_NEW_ORG = "exporter_registered_new_org"
```
saruniitr, line 9: I know `ECJU_CREATED` was already there but it is not clear what that is representing, is it ECJU query created? may be better to rename it.

saruniitr (approved):
looks good to me

PR #1043, opened 2022-07-12
Add explicit tests for date based functions to handle discrepancies on the date the tests are run

kevincarrogan:
We were getting non-deterministic code coverage results, even when testing against an empty commit there were cases where the code coverage went down.

This was due to the fact that there were functions being run that would change execution path depending on the day of the week, in this case if the day before was a bank holiday or a weekend it would affect the code coverage.

This pull request adds explicit tests to cover all of the possible execution paths of these date based functions.

### DIFF
--- a/api/cases/tasks.py
+++ b/api/cases/tasks.py
@@ -48,16 +48,21 @@ def today(time=None):
     return datetime.combine(timezone.localtime(), time, tzinfo=tz(settings.TIME_ZONE))
 
 
-def yesterday(date=timezone.localtime(), time=None):
+def yesterday(date=None, time=None):
     """
     returns the previous working day from the date provided (defaults to now) at the time provided (defaults to now)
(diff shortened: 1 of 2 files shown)


api/cases/tasks.py
```diff
@@ -48,16 +48,21 @@ def today(time=None):
     return datetime.combine(timezone.localtime(), time, tzinfo=tz(settings.TIME_ZONE))
 
 
-def yesterday(date=timezone.localtime(), time=None):
+def yesterday(date=None, time=None):
     """
     returns the previous working day from the date provided (defaults to now) at the time provided (defaults to now)
     """
+    if not date:
```
currycoder, line 55: Good catch. I'm surprised this hasn't burned us sooner..

currycoder (approved):

PR #1045, opened 2022-07-13
[LTD-2511] Notify exporter added to organisation

currycoder:
This change ensures that an exporter user is notified by email when they are added to an organisation.

This uses the following template: https://www.notifications.service.gov.uk/services/f3d8fb42-a34b-4d85-8ac2-a62006a197dc/templates/c9b67dca-0916-453a-99c0-70ba563e1bdd

TODO on merge: Ensure that EXPORTER_BASE_URL and CASEWORKER_BASE_URL are set properly in lite-api's production vault variables.  The default will not work since the URL has private-beta in the name rather than {{env}}

### DIFF
--- a/api/conf/settings.py
+++ b/api/conf/settings.py
@@ -31,6 +31,7 @@ env = Env(
     STREAM_PAGE_SIZE=(int, 20),
     ENV=(str, "localhost"),
     EXPORTER_BASE_URL=(str, ""),
+    CASEWORKER_BASE_URL=(str, ""),
     GOV_NOTIFY_ENABLED=(bool, False),
     DOCUMENT_SIGNING_ENABLED=(bool, False),
 )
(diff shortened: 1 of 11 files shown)


api/users/views.py
```diff
@@ -111,13 +118,18 @@ def post(self, request):
...
-        data["organisation"] = get_request_user_organisation_id(request)
+        organisation = get_request_user_organisation(request)
+        data["organisation"] = organisation.pk
         data["role"] = UUID(data["role"])
 
         serializer = ExporterUserCreateUpdateSerializer(data=data)
 
         if serializer.is_valid():
             serializer.save()
+            notify.notify_exporter_user_added(
```
saruniitr, line 129: do we not need to add Exporter details to the payload?

saruniitr: I guess we are adding url to get the environment details?

api/conf/settings.py
```diff
@@ -373,11 +374,16 @@
...
 
-# If EXPORTER_BASE_URL is not provided, render the base_url using the environment
+# If EXPORTER_BASE_URL is not in env vars, build the base_url using the environment
 EXPORTER_BASE_URL = (
     env("EXPORTER_BASE_URL") if env("EXPORTER_BASE_URL") else f"https://exporter.lite.service.{ENV}.uktrade.digital"
 )
 
+# If CASEWORKER_BASE_URL is not in env vars, build the base_url using the environment
+CASEWORKER_BASE_URL = (
+    env("CASEWORKER_BASE_URL") if env("CASEWORKER_BASE_URL") else f"https://internal.lite.service.{ENV}.uktrade.digital"
```
alixedi, line 384: (Please double check) I reckon you could get away with -

```
env("CASEWORKER_BASE_URL") or f"https://internal.lite.service.{ENV}.uktrade.digital"
```

alixedi (approved):
One minor comment that I will leave with you,.

sekharpanja (approved):

PR #1046, opened 2022-07-13
[LTD-2514] Notify exporter of ecju query

currycoder:
This change ensures that exporter users are notified by email when a query is opened on their application.

The following template is used: https://www.notifications.service.gov.uk/services/f3d8fb42-a34b-4d85-8ac2-a62006a197dc/templates/84431173-72a9-43a1-8926-b43dec7871f9

### DIFF
--- a/api/goods/tests/test_serializers.py
+++ b/api/goods/tests/test_serializers.py
@@ -4,8 +4,8 @@ from api.applications.tests.factories import (
     GoodOnApplicationFactory,
     StandardApplicationFactory,
 )
-from api.parties.tests.factories import PartyFactory
 from api.staticdata.countries.factories import CountryFactory
+from api.parties.tests.factories import PartyFactory
 from test_helpers.clients import DataTestClient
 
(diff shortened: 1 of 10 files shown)


api/goods/tests/test_serializers.py
```diff
@@ -42,6 +42,9 @@ def test_get_destinations(self):
 
         party_on_application = PartyOnApplicationFactory.create(
             application=self.application,
+            party__country__id="IT",
```
currycoder, line 45: Seems like for whatever reason, removing some testcases from a different app caused the assertion at line 64 below to fail with `"Italy" == "Italy"`.  Not 100% sure why but I am guessing something non-deterministic was happening with factory boy's iterator which lead to the same country being set between these two PartyOnApplication objects (and the test explicitly depended on them being different countries).

In any case, I have made the selection of two different countries explicit here.

alixedi (approved):
Beautiful. Thank you!

One minor comment that I will leave with you - would it help if we factor out the query into a `get_case_info` function?

PR #1050, opened 2022-07-18
[LTD-2526] Add celery

currycoder:
This change adds celery to lite-api.  

There are a few categories of changes here:
- Redis has been added as a dependency for lite-api.
- Celery library has been added, with a runner which automatically detects celery tasks under `{APP_NAME}/celery_tasks.py` (this avoids collisions with tasks.py which we are already using for django-background-tasks)
- This has been added to the lite-api-dev PAAS environment so far (see instructions for spinning redis/celery up in PAAS here: https://uktrade.atlassian.net/browse/LTD-2526)
- A new celery container will run when running the app through docker-compose.
- Instructions have been added for running celery/redis "on the metal" for non-docker working copies.
- Local running is managed by "watchmedo" which has been installed as a dev dependency; it gives us automatic reloads for celery on source changes just like manage.py runserver.
- Some simple debug tasks were added to allow testing of celery across all environments.

To play with this on the lite-api-dev PAAS environment, follow the instructions for running debug async tasks in the final bullet here: https://uktrade.atlassian.net/browse/LTD-2526

To play with this locally:
- Run `docker-compose build` to install new pip deps
- Run `docker-compose up -d` to bring the new container up
- Run `docker exec -it lite-api_api_1 python manage.py shell` and then;
```
from api.core import celery_tasks
res = celery_tasks.debug_add.apply_async([1,2])
res.get() # should return 3
res = celery_tasks.debug_count_cases.apply_async()
res.get() # should return the count of Case objects for this environment
res = celery_tasks.debug_exception.apply_async()
res.get() # should raise an exception
```

**Rollout**
To roll this out across environments, we will need to do a slightly special set of deployments which follow the bullets laid out in the comment here: https://uktrade.atlassian.net/browse/LTD-2526

### DIFF
--- a/api/conf/settings.py
+++ b/api/conf/settings.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+from urllib.parse import urlencode
 
 from environ import Env
 import sentry_sdk
@@ -246,6 +247,29 @@ else:
     AWS_SECRET_ACCESS_KEY = env("AWS_SECRET_ACCESS_KEY")
     AWS_REGION = env("AWS_REGION")
     AWS_STORAGE_BUCKET_NAME = env("AWS_STORAGE_BUCKET_NAME")
+
+if "redis" in VCAP_SERVICES:
+    REDIS_BASE_URL = VCAP_SERVICES["redis"][0]["credentials"]["uri"]
+else:
+    REDIS_BASE_URL = env("REDIS_BASE_URL", default=None)
+
+
+def _build_redis_url(base_url, db_number, **query_args):
+    encoded_query_args = urlencode(query_args)
+    return f"{base_url}/{db_number}?{encoded_query_args}"
+
+
+if REDIS_BASE_URL:
+    # Give celery tasks their own redis DB - future uses of redis should use a different DB
+    REDIS_CELERY_DB = env("REDIS_CELERY_DB", default=0)
+    is_rediss = REDIS_BASE_URL.startswith("rediss://")
+    url_args = {"ssl_cert_reqs": "CERT_REQUIRED"} if is_rediss else {}
+
+    CELERY_BROKER_URL = _build_redis_url(REDIS_BASE_URL, REDIS_CELERY_DB, **url_args)
+    CELERY_RESULT_BACKEND = CELERY_BROKER_URL
+
+CELERY_TASK_ALWAYS_EAGER = env.bool("CELERY_TASK_ALWAYS_EAGER", False)
+CELERY_TASK_SEND_SENT_EVENT = env.bool("CELERY_TASK_SEND_SENT_EVENT", True)
 
 S3_CONNECT_TIMEOUT = 60  # Maximum time, in seconds, to wait for an initial connection
 S3_REQUEST_TIMEOUT = 60  # Maximum time, in seconds, to wait between bytes of a response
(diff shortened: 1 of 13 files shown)


api/core/celery_tasks.py
```diff
@@ -0,0 +1,26 @@
+from celery import shared_task
+from api.cases.models import Case
```
currycoder, line 2: Since these are debug tasks I would like to mark them as no-cover to get rid of coverage complaints...  But I can't see a way at the moment to mark a part of a file as ignore coverage.  Would appreciate any ideas here?  I don't want to add the whole file to .coveragerc

kevincarrogan: `# pragma: no cover` can do individual lines or code blocks e.g. functions or if statements.

kevincarrogan (approved):

api/conf/settings.py
```diff
@@ -246,6 +247,29 @@
...
+
+def _build_redis_url(base_url, db_number, **query_args):
+    encoded_query_args = urlencode(query_args)
+    return f"{base_url}/{db_number}?{encoded_query_args}"
+
+
+if REDIS_BASE_URL:
+    # Give celery tasks their own redis DB - future uses of redis should use a different DB
+    REDIS_CELERY_DB = env("REDIS_CELERY_DB", default=0)
+    is_rediss = REDIS_BASE_URL.startswith("rediss://")
```
alixedi, line 264: `is_redis`?

alixedi: Generally the intricacy here makes me slightly nervous. 

I wonder if there is room for a fallback mechanism that guarantees that a misconfiguration would just mean Celery won't run and not a deployment failure or worst - the application won't run.

I also wonder if we can perhaps move some of this off settings and in [here](https://github.com/uktrade/lite-api/pull/1050/commits/858b3fa8300aab452c84e59fec22bec38cb784e9#diff-ddcb575c7a38824b2cad80b976f877fcc39f1a409328cf71393d17113f403b0eR8) by passing the URL directly into the `Celery` constructor? (not sure if it is supported).

currycoder: This is a good question.  I will have a look at what PAAS does if the redis URL is mangled in some way and get back to you.

But my understanding is that the following will happen:

For our web process (which runs in a separate cf instance), the app will start perfectly fine.  Calls to `task.apply_asnyc()` will make the web request being serviced yield a 500 error.  It's up to us to handle the exception in some way (or not)

For the celery process (again, a separate/airgapped cf instance), the app will fail to start.  Here's where I'm unclear on how PAAS will deal with things.

As for moving settings around, what benefit do you think that might bring?  I kind of like settings.py as the "one stop shop" for these bits.

currycoder: > `is_redis`?

`is_rediss` is a bit :snake: - I'll make it `is_ssl_connection` to be more readable :)

currycoder: So I tried this out @alixedi - I deliberately mangled the REDIS_BASE_URL in the dev environment and found the following:
- The API app deployed successfully and worked fine - I was able to browse around caseworker/exporter apps with no issues.
- `celeryworker` was stuck in a "Running" -> "Crashed" -> "Starting" loop according to cf CLI.  Cloudfoundry just repeatedly brought the doomed celeryworker instance back up.  We also got an error logged to sentry.

I would argue that based on this testing, this is bulletproof enough for what we need.  The api app is not clobbered (although we might want to think about exceptions raised by `apply_async()`) and we have signals in sentry to help us debug the issue.

What do you think?

kevincarrogan: "We also got an error logged to sentry." I think that would be enough for me.

PR #1054, opened 2022-07-21
Send a notification when an application is opened for editing to expo…

kevincarrogan:
…rter

### DIFF
--- a/api/applications/notify.py
+++ b/api/applications/notify.py
@@ -1,0 +1,25 @@
+from api.core.helpers import get_exporter_frontend_url
+from gov_notify.enums import TemplateType
+from gov_notify.payloads import ExporterCaseOpenedForEditing
+from gov_notify.service import send_email
+
+
+def _notify_exporter_case_opened_for_editing(email, data):
+    payload = ExporterCaseOpenedForEditing(**data)
+    send_email(
+        email,
+        TemplateType.EXPORTER_CASE_OPENED_FOR_EDITING,
+        payload,
+    )
+
+
+def notify_exporter_case_opened_for_editing(application):
+    exporter = application.submitted_by
+    _notify_exporter_case_opened_for_editing(
+        exporter.email,
+        {
+            "user_first_name": exporter.first_name,
+            "application_reference": application.reference_code,
+            "exporter_frontend_url": get_exporter_frontend_url("/"),
+        },
+    )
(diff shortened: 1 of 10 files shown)


api/applications/views/applications.py
```diff
@@ -526,10 +527,12 @@ def put(self, request, pk):
             },
         )
 
-        # Case routing rules
         if old_status != application.status:
             run_routing_rules(case=application, keep_status=True)
 
+            if application.status == get_case_status_by_status(CaseStatusEnum.REOPENED_FOR_CHANGES):
```
currycoder, line 533: Is there a field we can check rather than doing a whole new DB query (which I believe get_case_status_by_status is doing)

api/applications/tests/test_application_status.py
```diff
@@ -322,6 +322,21 @@ def test_gov_set_status_when_they_have_permission_to_reopen_closed_cases_success
             self.standard_application.status, get_case_status_by_status(CaseStatusEnum.REOPENED_FOR_CHANGES)
         )
 
+    @mock.patch("api.applications.views.applications.notify_exporter_case_opened_for_editing")
```
currycoder, line 325: Something for discussion. I'm curious about why you chose to create a whole new test for this instead of extending the one above?

api/applications/notify.py
```diff
@@ -0,0 +1,25 @@
...
+    payload = ExporterCaseOpenedForEditing(**data)
+    send_email(
+        email,
+        TemplateType.EXPORTER_CASE_OPENED_FOR_EDITING,
+        payload,
+    )
+
+
+def notify_exporter_case_opened_for_editing(application):
+    exporter = application.submitted_by
```
currycoder, line 17: As Gurdeep flagged in slack today, should we check the user's status here before sending the notification?

api/applications/tests/test_application_status.py
kevincarrogan: I think I just wanted to write one that had a better name and explicitly showed that a notification was sent instead of overloading the previous test.

api/applications/views/applications.py
kevincarrogan: I don't think that extra query is necessarily something to worry about and this is kind of consistent with how it's done elsewhere. I can change it though but it feels like an over optimisation.

api/applications/notify.py
kevincarrogan: I thought we were doing that as a separate ticket (and kind of hoping it wouldn't be something we do individually anyway but have a way of checking that without the caller having to check it each time).

api/applications/views/applications.py
currycoder: I'm fairly easy.  I think the thing that switched me on to it was that it just looked a bit odd.  But I'm happy to defer to your preference; especially with it being consistent in other areas.

api/applications/notify.py
currycoder: That's an approach I hadn't considered, fair enough :+1:

currycoder (approved):

PR #1057, opened 2022-07-21
LTD-2513-email-account-registration

depsiatwal:


api/organisations/tests/test_organisations.py
```diff
@@ -160,6 +164,8 @@ def test_create_commercial_organisation_as_internal_success(self, mocked_notify_
 
         # Ensure that exporters are not notified when an internal user creates an organisation
         assert not mocked_notify_exporter_registration.called
+        # Caseworkers are still notified
+        assert mocked_caseworker_new_registration.called
```
kevincarrogan, line 168: Can you bulk this out to be an `assert_called_with`?

api/organisations/views/organisations.py
```diff
@@ -86,6 +86,14 @@ def post(self, request):
             if not validate_only:
                 serializer.save()
 
+                # Notify caseworkers of new registration
+                notify.notify_caseworker_new_registration(
+                    {
+                        "applicant_email": serializer.validated_data["user"]["email"],
```
currycoder, line 92: Is it OK from a data protection perspective to pop this email in to the notification? My feeling is that when we send this, we no longer have a control on the deletion of this data etc. Could be better to just supply a link to the thing that the caseworker needs to action

depsiatwal: sorry not sure I understand the point 
just for clarification this is the payload to the gov notify template application_email is used in the template

api/organisations/tests/test_organisations.py
depsiatwal: ahh yeah was thinking that too put got lazy as we do it else where 
DONE

api/conf/settings.py
```diff
@@ -443,3 +443,4 @@
         "uk_sanctions_file": "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1071293/UK_Sanctions_List.xml",
     },
 )
+LITE_INTERNAL_NOTIFICATION_EMAILS = env.json("LITE_INTERNAL_NOTIFICATION_EMAILS", {})
```
saruniitr, line 446: Further below https://github.com/uktrade/lite-api/pull/1057/files#diff-7442d34a43e5b7697f726fabcc17b86963d12359a0a14ca8756f1586a2208468R63 it is expected as a `list` but the default here is `{}`

gov_notify/enums.py
```diff
@@ -8,8 +8,9 @@ class TemplateType(Enum):
     EXPORTER_LICENCE_ISSUED = "exporter_licence_issued"
     EXPORTER_ORGANISATION_APPROVED = "exporter_organisation_approved"
     EXPORTER_ORGANISATION_REJECTED = "exporter_organisation_rejected"
-    EXPORTER_ECJU_QUERY = "exporter_ecju_query"
+    CASEWORKER_REGISTERED_NEW_ORG = "caseworker_registered_new_org"
```
saruniitr, line 11: Not sure if the naming represents the requirement correctly - Exporter registered new organisation and Caseworker reviewed and approved it but this seems to imply Caseworker registered it.

depsiatwal: or it could imply caseworker is receiving new registration notification 
I would say this doesn't imply the action but the notification guess there is no official naming standard

gov_notify/enums.py
```diff
@@ -10,8 +10,9 @@ class TemplateType(Enum):
     EXPORTER_LICENCE_REVOKED = "exporter_licence_revoked"
     EXPORTER_ORGANISATION_APPROVED = "exporter_organisation_approved"
     EXPORTER_ORGANISATION_REJECTED = "exporter_organisation_rejected"
-    EXPORTER_ECJU_QUERY = "exporter_ecju_query"
+    CASEWORKER_REGISTERED_NEW_ORG = "caseworker_registered_new_org"
     EXPORTER_CASE_OPENED_FOR_EDITING = "exporter_editing"
+    EXPORTER_ECJU_QUERY = "exporter_ecju_query"
```
saruniitr, line 15: this is not relevant to this, why is this added in this?

depsiatwal: this was a result of a bad merge so was removed by accident and guess PR is picking up as added

saruniitr (approved):

PR #1058, opened 2022-07-22
LTD-2529: Add Exporter notification when licence is revoked

saruniitr:


api/cases/notify.py
```diff
@@ -51,6 +51,28 @@ def notify_exporter_licence_refused(case):
     )
 
 
+def _notify_exporter_licence_revoked(email, data):
```
depsiatwal, line 54: any rational of why a private method is required ?

depsiatwal (approved):
just small comment and need tests fixing other then that LGTM

api/cases/notify.py
saruniitr: I've just followed the convention from previous examples.

depsiatwal: yeah seen that in other PR still wonder why the _ methods :-)

PR #1061, opened 2022-07-25
Only send licence issued notification when the recommendation was "ap…

kevincarrogan:
This only sends licence issue notification when a licence is actually approved.

There are other scenarios where our code effectively "issues" a licence but it may have been for a different reason where the licence, in real terms, isn't actually issued.

This makes it explicit that we only want to send this notification when the licence is issued (approved in this case).

### DIFF
--- a/api/cases/views/views.py
+++ b/api/cases/views/views.py
@@ -801,9 +801,6 @@ class FinaliseView(UpdateAPIView):
             licence.decisions.set([Decision.objects.get(name=decision) for decision in required_decisions])
             licence.issue()
 
-            # Send email notification for licence issued
-            notify_exporter_licence_issued(licence)
-
             return_payload["licence"] = licence.id
             if Licence.objects.filter(case=case).count() > 1:
                 audit_trail_service.create(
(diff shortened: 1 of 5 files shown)


api/cases/views/views.py
```diff
@@ -825,6 +822,16 @@ def put(self, request, pk):
         case.save()
 
         decisions = required_decisions.copy()
+
+        if AdviceType.REFUSE in decisions:
```
currycoder, line 826: Just to check, can multiple different AdviceTypes be present in decisions?  e.g. both NLR and Approve..?

currycoder (approved):

api/cases/views/views.py
kevincarrogan: Yes. There can be multiple different decisions for different products on a licence.

At the moment, given the "optimise for speed" delivery approach we're taking this could result in a user getting multiple separate emails (it was the easiest to code).

There is a ticket for us to actually just generate a single email with different content depending on the combination of decisions made but we can tackle that in the future (Notify also gives us pretty limited capabilities in dynamically deciding what to show and what not to show so not sure exactly what might be possible here).

saruniitr: Yes, currently it is either "Approve all" or "Refuse all" and the remaining are NLR.
The other option is "Approve some" and "Refuse some" which is not implemented yet.

currycoder: Thanks both, that helps my understanding

PR #1062, opened 2022-07-25
LTD-2541: Reset products verification status when an application is copied

saruniitr:
## Change description

When we create new applications by making a copy of an existing application
currently we are retaining the product's verification status. Because of this
when this application is submitted, the products are already shown as
verified without TAU assessing them. Fix this by resetting the verification
status when an application is copied.

### DIFF
--- a/api/applications/views/applications.py
+++ b/api/applications/views/applications.py
@@ -73,6 +73,7 @@ from api.core.helpers import convert_date_to_string, str_to_bool
 from api.core.permissions import assert_user_has_permission
 from api.flags.enums import FlagStatuses, SystemFlags
 from api.flags.models import Flag
+from api.goods.enums import GoodStatus
 from api.goods.serializers import GoodCreateSerializer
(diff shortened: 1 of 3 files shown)


api/applications/views/applications.py
```diff
@@ -758,6 +760,8 @@ def post(self, request, pk):
         self.new_application.last_closed_at = None
         self.new_application.sla_updated_at = None
 
+        self.reset_products_assessment_status(self.new_application)
```
currycoder, line 763: Not 100% sure but could the testcase issue be a genuine problem?  

A few lines later in 769 you save the new application - I'm a little rusty on exactly what that triggers but if it saves related goods then it will save them at the state they were when new_application was copied - e.g. the source application's goods.

So `reset_products_assessment_status()` would write those good objects with DRAFT status, but the `self.new_application.save()` call overwrites that again?

saruniitr:
We are not modifying this behaviour, for nw.

PR #1064, opened 2022-08-01
new cloud build file release GCR trigger

sekharpanja:
This creates a new Cloud build file to use within new GCR trigger to allow adding tag to lite-api image during git release


ejayesh (approved):
approved

PR #1069, opened 2022-08-08
integrate celery and notifications

sekharpanja:
integrate celery and notifications

### DIFF
--- a/api/core/celery_tasks.py
+++ b/api/core/celery_tasks.py
@@ -1,5 +1,11 @@
+import logging
+
 from celery import shared_task
 from api.cases.models import Case
+
+from django.conf import settings
+from notifications_python_client import NotificationsAPIClient
+from notifications_python_client.errors import HTTPError
 
 
 @shared_task
@@ -24,3 +30,14 @@ def debug_exception():
     Debug task which raises an exception.
     """
     raise Exception("debug_exception task")
+
+
+@shared_task
+def send_email(email_address, template_id, data):
+    if not settings.GOV_NOTIFY_ENABLED:
+        logging.info({"gov_notify": "disabled"})
+        return
+
+    return NotificationsAPIClient(settings.GOV_NOTIFY_KEY).send_email_notification(
+        email_address=email_address, template_id=template_id, personalisation=data
+    )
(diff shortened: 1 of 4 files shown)


gov_notify/tests/test_service.py
```diff
@@ -18,6 +21,14 @@ def test_send_email(self, mock_client):
 
         service.send_email(email_address=email, template_type=template_type, data=organisation_status_data)
 
-        mock_client.send_email.assert_called_with(
-            email_address=email, template_id=TemplateType.EXPORTER_REGISTERED_NEW_ORG.template_id, data=data
-        )
+        assert mock_send_email.called
```
kevincarrogan, line 24: I think you could set this back to `assert_called_with` here.

gov_notify/service.py
```diff
@@ -1,18 +1,20 @@
...
+from api.core.celery_tasks import send_email as celery_send_email
 
 
 def send_email(email_address, template_type, data: Optional[EmailData] = None):
     """
-    Send an email using the gov notify service.
+    Send an email using the gov notify service via celery.
     """
+    if not settings.GOV_NOTIFY_ENABLED:
+        logging.info({"gov_notify": "disabled"})
```
kevincarrogan, line 15: Using a named logger is usually better i.e. `logger = logging.getLogger(__name__)`.

api/core/celery_tasks.py
```diff
@@ -24,3 +27,10 @@ def debug_exception():
     Debug task which raises an exception.
     """
     raise Exception("debug_exception task")
+
+
+@shared_task
+def send_email(email_address, template_id, data):
+    return NotificationsAPIClient(settings.GOV_NOTIFY_KEY).send_email_notification(
+        email_address=email_address, template_id=template_id, personalisation=data
+    )
```
kevincarrogan, line 36: Codecov is complaining that this isn't covered. I'm presuming because we've mocked this and because it's a celery task. Is there a way we might be able to put this under test somehow? It seems like kind of an important thing to verify that we're actually calling the `NotificationsAPIClient` correctly.

kevincarrogan (approved):

api/core/celery_tasks.py
sekharpanja: I will add this in a separate PR as discussed.

PR #1073, opened 2022-08-10
LTD-2641: Return current node if there is no parent node

saruniitr:
## Change description

We have a helper function that returns all parent nodes of a given
CLE rating but if the given rating itself is a parent and it doesn't 
have a parent then it doesn't return anything. 
This makes flagging rules fail if that parent is specified as matching_group,
in which case rule doesn't get applied.


saruniitr:
we are going to revisit this later.

PR #1075, opened 2022-08-16
LTD-2645: Create config data statically

saruniitr:
## Change description

Some changes required on API to support creating config data statically in lite_routing


kevincarrogan (approved):

currycoder:
Presumably this will need to merge after https://github.com/uktrade/lite-routing/pull/16 and include a bump to the submodule SHA in order for these migrations to actually take effect

saruniitr:
> Presumably this will need to merge after [uktrade/lite-routing#16](https://github.com/uktrade/lite-routing/pull/16) and include a bump to the submodule SHA in order for these migrations to actually take effect

yes that's correct. Once `lite-routing` PR is merged then I will update `lite-api`.

api/flags/tests/test_case_view_flag_ordering.py
```diff
@@ -34,18 +37,14 @@ def test_gov_user_can_see_all_flags(self):
 
         self.organisation.flags.set([flag for flag in flags if flag.level == FlagLevels.ORGANISATION])
 
-        ordered_flags = get_ordered_flags(case, self.team)
```
saruniitr, line 37: This test was checking the order of flags applied to the case.
The rules in migration started adding two new flags which affected the order.
I am not sure why the order is checked in this test which doesn't really matter as long as flags are present so updated the test to just check for presence of all flags.

api/teams/tests/factories.py
```diff
@@ -3,12 +3,7 @@
 
 
 class DepartmentFactory(factory.django.DjangoModelFactory):
-    name = factory.Iterator(["HMRC", "MOD", "BEIS", "DIT", "NCSC"])
```
saruniitr, line 6: it was complaining that department names already exists (as we create in migrations)

PR #1081, opened 2022-08-22
LTD-2696: Add migration to retrospectively add missing Case UPDATED_STATUS event

saruniitr:
## Change description

Audit UPDATED_STATUS was removed as part of Notes and timeline changes
and for sometime this event is not emitted when a case is finalised.
Reporting relies on this event so we added the event back but for the
missing data this migration retrospectively adds for the cases finalised
between the dates.


api/audit_trail/migrations/0012_add_missing_case_updated_status_events.py
```diff
@@ -0,0 +1,60 @@
...
+                created_at=event.created_at,
+                updated_at=event.updated_at,
+                actor_object_id=event.actor_object_id,
+                actor_content_type=event.actor_content_type,
+                verb=AuditType.UPDATED_STATUS,
+                target_object_id=event.target_object_id,
+                target_content_type=event.target_content_type,
+                payload={
+                    "status": {"new": CaseStatusEnum.FINALISED, "old": CaseStatusEnum.UNDER_FINAL_REVIEW},
+                    "additional_text": "Added retrospectively by LITE",
```
saruniitr, line 45: this `additional_text` is option but added to make it clear that this is added later by the system.
For this reason the `actor` attributes are also reused from `event`. Actor is basically the user that performed this action which in this case issuing licence. Technically this `UPDATED_STATUS` is not from that user as we are adding it later so `additional_text` clarifies this.

api/audit_trail/migrations/0012_add_missing_case_updated_status_events.py
```diff
@@ -0,0 +1,60 @@
...
+        # Reporting relies on this event so we added the event back but for the
+        # missing data this migration retrospectively adds for the cases finalised
+        # between these dates.
+        start_date = datetime.strptime("28-06-2022", "%d-%m-%Y")
+        end_date = datetime.strptime("20-08-2022", "%d-%m-%Y")
+
+        finalised_cases_events = Audit.objects.filter(
+            Q(created_at__date__gte=start_date)
+            & Q(created_at__date__lte=end_date)
+            & Q(verb=AuditType.CREATED_FINAL_RECOMMENDATION)
```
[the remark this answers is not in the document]
kevincarrogan: You don't actually need Q objects for this, just a filter with multiple arguments is the same as `and`. Doesn't really need to be changed though.

api/audit_trail/migrations/0012_add_missing_case_updated_status_events.py
kevincarrogan: I'm presuming this `additional_text` isn't displayed anywhere?

This is a good idea though.

kevincarrogan (approved):

api/audit_trail/migrations/0012_add_missing_case_updated_status_events.py
saruniitr: It is actually displayed in Notes and timeline in a box with grey background.

If we don't want this to be displayed then I think we need to change the `actor` to be of LITE system user instead of `event.actor`

kevincarrogan: Ok. I was just wondering if this is going to confuse users.

Maybe switching the actor is the better thing then.

api/audit_trail/migrations/0012_add_missing_case_updated_status_events.py
```diff
@@ -0,0 +1,64 @@
...
+    with transaction.atomic():
+        # Audit UPDATED_STATUS was removed as part of Notes and timeline changes
+        # and for the dates below this event is not emitted when a case is finalised.
+        # Reporting relies on this event so we added the event back but for the
+        # missing data this migration retrospectively adds for the cases finalised
+        # between these dates.
+        start_date = datetime.strptime("28-06-2022", "%d-%m-%Y")
+        end_date = datetime.strptime("20-08-2022", "%d-%m-%Y")
+
+        system_user_event = Audit.objects.filter(actor_object_id=SystemUser.id).last()
```
saruniitr, line 29: this is required for the `actor_content_type`

PR #1087, opened 2022-08-26
LTD-2716: Invoke flagging rules with python criteria

saruniitr:


api/workflow/flagging_rules_automation.py
```diff
@@ -15,10 +15,12 @@
 from api.staticdata.statuses.enums import CaseStatusEnum
 from api.staticdata.control_list_entries.helpers import get_clc_child_nodes, get_clc_parent_nodes
 
+from lite_routing.routing_rules_internal.flagging_rules_criteria import run_flagging_rules_criteria
+
 
 def get_active_flagging_rules_for_level(level):
     return FlaggingRule.objects.prefetch_related("flag").filter(
-        status=FlagStatuses.ACTIVE, flag__status=FlagStatuses.ACTIVE, level=level
+        status=FlagStatuses.ACTIVE, flag__status=FlagStatuses.ACTIVE, level=level, is_python_criteria=True
```
currycoder, line 23: Won't this mean that **only** `is_python_criteria=True` rules will run?

api/workflow/tests/test_flagging_rules.py
```diff
@@ -376,6 +379,41 @@ def test_apply_verified_goods_only_flagging_rule_to_open_cases_success(self):
         case.refresh_from_db()
         self.assertNotIn(flag, case.flags.all())
 
+    def test_flagging_rule_with_python_criteria_not_met(self):
+        application = self.create_standard_application_case(self.organisation)
+        good = GoodOnApplication.objects.filter(application_id=application.id).first().good
+        good.control_list_entries.set([ControlListEntry.objects.get(rating="ML5a")])
+        good.flags.clear()
+        flagging_rule = FlaggingRule.objects.get(id=FlaggingRulesEnum.SMALL_ARMS)
+        flagging_rule.is_python_criteria = True
```
currycoder, line 388: Can't a simple extra rule be added just for this test case rather than repurposing an existing rule?  In the routing rule cases I believe that I just mocked out the criteria call to override the criteria being met/not met

currycoder: My concern with using an existing rule is that we might want to get rid of it somewhere down the line

api/workflow/flagging_rules_automation.py
saruniitr: yeah, this is not required, removed.

api/workflow/tests/test_flagging_rules.py
saruniitr: ah this is the new rule created for the additional criteria of adding small arms flag. it is not an existing rule. Since the new rule is added in migrations I am retrieving that using the `id`.

currycoder: Yep understood. My point here is that this is unit tests for the flagging rules functions themselves, so it feels a little messy to test them by using a real rule and not a fake one/mocks. Does that make sense?

currycoder: I also realised that by using a "real rule" to test this, we are leaking what the rules do in to this repository which we don't want to do!

api/workflow/flagging_rules_automation.py
```diff
@@ -122,6 +144,12 @@ def apply_goods_rules_for_good(good, flagging_rules: QuerySet = None):
 
     flags = flagging_rules.values_list("flag_id", flat=True)
 
+    flags = list(flags)
```
currycoder, line 147: So these three new blocks are fairly repeated.  Could there be a clean way to factor them out to a helper?  Something like `flags.append(_apply_python_flagging_rules(good, level=FlagLevels.GOOD))`

api/workflow/tests/test_flagging_rules.py
```diff
@@ -376,6 +385,56 @@ def test_apply_verified_goods_only_flagging_rule_to_open_cases_success(self):
...
+            level=FlagLevels.CASE,
+            status=FlagStatuses.ACTIVE,
+            is_python_criteria=True,
+        )
+        case = self.create_standard_application_case(self.organisation)
+        case.flags.clear()
+        self.assertEqual(case.flags.count(), 0)
+        apply_case_flagging_rules(case)
+        assert mocked_criteria_function.called
+        self.assertEqual(case.flags.count(), 1)
```
currycoder, line 403: Could it be asserted that the specific flag is present, not just the count?

api/workflow/tests/test_flagging_rules.py
```diff
@@ -376,6 +385,56 @@ def test_apply_verified_goods_only_flagging_rule_to_open_cases_success(self):
...
+            flag=Flag.objects.first(),
+            level=FlagLevels.GOOD,
+            status=FlagStatuses.ACTIVE,
+            is_python_criteria=True,
+        )
+        good = GoodFactory(organisation=self.organisation)
+        self.assertEqual(good.flags.count(), 0)
+        apply_goods_rules_for_good(good)
+        assert mocked_criteria_function.called
+        self.assertEqual(good.flags.count(), 2)
```
currycoder, line 419: Is this 2 because another flag has applied?  I would have expected it to be 1.

Same comment as above RE: asserting the specific flag

api/workflow/tests/test_flagging_rules.py
```diff
@@ -376,6 +385,56 @@ def test_apply_verified_goods_only_flagging_rule_to_open_cases_success(self):
...
+            level=FlagLevels.DESTINATION,
+            status=FlagStatuses.ACTIVE,
+            is_python_criteria=True,
+        )
+        party_on_application = PartyOnApplicationFactory(party__country__id="US")
+        party = party_on_application.party
+        self.assertEqual(party.flags.count(), 0)
+        apply_destination_rule_on_party(party)
+        assert mocked_criteria_function.called
+        self.assertEqual(party.flags.count(), 2)
```
currycoder, line 436: Just marking that this one is also asserting 2 flags..  I expect the answer would be the same as above

api/workflow/tests/test_flagging_rules.py
saruniitr: I will clarify, usually I add comments but forgot in this instance

api/workflow/tests/test_flagging_rules.py
saruniitr: :+1:

api/workflow/tests/test_flagging_rules.py
saruniitr: It is adding two flags here, one from the rule itself and the other from existing rules. `GoodFactory()` creating `Good` with `ML1a` rating which matches with one of the existing rules and applying that flag

api/workflow/tests/test_flagging_rules.py
saruniitr: Same here, one flag from the rule and the other flag is `Green Countries` flag from the existing rules because of the country used in `PartyOnApplicationFactory()`

api/workflow/flagging_rules_automation.py
saruniitr: :+1:

api/workflow/tests/test_flagging_rules.py
currycoder: Well the problem with this is that as we add more python flagging rules this number needs to increase.  I think this is a signal that this test case isn't very well isolated.

Could it be possible to deactivate all other rules first so that only the dummy rule that is added is active?  I don't think that would have an effect on other tests as the data is reset between test cases.

api/workflow/tests/test_flagging_rules.py
currycoder: Same comment as above

currycoder (approved):

PR #1092, opened 2022-08-31
Ensure queue_id recorded for 'move case' audit event

currycoder:
Record queue_ids for move case audit event (so that we can use this for flagging rules).


api/cases/signals.py
```diff
@@ -0,0 +1,31 @@
...
+
+    original = None
+    try:
+        original = Case.objects.get(pk=instance.id)
+    except Case.DoesNotExist:
+        return
+
+    status_changed = original.status != instance.status
+    status_draft = instance.status == get_case_status_by_status(CaseStatusEnum.DRAFT)
+    if status_changed and not status_draft:
```
saruniitr, line 30: besides 'draft' should we also exclude any of the terminal states?

saruniitr (commented):
just one query, otherwise changes looks good to me.

api/cases/signals.py
currycoder: Yeah that probably makes sense, I will adjust this.

saruniitr (approved):

PR #1100, opened 2022-09-08
LTD-2732: Update test to not rely on static count

saruniitr:


api/queues/tests/test_queues_view.py
```diff
@@ -30,7 +30,7 @@ def test_list_all_queues(self):
         data = response.json()
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(len(data), 54)
+        self.assertEqual(len([q for q in data if q["team"]]), Queue.objects.count())
```
saruniitr, line 33: Filter system queues as `Queue` doesn't contain system queue.

currycoder (approved):

PR #1101, opened 2022-09-09
LTD-2737: Add factories for Denial Matches

saruniitr:


kevincarrogan (approved):

api/staticdata/management/commands/tests/test_commands.py
```diff
@@ -125,3 +127,11 @@ def test_seed_decisions(self):
             self.assertTrue(
                 Decision.objects.filter(id=AdviceType.ids[key], name=key).exists(), f"Decision {key} does not exist"
             )
+
+    @pytest.mark.seeding
+    @mock.patch.dict(os.environ, {"INTERNAL_USERS": "[]"})
```
kevincarrogan, line 132: I wonder if `override_settings` might be better here and might also fix the test that's failing.

kevincarrogan: https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.override_settings

saruniitr: I tried with that initially but for some reason it is only failing on CI, not clear why.

PR #1106, opened 2022-09-15
 LTD-2816-db-changes-security-approval

depsiatwal:


api/applications/models.py
```diff
@@ -226,7 +227,17 @@ class StandardApplication(BaseApplication):
     goods_recipients = models.TextField(choices=GOODS_RECIPIENTS_CHOICES, default="")
     goods_starting_point = models.TextField(choices=GOODS_STARTING_POINT_CHOICES, default="")
 
+    # MOD Security approval fields
     is_mod_security_approved = models.BooleanField(blank=True, default=None, null=True)
```
currycoder, line 231: Just triple checking, but this is where the answer for the first security approval question will be stored, right?

depsiatwal: that's correct :-) it's the one you created

kevincarrogan: Isn't this actually a little redundant though? Can't the value of this be inferred from `security_approvals`?

api/applications/models.py
```diff
@@ -226,7 +227,17 @@ class StandardApplication(BaseApplication):
...
+    security_approvals = ArrayField(
+        models.CharField(choices=SecurityClassifiedApprovalsType.choices, max_length=255), blank=True, null=True
+    )
+    f680_reference_number = models.TextField(default=None, blank=True, null=True, max_length=100)
+    f1686_contracting_authority = models.TextField(default=None, blank=True, null=True, max_length=100)
+    is_f1686_approval_document_available = models.BooleanField(blank=True, default=None, null=True)
+    f1686_reference_number = models.TextField(default=None, blank=True, null=True, max_length=100)
+    f1686_approval_date = models.DateField(blank=False, null=True)
+    other_security_approval_details = models.TextField(default=None, blank=True, null=True, max_length=2500)
```
kevincarrogan, line 240: I think we've tried to move away from having `max_length` for `TextField` entries.

api/applications/models.py
```diff
@@ -226,7 +227,17 @@ class StandardApplication(BaseApplication):
...
 
+    # MOD Security approval fields
     is_mod_security_approved = models.BooleanField(blank=True, default=None, null=True)
+    security_approvals = ArrayField(
+        models.CharField(choices=SecurityClassifiedApprovalsType.choices, max_length=255), blank=True, null=True
+    )
+    f680_reference_number = models.TextField(default=None, blank=True, null=True, max_length=100)
+    f1686_contracting_authority = models.TextField(default=None, blank=True, null=True, max_length=100)
+    is_f1686_approval_document_available = models.BooleanField(blank=True, default=None, null=True)
+    f1686_reference_number = models.TextField(default=None, blank=True, null=True, max_length=100)
```
kevincarrogan, line 238: I think this makes more sense to be a `CharField`.

api/applications/models.py
```diff
@@ -226,7 +227,17 @@ class StandardApplication(BaseApplication):
     goods_recipients = models.TextField(choices=GOODS_RECIPIENTS_CHOICES, default="")
     goods_starting_point = models.TextField(choices=GOODS_STARTING_POINT_CHOICES, default="")
 
+    # MOD Security approval fields
     is_mod_security_approved = models.BooleanField(blank=True, default=None, null=True)
+    security_approvals = ArrayField(
+        models.CharField(choices=SecurityClassifiedApprovalsType.choices, max_length=255), blank=True, null=True
+    )
+    f680_reference_number = models.TextField(default=None, blank=True, null=True, max_length=100)
```
kevincarrogan, line 235: I think this makes more sense to be a `CharField`.

api/applications/enums.py
```diff
@@ -196,3 +196,30 @@ def get_text(cls, choice):
...
+
+    choices = (
+        (F680, "F680"),
+        (F1686, "F1686"),
+        (OTHER, "Other"),
+    )
+
+    @classmethod
+    def keys(cls):
+        return [key for key, _ in cls.choices]
```
kevincarrogan, line 225: For these new enums I think we should just start using `models.TextChoices` https://docs.djangoproject.com/en/4.1/ref/models/fields/#enumeration-types

api/applications/models.py
depsiatwal: I did think about this and for the other we refers this

security_approvals is an array I just think it's cleaner and easier for UI to set this and BE logic to use this field.

api/applications/models.py
depsiatwal: done

api/applications/models.py
depsiatwal: ahh yes changed except details field

api/applications/models.py
depsiatwal: done

kevincarrogan (approved):

PR #1115, opened 2022-09-21
LTD-2741: Don't clear flags on good when TAU gives their assessment

saruniitr:
## Change description

When TAU gives their assessment then any flags set on good are cleared. There is no reason specified but it is done when the given rating is different from the rating specified by Exporter.

Clearing existing flags has not been an issue as the teams before TAU have not applied any flags but this is not true with upcoming routing changes. If these are cleared then the Case doesn't get routed correctly, hence don't clear these flags.


api/goods/views.py
```diff
@@ -143,8 +143,6 @@ def post(self, request, case_pk):
                     or new_is_controlled != old_is_controlled
                     or report_summary_updated
                 ):
-                    if isinstance(good, GoodOnApplication):
```
currycoder, line 146: Later on in this function we have `apply_good_flagging_rules_for_case()` - I would expect that to re-apply the pertinent flags.  Did you find that this got in the way of something working for the NLR flow?

saruniitr: Correct but for these flags there are no flagging rules as Enforcement unit manually applying these flags so we decided to not have any flagging rules.

Yeah I found this issue when working with the "Refer to FCDO/DI" flag scenarios. As the flag is getting cleared the routing rule is not triggering and not routing to the expected queues.
Does this clarify?

currycoder: Thanks Arun, yes makes absolute sense.

currycoder (approved):

PR #1116, opened 2022-09-22
LTD-2741: Enable Open search when running tests on CI

saruniitr:
## Change description

One of the routing test related to DN requires Open search enabled to work


.circleci/config.yml
```diff
@@ -94,7 +94,7 @@ jobs:
     working_directory: ~/lite-api
     environment:
       <<: *common_env_vars
-      LITE_API_ENABLE_ES: False
+      LITE_API_ENABLE_ES: True
```
currycoder, line 97: You say opensearch, but this would be elasticsearch right...?  Would it still work?

currycoder (approved):

.circleci/config.yml
saruniitr: Previously it was Elastic search but because of licence restrictions we moved to Open search which is a forked project from ES and compatible

saruniitr: Previously it was Elastic search but because of licence restrictions we moved to Open search. It is compatible as it is a forked project of ES.

PR #1119, opened 2022-09-22
Ltd 2822 regimes models

kevincarrogan:
This makes setting regime data on a good independent of the flags.

For wassenaar this was simply just handled by setting a flag and then basing the regime on that flag, now we're expanding the available regimes then this isn't going to be a viable strategy going forward.

To cater for this the regimes are more explicitly modelled in the system and then attached to the good_on_application to be added when assessed. Flagging rules can then be used to set the correct flags based on what regimes have been assigned.


api/applications/migrations/0063_merge_0062_auto_20220921_1148_0062_auto_20220921_1632.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("applications", "0062_auto_20220921_1148"),
+        ("applications", "0062_auto_20220921_1632"),
+    ]
+
+    operations = []
```
currycoder, line 13: Do we really need a merge migration?  I would have expected the dependency migration could be changed in the new migration above..?

api/applications/migrations/0062_auto_20220921_1632.py
```diff
@@ -0,0 +1,43 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name="GoodOnApplicationRegimeEntry",
+            fields=[
+                ("id", models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
+                (
+                    "good_on_application",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE,
```
currycoder, line 23: On deletion of the good on application, the GoodOnApplicationRegimeEntry is deleted?  Just checking my rusty cascade understanding

api/staticdata/regimes/migrations/0002_add_mtcr_regimes.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+
+from django.db import migrations, transaction
+
+from api.staticdata.regimes.enums import RegimesEnum, RegimeSubsectionsEnum
+
+
+def create_entries(RegimeEntry, subsection, entries_csv_filename):
+    data_path = os.path.dirname(os.path.abspath(__file__))
+    file_path = os.path.join(data_path, "data", entries_csv_filename)
```
currycoder, line 12: Could this folder be called `migration_data` or something?  I'm concerned that the data used for a data migration should be immutable and I want us to have all the signals in a future PR that innocently tries to change one of these CSV files...

api/staticdata/regimes/migrations/0003_auto_20220922_1526.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.2.15 on 2022-09-22 14:26
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
currycoder, line 7: Could this be collapsed in to that initial migration?

api/staticdata/regimes/migrations/data/MTCR_Cat_2_regimes.csv
```diff
@@ -0,0 +1,241 @@
+M10A1,ACTIVE,4,UK,Cat 2
```
currycoder, line 1: I see string sort

api/staticdata/regimes/views.py
```diff
@@ -0,0 +1,23 @@
...
+    authentication_classes = (HawkOnlyAuthentication,)
+
+    def get(self, request):
+        entries = (
+            RegimeEntry.objects.filter(subsection__regime=RegimesEnum.MTCR).order_by("name").values_list("pk", "name")
+        )
+
+        return JsonResponse(
+            data={
+                "entries": list(entries),
```
currycoder, line 21: I know this is a little bonus you wanted to build on top of, but a serializer here would be handy IMO

api/applications/migrations/0062_auto_20220921_1632.py
kevincarrogan: Yes. That is correct.

api/staticdata/regimes/migrations/0002_add_mtcr_regimes.py
kevincarrogan: It's actually in the migration folder so not sure if that makes you change your mind about renaming it?

api/staticdata/regimes/views.py
kevincarrogan: It's because I'm waiting until I put in wassenaar to work out what this is really going to look like so it's a bit quick and dirty for now. I don't think this view is going to last long in its current state.

api/staticdata/regimes/migrations/0002_add_mtcr_regimes.py
currycoder: oh yeah fair enough, let's leave that then

api/staticdata/regimes/migrations/data/MTCR_Cat_2_regimes.csv
kevincarrogan: Is this in regards to how I'm sorting this in the view? It's actually kind of pointless really as we never present a big list to the use anyway it just felt slightly nicer than just dumping it out in whatever sort order the database decided to do it in.

currycoder: It's not important as you say this isn't really used anywhere other than the migration.

currycoder (approved):

PR #1126, opened 2022-09-27
Handle case where `regime_entries` isn't sent through

kevincarrogan:
This is a non-mandatory field and as such we can't guarantee it will exist due to an frontend with older code


depsiatwal (approved):
lgtm

PR #1142, opened 2022-10-12
LTD-2907: Enable filtering Denials search results based on entity country

saruniitr:
## Change description

This is achieved by adding the model field to `filter_fields`. This will allow the filtering backend to parse the query params for this field and filter the documents. Also added lowercase analyser for this field in case if the country name is in lowercase.


depsiatwal (commented):
LGTM 
I think tests might need updating ?

kevincarrogan (approved):

PR #1146, opened 2022-10-13
 LTD-2939-add-regime-reg-ref-denials

depsiatwal:


api/external_data/management/commands/ingest_denials.py
```diff
@@ -64,6 +65,8 @@ def load_denials(self, filename):
                     **{field: row.pop(field, None) for field in self.required_headers},
                 }
             )
+            if i == 100:
+                break
```
kevincarrogan, line 69: I don't think you've meant to do this.

depsiatwal: at least it solves the issues of display to many denials :-)

PR #1148, opened 2022-10-13
Only add to countersign queue if case is present on no other feeder queues for it

currycoder:
This PR ensures that a case will only go to a countersigning queue if it is not present on any other "feeder" queues for that countersigning queue.

This improves the countersigning mechanism as before a case would visit the countersign queue any time that it left a feeder queue for it.  This meant in a situation where a case was on 4 feeder queues, it would reach the countersigning queue as soon as it left any one of those feeder queues.  A better workflow for everybody is that the case reaches the countersigning queue after it has left *all* of the feeder queues.


api/workflow/user_queue_assignment.py
```diff
@@ -74,19 +74,25 @@ def user_queue_assignment_workflow(queues: [Queue], case: Case):
...
-                    "queues": queue.countersigning_queue.name,
-                    "queue_ids": [str(queue.countersigning_queue_id)],
-                    "case_status": case.status.status,
-                },
-            )
+            remaining_feeder_queues = [
+                remaining_feeder_queue
+                for remaining_feeder_queue in case.queues.all()
+                if remaining_feeder_queue.countersigning_queue_id == queue.countersigning_queue_id
+            ]
```
kevincarrogan, line 81: Isn't this just a `.filter` really? Maybe `case.queues.filter(countersigning_queue_id=queue.countersigning_queue_id)`

kevincarrogan (approved):

PR #1150, opened 2022-10-14
Ltd 2942 ars entries

kevincarrogan:
This adds new picklist entries for summaries without the prefix.

There are also changes to the search results to try to produce better matches or at least give the user the ability to find the better matches.

  - Orders matches with priority of the name beginning with the searched string
  - Orders by name after that
  - Increases the number of results to 100


depsiatwal (approved):
looking good

PR #1165, opened 2022-10-31
LTD-2973: Add management command to truncate End-user name

saruniitr:
## Change description

HMRC spec only allows 80 chars for End-user name but currently we don't have validation in UI or API to prevent this. We are only validating in `lite-hmrc` when licence details are sent to be forwarded to HMRC.

Because of this we failed to send few licence details. Using this mgmt command we can resend those details to `lite-hmrc` until we add these restrictions in the UI and API.

### DIFF
--- a/api/support/management/commands/truncate_end_user_name.py
+++ b/api/support/management/commands/truncate_end_user_name.py
@@ -1,0 +1,71 @@
+import logging
+
+from rest_framework import status
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.utils import timezone
+
+from api.core.requests import post
+from api.licences.models import Licence
+from api.licences.serializers.hmrc_integration import HMRCIntegrationLicenceSerializer
+from api.licences.libraries.hmrc_integration_operations import HMRCIntegrationException, SEND_LICENCE_ENDPOINT
+
+
+class Command(BaseCommand):
+    help = """
+    Command to truncate End-user name (to 80 chars) and sends licence details to HMRC
+
+    This is required because HMRC spec allows only 80 chars otherwise we won't be able to
+    send licence details to HMRC
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "licence_reference",
+            type=str,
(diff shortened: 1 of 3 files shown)


api/support/management/commands/truncate_end_user_name.py
```diff
@@ -0,0 +1,71 @@
...
+from django.db import transaction
+from django.utils import timezone
+
+from api.core.requests import post
+from api.licences.models import Licence
+from api.licences.serializers.hmrc_integration import HMRCIntegrationLicenceSerializer
+from api.licences.libraries.hmrc_integration_operations import HMRCIntegrationException, SEND_LICENCE_ENDPOINT
+
+
+class Command(BaseCommand):
```
currycoder, line 16: Would recommend calling this command (file name) something more meaningful.  It's not immediately obvious to me that it will send the licence from the name alone

api/support/management/commands/truncate_end_user_name.py
```diff
@@ -0,0 +1,71 @@
...
+            "licence_reference",
+            type=str,
+            help="Reference number of the licence generated",
+        )
+
+    def handle(self, *args, **options):
+        licence_ref = options.pop("licence_reference")
+        logging.info("Given licence reference is: %s", licence_ref)
+
+        with transaction.atomic():
```
currycoder, line 35: Small thought, but if the sending code itself was updated with the truncate code, the command could be simplified to just be a "resend licence" command.  What do you think?

currycoder (commented):
A couple of comments inline.  I agree with codecov that those exception lines should probably be covered, could that be done too?

api/support/management/commands/truncate_end_user_name.py
saruniitr: yeah it was getting too long, I will try again.

api/support/management/commands/truncate_end_user_name.py
saruniitr: This command is only temporary to handle the failed licences manually.
Once we introduce the restrictions we should not need this command and can even be removed.

PR #1167, opened 2022-11-01
LTD-2968: Select only rules that doesn't have python criteria

saruniitr:
## Change description

When DB based rules are executed ensure that we only select the rules that doesn't have python criteria. Eventually when we move all rules to have python criteria this step can be skipped.


api/workflow/flagging_rules_automation.py
```diff
@@ -42,7 +42,7 @@ def _apply_python_flagging_rules(level, object):
 
 def get_active_flagging_rules_for_level(level):
```
currycoder, line 43: Could it be worth renaming this now to e.g. `get_active_legacy_flagging_rules_for_level` or similar?

saruniitr: yeah it makes it clear that these are legacy but once we migrated all of them we will remove this function completely. I will rename it anyway till that happens.

currycoder (approved):

PR #1174, opened 2022-11-08
 LTD-2894-shift-generic-fields-to-GoodOnApplication 

depsiatwal:


depsiatwal:
The following fields 
            "is_onward_exported"
            "is_onward_altered_processed"
            "is_onward_altered_processed_comments"
            "is_onward_incorporated"
            "is_onward_incorporated_comments"

Added to firearm_details originally which is an extension of good when an item is a firearm.  When adding non-firearm products during C5 it became these transparent these fields are require for other products also. In addition these fields can be set each time a good is added to any application. Hence they were added to 'GoodOnApplication' 

These PR shifts the firearms fields from "Goods" and shifts them to GoodOnapplication as the fields are removed. The UI has changed so that they are set/saved using the relevent payload. 

This PR also migrates the values and removes the fields.

api/goods/migrations/0023_sync_onward_exported_fields.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def synchronise_onward_exported_fields(apps, schema_editor):
+    FirearmGoodDetails = apps.get_model("goods", "FirearmGoodDetails")
+    GoodOnApplication = apps.get_model("applications", "GoodOnApplication")
+
+    firearm_good_details = FirearmGoodDetails.objects.all()
+
+    for firarm_detail in firearm_good_details:
+        GoodOnApplication.objects.filter(firearm_details_id=firarm_detail.id).update(
```
saruniitr, line 11: Where these are part of applications that have reached terminal state, it is better to exclude them so as to not change their history.

saruniitr: there is a typo in `firarm_detail.id`

api/goods/migrations/0024_auto_20221108_1050.py
```diff
@@ -0,0 +1,33 @@
...
+        ),
+        migrations.RemoveField(
+            model_name="firearmgooddetails",
+            name="is_onward_incorporated",
+        ),
+        migrations.RemoveField(
+            model_name="firearmgooddetails",
+            name="is_onward_incorporated_comments",
+        ),
+    ]
```
currycoder, line 33: Just a thought but is it worth saving this particular part for a separate PR/separate release?  I'm thinking if for whatever reason the data migration doesn't work as we are expecting, this data is gone gone.  So by delaying the deletion to a further release we get confidence that everything has moved over as expected.  Also, if the firearms detail API is unchanged in this first release we can do a release sequence which avoids the chance for 
backward incompatibility errors for our users; 1) Release API, then 2) Release frontend

api/goods/migrations/0023_sync_onward_exported_fields.py
depsiatwal: fixed

currycoder (approved):

PR #1176, opened 2022-11-09
Ltd-2990-enable-hawk

depsiatwal:


sekharpanja (approved):

api/conftest.py
```diff
@@ -68,6 +68,11 @@ def {fixture_name}(db):
         print(f"pytest.fixture: {fixture_name}")
 
 
+@pytest.fixture(autouse=True)
+def disable_hawk():
```
kevincarrogan, line 72: Use the `settings` fixture here instead of directly manipulating the imported `settings`. 

`def disable_hawk(settings):`

PR #1184, opened 2022-11-17
LTD-2878 switch workflow trigger to post case save

MarkThomasLite:


api/cases/signals.py
```diff
@@ -19,13 +19,7 @@ def case_pre_save_handler(sender, instance, raw=False, **kwargs):
...
         return
 
-    original = None
-    try:
-        original = Case.objects.get(pk=instance.id)
-    except Case.DoesNotExist:
-        return
-
-    status_changed = original.status != instance.status
+    status_changed = instance.previous_status != instance.status
```
currycoder, line 22: In the case of successive saves, shouldn't this function set `instance.previous_status` after it is done checking it?

currycoder: Might be worth adding a test for this situation too

api/cases/models.py
```diff
@@ -101,6 +101,7 @@ class Case(TimestampableModel):
     sla_remaining_days = models.SmallIntegerField(null=True)
     sla_updated_at = models.DateTimeField(null=True)
     additional_contacts = models.ManyToManyField("parties.Party", related_name="case")
+    previous_status = None
```
currycoder, line 104: Might be worth prefixing this with an underscore to signal that it's something internal.  And also popping a comment here to explain the attribute's usage

api/cases/signals.py
MarkThomasLite: it was in there but I guess I removed it during cleanup, adding it + the extra test

currycoder (approved):

Harriethw (approved):

saruniitr (approved):

PR #1194, opened 2022-11-29
Consolidate regime entries into a single view

kevincarrogan:
This still keeps the previous URLs intact but redirects to the new URL for backwards compatibility with our frontend until we explicitly make the same change on the frontend


Harriethw (approved):
LGTM but might be worth a more experienced person taking a look ☺️ 

Out of curiosity, where are these regimes used on the frontend? trying to dig for them and can see they are on the TAU app, but I can't see how to get there (or what TAU stands for!)

saruniitr (approved):

PR #1199, opened 2022-12-02
beis cw cles

MarkThomasLite:


api/staticdata/control_list_entries/migrations/0004_controllistentry_new_entries_20221130.py
```diff
@@ -0,0 +1,50 @@
...
+            "1C35085",
+            "1C35086",
+            "1C35087",
+            "1C35088",
+            "1C35089",
+        )
+        ControlListEntry = apps.get_model("control_list_entries", "ControlListEntry")
+        for rating in ratings:
+            ControlListEntry.objects.create(
+                rating=rating, text=rating, parent_id=None, category="End-Use", controlled=True
```
saruniitr, line 41: Is the `category` correct for these entries?

saruniitr, line 41: We already have CLEs that start with `1C350*` and their `parent_id` is not None, do we need to assign the same parent for these as well?

PR #1214, opened 2022-12-12
LTD-3098 refusal letter update

MarkThomasLite:


assets/css/ecju.css
```diff
@@ -15,9 +15,9 @@
 .document #recipient-address {
   width: 90mm;
 }
-.document #sender-address {
+/* .document #sender-address {
   margin-left: 10%;
-}
+} */
```
kevincarrogan, line 20: Can this be deleted instead of commenting it out?

PR #1222, opened 2022-12-14
LTD-3127: Update trigger list assessment endpoint to support multiple products

saruniitr:
## Change description

BEIS can select multiple products to assess for trigger list. Currently the endpoint allows for only one product so update it to support multiple products.


api/applications/tests/test_retrieve_good_on_applications.py
```diff
@@ -26,7 +26,7 @@ def test_caseworker_retrieve_a_good_on_application(self):
         self.good_on_application.good.save()
 
         url = reverse(
-            "applications:good_on_application_internal",
+            "applications:good_on_application",
```
currycoder, line 29: Does this other endpoint have the same format as `good_on_application_internal`?

api/applications/views/goods.py
```diff
@@ -223,41 +223,39 @@ def delete(self, request, obj_pk):
         return JsonResponse(data={"status": "success"}, status=status.HTTP_200_OK)
 
 
-class ApplicationGoodOnApplicationInternal(APIView):
+class ApplicationGoodOnApplicationTriggerListAssessment(APIView):
     """Good on a standard application for internal user edits"""
```
currycoder, line 227: I think this docstring needs updating to reflect what the endpoint is now

api/applications/views/goods.py
```diff
@@ -223,41 +223,39 @@ def delete(self, request, obj_pk):
...
-        return get_object_or_404(GoodOnApplication.objects.all(), pk=self.kwargs["obj_pk"])
+    def get_application(self):
+        return get_object_or_404(StandardApplication.objects.all(), pk=self.kwargs["pk"])
 
-    def get(self, request, **kwargs):
-        good_on_application = self.get_object()
-        serializer = self.serializer_class(good_on_application, context={"include_audit_trail": True})
-        return JsonResponse(serializer.data)
+    def get_queryset(self):
+        good_on_application_ids = self.request.data["goods"]
```
currycoder, line 236: Do we want to use a serializer for the input here?  It's not immediately obvious to me what the format should be

api/applications/views/goods.py
```diff
@@ -223,41 +223,39 @@ def delete(self, request, obj_pk):
...
-        data = request.data.copy()
-        serializer = self.serializer_class(instance=good_on_application, data=data, partial=True)
-        if not serializer.is_valid():
-            return JsonResponse(
-                data={"errors": serializer.errors},
-                status=status.HTTP_400_BAD_REQUEST,
-            )
+        for good_on_application in self.get_queryset():
+            data = request.data.copy()
+            serializer = self.serializer_class(instance=good_on_application, data=data, partial=True)
```
currycoder, line 249: IMO we definitely want a serializer for input; data is just whatever the request sends in right now - I'd imagine we want to restrict it, right?

api/applications/views/goods.py
```diff
@@ -223,41 +223,39 @@ def delete(self, request, obj_pk):
...
 
-        data = request.data.copy()
-        serializer = self.serializer_class(instance=good_on_application, data=data, partial=True)
-        if not serializer.is_valid():
-            return JsonResponse(
-                data={"errors": serializer.errors},
-                status=status.HTTP_400_BAD_REQUEST,
-            )
+        for good_on_application in self.get_queryset():
+            data = request.data.copy()
```
currycoder, line 248: This is essentially a bulk update; for each matching good on application we update it's fields from one static set of data.
Just talking this through but another way to structure the request would be e.g.
```
[
    {"id": "..", "nsg_list_type": "..."}, {"id": "..", "some_other_field": "..."}
]
```

This might end up being a bit more general purpose as many good on applications could be updated with differing values.

api/applications/views/goods.py
```diff
@@ -223,41 +223,39 @@ def delete(self, request, obj_pk):
...
+                return JsonResponse(
+                    data={"errors": serializer.errors},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
 
-        serializer.save()
-        return JsonResponse(status=status.HTTP_200_OK, data=serializer.data)
+            serializer.save()
+
+        return JsonResponse(status=status.HTTP_200_OK, data={})
```
currycoder, line 258: Should the response body be something here?  e.g. a list of all updated good on applications fully expanded

api/applications/views/goods.py
```diff
@@ -223,41 +223,39 @@ def delete(self, request, obj_pk):
         return JsonResponse(data={"status": "success"}, status=status.HTTP_200_OK)
 
 
-class ApplicationGoodOnApplicationInternal(APIView):
+class ApplicationGoodOnApplicationTriggerListAssessment(APIView):
```
currycoder, line 226: I'd argue that as this stands this is a misnomer, it's a bulk update endpoint going far beyond trigger list assessments

api/applications/urls.py
```diff
@@ -51,9 +51,9 @@
     # Goods
     path("<uuid:pk>/goods/", goods.ApplicationGoodsOnApplication.as_view(), name="application_goods"),
     path(
-        "good-on-application/internal/<uuid:obj_pk>/",
-        goods.ApplicationGoodOnApplicationInternal.as_view(),
-        name="good_on_application_internal",
+        "<uuid:pk>/trigger-list/",
```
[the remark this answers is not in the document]
saruniitr: this is totally new. I created it as part of this ticket before working on the UI.
While doing the UI changes realised that it needs to support multiple products hence changing it.

api/applications/tests/test_retrieve_good_on_applications.py
saruniitr: before the new endpoint is introduced it was using `applications:good_on_application` only.
As the new endpoint was meant for internal use only so I switched in the tests but with the latest changes it cannot provide `get` method for a single object so went back to how it was before.

api/applications/views/goods.py
saruniitr: :+1:

api/applications/views/goods.py
saruniitr: Serializer is used further down below when we save the updates.

api/applications/views/goods.py
saruniitr: `data` is used in a serializer currently and if there are any unexpected fields it fails validation.
There is also a test for that as codecov/patch complained

api/applications/views/goods.py
saruniitr: it can be but UI doesn't use that.

api/applications/views/goods.py
saruniitr: hmm, I see what you mean.
I have named it like this because this is the only place where it is used and unlikely to be used in other places. I will update to keep it generic.

api/applications/views/goods.py
saruniitr: hmm this is better, I will see if it can be updated like this.

api/applications/views/goods.py
currycoder: right you are! For some reason I missed that this *is the serializer*!

api/applications/views/goods.py
currycoder: Agreed, but I feel like it could be good in terms of API design and future uses for this endpoint

saruniitr: agreed

api/applications/serializers/good.py
```diff
@@ -236,6 +237,41 @@ def create(self, validated_data):
...
+class GoodOnApplicationUpdateListSerializer(serializers.ListSerializer):
+    def update(self, instance, validated_data):
+        good_on_application_mapping = {good_on_application.id: good_on_application for good_on_application in instance}
+        data_mapping = {item["id"]: item for item in validated_data}
+
+        ret = []
+        for id, data in data_mapping.items():
+            good_on_application = good_on_application_mapping.get(id, None)
+            if good_on_application is None:
+                raise NotImplementedError("Serializer doesn't support creating multiple goods on application")
```
saruniitr, line 249: Unlikely to hit this and can be removed, only included as precautionary.

api/applications/serializers/good.py
```diff
@@ -236,6 +237,41 @@ def create(self, validated_data):
         return super().create(validated_data)
 
 
+class GoodOnApplicationUpdateListSerializer(serializers.ListSerializer):
```
saruniitr, line 240: default `ListSerializer` doesn't support multiple create/update hence a custom one is created.

api/applications/views/goods.py
```diff
@@ -223,41 +224,39 @@ def delete(self, request, obj_pk):
...
         if application.status.status in get_case_statuses(read_only=True):
             return JsonResponse(
                 data={"errors": [strings.Applications.Generic.READ_ONLY]},
                 status=status.HTTP_403_FORBIDDEN,
             )
 
-        data = request.data.copy()
-        serializer = self.serializer_class(instance=good_on_application, data=data, partial=True)
+        good_on_application_ids = [item["id"] for item in request.data]
+        serializer = self.serializer_class(
```
currycoder, line 245: So if for whatever reason an ID that the request is trying to update doesn't exist, we would raise NotImplementedError i.e. 500.  I feel like in that case 404 would be better.

currycoder (approved):

PR #1225, opened 2022-12-16
Add trigger list to advanced filters

antroy-madetech:
Introduce a filter for cases with a nsg list type of "TRIGGER_LIST"

<del>includes commits from #1224  so waiting for that to be merged first.</del>


depsiatwal (approved):
looks sensible to me

PR #1226, opened 2022-12-19
adding regime entry search to advanced search on internal site

sekharpanja:
As per LTD-3121, this PR add option within advanced search on internal site

### DIFF
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -106,7 +106,7 @@ class CaseQuerySet(models.QuerySet):
         )
 
     def with_regime_entry(self, regime_entry):
-        return self.filter(Q(baseapplication__goods__regime_entries__id__in=[regime_entry]))
+        return self.filter(baseapplication__goods__regime_entries__id=regime_entry)
 
     def with_flags(self, flags):
         case_flag_ids = self.filter(flags__id__in=flags).values_list("id", flat=True)
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -105,6 +105,12 @@ class CaseQuerySet(models.QuerySet):
             | Q(baseapplication__goods_type__control_list_entries__rating__in=[control_list_entry])
         )
 
+    def with_regime_entry(self, control_list_entry):
+        return self.filter(
+            Q(baseapplication__goods__good__regimes__id__in=[control_list_entry])
+            | Q(baseapplication__goods_type__regimes__id__in=[control_list_entry])
+        )
+
     def with_flags(self, flags):
         case_flag_ids = self.filter(flags__id__in=flags).values_list("id", flat=True)
         org_flag_ids = self.filter(organisation__flags__id__in=flags).values_list("id", flat=True)
@@ -231,6 +237,7 @@ class CaseManager(models.Manager):
         exporter_site_name=None,
         exporter_site_address=None,
         control_list_entry=None,
+        regime_entry=None,
         flags=None,
         country=None,
         team_advice_type=None,
@@ -318,6 +325,9 @@ class CaseManager(models.Manager):
 
         if control_list_entry:
             case_qs = case_qs.with_control_list_entry(control_list_entry)
+
+        if regime_entry:
+            case_qs = case_qs.with_regime_entry(regime_entry)
 
         if flags:
             case_qs = case_qs.with_flags(flags)
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -105,10 +105,9 @@ class CaseQuerySet(models.QuerySet):
             | Q(baseapplication__goods_type__control_list_entries__rating__in=[control_list_entry])
         )
 
-    def with_regime_entry(self, control_list_entry):
+    def with_regime_entry(self, regime_entry):
         return self.filter(
-            Q(baseapplication__goods__good__regimes__id__in=[control_list_entry])
-            | Q(baseapplication__goods_type__regimes__id__in=[control_list_entry])
+            Q(baseapplication__goods__regime_entries__name__in=[regime_entry])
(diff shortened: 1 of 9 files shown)


api/staticdata/regimes/views.py
```diff
@@ -11,6 +13,20 @@
...
+
+    def get_queryset(self):
+        return RegimeEntry.objects.all()
+
+    def get(self, request):
+        """
+        Returns list of all Regimes
+        """
+        queryset = self.get_queryset()
+        return JsonResponse(data={"regime_entires": list(queryset.values("id", "shortened_name", "name"))})
```
kevincarrogan, line 27: There is a typo here. It should be `regime_entries`.

kevincarrogan: Instead of writing your own `get` method you should instead use the `RegimeEntrySerializer`.

api/staticdata/regimes/views.py
```diff
@@ -11,6 +13,20 @@
 from .serializers import RegimeEntrySerializer
 
 
+class RegimeEntriesListView(APIView):
+    authentication_classes = (HawkOnlyAuthentication,)
+
+    def get_queryset(self):
+        return RegimeEntry.objects.all()
```
kevincarrogan, line 20: You can use the `queryset` attribute instead of declaring a function here.

api/staticdata/regimes/views.py
```diff
@@ -11,6 +13,20 @@
 from .serializers import RegimeEntrySerializer
 
 
+class RegimeEntriesListView(APIView):
```
kevincarrogan, line 16: Could this maybe be a `ListAPIView`?

api/cases/managers.py
```diff
@@ -105,6 +105,9 @@ def with_control_list_entry(self, control_list_entry):
             | Q(baseapplication__goods_type__control_list_entries__rating__in=[control_list_entry])
         )
 
+    def with_regime_entry(self, regime_entry):
+        return self.filter(Q(baseapplication__goods__regime_entries__id__in=[regime_entry]))
```
kevincarrogan, line 109: You don't need to use a `Q` object here.

kevincarrogan: This doesn't need to be `__in=[regime_entry]` it can just be `self.filter(baseapplication__goods__regime_entries__id=regime_entry)`.

api/cases/tests/test_case_search_advanced.py
```diff
@@ -125,6 +126,21 @@ def test_filter_by_good_control_list_entry(self):
...
+
+        T1 = RegimeEntry.objects.get(name="T1")
+        T5 = RegimeEntry.objects.get(name="T5")
+        qs_1 = Case.objects.search(regime_entry="")
+        qs_2 = Case.objects.search(regime_entry=T5.id)
+        qs_3 = Case.objects.search(regime_entry=T1.id)
+
+        self.assertEqual(qs_1.count(), 1)
+        self.assertEqual(qs_2.count(), 0)
+        self.assertEqual(qs_3.count(), 1)
```
kevincarrogan, line 142: Can we check that this is returning the right Case objects instead of just doing a count?

api/staticdata/regimes/tests/test_helpers.py
```diff
@@ -0,0 +1,18 @@
...
+    def setUp(self):
+        super().setUp()
+
+    def test_get_regime_entry(self):
+
+        regime = get_regime_entry(name="T1")
+        self.assertEqual("T1", regime.name)
+
+    def test_get_regime_entry_not_found(self):
+        with pytest.raises(Exception):
```
kevincarrogan, line 17: We can be more explicit in the exception that we're checking for here.

api/staticdata/regimes/tests/test_helpers.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from test_helpers.clients import DataTestClient
+
+from api.staticdata.regimes.helpers import get_regime_entry
+
+
+class RegimeHelperTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
```
kevincarrogan, line 9: This doesn't need to be done as you're not doing anything else in the `setUp`.

api/staticdata/regimes/urls.py
```diff
@@ -6,6 +6,7 @@
 app_name = "regimes"
 
 urlpatterns = [
+    path("", views.RegimeEntriesListView.as_view(), name="entries_list"),
```
kevincarrogan, line 9: I think this should be under `/entries/` instead of just being at the top level.

api/staticdata/regimes/views.py
```diff
@@ -11,6 +12,19 @@
...
+    model = RegimeEntry
+    serializer_class = RegimeEntrySerializer
+    queryset = RegimeEntry.objects.all()
+
+    def get(self, request):
+        """
+        Returns list of all Regime Entries
+        """
+        return JsonResponse(data={"regime_entries": list(self.queryset.values("pk", "shortened_name", "name"))})
+
```
kevincarrogan, line 26: Now that you're defining the queryset and serializer you don't need to write your own `get` method (in fact you writing this means it's ignoring those attributes).

It would be better to just let the declared attributes do this for you. This probably means you'll need to change the frontend as it won't be under the `regime_entries` key but the benefit is that this is just using standard DRF stuff which should make things more consistent across our application.

kevincarrogan: I would also double check whether you need to disable the `pagination_class` the same as is done in the `EntriesView` once you get rid of this `get` method.

kevincarrogan (approved):

PR #1233, opened 2022-12-22
Allow goods to be edited from internal endpoint as long as applicatio…

kevincarrogan:
This allows the goods on a case to be edited as long as the status of the case is not terminal.

This differs from the other endpoint as this is for the internal application and we allow editing during different statuses to the exporter endpoint.


depsiatwal (approved):
lgtm

saruniitr (approved):

PR #1238, opened 2022-12-30
Remove additional commands that add demo data and routing rules

kevincarrogan:
These have all now been supersceded by migrations

In the case of the internal demo data we also don't want to add data that's not the same as production in regards to Teams, Queues, Flags or Routing Rules so this command has also been removed


api/staticdata/management/commands/tests/test_commands.py
```diff
@@ -94,22 +89,6 @@ def test_seed_role_permissions(self):
         self.seed_command(seedrolepermissions.Command)
         self.assertTrue(Permission.objects.count() >= len(GovPermissions) + len(ExporterPermissions))
 
-    @pytest.mark.seeding
-    def test_seed_demo_data(self):
-        self.seed_command(seedadminteam.Command)
-        self.seed_command(seedinternaldemodata.Command)
-        for team in deserialize_csv_from_string(settings.LITE_API_DEMO_TEAMS_CSV):
```
saruniitr, line 101: we can remove these env variables from settings also and later in vault

saruniitr (approved):

PR #1240, opened 2023-01-05
LTD-3108-end-party-url-validation 

depsiatwal:
This is to fix an error where end party url was breaking during validation with a 500 error. 
When a user enters HTTP:// the code would add an additional http:// resulting in a validation error.

The whole string is converted to lower case as urls are inherently lower case anyway

### DIFF
--- a/api/parties/tests/test_serializers.py
+++ b/api/parties/tests/test_serializers.py
@@ -9,16 +9,14 @@ from django.core.exceptions import ValidationError
     [
         ("http://workingexample.com", "http://workingexample.com"),
         ("http://www.workingexample.com", "http://www.workingexample.com"),
-        ("http://WWW.workingexample.com", "http://www.workingexample.com"),
+        ("http://WWW.workingexample.com", "http://WWW.workingexample.com"),
         ("http://workingexample.com", "http://workingexample.com"),
         ("workingexample.com", "https://workingexample.com"),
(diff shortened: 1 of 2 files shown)


kevincarrogan:
“ The whole string is converted to lower case as urls are inherently lower case anyway” that’s not true. The domain and protocol are case insensitive and could be lower cased the path in a url is very much case sensitive.

api/parties/tests/test_serializers.py
```diff
@@ -0,0 +1,26 @@
...
+        ("http://www.workingexample.com", "http://www.workingexample.com"),
+        ("http://WWW.workingexample.com", "http://www.workingexample.com"),
+        ("http://workingexample.com", "http://workingexample.com"),
+        ("workingexample.com", "https://workingexample.com"),
+        ("workingexample.com", "https://workingexample.com"),
+        ("HTTPS://workingexample.com", "https://workingexample.com"),
+    ]
+)
+def test_party_serializer_validate_website_valid(url_input, url_output):
+    assert url_input
```
kevincarrogan, line 20: You’re not actually testing anything here.

depsiatwal: opps forgot to remove the comment

Harriethw (approved):
LGTM!

PR #1243, opened 2023-01-05
LTD-2286: Relaxed validation of phone numbers.

antroy-madetech:
Default Country Code is now GB.
Changed instances of 'phone' to 'telephone'


api/organisations/serializers.py
```diff
@@ -1,8 +1,9 @@
 import re
 
+import phonenumbers
```
Harriethw, line 3: it looks like this import isn't being used, is it still needed?

api/organisations/serializers.py
```diff
@@ -189,7 +198,7 @@ class OrganisationCreateUpdateSerializer(serializers.ModelSerializer):
             "max_length": Organisations.Create.LENGTH_REGISTRATION_NUMBER,
         },
     )
-    phone_number = PhoneNumberField(required=True, allow_blank=True)
+    phone_number = PhoneNumberField()
```
Harriethw, line 201: it looks like previously the field could be blank (`allow_blank=True`), is that still the case?

.pre-commit-config.yaml
```diff
@@ -1,3 +1,6 @@
+default_language_version:
```
Harriethw, line 1: minor/suggestion: would it be worth splitting out the changes to pre-commit and docker-compose into a separate commit with a bit of explanation about why they were changed?

api/organisations/serializers.py
```diff
@@ -145,6 +146,14 @@ def update(self, instance, validated_data):
         return instance
 
 
+class PhoneNumberField(serializers.CharField):
+    def to_internal_value(self, data):
+        phone_number = to_python(data, "GB")
+        if phone_number and not phone_number.is_valid():
+            raise serializers.ValidationError("Enter a valid telephone number.")
+        return phone_number
```
kevincarrogan, line 154: Do we actually need to create our own field here? I think this could all be achieved by just changing the configuration of the field itself that is provided by `phonenumber_field`.

antroy-madetech: For some reason the phonenumber_field.serializers.PhoneNumberField doesn't support having a default region. Looks like the latest version of the module does though, so I'll try that.

api/organisations/serializers.py
antroy-madetech: No. I'll remove it.

.pre-commit-config.yaml
antroy-madetech: Done

kevincarrogan (approved):

kevincarrogan (approved):

PR #1244, opened 2023-01-09
[LTD-3155] Create routing history model

currycoder:
This PR adds a new `RoutingHistory` model which is the first step in a chunk of work to record routing events that occurred for cases.  e.g. "flag X was added by routing engine", "case was added to queue Y by user Z"

The following work for routing history is captured in the epic (https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3154) and design doc (https://docs.google.com/document/d/1zWoQIsgTuwd2h2nBhfQvGF68LG5e0R_eqq77B72qOuE/edit)


api/workflow/routing_rules/models.py
```diff
@@ -117,3 +120,52 @@ def is_python_criteria_satisfied(self, case):
         return run_criteria_function(self.id, case)
 
     natural_key.dependencies = ["teams.Team", "queues.Queue", "users.GovUser", "countries.Country"]
+
+
+class RoutingHistory(TimestampableModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    case = models.ForeignKey("cases.Case", on_delete=models.CASCADE, related_name="routing_history_records")
+    entity_limit = models.Q(app_label="queues", model="Queue") | models.Q(app_label="flags", model="Flag")
```
saruniitr, line 128: Could you clarify the purpose of this field? is it worth adding help text?

currycoder: Yeah I think it's worth a comment; it's not a field that is persisted to the DB just a variable to ensure that the `entity` generic foreign key can only refer to Queue or Flag models.  I'll make this explicit with a comment (and maybe prefix this one with an underscore)

saruniitr (approved):

PR #1245, opened 2023-01-10
Update README with pre-commit instructions

hnryjmes:


README.md
```diff
@@ -63,14 +63,16 @@ Service for handling backend calls in LITE.
...
-- Install pre-commit (e.g. on macOS `pip install pre-commit`)
-- `pre-commit install`
-
-* to run on its own
-  - `pre-commit run`
-
-If your changes incorrectly set off the PII / secrets warning, it can be added to `pii-secret-exclude.txt`
+## Git pre-commit setup
+
+- Install pre-commit (e.g `pip install pre-commit`)
```
kevincarrogan, line 68: We can probably link out to the `pre-commit` installation instructions themselves here. https://pre-commit.com/#install

README.md
```diff
@@ -63,14 +63,16 @@ Service for handling backend calls in LITE.
...
+- Run `pre-commit install`
+- Run following to scan all files for issues
+  - `pre-commit run --all-files`
+- If your changes incorrectly set off the PII / secrets warning, it can be added to `pii-secret-exclude.txt`
```
kevincarrogan, line 72: I actually think this is bad advice. We should remove this.

README.md
```diff
@@ -63,14 +63,16 @@ Service for handling backend calls in LITE.
...
+- After this initial setup, pre-commit should run automatically whenever you run `git commit` locally.
+- All developers should use the pre-commit hooks for the project to make routine tasks easier (e.g. linting, style checking) and to help ensure secrets and personally identifiable information (PII) are not leaked.
```
kevincarrogan, line 74: We should say "must" not "should" here.

hnryjmes:
@kevincarrogan thanks I've made the changes you suggested plus minor rewording, and the same is done on both PRs, so ready for re-review

kevincarrogan:
> have you removed If your changes incorrectly set off the PII / secrets warning, it can be added to `pii-secret-exclude.txt` ?

That was my suggestion. I don't think it's actually good advice. We shouldn't suggest just adding files to this as the default approach to handling this.

kevincarrogan (approved):

PR #1247, opened 2023-01-13
filter cases with open ECJU queries

Harriethw:
### aim

In this frontend PR https://github.com/uktrade/lite-frontend/pull/1049 we are adding tabs beneath the filters on the cases views (both all cases and for specific queues), one tab shows all cases (with whatever filters have been applied) the other tab shows cases that have an ECJU Query that has NOT been responded to by the user.

The tabs will return a query param on the request, and when it is true we want to apply the filter.

### DIFF
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -181,6 +181,9 @@ class CaseQuerySet(models.QuerySet):
     def with_trigger_list(self):
         return self.filter(baseapplication__goods__nsg_list_type=NSGListType.TRIGGER_LIST)
 
+    def only_open_queries(self):
+        return self.filter(case_ecju_query__isnull=False, case_ecju_query__response__isnull=True)
+
     def order_by_date(self, order="-"):
         """
         :param order: ('', '-')
@@ -252,6 +255,7 @@ class CaseManager(models.Manager):
         sla_days_elapsed=None,
         is_nca_applicable=None,
         is_trigger_list=None,
+        only_open_queries=False,
         **kwargs,
     ):
         """
@@ -364,6 +368,9 @@ class CaseManager(models.Manager):
 
         if is_trigger_list:
             case_qs = case_qs.with_trigger_list()
+
+        if only_open_queries:
+            case_qs = case_qs.only_open_queries()
 
         if is_work_queue:
             case_qs = case_qs.annotate(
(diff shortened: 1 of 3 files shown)


api/cases/tests/test_case_search.py
```diff
@@ -249,6 +250,35 @@ def test_get_submitted_status_and_clc_type_cases(self):
...
+            query_type=PicklistType.ECJU,
+        )
+        ecju_query_with_response.save()
+        url = f"{self.url}?has_open_queries_filter=True"
+
+        response = self.client.get(url, **self.gov_headers)
+
+        response_data = response.json()
+
+        self.assertEqual(response_data["count"], 1)
```
currycoder, line 279: I'm a little unsure looking at the testcase as to whether there is another Case object present when the request is made?  i.e. some proof that we return just the one case with an open query and not another case that is in the DB which does not have an open query.

If that other case needs adding I would suggest using `api.cases.tests.CaseFactory` to add it.

Harriethw: You can see in the setup for the tests in this class that it creates a bunch of cases. On line 263 I'm add a resolved ECJU query to one of those cases.

When we add the count to the deselected tab we can check that the correct number of cases was returned for the other query.

currycoder: Brilliant, thanks for clarifying

api/cases/managers.py
```diff
@@ -181,6 +181,9 @@ def with_nca_applicable(self):
     def with_trigger_list(self):
         return self.filter(baseapplication__goods__nsg_list_type=NSGListType.TRIGGER_LIST)
 
+    def with_open_queries(self):
```
currycoder, line 184: Minor thing but could this method make better sense being named `only_open_queries` to reflect the naming that we landed on?  `with_` feels like this is including open queries in addition to other cases whereas this is a reduction of the result set

currycoder (approved):

api/cases/managers.py
```diff
@@ -365,6 +369,9 @@ def search(  # noqa
         if is_trigger_list:
             case_qs = case_qs.with_trigger_list()
 
+        if only_open_queries == "True":
```
currycoder, line 372: I feel like the data should be converted in the view instead; that way our internal manager search method has a fairly sensible interface.

I'm betting that this wasn't an issue for the other params as they will end up `""` or `None` in the case where they are unset.

Harriethw: yep that makes sense, I've converted to a boolean in the view

api/cases/managers.py
Harriethw: makes sense! have changed the method name

PR #1249, opened 2023-01-18
LTD-3189: add HEAD request for cases count

Harriethw:
Waiting for #1247 to be merged into master.

We would like to show a count of cases with filters possibly applied, on the queue view. This HEAD request allows us to just return the count and avoid extra processing of data.

### DIFF
--- a/api/cases/tests/test_case_search.py
+++ b/api/cases/tests/test_case_search.py
@@ -300,6 +300,49 @@ class FilterAndSortTests(DataTestClient):
         cases_returned = [x["id"] for x in response_data]
         self.assertIn(str(case_with_open_query.id), cases_returned)
 
+    def test_head_request(self):
+        """
+        Given there is are cases
+        When a HEAD request is sent to view cases with no params
+        Then the count for all cases is returned in the header
+        """
+        all_cases = self.application_cases + self.clc_cases
+
+        response = self.client.head(self.url, **self.gov_headers)
+
+        response_headers = response.headers
+
+        self.assertEqual(response_headers["Resource-Count"], f"{len(all_cases)}")
+
+    def test_head_request_open_queries(self):
+        """
+        Given there is a case with an ECJU Query that has not been responded to
+        When a HEAD request is sent to view cases with open queries
+        Then the count for open queries is returned in the header
+        """
+
+        ## create an ecju query for a case that should appear in tab
+        case_with_open_query = self.application_cases[0]
+        self.create_ecju_query(case_with_open_query, gov_user=self.gov_user)
+
+        ## create an ecju query with a response so it should not appear in count
+        ecju_query_with_response = EcjuQuery(
+            question="ECJU Query 2",
+            case=self.application_cases[1],
+            response="I have a response",
+            raised_by_user=self.gov_user,
+            responded_by_user=self.exporter_user,
+            query_type=PicklistType.ECJU,
+        )
+        ecju_query_with_response.save()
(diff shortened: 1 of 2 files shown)


api/cases/views/search/views.py
```diff
@@ -114,3 +115,31 @@ def get(self, request, *args, **kwargs):
...
+        # we include hidden cases in non work queues (all cases, all open cases)
+        # and if the flag to include hidden is added
+        include_hidden = not is_work_queue or str_to_bool(request.GET.get("hidden"))
+        filters = {key: value for key, value in request.GET.items() if key not in ["hidden", "queue_id", "flags"]}
+
+        filters["flags"] = request.GET.getlist("flags", [])
+        filters["submitted_from"] = make_date_from_params("submitted_from", filters)
+        filters["submitted_to"] = make_date_from_params("submitted_to", filters)
+        filters["finalised_from"] = make_date_from_params("finalised_from", filters)
+        filters["finalised_to"] = make_date_from_params("finalised_to", filters)
```
currycoder, line 133: Looks like this has a fair overlap with the `get()` method above...  Could it be factored out in to a `sanitize_filters()` helper method or similar?

api/cases/views/search/views.py
```diff
@@ -114,3 +115,31 @@ def get(self, request, *args, **kwargs):
...
+
+        count = Case.objects.search(
+            queue_id=queue_id,
+            is_work_queue=is_work_queue,
+            user=user,
+            include_hidden=include_hidden,
+            **filters,
+        ).count()
```
currycoder, line 141: A helper might keep things DRY'er e.g. 
```
search_queryset = self.get_search_queryset()
count = search_queryset.count()
```

api/cases/tests/test_case_search.py
```diff
@@ -300,6 +300,35 @@ def test_only_open_queries_false_search(self):
...
+            query_type=PicklistType.ECJU,
+        )
+        ecju_query_with_response.save()
+        url = f"{self.url}?only_open_queries=True"
+
+        response = self.client.head(url, **self.gov_headers)
+
+        response_headers = response.headers
+
+        self.assertEqual(response_headers["Resource-Count"], "1")
```
currycoder, line 330: Could there also be a testcase for when a bunch of cases are created (maybe use `CaseFactory`) and asserting that an unfiltered head request gets the right count back?

Harriethw: good point, added above! (and renamed this test)

api/cases/views/search/views.py
Harriethw: good point, I've called it `get_filters` as that seems to fit more with the naming conventions elsewhere in the app, lmk your thoughts

api/cases/views/search/views.py
Harriethw: added a version of this! lmk what you think

api/cases/views/search/views.py
```diff
@@ -114,3 +102,41 @@ def get(self, request, *args, **kwargs):
                 "queue": queue,
             }
         )
+
+    def head(self, request, *agrs, **kwargs):
+        user = request.user.govuser
```
Harriethw, line 107: leaving some repetition from the `get` request so that it's not completely abstracted/more readable as I found the `non_work_queue`/`hidden` logic a bit confusing without knowing the domain that well. but open to feedback if we think this could also be put in a shared method.

currycoder (approved):

PR #1252, opened 2023-01-19
 LTD-3029-hide-queue-entries-menu

depsiatwal:
This removed 'my assigned cases' and 'my caseload' from the menu in caseworker. 

It took some time to figure out if removing from SYSTEM_QUEUES was going to have an impact. 

It's for the menu it then also makes sense to remove the tests because search case can never be used to filter for this queue_id

hopefully most of this is display specific on the menu the system will require some testing which I will update on the ticket.


api/queues/constants.py
```diff
@@ -19,8 +19,6 @@
     OPEN_CASES_QUEUE_ID: OPEN_CASES_QUEUE_NAME,
     MY_TEAMS_QUEUES_CASES_ID: MY_TEAMS_QUEUES_CASES_NAME,
     UPDATED_CASES_QUEUE_ID: UPDATED_CASES_QUEUE_NAME,
-    MY_ASSIGNED_CASES_QUEUE_ID: MY_ASSIGNED_CASES_QUEUE_NAME,
-    MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_ID: MY_ASSIGNED_AS_CASE_OFFICER_CASES_QUEUE_NAME,
```
currycoder, line 23: Visible queues are entirely defined by the backend?  That's interesting - I would have assumed there would be some FE logic to remove too

currycoder (approved):

api/queues/constants.py
depsiatwal: yes surprised me too but FE just displays the buckets returned by the BW

PR #1253, opened 2023-01-20
wip

hnryjmes:
Update report summary prefixes file


pii-secret-exclude.txt
```diff
@@ -112,3 +112,5 @@ api/conftest.py
 api/staticdata/regimes/migrations/data/NSG_Dual_use.csv
 api/staticdata/regimes/migrations/data/NSG_Potential_trigger_list.csv
 api/applications/migrations/0065_auto_20221125_1118.py
+api/staticdata/report_summaries/migrations/data/0002_add_report_summaries/report_summary_prefixes.csv
+api/staticdata/report_summaries/migrations/data/0002_add_report_summaries/report_summary_subjects.csv
```
saruniitr, line 116: Does CI fails if these are not excluded? or is it possible to fix the errors without excluding?

api/staticdata/tests/test_staticdata.py
```diff
@@ -0,0 +1,30 @@
+import csv
+import os
+
+from contextlib import contextmanager
+
+DATA_PATH = os.path.join(
+    os.path.split(os.path.split(os.path.abspath(__file__))[0])[0],
```
saruniitr, line 7: Instead of splitting you can take `dirname` and join accordingly,
eg https://github.com/uktrade/lite-api/blob/master/api/staticdata/report_summaries/migrations/0002_add_report_summaries.py#L10

pii-secret-exclude.txt
hnryjmes: I don't think this is possible in this case. If the pii checker runs on the CI at any point then every id would be flagged as a postcode. As it is a csv file we cannot have a per-line excluding comment as we can for python files

api/staticdata/tests/test_staticdata.py
hnryjmes: Done

pii-secret-exclude.txt
saruniitr: :+1:

saruniitr (approved):

PR #1255, opened 2023-01-23
Ltd 3260 sanctions files

kevincarrogan:
This fixes an issue where the date format across the various sanctions files was different and this was causing OpenSearch to throw an exception when indexing some documents. 

It appears that whatever format is used first in the index is then used as the default for all other documents. 

The fix here is to make sure that the date format is consistent across all documents.


depsiatwal (approved):
lgtm

PR #1256, opened 2023-01-23
Ltd 3163 prefix summary

antroy-madetech:


api/applications/models.py
```diff
@@ -466,6 +471,11 @@ def save(self, *args, **kwargs):
         is saved with properties that seem to be copied from the
         associated Good.
         """
+        if self.subject and self.prefix:
+            self.report_summary = f"{self.prefix.name}: {self.subject.name}"
+        elif self.subject:
+            self.report_summary = self.subject.name
+
```
saruniitr, line 478: As discussed please update report summary in the view.
The changes in that field are recorded in an audit event also.

.gitmodules
```diff
@@ -5,4 +5,4 @@
 [submodule "lite_routing"]
 	path = lite_routing
 	url = git@github.com:uktrade/lite-routing.git
-	branch = main
+	branch = LTD-3163-Prefix_summary
```
saruniitr, line 8: this should always be pointing to `main`

api/cases/migrations/0036_case_additional_contacts.py
```diff
@@ -27,6 +27,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ("parties", "0007_auto_20200317_1730"),
+        ("applications", "0003_auto_20200210_1326"),
```
saruniitr, line 30: is this required?

api/goods/serializers.py
```diff
@@ -894,6 +895,18 @@ def get_case_officer(self, instance):
...
+        model = ReportSummarySubject
+        fields = ("id",)
+
+
+class ReportSummaryPrefixSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ReportSummaryPrefix
+        fields = ("id",)
+
+
```
saruniitr, line 909: These are not required.
We are sending these values as `id` so we need to use `PrimaryKeyRelatedField()` in the serializer

api/goods/views.py
```diff
@@ -74,6 +75,22 @@
...
+def get_new_report_data(request):
+    data = request.data
+    summary = data.get("report_summary", None)
+    rs_subject = None
+
+    if "report_summary_subject_id" in data:
+        rs_subject = ReportSummarySubject.objects.get(id=data["report_summary_subject_id"])
+
+    if rs_subject is not None and "report_summary_prefix_id" in data:
+        rs_prefix = ReportSummaryPrefix.objects.get(id=data["report_summary_prefix_id"])
```
saruniitr, line 87: they raise exception if the id is not present, we need to handle that and return error accordingly

api/goods/views.py
```diff
@@ -74,6 +75,22 @@
...
+
+        summary = " ".join([part.name for part in [rs_prefix, rs_subject] if part and part.name])
```
saruniitr, line 89: it is probably more readable if we do it as `if-else`

api/goods/views.py
```diff
@@ -120,26 +138,38 @@ def post(self, request, case_pk):
...
+            old_report_summary = good.report_summary
             old_control_list_entries = list(good.control_list_entries.values_list("rating", flat=True))
             old_is_controlled = good.is_good_controlled
-            old_report_summary = good.report_summary
             old_regime_entries = list(good.regime_entries.values_list("name", flat=True))
-            obj = serializer.save()
 
-            report_summary_updated = request.data["report_summary"] != old_report_summary
+            data["report_summary_prefix"] = {"id": request.data.get("report_summary_prefix_id")}
+            data["report_summary_subject"] = {"id": request.data.get("report_summary_subject_id")}
```
saruniitr, line 162: Once the serializer is updated to use `PrimaryKeyRelatedField()` then this won't be required

.gitmodules
antroy-madetech: I know, but there are changes on that branch I need for this to work. I'll revert to main once the branch on lite_routing is merged.

api/cases/migrations/0036_case_additional_contacts.py
antroy-madetech: This is one of the dependencies Brendan added in order to get the new migration working. Once the ticket is functionally complete, I can try removing these one by one and re-running the migration to ensure we have minimal deps added?

api/goods/views.py
antroy-madetech: Return a 400 with appropriate message to the client?

api/goods/views.py
```diff
@@ -87,7 +122,8 @@ def get_queryset(self):
             pks = [pks]
         if self.application.case_type.sub_type in [CaseTypeSubTypeEnum.OPEN, CaseTypeSubTypeEnum.HMRC]:
             return GoodsType.objects.filter(pk__in=pks)
-        return GoodOnApplication.objects.filter(application_id=self.kwargs["case_pk"], good_id__in=pks)
+        objects_filter = GoodOnApplication.objects.filter(application_id=self.kwargs["case_pk"], good_id__in=pks)
+        return objects_filter
```
saruniitr, line 126: `objects_filter` is not used for anything other than returning, I think this change is not necessary.

api/goods/views.py
```diff
@@ -73,6 +74,40 @@
...
+        summary = rs_subject.name
+
+    if prefix_id is not None:
+        try:
+            rs_prefix = ReportSummaryPrefix.objects.get(id=prefix_id)
+        except ReportSummaryPrefix.DoesNotExist:
+            raise ValidationError(GOOD_ON_APP_BAD_REPORT_SUMMARY_PREFIX)
+
+    if rs_subject is not None and rs_prefix is not None:
+        summary = " ".join([rs_prefix.name, rs_subject.name])
```
saruniitr, line 105: If we are in this `if` then it can be simply be,
`summary = f"{rs_prefix.name} {rs_subject.name}"`

With `join` it doesn't make it obvious.

api/goods/views.py
```diff
@@ -73,6 +74,40 @@
...
+
+    if summary is None:
+        raise ValidationError(GOOD_ON_APP_NO_REPORT_SUMMARY)
```
saruniitr, line 108: this can be at the beginning itself because if there is no `report_summary` we can bail out early

api/goods/serializers.py
```diff
@@ -903,6 +904,12 @@ class ControlGoodOnApplicationSerializer(GoodControlReviewSerializer):
         queryset=RegimeEntry.objects.all(),
         required=False,  # not required for backwards compatibility reasons so that the old UI will still work
     )
+    report_summary_prefix_id = PrimaryKeyRelatedField(
+        required=False, allow_null=True, queryset=ReportSummaryPrefix.objects.all()
+    )
+    report_summary_subject_id = PrimaryKeyRelatedField(
+        required=False, allow_null=True, queryset=ReportSummarySubject.objects.all()
+    )
```
saruniitr, line 912: I think `report_summary_prefix` and `report_summary_subject` are better.

api/goods/models.py
```diff
@@ -161,6 +162,10 @@ class Good(TimestampableModel):
     grading_comment = models.TextField(default=None, blank=True, null=True, max_length=2000)
     # max length same as picklist
     report_summary = models.TextField(default=None, blank=True, null=True, max_length=5000)
+    # Report Summary prefix and subject
+    report_summary_prefix = models.ForeignKey(ReportSummaryPrefix, on_delete=models.CASCADE, blank=False, null=True)
+    report_summary_subject = models.ForeignKey(ReportSummarySubject, on_delete=models.CASCADE, blank=False, null=True)
+
```
saruniitr, line 168: migration file for this is not added yet

api/goods/tests/test_control_codes.py
```diff
@@ -306,6 +314,149 @@ def test_cannot_set_control_list_entries_when_application_in_terminal_state(self
         response = self.client.post(self.url, data, **self.gov_headers)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
+    def test_missing_report_summary_and_subject_gives_400_bad_request(self):
```
saruniitr, line 317: Also add a test where the assessments are cleared

api/goods/views.py
saruniitr: Actually this is causing issues.
When you do "Clear assessments" from the UI, `report_summary` comes as None and this causes it to fail.

api/goods/views.py
antroy-madetech: No - I'd put a breakpoint in between and not reverted.

api/goods/views.py
```diff
@@ -73,6 +74,40 @@
 
 good_overview_put_deletion_logger = logging.getLogger(settings.GOOD_OVERVIEW_PUT_DELETION_LOGGER)
 
+GOOD_ON_APP_NO_REPORT_SUMMARY = "Select a report summary subject"
+GOOD_ON_APP_BAD_REPORT_SUMMARY_PREFIX = "Select a valid report summary prefix"
+GOOD_ON_APP_BAD_REPORT_SUMMARY_SUBJECT = "Select a valid report summary subject"
+
+
+def get_new_report_data(request):
```
saruniitr, line 82: `get_new_report_summary_data()?`

api/staticdata/report_summaries/migrations/0003_update_tau_picklist.py
```diff
@@ -53,6 +53,8 @@ class Migration(migrations.Migration):
     dependencies = [
         ("report_summaries", "0002_add_report_summaries"),
         ("picklists", "0002_auto_20220110_1430"),
+        ("teams", "0007_auto_20220531_1344"),
+        ("lite_routing", "0001_initial"),
```
saruniitr, line 57: `lite-routing` submodule change below should be a separate change.

antroy-madetech: I think this has just updated to the latest HEAD of main after a merge.

saruniitr (approved):

PR #1260, opened 2023-01-27
LTD-3239: Add a 'my cases' tab/filter

Harriethw:
### aim

[LTD-3239](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3239)

The frontend for the cases/queues view is going to have several tabs that aim to filter out cases. We have already introduced the 'all cases' and 'open queries' tabs, but now that one more is being added ('my cases') we need to change the way the logic for filtering is handled (it's not either/or anymore).

[Together with this Frontend PR](https://github.com/uktrade/lite-frontend/pull/1064) I'm proposing:
 
_on the frontend_
1. instead of a param for each tab (e.g. `only_open_queries`) we have a param whose value is the selected tab e.g. `selected_tab=open_queries`
2. in the view, to get a count for each tab, we the tab names and make a HEAD request to get the count like we do currently, and save that in a dict for each tab, as well as its URL and whether it is selected.

_on the backend_
1. the view takes a look at the `selected_tab` value and passes the manager `True` for the name of the selected tab. e.g. `open_queries=True`
2. the manager can then apply the right filter. 

Would be great to get feedback on the structure/naming of this approach.

[LTD-3239]: https://uktrade.atlassian.net/browse/LTD-3239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -184,6 +184,12 @@ class CaseQuerySet(models.QuerySet):
     def only_open_queries(self):
         return self.filter(case_ecju_query__isnull=False, case_ecju_query__response__isnull=True)
 
+    def only_my_cases(self, user):
+        assigned_as_case_officer_case_ids = get_assigned_as_case_officer_case_ids(user)
+        assigned_to_user_case_ids = get_assigned_to_user_case_ids(user)
+        cases = assigned_to_user_case_ids.union(assigned_as_case_officer_case_ids)
+        return self.filter(id__in=cases)
+
     def order_by_date(self, order="-"):
         """
         :param order: ('', '-')
@@ -255,7 +261,8 @@ class CaseManager(models.Manager):
         sla_days_elapsed=None,
         is_nca_applicable=None,
         is_trigger_list=None,
-        only_open_queries=False,
+        open_queries=None,
+        my_cases=None,
         **kwargs,
     ):
         """
@@ -369,8 +376,11 @@ class CaseManager(models.Manager):
         if is_trigger_list:
             case_qs = case_qs.with_trigger_list()
 
-        if only_open_queries:
+        if open_queries:
             case_qs = case_qs.only_open_queries()
+
+        if my_cases:
+            case_qs = case_qs.only_my_cases(user)
 
         if is_work_queue:
             case_qs = case_qs.annotate(
--- a/api/cases/views/search/views.py
+++ b/api/cases/views/search/views.py
@@ -132,11 +132,15 @@ class CasesSearchView(generics.ListAPIView):
     def get_filters(self, request):
         filters = {key: value for key, value in request.GET.items() if key not in ["hidden", "queue_id", "flags"]}
 
+        search_tabs = ("my_cases", "open_queries")
+        selected_tab = request.GET.get("selected_tab")
+        if selected_tab and selected_tab in search_tabs:
+            filters[selected_tab] = True
+
         filters["flags"] = request.GET.getlist("flags", [])
         filters["submitted_from"] = make_date_from_params("submitted_from", filters)
         filters["submitted_to"] = make_date_from_params("submitted_to", filters)
         filters["finalised_from"] = make_date_from_params("finalised_from", filters)
         filters["finalised_to"] = make_date_from_params("finalised_to", filters)
-        filters["only_open_queries"] = True if request.GET.get("only_open_queries") == "True" else False
 
         return filters
--- a/api/cases/tests/test_case_search.py
+++ b/api/cases/tests/test_case_search.py
@@ -248,10 +248,31 @@ class FilterAndSortTests(DataTestClient):
             case_type_reference = Case.objects.filter(pk=case["id"]).values_list("case_type__reference", flat=True)[0]
             self.assertEqual(case_type_reference, CaseTypeEnum.GOODS.reference)
 
-    def test_only_open_queries_gets_cases(self):
+    def test_tab_all_cases_search(self):
         """
         Given there is a case with an ECJU Query that has not been responded to
-        When a user requests to view cases with open queries
+        When the tab is 'all_cases' and a team queue is not chosen
+        Then all cases are returned
+        """
+        all_cases = self.application_cases + self.clc_cases
+
+        ## create an open ecju query that should be returned
+        case_with_open_query = self.application_cases[0]
+        self.create_ecju_query(case_with_open_query, gov_user=self.gov_user)
+
+        url = f"{self.url}?selected_tab=all_cases"
+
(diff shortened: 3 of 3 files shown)


api/cases/views/search/views.py
```diff
@@ -132,11 +132,15 @@ def get_case_queryset(self, user, queue_id, is_work_queue, include_hidden, filte
     def get_filters(self, request):
         filters = {key: value for key, value in request.GET.items() if key not in ["hidden", "queue_id", "flags"]}
 
+        search_tabs = ["all_cases", "my_cases", "open_queries"]
+        selected_tab = request.GET.get("selected_tab")
+        if selected_tab and selected_tab in search_tabs:
+            filters[selected_tab] = True
```
Harriethw, line 138: should I perhaps append something to the tab name here so that the names make more sense in the `manager`? e.g. `f"only_{selected_tab}"` or `f"{selected_tab}_tab"`

api/cases/managers.py
```diff
@@ -255,7 +261,8 @@ def search(  # noqa
         sla_days_elapsed=None,
         is_nca_applicable=None,
         is_trigger_list=None,
-        only_open_queries=False,
+        open_queries=None,
```
Harriethw, line 264: see below comment with question about naming for these params

api/cases/managers.py
```diff
@@ -184,6 +184,12 @@ def with_trigger_list(self):
     def only_open_queries(self):
         return self.filter(case_ecju_query__isnull=False, case_ecju_query__response__isnull=True)
 
+    def only_my_cases(self, user):
+        assigned_as_case_officer_case_ids = get_assigned_as_case_officer_case_ids(user)
+        assigned_to_user_case_ids = get_assigned_to_user_case_ids(user)
+        cases = assigned_to_user_case_ids.union(assigned_as_case_officer_case_ids)
```
saruniitr, line 190: Is it going to be clear for the user in the UI that this tab returns cases that are assigned to him as well as he is a case officer for some of those cases?
Assigned as case officer means he is responsible till the case reaches the end of its journey?

api/cases/managers.py
```diff
@@ -369,9 +376,13 @@ def search(  # noqa
         if is_trigger_list:
             case_qs = case_qs.with_trigger_list()
 
-        if only_open_queries:
+        if open_queries:
+            print("here in open queries")
```
saruniitr, line 380: debug statement

api/cases/views/search/views.py
```diff
@@ -132,11 +132,15 @@ def get_case_queryset(self, user, queue_id, is_work_queue, include_hidden, filte
     def get_filters(self, request):
         filters = {key: value for key, value in request.GET.items() if key not in ["hidden", "queue_id", "flags"]}
 
+        search_tabs = ["all_cases", "my_cases", "open_queries"]
```
saruniitr, line 135: this can be a tuple as we are not expecting to modify this

api/cases/views/search/views.py
```diff
@@ -132,11 +132,15 @@ def get_case_queryset(self, user, queue_id, is_work_queue, include_hidden, filte
...
+        selected_tab = request.GET.get("selected_tab")
```
saruniitr, line 136: is it worth giving a default value?

api/cases/managers.py
kevincarrogan, line 380: There's some debug lurking about here.

api/cases/managers.py
```diff
@@ -184,6 +184,12 @@ def with_trigger_list(self):
     def only_open_queries(self):
         return self.filter(case_ecju_query__isnull=False, case_ecju_query__response__isnull=True)
 
+    def only_my_cases(self, user):
+        assigned_as_case_officer_case_ids = get_assigned_as_case_officer_case_ids(user)
+        assigned_to_user_case_ids = get_assigned_to_user_case_ids(user)
+        cases = assigned_to_user_case_ids.union(assigned_as_case_officer_case_ids)
+        return self.filter(id__in=cases)
```
kevincarrogan, line 191: Do we have any idea how this query might end up performing?

api/cases/views/search/views.py
Harriethw: the default is set on the front end as `all_cases`, and we don't apply any filters if it is `all_cases`, so if we were to set it here we would then not do anything with it... so doesn't seem worth setting a default here, but can do if it makes it more readable?

api/cases/managers.py
Harriethw: I guess that's a question for product? I'm just implementing what I understand from the story - these should be cases where I'm assigned `as a case adviser or Licensing Unit case officer`, is that what this is doing here?

api/cases/views/search/views.py
saruniitr: Yeah I noticed the default in FE.

api/cases/managers.py
saruniitr: my understanding matches with yours so this changes looks good to me.

api/cases/views/search/views.py
Harriethw: I'm actually thinking I'll take out `all_cases` from line 135 as we currently don't do anything in the scenario that it's selected

api/cases/managers.py
Harriethw: I'm not sure tbh, I'm combining filters that are used elsewhere in the app and it doesn't seem too slow on local so was just crossing my fingers that it would be fine :)  would there be a more optimal way of filtering, e.g. doing it in one rather than the two separate ones?

kevincarrogan: I don't necessarily have a different recommendation for doing this, what you've written is right I've just seen Django _sometimes_ do silly things with queries like this and I've had to rewrite them as a result.

api/cases/views/search/views.py
Harriethw: I've changed line 135 to `search_tabs = ("my_cases", "open_queries")` and as I said I don't think we should set a default here because the default behaviour is to do nothing, so would be adding a key that is never read or used?

api/cases/managers.py
Harriethw: been running it locally and speed seems ok, sometimes a millisecond or so slower than current load time, sometimes a  millisecond faster or the same

api/cases/views/search/views.py
saruniitr: :+1:

saruniitr (approved):
looks good to me

PR #1262, opened 2023-01-30
add next_workflow_status to CaseStatus model

MarkThomasLite:
CaseStatus model update with new field next_workflow_status with foreign key linked to the same model.


api/staticdata/statuses/models.py
```diff
@@ -15,6 +15,7 @@ class CaseStatus(models.Model):
     workflow_sequence = models.PositiveSmallIntegerField(null=True)
     is_read_only = models.BooleanField(blank=False, null=True)
     is_terminal = models.BooleanField(blank=False, null=True)
+    next_workflow_status = models.ForeignKey("CaseStatus", on_delete=models.DO_NOTHING, null=True)
```
saruniitr, line 18: you also need `blank=True`

saruniitr (approved):

PR #1271, opened 2023-02-02
API endpoint for getting Report Summary data

antroy-madetech:
In order for the frontend to populate the report summary prefix and subject search fields, we need endpoints on the API.

### DIFF
--- a/api/staticdata/report_summaries/tests/test_views.py
+++ b/api/staticdata/report_summaries/tests/test_views.py
@@ -1,0 +1,28 @@
+from rest_framework.reverse import reverse
+
+from api.staticdata.report_summaries.models import ReportSummaryPrefix, ReportSummarySubject
+from test_helpers.clients import DataTestClient
+
+
+class ReportSummaryPrefixesTests(DataTestClient):
+    def test_get_report_summary_prefixes_OK(self):
+        url = reverse("staticdata:report_summaries:prefix")
+        response = self.client.get(url, **self.exporter_headers)
+        prefixes = response.json()["report_summary_prefixes"]
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(len(prefixes) > 0)
+        self.assertEqual(len(prefixes), ReportSummaryPrefix.objects.count())
+        self.assertEqual(prefixes[0]["name"], ReportSummaryPrefix.objects.get(id=(prefixes[0]["id"])).name)
+
+
+class ReportSummarySubjectsTests(DataTestClient):
+    def test_get_report_summary_subjects_OK(self):
+        url = reverse("staticdata:report_summaries:subject")
+        response = self.client.get(url, **self.exporter_headers)
+        subjects = response.json()["report_summary_subjects"]
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(len(subjects) > 0)
+        self.assertEqual(len(subjects), ReportSummarySubject.objects.count())
+        self.assertEqual(subjects[0]["name"], ReportSummarySubject.objects.get(id=(subjects[0]["id"])).name)
--- a/api/staticdata/report_summaries/tests/test_views.py
+++ b/api/staticdata/report_summaries/tests/test_views.py
@@ -7,22 +7,34 @@ from test_helpers.clients import DataTestClient
 class ReportSummaryPrefixesTests(DataTestClient):
     def test_get_report_summary_prefixes_OK(self):
         url = reverse("staticdata:report_summaries:prefix")
-        response = self.client.get(url, **self.exporter_headers)
-        prefixes = response.json()["report_summary_prefixes"]
+        response = self.client.get(url, **self.gov_headers)
 
         self.assertEqual(response.status_code, 200)
+
+        prefixes = response.json()["report_summary_prefixes"]
         self.assertTrue(len(prefixes) > 0)
         self.assertEqual(len(prefixes), ReportSummaryPrefix.objects.count())
-        self.assertEqual(prefixes[0]["name"], ReportSummaryPrefix.objects.get(id=(prefixes[0]["id"])).name)
+
+        for prefix in prefixes:
+            db_prefix = ReportSummaryPrefix.objects.get(id=(prefix["id"]))
+
+            self.assertEqual(prefix["id"], str(db_prefix.id))
+            self.assertEqual(prefix["name"], db_prefix.name)
 
 
 class ReportSummarySubjectsTests(DataTestClient):
     def test_get_report_summary_subjects_OK(self):
         url = reverse("staticdata:report_summaries:subject")
-        response = self.client.get(url, **self.exporter_headers)
-        subjects = response.json()["report_summary_subjects"]
+        response = self.client.get(url, **self.gov_headers)
 
         self.assertEqual(response.status_code, 200)
+
+        subjects = response.json()["report_summary_subjects"]
         self.assertTrue(len(subjects) > 0)
         self.assertEqual(len(subjects), ReportSummarySubject.objects.count())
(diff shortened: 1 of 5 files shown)


api/staticdata/report_summaries/views.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class ReportSummaryPrefixView(APIView):
+    authentication_classes = (SharedAuthentication,)
+
+    def get(self, request):
+        """
+        Returns report summary prefixes
+        """
+        prefixes = ReportSummaryPrefixSerializer(ReportSummaryPrefix.objects.all(), many=True).data
```
currycoder, line 16: How many prefixes do we expect there to be?

api/staticdata/report_summaries/views.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class ReportSummarySubjectView(APIView):
+    authentication_classes = (SharedAuthentication,)
+
+    def get(self, request):
+        """
+        Returns report summary subjects
+        """
+        subjects = ReportSummarySubjectSerializer(ReportSummarySubject.objects.all(), many=True).data
```
currycoder, line 27: How many subjects do we expect there to be?

api/staticdata/report_summaries/tests/test_views.py
```diff
@@ -0,0 +1,28 @@
+from rest_framework.reverse import reverse
+
+from api.staticdata.report_summaries.models import ReportSummaryPrefix, ReportSummarySubject
+from test_helpers.clients import DataTestClient
+
+
+class ReportSummaryPrefixesTests(DataTestClient):
+    def test_get_report_summary_prefixes_OK(self):
+        url = reverse("staticdata:report_summaries:prefix")
```
currycoder, line 9: The records themselves are created in the DB by migration, right?

api/staticdata/report_summaries/tests/test_views.py
```diff
@@ -0,0 +1,28 @@
...
+        response = self.client.get(url, **self.exporter_headers)
+        prefixes = response.json()["report_summary_prefixes"]
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(len(prefixes) > 0)
+        self.assertEqual(len(prefixes), ReportSummaryPrefix.objects.count())
+        self.assertEqual(prefixes[0]["name"], ReportSummaryPrefix.objects.get(id=(prefixes[0]["id"])).name)
```
currycoder, line 16: Could the id also be checked here?  Could be good for the tests to assert the whole format for the response

api/staticdata/report_summaries/tests/test_views.py
```diff
@@ -0,0 +1,28 @@
...
+class ReportSummarySubjectsTests(DataTestClient):
+    def test_get_report_summary_subjects_OK(self):
+        url = reverse("staticdata:report_summaries:subject")
+        response = self.client.get(url, **self.exporter_headers)
+        subjects = response.json()["report_summary_subjects"]
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(len(subjects) > 0)
+        self.assertEqual(len(subjects), ReportSummarySubject.objects.count())
+        self.assertEqual(subjects[0]["name"], ReportSummarySubject.objects.get(id=(subjects[0]["id"])).name)
```
currycoder, line 28: Could the id also be checked here? Could be good for the tests to assert the whole format for the response

api/staticdata/report_summaries/views.py
```diff
@@ -0,0 +1,28 @@
+from django.http import JsonResponse
+from rest_framework.views import APIView
+
+from api.core.authentication import SharedAuthentication
+from api.staticdata.report_summaries.models import ReportSummaryPrefix, ReportSummarySubject
+from api.staticdata.report_summaries.serializers import ReportSummaryPrefixSerializer, ReportSummarySubjectSerializer
+
+
+class ReportSummaryPrefixView(APIView):
+    authentication_classes = (SharedAuthentication,)
```
currycoder, line 10: This is down as shared auth...  My understanding is that both exporters and caseworkers will be able to access this URL.  Is there any need for exporter users to access it?

api/staticdata/report_summaries/views.py
```diff
@@ -0,0 +1,28 @@
...
+    def get(self, request):
+        """
+        Returns report summary prefixes
+        """
+        prefixes = ReportSummaryPrefixSerializer(ReportSummaryPrefix.objects.all(), many=True).data
+        return JsonResponse(data={"report_summary_prefixes": prefixes})
+
+
+class ReportSummarySubjectView(APIView):
+    authentication_classes = (SharedAuthentication,)
```
currycoder, line 21: This is down as shared auth...  My understanding is that both exporters and caseworkers will be able to access this URL.  Is there any need for exporter users to access it?

api/staticdata/report_summaries/views.py
saruniitr: This is only required for caseworkers so `GovAuthentication` is enough

api/staticdata/report_summaries/views.py
saruniitr: Same as above, only caseworkers need to access this

api/staticdata/report_summaries/views.py
saruniitr: around 40,
https://github.com/uktrade/lite-api/blob/dev/api/staticdata/report_summaries/migrations/data/0002_add_report_summaries/report_summary_prefixes.csv

api/staticdata/report_summaries/views.py
saruniitr: these are about 1000

api/staticdata/report_summaries/tests/test_views.py
saruniitr: correct and they already exist

api/staticdata/report_summaries/views.py
antroy-madetech: OK, will change that.

api/staticdata/report_summaries/tests/test_views.py
antroy-madetech: The id is already checked here, as we are looking up the object by id. So it is testing that the id and name matches. I can make this more explicit if you want.

api/staticdata/report_summaries/tests/test_views.py
antroy-madetech: The id is already checked here, as we are looking up the object by id. So it is testing that the id and name matches. I can make this more explicit if you want.

api/staticdata/report_summaries/tests/test_views.py
antroy-madetech: Actually I expanded the test. Takes a second longer to check all entries.

api/staticdata/report_summaries/tests/test_views.py
antroy-madetech: Actually I expanded the test. Takes a second longer to check all entries.

api/staticdata/report_summaries/views.py
antroy-madetech: Done

api/staticdata/report_summaries/views.py
currycoder: Out of interest, how quickly does this web request return?  Just thinking through whether we might end up tying up our app/DB in some way without a bit of caching.  Maybe it's fine.

antroy-madetech: 60ms - seems fine :-)

currycoder (approved):

saruniitr (approved):

PR #1272, opened 2023-02-02
LTD-3303: Ensure products on licence are correctly updated after product assessment changes

saruniitr:
## Change description

Cases are sometimes sent back to update product assessment. If we try to
finalise them without consolidating the advice again then previously added
products will still remain on the licence. This can be an issue if they
have become NLR after product assessment changes. Identify this case
and exclude any NLR items when finalising the application.


manifest.yml
```diff
@@ -1,7 +1,7 @@
 applications:
   - buildpacks:
       - https://github.com/cloudfoundry/apt-buildpack.git
-      - python_buildpack
+      - https://github.com/cloudfoundry/python-buildpack.git#v1.8.3
```
saruniitr, line 4: please ignore, tot sure why this is showing up here, don't see this change locally

PR #1273, opened 2023-02-02
LTD-3303: Ensure products on licence are correctly updated after product assessment changes

saruniitr:
## Change description

Cases are sometimes sent back to update product assessment. If we try to finalise them without consolidating the advice again then previously added products will still remain on the licence. This can be an issue if they have become NLR after product assessment changes. Identify this case and exclude any NLR items when finalising the application.

### DIFF
--- a/api/applications/helpers.py
+++ b/api/applications/helpers.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.conf import settings
 
 from elasticsearch_dsl import Search, Q
@@ -47,6 +49,8 @@ from api.documents.libraries import s3_operations
 from api.external_data.models import SanctionMatch
 from api.licences.models import GoodOnLicence
 from lite_content.lite_api import strings
+
+logger = logging.getLogger(__name__)
 
 
 def get_application_view_serializer(application: BaseApplication):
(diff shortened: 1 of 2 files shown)


api/applications/helpers.py
```diff
@@ -177,6 +181,24 @@ def validate_and_create_goods_on_licence(application_id, licence_id, data):
...
 
+    # Exclude any NLR products that are associated with the licence
+    # Ideally these won't exist but in case of product assessment changes when the
+    # Case is sent back then it is possible
+    # Eg if we try to finalise it before consolidating final advice. In such cases
+    # initially a draft licence gets created and NLR products are also associated with
+    # the licence. After consolidation they would still remain because we update the
+    # draft if it exists hence we need to delete them manually
+    good_on_applications_ids = good_on_applications.values_list("id", flat=True)
+    nlr_products = GoodOnLicence.objects.filter(licence_id=licence_id).exclude(good_id__in=good_on_applications_ids)
```
kevincarrogan, line 192: Is `GoodOnLicence` here a many-to-many model? What does it link to and from?

Have we double checked that deleting these things doesn't cause some cascade deletes we don't want to happen?

kevincarrogan:
```
In [1]: from api.licences.models import GoodOnLicence

In [2]: GoodOnLicence.objects.count()
Out[2]: 1517

In [3]: import random

In [4]: good_on_licence = random.choice(GoodOnLicence.objects.all())

In [5]: good_on_licence
Out[5]: <GoodOnLicence: GoodOnLicence object (13390fdf-dd9d-4364-a2d2-96171df3b416)>

In [6]: good_on_licence.delete()
Out[6]: (1, {'licences.GoodOnLicence': 1})
```

That looks fine.

kevincarrogan (approved):

api/applications/helpers.py
saruniitr: `GoodOnLicence` is similar to `GoodOnApplication` but on the Licence.
I verified and there are no cascade deletes.

I also checked that in all the finalised applications, the count of products that require licence matches with the count of products on the licence (as it should),

```
     for case in Case.objects.filter(status__status='finalised').order_by('reference_code'):
         sa = StandardApplication.objects.get(id=case.id)
         licenced_goods = sa.goods.filter(is_good_controlled=True)
         for licence in case.licences.all():
             print(f"{sa.reference_code}, {licence.reference_code},   {licenced_goods.count() == licence.goods.count()}")
```

It was all `True` except in very old and early licences but that was mainly because there were few bugs before which are fixed now.

PR #1276, opened 2023-02-03
[LTD-3234] Remove case adviser

currycoder:
This adds a new endpoint to remove a case adviser.  This is located under the queues URL (case assignment is an assignment of a caseworker for a case located on a particular queue).  This keeps it close to an existing case assignment endpoint.

The endpoint creates an AuditTrail object when used.  In addition, audit trail text for assignments/removed users was updated.

This additionally ensures that the assignment id can be interrogated from the case detail endpoint (as this is used for showing assignments in the frontend, and we need to know the assignment ID to be able to call the new removal endpoint).


api/queues/views/case_assignments.py
```diff
@@ -62,3 +65,34 @@ def put(self, request, pk):
 
         # Return the newly set case assignments
         return self.get(request, pk)
+
+
+class CaseAssignmentDetail(views.APIView):
+    authentication_classes = (GovAuthentication,)
+
+    def _create_assignment_removal_audit_entry(self, assignment, actor):
+        removed_user_name = assignment.user.email
```
depsiatwal, line 74: is this mean't to be email or user_name ?

api/queues/views/case_assignments.py
```diff
@@ -62,3 +65,34 @@ def put(self, request, pk):
...
+            removed_user_name = f"{assignment.user.first_name} {assignment.user.last_name}"
+        audit_trail_service.create(
+            actor=actor,
+            verb=AuditType.REMOVE_USER_FROM_CASE,
+            target=assignment.case,
+            payload={
+                "removed_user_queue_id": str(assignment.queue.id),
+                "removed_user_queue_name": assignment.queue.name,
+                "removed_user_id": str(assignment.user.baseuser_ptr.id),
+                "removed_user_name": removed_user_name,
```
depsiatwal, line 85: oh I see it's called use_name in audit

api/queues/views/case_assignments.py
```diff
@@ -62,3 +65,34 @@ def put(self, request, pk):
...
         # Return the newly set case assignments
         return self.get(request, pk)
+
+
+class CaseAssignmentDetail(views.APIView):
+    authentication_classes = (GovAuthentication,)
+
+    def _create_assignment_removal_audit_entry(self, assignment, actor):
+        removed_user_name = assignment.user.email
+        if assignment.user.first_name and assignment.user.last_name:
```
antroy-madetech, line 75: Trivial thing, but I'd prefer this to be "if first and last: else: email"  to make it clearer that the email is the fallback name. Not precious about this, so approving and leaving it to you to decide.

antroy-madetech (approved):

api/queues/views/case_assignments.py
currycoder: Updated - how's that Ant?

api/queues/tests/test_case_assignment.py
```diff
@@ -1,3 +1,6 @@
+import pytest
+from parameterized import parameterized
+
```
antroy-madetech, line 3: These don't seem to be used

depsiatwal (approved):
LGTM

api/cases/tests/test_case_assignment_views.py
```diff
@@ -0,0 +1,68 @@
...
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+        assert response.json() == {"error": "No such case assignment"}
+
+    @parameterized.expand(
+        [
+            ("foo", "bar"),
+            ("", "bar"),
+            ("", ""),
+        ]
+    )
```
antroy-madetech, line 29: I find it's nice to provide a "name" param to more easily see which test out of a parameterized test method has failed.

api/cases/tests/test_case_assignment_views.py
```diff
@@ -0,0 +1,68 @@
...
+        assert not CaseAssignment.objects.filter(id=assignment.id).exists()
+
+        # Ensure removal audit entry was added
+        audit_entry = Audit.objects.get(verb=AuditType.REMOVE_USER_FROM_CASE)
+        assert audit_entry.actor == self.gov_user
+        assert audit_entry.target == assignment.case
+        expected_identifier = (
+            assignment.user.first_name + " " + assignment.user.last_name
+            if assignment.user.baseuser_ptr.first_name and assignment.user.baseuser_ptr.last_name
+            else assignment.user.email
```
antroy-madetech, line 61: I think expected_identifier would be better passed in as a parameter rather than duplicating the logic here.

antroy-madetech (approved):
LGTM

PR #1282, opened 2023-02-06
delete removed end user parties if application in draft state

MarkThomasLite:


api/applications/models.py
```diff
@@ -565,8 +569,11 @@ def __repr__(self):
         )
 
     def delete(self, *args, **kwargs):
-        self.deleted_at = timezone.now()
-        self.save()
+        if "is_draft" in kwargs:
+            super().delete()
```
kevincarrogan, line 573: What if someone passed in `is_draft=False`. What would expect to happen?

api/applications/tests/test_consignee.py
```diff
@@ -322,3 +322,74 @@ def test_consignee_copy_of_success(self):
...
+        super().setUp()
+        self.app = self.create_draft_standard_application(self.organisation)
+        self.submit_application(self.app)
+        self.app.status = get_case_status_by_status("applicant_editing")
+        self.app.save()
+        self.url = reverse("applications:parties", kwargs={"pk": self.app.id})
+        self.app.refresh_from_db()
+
+    @mock.patch("api.documents.models.Document.delete_s3")
+    def test_consignee_deleted_when_new_one_added(self, delete_s3_function):
```
kevincarrogan, line 338: Why is this checking stuff against S3?

api/applications/tests/test_consignee.py
```diff
@@ -322,3 +322,74 @@ def test_consignee_copy_of_success(self):
...
+        And the submitted aPP contains a consignee document
+        When there is an attempt to delete the consignee
+        Then 204 NO CONTENT is returned
+        """
+        consignee = PartyOnApplication.objects.get(
+            application=self.app, party__type=PartyType.CONSIGNEE, deleted_at__isnull=True
+        ).party
+        url = reverse("applications:party", kwargs={"pk": self.app.id, "party_pk": consignee.pk})
+        response = self.client.delete(url, **self.exporter_headers)
+        print(f"response:{response.json()}")
```
kevincarrogan, line 387: There's some debug here.

api/applications/tests/test_consignee.py
```diff
@@ -322,3 +322,74 @@ def test_consignee_copy_of_success(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
kevincarrogan, line 388: The comment says `Then 204 NO CONTENT is returned` but this is checking for a 200.

api/applications/tests/test_consignee.py
```diff
@@ -234,9 +235,8 @@ def test_delete_consignee_document_success(self, delete_s3_function, scan_docume
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
         delete_s3_function.assert_called_once()
 
-    @mock.patch("api.documents.tasks.scan_document_for_viruses.now")
     @mock.patch("api.documents.models.Document.delete_s3")
-    def test_delete_consignee_success(self, delete_s3_function, scan_document_for_viruses_function):
```
kevincarrogan, line 239: Why has the scan mock been removed?

api/applications/tests/test_end_user.py
```diff
@@ -140,7 +140,11 @@ def test_end_user_is_deleted_when_new_one_added(self, delete_s3_function):
         )
 
         self.client.post(self.url, self.new_end_user_data, **self.exporter_headers)
-        poa.refresh_from_db()
+        try:
+            poa.refresh_from_db()
+            self.fail(f"poa not deleted: {poa}")
```
kevincarrogan, line 145: Can we just write `PartyOnApplication` here? I think it's more readable to return that in the error than an acronym.

api/applications/tests/test_end_user.py
```diff
@@ -403,3 +407,28 @@ def test_end_user_copy_of_success(self):
...
+        Deletes the poa instead of expiring
+        """
+        poa = PartyOnApplication.objects.get(
+            application=self.submitted, party__type=PartyType.END_USER, deleted_at__isnull=True
+        )
+
+        poa.delete()
+        try:
+            poa = PartyOnApplication.objects.get(
+                application=self.submitted, party__type=PartyType.END_USER, deleted_at__isnull=True
```
kevincarrogan, line 430: I think it might be better to just use the `pk` here instead of these additional filters so we know it's definitely been deleted as opposed to a different value changing (we could have just set `deleted_at` here and the test would have passed and this is explicitly what we're testing).

hnryjmes (approved):
Approving with recommendation to rebase and squash prior to merging :)

api/applications/tests/test_end_user.py
```diff
@@ -403,3 +407,33 @@ def test_end_user_copy_of_success(self):
...
+        response = self.client.delete(url, **self.exporter_headers)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(
+            PartyOnApplication.objects.filter(
+                application=self.app, party__type=PartyType.END_USER, deleted_at__isnull=True
+            ).count(),
+            0,
+        )
+        try:
+            party_on_application = PartyOnApplication.objects.get(pk=end_user.id)
```
kevincarrogan, line 436: You're using `end_user.id` here but the `end_user` is actually the `party` object from line 423 and not the PartyOnApplication.

PR #1286, opened 2023-02-06
Report Summary endpoint filtering

antroy-madetech:
This PR adds name filtering to the report_summary prefix and subject endpoints.

Ordering is by prefixed results first and then alphabetically.

### DIFF
--- a/api/staticdata/report_summaries/helpers.py
+++ b/api/staticdata/report_summaries/helpers.py
@@ -1,0 +1,23 @@
+from django.db.models import BinaryField, Case, When
+
+
+def filtered_ordered_queryset(model_class, part_of_name):
+    """
+    Creates a queryset to get rows for the model where the
+    'name' column contains the part_of_name parameter. The results
+    are sorted so that entries where the name is prefixed by
+    the part_of_name appear higher in the rankings and
+    alphabetically after that.
+    """
+    queryset = model_class.objects.all()
+    if part_of_name:
+        queryset = queryset.filter(name__icontains=part_of_name)
+        queryset = queryset.annotate(
+            is_prefixed=Case(
+                When(name__istartswith=part_of_name.lower(), then=True),
+                default=False,
+                output_field=BinaryField(),
+            ),
+        )
+        queryset = queryset.order_by("-is_prefixed", "name")
+    return queryset
--- a/api/staticdata/report_summaries/helpers.py
+++ b/api/staticdata/report_summaries/helpers.py
@@ -1,15 +1,13 @@
 from django.db.models import BinaryField, Case, When
 
 
-def filtered_ordered_queryset(model_class, part_of_name):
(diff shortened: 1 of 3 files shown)


api/staticdata/report_summaries/views.py
```diff
@@ -13,8 +14,22 @@ def get(self, request):
...
+        if name:
+            prefixes = prefixes.filter(name__icontains=name)
+            prefixes = prefixes.annotate(
+                is_prefixed=Case(
+                    When(name__istartswith=name.lower(), then=True),
+                    default=False,
+                    output_field=BinaryField(),
+                ),
+            )
+            prefixes = prefixes.order_by("-is_prefixed", "name")
```
saruniitr, line 29: best to move this into `get_queryset()` method.

For my understanding, `is_prefixed` - is this to determine if the given string is in the middle or at the beginning?

antroy-madetech: Yes - exactly. It's so that the text prefixed by the letters typed appear at the top of the list, but other matches also appear but lower down. Will extract a method for this.

api/staticdata/report_summaries/views.py
```diff
@@ -6,15 +7,40 @@
...
+        queryset = queryset.filter(name__icontains=part_of_name)
+        queryset = queryset.annotate(
+            is_prefixed=Case(
+                When(name__istartswith=part_of_name.lower(), then=True),
+                default=False,
+                output_field=BinaryField(),
+            ),
+        )
+        queryset = queryset.order_by("-is_prefixed", "name")
+    return queryset
```
saruniitr, line 29: Actually there is a method in the base class `get_queryset(self)`, I meant to use that.
In this case you still need to call that like you are calling this method.

https://github.com/uktrade/lite-api/blob/dev/api/picklists/views.py#L45
https://github.com/uktrade/lite-api/blob/dev/api/goods/views.py#L142

api/staticdata/report_summaries/helpers.py
```diff
@@ -0,0 +1,23 @@
+from django.db.models import BinaryField, Case, When
+
+
+def filtered_ordered_queryset(model_class, part_of_name):
```
saruniitr, line 4: IMHO this function is not required, the implementation can directly go within `get_queryset()` method.
This would duplicate things in two views but that is fine in this case as they are separate endpoints. Reusing a helper function in this case just increases the coupling. They happen to use the same query param but if that changes then it breaks.

We can also specify `serializer_class` and remove `get()` (UI may need some changes as the result in included a specific key)

Interested to know what others think of this.

Harriethw (approved):
approved but agree with the convo about the naming of the function  :)

api/staticdata/report_summaries/views.py
antroy-madetech: Refactored.

saruniitr (approved):

PR #1291, opened 2023-02-07
Denials pagination

Harriethw:
## aim

[Ticket](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3294) 

The Denials search was only returning the first 10 results, this adds pagination to the view. 
~~The lite-routing test needs to be skipped first: https://github.com/uktrade/lite-routing/pull/131 ✅~~
~~The lite-routing test needs to temporarily accept both types of json results https://github.com/uktrade/lite-routing/pull/133 ✅~~
this lite routing test fix needs merging https://github.com/uktrade/lite-routing/pull/134/files ✅  
Then the SHA needs to be updated: https://github.com/uktrade/lite-api/pull/1304 ✅    
Then this PR merged ✖️ 
And frontend merged: https://github.com/uktrade/lite-frontend/pull/1095 ✖️ 
Then the test needs to be fixed: https://github.com/uktrade/lite-routing/pull/129/files  ✖️ 
And SHA updated again  ✖️ 

* although  `MaxPageNumberPagination` is set as the default class in settings, I've had to specify it here, maybe to override for the `DocumentViewSet`? We need to use that class because the frontend relies on the keys returned. 
* moved the `entity_type` key method to the Serializer
* this change would break any code relying on the old response format from the `let` method, but as far as I can see it's only be called once on the frontend, and we will change that.

### DIFF
--- a/api/external_data/serializers.py
+++ b/api/external_data/serializers.py
@@ -102,9 +102,25 @@ class DenialFromCSVFileSerializer(serializers.Serializer):
 
 
 class DenialSearchSerializer(DocumentSerializer):
+    entity_type = serializers.SerializerMethodField()
+
     class Meta:
         document = documents.DenialDocumentType
-        fields = ("denied_name",)
+        fields = (
+            "id",
+            "address",
+            "country",
+            "end_use",
+            "item_description",
+            "item_list_codes",
+            "name",
+            "notifying_government",
+            "reference",
+            "regime_reg_ref",
+        )
+
+    def get_entity_type(self, obj):
+        return get_denial_entity_type(obj.data.to_dict())
 
 
 class SanctionMatchSerializer(serializers.ModelSerializer):
--- a/api/external_data/tests/test_views.py
+++ b/api/external_data/tests/test_views.py
@@ -1,6 +1,7 @@
 import os
 
 from elasticsearch_dsl import Index
+from parameterized import parameterized
 import pytest
 
 from django.core.management import call_command
@@ -114,7 +115,13 @@ class DenialViewSetTests(DataTestClient):
 
 class DenialSearchView(DataTestClient):
     @pytest.mark.elasticsearch
-    def test_search(self):
+    @parameterized.expand(
+        [
+            ({},),
+            ({"page": 1},),
+        ]
+    )
+    def test_search_denials(self, page_query):
         call_command("search_index", models=["external_data.denial"], action="rebuild", force=True)
         # given some denials exist
         url = reverse("external_data:denial-list")
@@ -135,9 +142,12 @@ class DenialSearchView(DataTestClient):
         # then only 2 denials will be returned when searching
         url = reverse("external_data:denial_search-list")
 
-        response = self.client.get(url, {"search": "Example"}, **self.gov_headers)
+        response = self.client.get(url, {**page_query, "search": "Example"}, **self.gov_headers)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(response.json()["hits"]["hits"]), 2)
+        response_json = response.json()
+        self.assertEqual(len(response_json["results"]), 2)
+        self.assertEqual(response_json["total_pages"], 1)
+        assert "entity_type" in response_json["results"][0]
 
 
 class DenialSearchViewTests(DataTestClient):
--- a/api/external_data/views.py
+++ b/api/external_data/views.py
@@ -9,8 +9,8 @@ from django_elasticsearch_dsl_drf.viewsets import DocumentViewSet
 from django.conf import settings
 
 from api.core.authentication import GovAuthentication
+from api.conf.pagination import MaxPageNumberPagination
 from api.external_data import documents, models, serializers
-from api.external_data.helpers import get_denial_entity_type
 
 
 class DenialViewSet(viewsets.ModelViewSet):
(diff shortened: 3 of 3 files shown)


api/external_data/serializers.py
```diff
@@ -102,9 +102,26 @@ def add_bulk_errors(errors, row_number, line_errors):
...
+    entity_type = serializers.SerializerMethodField()
+
     class Meta:
         document = documents.DenialDocumentType
-        fields = ("denied_name",)
+        fields = (
+            "id",
+            "address",
+            "country",
+            "denied_name",
```
saruniitr, line 113: `denied_name` is probably the same as `name` and not sure this is required anymore. It is not there in the document also. If it is not used in the UI then best to remove this.

api/external_data/serializers.py
```diff
@@ -102,9 +102,26 @@ def add_bulk_errors(errors, row_number, line_errors):
...
 class DenialSearchSerializer(DocumentSerializer):
+    entity_type = serializers.SerializerMethodField()
+
     class Meta:
         document = documents.DenialDocumentType
-        fields = ("denied_name",)
+        fields = (
+            "id",
+            "address",
+            "country",
```
saruniitr, line 112: Could you order these fields similar to the document?

api/external_data/views.py
```diff
@@ -53,13 +54,6 @@ class DenialSearchView(DocumentViewSet):
...
-    def list(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
-        results = queryset.execute().to_dict()
-        for denial in results["hits"]["hits"]:
-            denial["_source"]["entity_type"] = get_denial_entity_type(denial["_source"]["data"])
-        return Response(results)
-
     def filter_queryset(self, queryset):
         queryset = queryset.filter("term", is_revoked=False)
         return super().filter_queryset(queryset)
```
saruniitr, line 59: looks like the `lite_routing` update below got added by mistake?

Harriethw: there needs to be changes to the lite_routing tests that use this endpoint

api/external_data/serializers.py
Harriethw: oh, but alphabetical is so much nicer 🥲

api/external_data/views.py
Harriethw: oh i see what you mean it's the commit update... the PR is still in draft as I am still adding and tidying up :)

api/external_data/tests/test_views.py
```diff
@@ -135,9 +135,12 @@ def test_search(self):
...
         url = reverse("external_data:denial_search-list")
 
-        response = self.client.get(url, {"search": "Example"}, **self.gov_headers)
+        response = self.client.get(url, {"search": "Example", "page": 1}, **self.gov_headers)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(response.json()["hits"]["hits"]), 2)
+        response_json = response.json()
+        self.assertEqual(len(response_json["results"]), 2)
+        self.assertEqual(response_json["total_pages"], 1)
+        assert "entity_type" in response_json["results"][0]
```
antroy-madetech, line 144: It would be good to see some more tests around this. e.g. what happens if "page" param is omitted. Here there are only 2 results - would be good to see tests where there are more than page size results and assert that a full page is returned for page 1 and some other number for the final page.

api/external_data/views.py
```diff
@@ -53,13 +54,6 @@ class DenialSearchView(DocumentViewSet):
         }
     }
 
-    def list(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
-        results = queryset.execute().to_dict()
-        for denial in results["hits"]["hits"]:
-            denial["_source"]["entity_type"] = get_denial_entity_type(denial["_source"]["data"])
-        return Response(results)
-
```
antroy-madetech, line 62: How come this is no longer needed?

api/external_data/tests/test_views.py
Harriethw: The pagination functionality is part of `MaxPageNumberPagination`, which inherits from the Django `PageNumberPagination` - so we'd be testing the framework logic, which I feel is out of scope for this test? By checking that we get `total_page`  were testing that it talks to the pagination side of things, and we can trust that to do its work?  but I agree that maybe a query without `page` in it could be good.

api/external_data/views.py
Harriethw: from what I understand of the `DocumentViewSet` and what Django is doing... if we keep `list` it overrides the `get_paginated_response` that would kick in by using `MaxPageNumberPagination` (even if we passed in `page` as query param). The logic to add the `entity_type` field has been moved to the serializer.

saruniitr: `list` returns paginated response if available otherwise just returns the queryset as list.
https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py#L39

This is one of the base class that `DocumentViewSet` inherits from.

saruniitr: The logic to add `entity_type` is in serializer now so this is no longer required. In the framework it will still be getting called, we just don't need to override anymore.

api/external_data/tests/test_views.py
saruniitr: yeah, a query without `page` param makes sense - that was the current behaviour anyway.

saruniitr (approved):

antroy-madetech (approved):
LGTM

PR #1292, opened 2023-02-07
LTD-3329: Add field to save trigger list status in GoodOnApplication

saruniitr:
## Change description

Currently we have `nsg_list_type` to save this data but the options provided in the UI are not correct hence replacing with a new boolean field.

Update serializers and tests.
`nsg_list_type` field will be removed in subsequent patches.


api/applications/tests/test_edit_good_on_applications.py
```diff
@@ -175,6 +178,7 @@ def test_edit_nsg_list_type_and_nca(self, non_terminal_status):
 
         for index, item in enumerate(response["data"], start=1):
             self.assertEqual(item["nsg_list_type"]["key"], NSGListType.TRIGGER_LIST)
+            self.assertTrue(item["is_trigger_list_guidelines_applicable"])
```
currycoder, line 181: Could it be possible to also go through the DB values and assert that they are now as expected?

api/applications/tests/test_edit_good_on_applications.py
```diff
@@ -140,6 +140,7 @@ def test_edit_nsg_list_type_and_nca(self, non_terminal_status):
 
         for good_on_application in [self.good_on_application, self.good_on_application2]:
             self.assertEqual(good_on_application.nsg_list_type, "")
+            self.assertIsNone(good_on_application.is_trigger_list_guidelines_applicable)
```
currycoder, line 143: Could it be possible to find/update some tests where `good_on_application.is_trigger_list_guidelines_applicable=False` AND where it is not set?  Would be good to see those along with assertions that the DB values are as expected.

api/applications/tests/test_edit_good_on_applications.py
saruniitr: :+1:

depsiatwal (approved):
also looks good to me

currycoder (approved):

PR #1302, opened 2023-02-10
LTD-3331: Update trigger list filter to use correct field

saruniitr:
## Change description

We have changed the options for trigger list assessment and introduced a new field to save that data. We also have a filter that needs to be updated to filter the cases correctly.


Harriethw (approved):

depsiatwal (approved):
lgtm

PR #1307, opened 2023-02-13
Dependabot/pip/cryptography 39.0.1

kevincarrogan:
This bumps the cryptography library to close off a dependabot alert.

When attempting to update the `cryptography` library there were various conflicts with the `pyOpenSSL` versions. To rectify this I have removed the need to use the `pyOpenSSL` library completely and swapped it out for the same functionality but exclusively using `cryptography`.


.circleci/config.yml
```diff
@@ -75,7 +75,9 @@ commands:
       - install_os_libraries
       - run:
           name: Install Dependencies
-          command: pipenv sync --dev
+          command: |
+            pipenv sync --dev
+            pipenv install endesive==1.5.9
```
currycoder, line 80: Remind me, why do we need to install this separately after the other pip deps?

api/documents/tests/test_certificate_download.py
```diff
@@ -1,13 +1,31 @@
...
...LGArG9v2H491tLqyM6avGGNqxNZ0YTy41ikjb6qPEebR0i63/YjNa1R0F0wdZZUf7xrwvEgpIQszIkfnmoz4sGHo5uqkUI8IhmoDsKqa8URe3gOh2z/+1y6CBydpUjRTpvJj9oemgfkSh8XaYUG+EZnPVsk4Pk5FX8Ys4qaCRXXw1sUO2fbSEB79CmjsLrJuAN5RXAX9/85qOaD5dDT1tOVCKohSs5+4N1DWcnh81xuy147GrLWnim1cNGSbS3zJcBq92FthEysE0ol3u7P+WE7QaF3PFiVQXU/DGawae7v6Q+mahwYLJkVG+t1m48qwJ3CB1Vzd1bGOkNuYVKAcEL7EIwzNWOcxzKDpTjjQNLXjwiJOPQcLH5ZRCjimqknhV3x4dkxj1ieNuwB+vQgSPDkWKvCx/ty89O9h1qMfEd4VOytMGAODQvYnVukgZOI98TwTm+IGMr1tkxRDNY+Rth5PXiU8LnsxfaF/VC96mMYx2sT0Uu/UIlEanP1k6K80pzOIiMEzXW2wrnqAzez27ivVYd1tRF7IywiV4jaZZICHhI7KEqBSmlCrJowD7Y/GK61QKbMMvS4k/y/aEz/W5pJ1zTy3sJ3SN9rqEiW8DIRuM4ld1oRyrRdRaeLA6lsDQD/fqusK4L25FO6wkcAMHq+VqnFAGiQ+qIGLfRiTfAVmQ7zZ3wmJV0RBF1yeTKbRT60Fwz8wr8F39j2jdKS4bdO59+Num0ChLUd1Y03DAFagsZcFqAlsuCpOJUgB9W516mH7glfkWecFnEwaeRwJPTuTFpq6u4pk/0ZFagBGo80lpSYRtdIl3W6qrQqSnElMu6KQcc1Aa421C+cd33h2h7+UMK2nyYsA1BsAOGsEq6bPwszne3L5G3PBXr5Z+2InbDMrLFodmH/v26maAG5qdm9xf19yMTsS+FOlqMjoH9Kr2dutScvrlNgx9VdaPw9FxkO/Zy+0VZ4pEH50mz0YepqgLG5zRFYLjGM+r/PSFmG1F9P/05qeRpNwRQzwEpg3bgT2DTy27CQIwho9EVFBlqh+wKGtUIptde46LTElMCMGCSqGSIb3DQEJFTEWBBQqZ0mSI6KXBV6SIvVhc4SWtnnvKjAxMCEwCQYFKw4DAhoFAAQUhkPBbGCaPNMXgdhTM8NnNrzrgjwECN/IAa9MiTnrAgIIAA==",  # /PS-IGNORE
```
currycoder, line 11: What's the source of this certificate?

Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN apt-get update --fix-missing
 RUN apt-get install -y libpq-dev gcc \
   build-essential python3-dev python3-pip python3-setuptools python3-wheel \
   python3-cffi libcairo2 libpango-1.0-0 libpangocairo-1.0-0 libgdk-pixbuf2.0-0 \
-  libffi-dev shared-mime-info swig git
+  libffi-dev shared-mime-info swig git imagemagick poppler-utils
```
currycoder, line 7: It's not immediately obvious to me where these new dependencies are used.  Could you explain?

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ commands:
           name: Install OS libraries
           command: |
             sudo apt-get update
-            sudo apt-get install swig libpango1.0-0 libcairo2 libpq-dev libpangocairo-1.0-0
+            sudo apt-get install swig libpango1.0-0 libcairo2 libpq-dev libpangocairo-1.0-0 imagemagick poppler-utils
```
currycoder, line 42: Is there a need to also update our jenkins builds to include these new dependencies?

Pipfile
```diff
@@ -52,9 +53,7 @@ notifications-python-client = "~=5.7.1"
 dataclasses = "~=0.6"
 # endesive = "~=1.5.9" (not included here as installed in a later jenkins step. Uncomment if working locally)
 pypdf2 = "~=1.27.5"
-pyOpenSSL = "~=22.0.0"
-# The following dependency stops the cryptography package jumping by a major version which breaks pyOpenSSL 22.x
-cryptography = "~=38.0"
+cryptography = "~=39.0.1"
```
currycoder, line 56: Just thinking through other parts of the app where this may be in use..  Do the auth mechanisms all still work etc?

.circleci/config.yml
kevincarrogan: It’s not part of our Pipfile as it fails to install on cloudfoundry due to it trying to install things from source and cloudfoundry doesn’t have the required libraries to do that.

api/documents/tests/test_certificate_download.py
kevincarrogan: Me making it for this test. There are instructions in our Confluence on how to make it.

Dockerfile
kevincarrogan: The new pdf diff library requires them.

Pipfile
kevincarrogan: Yes. This library is only used here from what I can tell.

.circleci/config.yml
kevincarrogan: The end to end tests use that dependency image and they don’t run through the pdf diff tool so won’t need these additional dependencies, plus if I update it it’s going to break our version of chrome and all of our end to end tests.

Pipfile
kevincarrogan: I've also deployed and tested this on our `demo` environment and it's working fine. My main concern would have been whether this would have caused side effects in our HAWK authentication but this doesn't seem to have affected that in any way.

kevincarrogan: I've also inspected the `pipenv graph` and `cryptography` isn't a sub-dependency of any other libraries.

hnryjmes (approved):

Pipfile
currycoder: Awesome, thanks for being thorough here

currycoder (approved):

PR #1309, opened 2023-02-14
Added report_summary{prefix,subject} to the Good serializer

antroy-madetech:
The frontend needs access to the new prefix and subject fields in order to display those fields. This PR adds those to the good serializer to make them available.


Pipfile
```diff
@@ -73,6 +73,7 @@ importlib-metadata = "==4.8.1"  # Not needed for Python3.8+
 celery = "~=5.2.7"
 redis = "~=4.0.2"
 psycopg2-binary = "~=2.9.3"
+guid = "*"
```
saruniitr, line 76: why is this required? also not used.

antroy-madetech: Typo of uuid, and then the IDE helpfully brought the dependency in for me!

saruniitr (approved):

api/goods/tests/test_serializers.py
```diff
@@ -11,7 +12,8 @@
 from api.applications.tests.factories import PartyOnApplicationFactory
 
 from .factories import GoodFactory
-from ..serializers import GoodOnApplicationSerializer
+from ..serializers import GoodOnApplicationSerializer, GoodSerializerInternal
+from ...staticdata.report_summaries.models import ReportSummaryPrefix, ReportSummarySubject
```
currycoder, line 16: In general, we do absolute imports rather than relative - makes things a lot more verbose and easier to read, I feel.  (I realise you were probably just being consistent with the existing line, but I think that's worth changing too)

api/goods/tests/factories.py
```diff
@@ -1,9 +1,12 @@
+import uuid
+
 import factory
 from django.utils import timezone
 
 from api.goods import models
 from api.goods.enums import ItemCategory, Component, MilitaryUse, FirearmGoodType, GoodPvGraded
 from api.staticdata.control_list_entries.helpers import get_control_list_entry
+from api.staticdata.report_summaries.models import ReportSummarySubject, ReportSummaryPrefix
```
currycoder, line 9: Are these imports used?

api/data_workspace/tests/test_good_views.py
```diff
@@ -25,6 +25,8 @@ def test_goods(self):
             "comment",
             "is_good_controlled",
             "report_summary",
+            "report_summary_prefix",
```
currycoder, line 28: This is for the data workspace endpoint... I would expect there to be a test to adjust for the endpoint that you are consuming in the frontend too.  Could that be found and adjusted?

saruniitr: The FE for this is the Data workspace pipeline pulling these data but currently these pipelines don't use these fields yet so nothing breaks. These pipelines need `report_summary` which we are not removing yet for this reason.
Once all things in place we will be updating the pipelines to use subject/prefix and then remove `report_summary`

api/goods/tests/factories.py
saruniitr: I can remove these

api/goods/tests/test_serializers.py
saruniitr: I will change this as well

api/data_workspace/tests/test_good_views.py
currycoder: That's all fine, yeah.  But my point here is that I don't see tests for the other endpoints where these fields were added - only tests for the serializers.  So I was hoping that the tests for those other endpoints (which presumably the FE is consuming) would be adjusted to account for these new fields.

saruniitr: right, they are already updated in previous PRs, eg
https://github.com/uktrade/lite-api/blob/dev/api/goods/tests/test_control_codes.py#L363

This is a follow up PR where Ant noticed that these are not exposed in `Good` related serializers so this is mainly to expose these fields here.
These are mainly set in `GoodOnApplication` when product is assessed but similar to CLEs these are also updated in `Good` after assessment.

api/goods/tests/test_serializers.py
saruniitr: replaced with absolute imports

kevincarrogan (approved):

api/data_workspace/tests/test_good_views.py
currycoder: The tests for endpoints that use `GoodOnApplicationSerializer` are not updated.  So we have tests in this PR which unit test the serializers; that's fine..

But the goal of this PR is;
> The frontend needs access to the new prefix and subject fields in order to display those fields.

IMO, the tests here do not prove that this is functional in actual API calls.

I believe it would be a good thing to see some changes to some of the following tests that do prove that;
https://github.com/uktrade/lite-api/blob/dev/api/applications/tests/test_edit_application.py
https://github.com/uktrade/lite-api/blob/dev/api/applications/tests/test_retrieve_good_on_applications.py

saruniitr: The first link https://github.com/uktrade/lite-api/blob/dev/api/applications/tests/test_edit_application.py is mainly for the Exporter side editing applications.
`report_summary` is only visible in caseworker side. Exporter site is not aware of this and they don't supply this field.
I am not sure of the need to test for these fields here. These tests passing proves that these are independent of these changes.
Currently it uses the same serializer for both exporter and caseworker, infact we should have an exporter specific serializer which doesn't have these fields but that is a separate task altogether.

The second link https://github.com/uktrade/lite-api/blob/dev/api/applications/tests/test_retrieve_good_on_applications.py, perhaps we can update the tests to check if these fields exist in the response.
But these are too generic, we are also not checking for CLEs etc.

@currycoder  Do you agree that updating tests in second link is enough?

saruniitr: tests in the second link are updated

api/applications/tests/test_retrieve_good_on_applications.py
```diff
@@ -18,7 +18,12 @@ def test_exporter_retrieve_a_good_on_application(self):
 
         response = self.client.get(url, **self.exporter_headers)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertIn("audit_trail", response.json())
+        response = response.json()
+        self.assertTrue(
+            set(["audit_trail", "report_summary", "report_summary_prefix", "report_summary_subject"]).issubset(
```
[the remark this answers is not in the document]
saruniitr: hmm, to avoid force-pushing during review will update when squashing commits.

api/data_workspace/tests/test_good_views.py
currycoder: I mean... This kind of proves my point.

We innocently changed a serializer and are discovering during a PR conversation that doing this leads to leaking internal details on exporter API endpoints..  I've had a quick look through the frontend and it seems like data is validated well enough to defend against naughty users - but that might not always be the case and there's a hole in our API endpoints here.

If we only test serializers (low value IMO) and not the actual API endpoints that we are changing, then we aren't really going to have good visibility of our changes and be able to tell how our overall input/output of our application has changed and whether that is safe.

So overall, yeah I do really feel like we should adjust tests to prove the changed inputs/outputs of every endpoint this serializer gets used for!
More than that, we should probably close this hole right now rather than widening it.

What do you think?  I realise that there is an element of "this is debt", but if we don't change it now then when will we?

currycoder: We discussed this a bit offline and came to the conclusion that this would be covered in a follow up ticket.

currycoder (approved):

PR #1312, opened 2023-02-15
LTD-3028-Adds-property

seijihg:
Part 1
- add pending @property to model/serializer/endpoints based off of user.firstname
- swap out logic on frontend to use new pending field

Part 2
- swap out pending @property to be an actual DB field that’s updated when user first logs in

Related ticket to the FE:
https://github.com/uktrade/lite-frontend/pull/1101

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-3028

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/api/cases/views/search/service.py
+++ b/api/cases/views/search/service.py
@@ -3,8 +3,7 @@ from typing import List, Dict
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
-from django.db.models import Count, F, Q, Value, OuterRef
-from django.db.models.functions import Concat
+from django.db.models import Count, F, Q, OuterRef
 from django.utils import timezone
 
 from api.applications.models import HmrcQuery, PartyOnApplication
@@ -37,7 +36,7 @@ def get_gov_users_list():
             "id": i.baseuser_ptr.id,
             "pending": i.pending,
         }
-        for i in GovUser.objects.all()
+        for i in GovUser.objects.filter(status=UserStatuses.ACTIVE)
     ]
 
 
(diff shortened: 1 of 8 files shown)


api/cases/views/search/service.py
```diff
@@ -31,11 +29,14 @@ def get_case_type_type_list() -> List[Dict]:
 
 
 def get_gov_users_list():
-    return (
-        GovUser.objects.filter(status=UserStatuses.ACTIVE)
```
kevincarrogan, line 35: This used to filter by active users but now you’ve removed that filter. Is that right?

api/users/models.py
```diff
@@ -221,6 +221,12 @@ def has_permission(self, permission):
         user_permissions = self.role.permissions.values_list("id", flat=True)
         return permission.name in user_permissions
 
+    @property
+    def pending(self):
+        if self.baseuser_ptr.first_name:
```
saruniitr, line 226: When it is pending does the `first_name` will always be empty?
Can it be `None`? in that case also we consider this as pending?

kevincarrogan (approved):

api/users/models.py
seijihg: I cant be None at the moment since it gets "" if its None/Empty etc at the Authorisation part. Anyway this is just the first part of the ticket.

seijihg: Yea it seems so. It will be with Name or Empty because the field cannot be None due to not-null constraint in DB.

saruniitr: @seijihg then it can be simplified as below, you don't need `if-else`

```
def pending(self):
    return self.baseuser_ptr.first_name == ""   # this itself evaluates to True/False
```

depsiatwal (approved):
lgtm

currycoder (approved):

PR #1314, opened 2023-02-16
LTD-3341: add `alias` field to queue detail response

Harriethw:
### aim

We want to stop rendering content on the frontend based on what queue is being viewed - rather than using the UUID of the queue we would like to use the `alias` [(chat here)](https://ditdigitalteam.slack.com/archives/C015AHPB5QS/p1676463320645939?thread_ts=1676462912.099809&cid=C015AHPB5QS)

This PR adds an alias for the LU pre-circ queue and returns the alias field in the queue detail view, which is the endpoint the queues view on the frontend uses to determine what queue is being viewed.


api/queues/migrations/0006_populate_pre_circ_queue_alias.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.11 on 2022-01-27 09:48
+
+from django.db import migrations
+
+
+def populate_queue_alias(apps, schema_editor):
+    alias = "LU_PRE_CIRC_REVIEW"
+    Queue = apps.get_model("queues", "Queue")
+    pre_circ_queue = Queue.objects.filter(name__iexact="Licensing Unit Pre-circulation Cases to Review")
```
currycoder, line 9: Could I suggest just using the ID here instead of looking up by `name`?  ID will be consistent across environment AND if we change the name of this queue by a migration in the future, we won't need to worry too much about the sequence that the migrations execute in.  e.g. if we chose to rename this queue in `lite-routing` and somehow that migration gets to run before this one (unlikely but possible)

currycoder (approved):

api/queues/migrations/0006_populate_pre_circ_queue_alias.py
saruniitr: yes, using `id` is better than `name`

saruniitr (approved):
with the migration updated to use `id` it looks good to me

api/queues/migrations/0006_populate_pre_circ_queue_alias.py
Harriethw: I originally used ID, but then I saw that we were using name to check elsewhere, and the ID is from the `lite_routing` repo... if we're ok to use it though I will put back in

Harriethw: pls can you double check I've used the ID correctly 🙏

currycoder:
Yup looks good to go

PR #1321, opened 2023-02-23
Change CASCADE to PROTECT in the GoodOnApplication model

antroy-madetech:
CASCADE would delete the good on application if one of it's referenced prefix or subject report summary references was deleted. This change stops the prefix or summary being deleted if they are being used by a good on application.


kevincarrogan (approved):

hnryjmes (approved):
The code looks fine. The second sentence in your description is confusing me. Surely this `This change stops the prefix or summary being deleted if they are being used by a good on application.` should be something like, this change stops the good on application being deleted if it is referenced by a prefix or summary that is deleted

PR #1322, opened 2023-02-23
LTD-3389-3390: Model changes to record Countersigning data

saruniitr:
## Change description

For LU countersigning we need to record countersigning decision also as user can accept/reject original outcome.
The current countersigning fields in `Advice` model doesn't support this.
We can extend the above fields but for some cases double countersigning is required.
To support all these requirements and future proof it a separate model is introduced to save countersigning data.
This will support multiple countersigners and also give us the flexibility to add additional fields.

LU countersigning work is under a feature flag so to better isolate UI changes without affecting countersigning flow for other teams (eg FCDO) a new endpoint is also created that uses this new model. FCDO can continue to use existing flow and eventually they will be migrated to use this in future.

Add tests to cover possible cases.


api/cases/models.py
```diff
@@ -537,6 +537,16 @@ def equals(self, other):
         )
 
 
+class Countersign(TimestampableModel):
+    order = models.PositiveIntegerField(help_text="Indicates countersigning order")
+    outcome_accepted = models.BooleanField(default=None)
+    reasons = models.TextField(
```
saruniitr, line 543: User need to provide reasons whether they accept/reject recommendation. It is mandatory in both cases.
The plan is to use a single field but open to changing to separate fields also.

api/applications/serializers/advice.py
```diff
@@ -196,6 +197,23 @@ class Meta:
...
+        return result
+
+
+class CountersignAdviceWithDecisionSerializer(serializers.ModelSerializer):
+    countersigned_user = serializers.PrimaryKeyRelatedField(queryset=GovUser.objects.filter(status=UserStatuses.ACTIVE))
+    advice = serializers.PrimaryKeyRelatedField(queryset=Advice.objects.all())
+
+    class Meta:
+        model = Countersign
+        fields = "__all__"
```
kevincarrogan, line 213: Isn't this considered an anti-pattern and we should specify fields explicitly?

api/applications/serializers/advice.py
```diff
@@ -196,6 +197,23 @@ class Meta:
         list_serializer_class = CountersignAdviceListSerializer
 
 
+class CountersignAdviceWithDecisionListSerializer(serializers.ListSerializer):
+    def update(self, instances, validated_data):
+        instance_map = {index: instance for index, instance in enumerate(instances)}
+        result = [self.child.update(instance_map[index], data) for index, data in enumerate(validated_data)]
+        return result
```
kevincarrogan, line 204: What's the use case for us updating multiple decisions at once?

api/cases/models.py
kevincarrogan: Once we create this object will they ever change the outcome?

api/cases/views/views.py
```diff
@@ -1040,6 +1046,84 @@ def put(self, request, **kwargs):
...
+            department = "department"
+
+        audit_trail_service.create(
+            actor=request.user,
+            verb=AuditType.COUNTERSIGN_ADVICE,
+            target=case,
+            payload={"department": department},
+        )
+
+        return JsonResponse({"countersigned_advice": serializer.data}, status=status.HTTP_200_OK)
```
kevincarrogan, line 1124: There appears to be a lot of overlap here with the `post` method. Could this be consolidated in some way?

api/cases/models.py
saruniitr: Before moving case foward they will have the option to Edit their countersigning decision.
So initially they may have said "Yes" but they could change their mind to "No" and then move case forward.

api/applications/serializers/advice.py
saruniitr: explained in internal channel

antroy-madetech (commented):
Presumably we'll do the migration to copy the existing text field countersignatures into the new table when we do the work to move all other countersignatures to the new model?

antroy-madetech (commented):
Looks like codecov has picked up a few lines of code that are missing tests. Would be good to get those covered.

api/cases/views/views.py
```diff
@@ -1040,6 +1046,84 @@ def put(self, request, **kwargs):
...
+            department = department.name
+        else:
+            department = "department"
+
+        audit_trail_service.create(
+            actor=request.user,
+            verb=AuditType.COUNTERSIGN_ADVICE,
+            target=case,
+            payload={"department": department},
+        )
```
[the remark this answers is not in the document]
saruniitr: :+1:

api/cases/views/views.py
```diff
@@ -1040,6 +1046,71 @@ def put(self, request, **kwargs):
         return JsonResponse({"advice": serializer.data}, status=status.HTTP_200_OK)
 
 
+class CountersignAdviceV2(APIView):
```
depsiatwal, line 1049: V2 ? 
can we get a better name ?
are migrating off original one

saruniitr: @depsiatwal 
yeah open to suggestions.

The main difference is countersigners will be able to provide accept/refuse decision now so I included "WithDecision" for some of the serializer names so may similar can be done here?
Please suggest any other alternatives.

antroy-madetech (approved):

api/cases/views/views.py
depsiatwal: perhaps we can use CountersignDecisionAdvice

api/applications/serializers/advice.py
```diff
@@ -196,6 +197,33 @@ class Meta:
...
+    case = serializers.PrimaryKeyRelatedField(queryset=Case.objects.all())
+    advice = serializers.PrimaryKeyRelatedField(queryset=Advice.objects.all())
+
+    class Meta:
+        model = CountersignedAdvice
+        fields = ("id", "order", "outcome_accepted", "reasons", "countersigned_user", "case", "advice")
+        list_serializer_class = CountersignAdviceWithDecisionListSerializer
+
+
+class CountersignedAdviceWithDecisionViewSerializer(serializers.Serializer):
```
hnryjmes, line 218: Please can this be `CountersignAdviceWithDecisionViewSerializer` (Countersigned -> Countersign) if possible

api/cases/models.py
```diff
@@ -537,6 +537,18 @@ def equals(self, other):
         )
 
 
+class CountersignedAdvice(TimestampableModel):
```
hnryjmes, line 540: Please can this be `CountersignAdvice` (Countersigned -> Countersign) if possible

api/cases/views/views.py
```diff
@@ -1040,6 +1046,71 @@ def put(self, request, **kwargs):
...
+            CountersignedAdvice.objects.filter(id__in=countersign_ids), data=request.data, partial=True, many=True
+        )
+        if not serializer.is_valid():
+            return JsonResponse({"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        serializer.save()
+
+        self.audit_countersign(request, case)
+
+        return JsonResponse({"countersigned_advice": serializer.data}, status=status.HTTP_200_OK)
```
hnryjmes, line 1111: Please can this be `{"countersign_advice": serializer.data}` (countersigned -> countersign) if possible

api/cases/models.py
saruniitr: ok :+1:

api/cases/views/views.py
```diff
@@ -1040,6 +1046,71 @@ def put(self, request, **kwargs):
...
+
+        serializer = self.serializer_class(data=data, many=True)
+        if not serializer.is_valid():
+            return JsonResponse({"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        serializer.save()
+
+        self.audit_countersign(request, case)
+
+        return JsonResponse({"countersigned_advice": serializer.data}, status=status.HTTP_201_CREATED)
```
hnryjmes, line 1095: Please can this be `{"countersign_advice": serializer.data}` (countersigned -> countersign) if possible

api/cases/tests/test_get_case.py
```diff
@@ -131,6 +132,21 @@ def test_case_returns_has_advice(self):
         self.assertIn("my_team", has_advice_response_data)
         self.assertIn("final", has_advice_response_data)
 
+    def test_case_returns_countersigned_advice(self):
+        case = self.submit_application(self.standard_application)
+        url = reverse("cases:case", kwargs={"pk": case.id})
+        advice = FinalAdviceFactory(case=case, user=self.gov_user)
+        countersigned_advice = CountersignAdviceFactory(case=case, advice=advice)
```
hnryjmes, line 139: Please can this be `countersign_advice` (countersigned -> countersign) if possible

api/applications/serializers/advice.py
saruniitr: Actually this is already named as `Countersign...`
but following @depsiatwal suggestion above, I will update this as `CountersignDecisionAdviceViewSerializer`

api/cases/views/views.py
saruniitr: :+1:

api/cases/views/views.py
saruniitr: :+1:

api/cases/tests/test_get_case.py
saruniitr: :+1:

hnryjmes:
I think you already understand this @saruniitr  but for the avoidance of doubt, yes, everywhere we have `countersigned` as a variable name I think this should be `countersign` just because I feel strongly we should just pick one and make it consistent everywhere, even if 'countersigned' might sometimes be more appropriate in written English. I know this is a lot to change but I think future devs will appreciate it.

api/cases/models.py
```diff
@@ -537,6 +537,18 @@ def equals(self, other):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    order = models.PositiveIntegerField(help_text="Indicates countersigning order")
+    outcome_accepted = models.BooleanField()
+    reasons = models.TextField(
+        help_text="Reasons provided by the countersigner when they agree/disagree with the advice during countersigning",
+    )
+    countersigned_user = models.ForeignKey(GovUser, on_delete=models.DO_NOTHING, related_name="countersigned_user")
```
hnryjmes, line 547: not really a suggestion, more of a question for my own understanding - if we have `on_delete=models.DO_NOTHING` shown here, that means, when a GovUser is deleted, this field then contains a reference to a deleted object, or the field contains None ?

saruniitr: I think it becomes `None`. We want to keep this object in that case for auditing purposes.

api/cases/tests/test_create_advice.py
```diff
@@ -267,3 +271,151 @@ def test_countersign_advice_success(self):
         audit_qs = Audit.objects.filter(verb=AuditType.COUNTERSIGN_ADVICE)
         self.assertEqual(audit_qs.count(), 1)
         self.assertEqual(audit_qs.first().actor, self.gov_user)
+
+
+class CountersignAdviceWithDecisionTests(DataTestClient):
```
depsiatwal, line 276: does this now require renaming remove 'with'

hnryjmes (approved):

PR #1323, opened 2023-02-23
email notification for cases sent back by countersign

MarkThomasLite:
take application and send email notification to caseworker that needs to act on the comments

notify template: https://www.notifications.service.gov.uk/services/f3d8fb42-a34b-4d85-8ac2-a62006a197dc/templates/4d418015-cd7c-498f-a972-72e7ec4468cc

### DIFF
--- a/api/applications/notify.py
+++ b/api/applications/notify.py
@@ -28,7 +28,7 @@ def notify_exporter_case_opened_for_editing(application):
 
 
 def notify_caseworker_countersign_return(user_email, application, countersign_advice):
-    relative_url = reverse("cases:countersign_advice", kwargs={"pk": application.id})
+    relative_url = reverse("cases:countersign_decision_advice", kwargs={"pk": application.id})
     countersigner = countersign_advice.countersigned_user
     data = {
         "case_reference": application.reference_code,
--- a/api/applications/notify.py
+++ b/api/applications/notify.py
@@ -28,21 +28,16 @@ def notify_exporter_case_opened_for_editing(application):
     )
 
 
-def notify_caseworker_countersign_return(application):
-    advice = (
-        application.advice.filter(type=AdviceType.REFUSE, level=AdviceLevel.USER, countersigned_by__isnull=False)
-        .order_by("-updated_at")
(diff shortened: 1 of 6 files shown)


api/applications/notify.py
```diff
@@ -23,3 +26,25 @@ def notify_exporter_case_opened_for_editing(application):
...
         },
     )
+
+
+def notify_caseworker_countersign_return(application):
+    advice = (
+        application.advice.filter(type=AdviceType.REFUSE, level=AdviceLevel.USER, countersigned_by__isnull=False)
+        .order_by("-updated_at")
+        .first()
+    )
```
MarkThomasLite, line 36: need to review best way to get the advice as not keen on this, was originally going to add it as a method param as it would be easier to pass through

currycoder: I agree, it would be good to pass this in as the context from the place we are initiating the notification.  I don't see the bit where we call the notification function from the app in the PR yet though..

saruniitr, line 36: yeah this is most likely sent as an argument from where this notification is triggered.
In this ticket it is just enough to create a notification, once the UI changes where this is used is merged (currently in progress) then we can update it.

Also we will not be looking at 'refuse' type advice and the countersigning fields used above are not correct. For LU countersigning we added new model to save this data (in review) so we need to pull from those fields.

currycoder (approved):

api/applications/notify.py
```diff
@@ -23,3 +25,20 @@ def notify_exporter_case_opened_for_editing(application):
             "exporter_frontend_url": get_exporter_frontend_url("/"),
         },
     )
+
+
+def notify_caseworker_countersign_return(user_email, application, countersign_advice):
+    relative_url = reverse("cases:countersign_advice", kwargs={"pk": application.id})
```
saruniitr, line 31: this should be `"cases:countersign_decision_review"`

saruniitr: rest of the changes looks good to me

saruniitr (approved):

PR #1327, opened 2023-02-27
LTD-3419-Pending-field-in-model-part-2

seijihg:
This is part 2 of the below:

Part 1
- add pending @property to model/serializer/endpoints based off of user.firstname
- swap out logic on frontend to use new pending field

Part 2
- swap out pending @property to be an actual DB field that’s updated when user first logs in

Related ticket to the FE:
https://github.com/uktrade/lite-frontend/pull/1101

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-3419


api/gov_users/tests/test_authenticate.py
```diff
@@ -22,6 +23,9 @@ def test_authentication_success(self):
         response = self.client.post(self.url, data)
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        GovUser.objects.all()[1].first_name == self.gov_user.first_name
+        GovUser.objects.all()[1].last_name == self.gov_user.last_name
+        GovUser.objects.all()[1].pending == False
```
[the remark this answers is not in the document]
depsiatwal: as suggest please use is for boolean check

depsiatwal (commented):
small comment else LGTM

api/users/migrations/tests/test_0004_baseuser_pending.py
```diff
@@ -0,0 +1,22 @@
...
+        base_user = self.old_state.apps.get_model("users", "BaseUser")
+        base_user.objects.create(first_name="joe", email="test@1231.test.com")  # /PS-IGNORE
+        base_user.objects.create(first_name="", email="test@1232.test.com")  # /PS-IGNORE
+        base_user.objects.create(first_name="", last_name="", email="test@124.test.com")  # /PS-IGNORE
+        base_user.objects.create(first_name="", last_name="blogs", email="test@125.test.com")  # /PS-IGNORE
+        base_user.objects.create(last_name="test", email="test@126.test.com")  # /PS-IGNORE
+
+    def test_migration_0004_baseuser_pending(self):
+        base_user = self.new_state.apps.get_model("users", "BaseUser")
+        assert base_user.objects.filter(pending=True).count() == 4
```
kevincarrogan, line 22: Where do we test that the migration has correctly marked some of the pending values as False?

api/users/migrations/0004_baseuser_pending.py
```diff
@@ -0,0 +1,22 @@
...
+    BaseUser.objects.all().exclude(first_name="").update(pending=False)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("users", "0003_default_govuser_queue"),
+    ]
+
+    operations = [
```
kevincarrogan, line 15: This migration appears to be both a schema and a data migration. We said that we would always split these into two.

api/users/migrations/tests/test_0004_baseuser_pending.py
kevincarrogan: This is also just checking the count. We should check that it's actually set the right users to True.

saruniitr:
Question - I see `Pipfile.lock` is added but there is no associated change in `Pipfile`, so it this really required?

kevincarrogan (approved):
Remove the changes to the `Pipfile.lock` and then it's all good.

saruniitr (approved):

PR #1334, opened 2023-03-06
LTD-3449: Check LU countersigning requirements and validate countersignatures when finalising

saruniitr:
## Change description
A Case may need countersigning from LU depending on case attributes.
LU countersigning flags get applied to the case to indicate this. These
flags also block finalising to ensure necessary countersignatures are present
before the Case can be finalised.

In the current process countersiging is done by adding a case note and these
flags are manually removed to allow finalising the case.

In the new process countersigning is done using routing changes and we can
avoid the manual step of removing the flags by checking if necessary
countersignatures are present at the time of finalising it.

Some Cases need double countersigning so we check for both orders of
countersigning advice for this case.

[LTD-3449](https://uktrade.atlassian.net/browse/LTD-3449)

[LTD-3449]: https://uktrade.atlassian.net/browse/LTD-3449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/applications/views/advice.py
```diff
@@ -0,0 +1,60 @@
...
+    return Flag.objects.filter(
+        id__in=[
+            FlagsEnum.LU_SENIOR_MANAGER_CHECK_REQUIRED,
+            FlagsEnum.MANPADS,
+        ],
+        status=FlagStatuses.ACTIVE,
+    ).values_list("name", flat=True)
+
+
+def validate_lu_countersignatures(application, countersign_flags):
```
currycoder, line 38: Could this be renamed to something a little more descriptive like `is_lu_countersign_complete()`?

api/applications/views/advice.py
```diff
@@ -0,0 +1,60 @@
+from django.conf import settings
```
currycoder, line 1: This is in `views`, but it doesn't really contain any django views.  Could there be a better location for it?  My instant thought would be that this is a file containing views for advice.

api/applications/views/advice.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # We are here means this application needs countersigning so check
+    # if countersignatures of all orders are present and accepted
+    case = application.get_case()
+    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
+    sr_lu_mgr_countersign_flags = get_lu_sr_mgr_countersigning_flag_names()
+
+    countersign_orders = [1]
+    if countersign_flags.intersection(sr_lu_mgr_countersign_flags):
+        countersign_orders.append(2)
```
currycoder, line 50: I don't understand what the `CountersignAdvice.order` field is - could you explain?

api/applications/views/advice.py
```diff
@@ -0,0 +1,60 @@
...
+
+    countersign_advice = CountersignAdvice.objects.filter(
+        order__in=countersign_orders,
+        case=case,
+        advice__team=lu_team,
+        advice__level=AdviceLevel.FINAL,
+        advice__type__in=[AdviceType.APPROVE, AdviceType.PROVISO],
+    )
```
currycoder, line 58: Could it not be possible that the countersign advice is `AdviceType.REFUSE`.  In that case, wouldn't we waive this case through due to this function right now?

api/applications/views/advice.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def validate_lu_countersignatures(application, countersign_flags):
+    if not settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED:
+        return True
+
+    # We are here means this application needs countersigning so check
+    # if countersignatures of all orders are present and accepted
+    case = application.get_case()
+    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
```
currycoder, line 45: Can this ID be made to be an enum somewhere appropriate?

api/applications/views/applications.py
```diff
@@ -624,6 +630,19 @@ def put(self, request, pk):
             .order_by("name")
             .values_list("name", flat=True)
         )
+
+        # Check countersigning requirements and required countersignatures are present
+        if settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED:
+            lu_countersign_flags_all = get_lu_countersigning_flag_names()
+            countersign_flags = blocking_flags.intersection(lu_countersign_flags_all)
+            if countersign_flags:
+                countersign_valid = validate_lu_countersignatures(application, countersign_flags)
```
currycoder, line 639: Maybe rather than checking `if countersign_flags:`, this particular function shouldn't care at all and should defer that to `is_lu_countersign_complete()`.  That way we don't need to sprinkle the logic in here about countersign/lu countersign flags etc.

Instead this particular block could become something like;
```
if settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED:
    ensure_lu_countersign_complete(application) # raises MissingCounterSignature if countersign is not complete
```

api/applications/views/applications.py
```diff
@@ -624,6 +630,19 @@ def put(self, request, pk):
...
+                if not countersign_valid:
+                    raise MissingCounterSignature(
+                        "This applications requires countersigning and the required countersignatures are not available on the application"
+                    )
+                blocking_flags = blocking_flags.difference(countersign_flags)
```
currycoder, line 644: Not entirely sure what I feel about this fiddlyness of removing specific countersign flags from `blocking_flags`...  The thing is that we have a list of blocking flags - but an override for certain types of flags to be removed from that list.  Maybe that would be better tracked by keeping the contract a little cleaner and dropping them from the case when we know they have been satisfied..?

Could I suggest instead that `ensure_lu_countersign_complete()` removes the from the case itself?  And then the blocking flags lookup could be moved below this call...

api/applications/views/advice.py
currycoder: What if one LU countersign advice was to approve and the other to refuse?

api/applications/views/advice.py
saruniitr: yeah it can be renamed.

api/applications/views/advice.py
saruniitr: I meant to put helper functions in this file.
Does naming it as `advice_helpers.py` clarify? or create a directory `helpers` and move this file into that directory?

api/applications/views/advice.py
saruniitr: Yeah I should've explained this.
This model is created to handle multiple countersignatures. Currently for LU double countersign required in few cases but it can be more (similarly for other teams).
`order` basically represents the countersigning order,
order of 1 = first countersign
order of 2 = second countersign etc

api/applications/views/advice.py
saruniitr: If LU decides to refuse then they are handled using a separate business process (offline meetings etc).
All the cases that reach countersigning are the ones getting approved.

api/applications/views/advice.py
saruniitr: can be done but the ideal place is `lite-routing`. So first change needs to be merged there and then update sha etc

api/applications/views/applications.py
saruniitr: Yeah I considered that but if all countersigning is complete then we need to remove them from `blocking_flags` also hence made it to return status.

Reading together with the below comment, usually we don't remove flags once applied so was trying to keep it consistent with that pattern. Was not completely happy with the way it is split up but if flags can be removed within that function it can be simplified.

api/applications/views/advice.py
currycoder: Either works for me, thanks

api/applications/views/applications.py
currycoder: So what do we do with other `blocks_finalising` flags?  Do they get removed somehow by users or by some other mechanism?

api/applications/views/advice.py
currycoder: OK understood...  Right now, does this function prevent finalising for an application where only one out of the two countersigns have happened?  I think that below it only checks that there is *some* result for `countersign_advice` not that there are the correct number of results.

Any ideas?

api/applications/views/advice.py
currycoder: OK no problems; happy for that to follow

api/applications/views/advice.py
saruniitr: If double countersign required then we look for `order=2` objects as well otherwise only for first countersign.
So yes if a Case requires double countersign and only one of them is available then it blocks finalising (have a test for this as well)

api/applications/views/advice.py
saruniitr: So an additional check to see if there is any `CountersignAdvice` objects of type `REFUSE` and raise error?
Makes sense.
It will have to be a separate exception, I am thinking `CountersignInvalidAdviceTypeException`?

api/applications/views/applications.py
saruniitr: Yeah usually countersign flags are the ones that remain by this stage which currently they manually remove and finalise.
There is a chance that `Enforcement Check required` also present sometimes which again manually removed.

api/applications/views/helpers/advice.py
```diff
@@ -0,0 +1,129 @@
...
+        advice__type=AdviceType.REFUSE,
+    )
+    if refused_countersign_advice.exists():
+        raise CountersignInvalidAdviceTypeError(
+            "This application cannot be finalised as the countersigning has been refused"
+        )
+
+    # check countersignatures for the required orders
+    countersign_advice = CountersignAdvice.objects.filter(
+        order__in=countersign_orders,
```
currycoder, line 94: I'm still not convinced that this particular bit will work in the way that we want...

Say we have a case where countersign is needed from both groups, and say that only one group has created a  CountersignAdvice entry so far..

I believe that the `countersign_advice` will be a truth-y value in this case (one result returned) so we will continue on with completing the function rather than raising `CounterSignatureIncompleteError`

What do you think?

saruniitr: This is updated and included a test to cover this case.

api/applications/views/helpers/advice.py
```diff
@@ -0,0 +1,130 @@
...
+    countersign_process_flags = Flag.objects.filter(
+        id__in=[
+            FlagsEnum.LU_COUNTER_REQUIRED,
+            FlagsEnum.LU_SENIOR_MANAGER_CHECK_REQUIRED,
+        ],
+        status=FlagStatuses.ACTIVE,
+    )
+    flags_to_remove = countersign_flags.intersection(countersign_process_flags)
+    for party_on_application in application.parties.all():
+        party_on_application.party.flags.remove(*flags_to_remove)
```
currycoder, line 120: Interesting that these are party flags.  Just checking - is that accurate?  Could these flags be attached to case/good?

api/applications/views/helpers/advice.py
```diff
@@ -0,0 +1,130 @@
...
+        audit_trail_service.create_system_user_audit(
```
currycoder, line 121: The only problem with this is that there is *always* an audit trail entity for every single party on an application.  Is that accurate?  I would have thought that the process flags wouldn't be present on *every* party.

Could it be better to check that the flag was present on the party and then do the remove/audit trail entry?

api/applications/tests/test_finalise_application.py
```diff
@@ -307,6 +319,297 @@ def test_reissue_after_product_assessment_changes(self):
...
+            FlagsEnum.LU_COUNTER_REQUIRED,
+            FlagsEnum.LU_SENIOR_MANAGER_CHECK_REQUIRED,
+            FlagsEnum.MANPADS,
+            FlagsEnum.AP_LANDMINE,
+        ]
+    )
+    @override_settings(FEATURE_COUNTERSIGN_ROUTING_ENABLED=True)
+    def test_finalise_application_failure_with_countersigning_flags_but_no_countersignatures(self, flag_id):
+        flag = Flag.objects.get(id=flag_id)
+        self.standard_application.flags.add(flag)
```
currycoder, line 333: Shouldn't the first two flags get added to the party, not the case itself?

currycoder: Also for my understanding, is `self.standard_application` an `Application` object or a `Case` object?

api/applications/tests/test_finalise_application.py
```diff
@@ -307,6 +319,297 @@ def test_reissue_after_product_assessment_changes(self):
...
+        # Create Advice objects for all entities
+        case = Case.objects.get(id=self.standard_application.id)
+        advice_qs = Advice.objects.filter(case=case, level=AdviceLevel.FINAL, type=AdviceType.APPROVE)
+        for countersign in countersign_data:
+            if countersign.get("skip"):
+                continue
+            for index, advice in enumerate(list(advice_qs)):
+                outcome_accepted = True
+                if countersign.get("reject"):
+                    outcome_accepted = False if index % 2 else True  # reject alternate advice
```
currycoder, line 434: This is hard for me to understand - False if index is even, True otherwise..?  What's the intention?

currycoder (commented):
Looking good - just some hopefully minor amends now.

I have one extra question though; how will the routing BDD tests change to accommodate this?  Will there be extra steps introduced to ensure that LU countersign is also covered?

api/applications/views/helpers/advice.py
```diff
@@ -0,0 +1,130 @@
...
+
+def lu_countersigning_flags_all():
+    return Flag.objects.filter(
+        id__in=[
+            FlagsEnum.LU_COUNTER_REQUIRED,
+            FlagsEnum.LU_SENIOR_MANAGER_CHECK_REQUIRED,
+            FlagsEnum.MANPADS,
+            FlagsEnum.AP_LANDMINE,
+        ],
+        status=FlagStatuses.ACTIVE,
```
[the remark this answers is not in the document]
saruniitr: These flags are currently inactive and yes the routing activation management command is already updated to include these flags

api/applications/views/helpers/advice.py
saruniitr: No, these two are destination based so only attached to `Party`

api/applications/views/helpers/advice.py
saruniitr: hmm that is true. It is not a problem for the `remove()` but will add audit entry even if the flag is not there.
I will update.

api/applications/tests/test_finalise_application.py
saruniitr: Actually it doesn't matter here because there is no countersign advice. Setting on the correct level is mainly for removing them after ensuring countersignatures are present. While determining which flags are set I am using `case.parameter_set()` which includes flags from all levels hence didn't specify the level here as in other tests but I can update if it is confusing.

api/applications/tests/test_finalise_application.py
saruniitr: To simulate incomplete countersignatures we want atleast one rejection. Instead of fixing one I am just rejecting every alternate advice hence the mod2 division, nothing serious here :)

currycoder: Ahhhh thank you :D

saruniitr:
> Looking good - just some hopefully minor amends now.
> 
> I have one extra question though; how will the routing BDD tests change to accommodate this? Will there be extra steps introduced to ensure that LU countersign is also covered?

Correct, there will be an additional step that submits final decision and checks for countersignatures.
I will get that PR ready once this is merged.

currycoder (approved):

api/applications/views/helpers/advice.py
```diff
@@ -0,0 +1,133 @@
...
+        id__in=[
+            FlagsEnum.LU_COUNTER_REQUIRED,
+            FlagsEnum.LU_SENIOR_MANAGER_CHECK_REQUIRED,
+        ],
+        status=FlagStatuses.ACTIVE,
+    )
+    flags_to_remove = countersign_flags.intersection(countersign_process_flags)
+    for party_on_application in application.parties.all():
+        if not flags_to_remove.intersection(party_on_application.party.flags.all()):
+            continue
```
currycoder, line 121: Could be good to tackle this coverage complaint if it's easy (I think it should be a case of adjusting the tests slightly, right?) I'll leave this with you

saruniitr: yeah updated tests to cover this

hnryjmes (approved):
looks good, some minor typos or rewordings I would make in comments but not enough to block approval

api/applications/tests/test_finalise_application.py
```diff
@@ -307,6 +319,300 @@ def test_reissue_after_product_assessment_changes(self):
...
+                    {"id": FlagsEnum.LU_COUNTER_REQUIRED, "level": FlagLevels.DESTINATION},
+                    {"id": FlagsEnum.LU_SENIOR_MANAGER_CHECK_REQUIRED, "level": FlagLevels.DESTINATION},
+                    {"id": FlagsEnum.AP_LANDMINE, "level": FlagLevels.CASE},
+                    {"id": FlagsEnum.MANPADS, "level": FlagLevels.CASE},
+                ),
+            ],
+        ]
+    )
+    @override_settings(FEATURE_COUNTERSIGN_ROUTING_ENABLED=True)
+    def test_finalise_application_sucess_with_countersigning(self, advice_type, required_countersign, flags):
```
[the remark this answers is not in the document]
saruniitr: agree, will fix before merging

PR #1337, opened 2023-03-07
Run E2E tests on API changes

kevincarrogan:
### Aim

Run the e2e tests on API changes.

This is to handle the case where an API change may break the e2e tests.

We call the `lite-frontend` pipeline specifying the branch of the API changes and attempt to run them against the frontend `dev` branch.

To begin with I am going to keep this job as being not required on our CI checks so that we can see if this causes any issues.

[LTD-3455](https://uktrade.atlassian.net/browse/LTD-3455)

[LTD-3455]: https://uktrade.atlassian.net/browse/LTD-3455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -276,6 +276,56 @@ jobs:
                 fi
             fi
 
+  e2e_tests:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - run:
+          name: Kick off lite-frontend e2e pipeline
+          command: |
+            PIPELINE_ID=$(curl --location --request POST 'https://circleci.com/api/v2/project/github/uktrade/lite-frontend/pipeline' \
+            --header 'Content-Type: application/json' \
+            --header "Circle-Token: ${LITE_FRONTEND_E2E_CI_TOKEN}" \
+            --data-raw '{
+                "branch": "dev",
+                "parameters": {
+                  "run_ui_tests": true,
+                  "run_unit_tests": false,
+                  "api_git_tag": "'"$CIRCLE_BRANCH"'",
+                  "environment": "dev"
+                }
+              }' | jq -r '.id')
+
+            echo "export PIPELINE=$PIPELINE_ID" >> $BASH_ENV
+      - run:
+          name: Check lite-frontend e2e pipeline status
+          no_output_timeout: 30m
+          command: |
+            sleep 10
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -258,8 +258,57 @@ jobs:
...
+
+            while [ "$STATUS" == "running" ]; do
+              echo still running
+              sleep 10
+              STATUS=$(curl -s "https://circleci.com/api/v2/pipeline/$PIPELINE/workflow?circle-token=$LITE_FRONTEND_E2E_CI_TOKEN" | jq -r '.items[].status')
+            done
+
+            echo "finally done, checking final status: $STATUS"
+            if [[ "$STATUS" == "success" ]]; then
+              echo "Triggered lite-frontend e2e pipeline finished sucessfully"
```
hnryjmes, line 303: typo in successfully, should be: `"Triggered lite-frontend e2e pipeline finished successfully"`

.circleci/config.yml
```diff
@@ -258,8 +258,57 @@ jobs:
...
+          name: Kick off lite-frontend e2e pipeline
+          command: |
+            PIPELINE_ID=$(curl --location --request POST 'https://circleci.com/api/v2/project/github/uktrade/lite-frontend/pipeline' \
+            --header 'Content-Type: application/json' \
+            --header "Circle-Token: ${LITE_FRONTEND_E2E_CI_TOKEN}" \
+            --data-raw '{
+                "branch": "dev",
+                "parameters": {
+                  "run_ui_tests": true,
+                  "run_unit_tests": false,
```
hnryjmes, line 275: instead of writing ui tests, for some ui functionality we have been writing soup tests and these would come under unit tests. are we sure we don't want to run these? or is that not applicable here?

kevincarrogan: The unit tests, even the beautful soup ones, don't interact with the API at all, everything these is mocked.

The only tests that actually check our interaction between frontend and API are the UI tests.

Given that this is to ensure that any changes in the API don't break something the only tests that will be able to give us that answer are the UI tests.

hnryjmes (approved):

currycoder (approved):
Understand the idea of not making this required for a bit but my hunch is that this should be required; as we should generally only depend on changes in one direction e.g. new frontend changes depend on new API changes being merged.  If it fails it's probably an indicator that we are doing something the wrong way round.

PR #1343, opened 2023-03-10
Ltd 3453 output logging

kevincarrogan:
### Aim

Update the default configuration when running tests to make output less verbose. This is especially important when a test fails so that the information about the error is the easiest thing to find.

This turns off most of the default logging and capturing that was occurring when we were running the tests.

The following has been changed:

  - disable capturing and outputting warnings during tests
  - disable capturing and outputting stdin/stdout during tests
  - disable all loggers during tests
  - suppress output of seeding commands

[LTD-3453](https://uktrade.atlassian.net/browse/LTD-3453)

[LTD-3453]: https://uktrade.atlassian.net/browse/LTD-3453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/conf/settings_test.py
+++ b/api/conf/settings_test.py
@@ -4,3 +4,5 @@ from api.conf.settings import *
 ELASTICSEARCH_SANCTION_INDEX_ALIAS = "sanctions-alias-test"
 
 LOGGING = {"version": 1, "disable_existing_loggers": True}
+
+SUPPRESS_TEST_OUTPUT = True
--- a/api/conf/settings_test.py
+++ b/api/conf/settings_test.py
@@ -2,3 +2,5 @@ from api.conf.settings import *
 
 
 ELASTICSEARCH_SANCTION_INDEX_ALIAS = "sanctions-alias-test"
+
+LOGGING = {"version": 1, "disable_existing_loggers": True}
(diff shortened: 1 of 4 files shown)


Harriethw (approved):
nice! I can actually see why tests failed now :)

api/conf/settings_test.py
```diff
@@ -2,3 +2,7 @@
 
 
 ELASTICSEARCH_SANCTION_INDEX_ALIAS = "sanctions-alias-test"
+
+LOGGING = {"version": 1, "disable_existing_loggers": True}
+
+SUPPRESS_TEST_OUTPUT = True
```
currycoder, line 8: What does this do?

currycoder (approved):

api/conf/settings_test.py
kevincarrogan: Which bit? Or are you asking about both?

`LOGGING` - Removes all loggers from doing their thing.
`SUPPRESS_TEST_OUTPUT` - Stops all of the seeding commands from outputting stuff.

PR #1353, opened 2023-03-20
LTD-3508: Invalidate countersignatures on edit recommendation

saruniitr:
### Aim

In the LU countersigning journey, case-officers initially give their recommendation and submit for countersigning.
Countersigners can either accept or reject and send the case back for editing. When the case comes back again for countersigning then it is considered as if it has started the journey again because some other officer could pick this Case now rather than the previous officer. To differentiate new countersignatures from the old ones, they are marked as invalid whenever Caseofficer edits their recommendation. For this purpose a new field `valid` is introduced in `CountersignAdvice` model.
We are not deleting old advice because we want to retain history.
This is only done if there is a rejected countersignature, if all are approved then Caseofficer would be finalising the case.

We have two situations where we need to invalidate existing countersignatures.
- When a Case requires only licensing manager countersigning and he rejects it - in this case when caseofficer edits their recommendation then we invalidate first order countersign.
- When a Case requires Senior licensing manager countersigning and he rejects it - in this case Licensing manager would've approved it but since Senior manager rejected it we need invalidate both orders.

Updated API `put` method for final advice to handle above scenarios.

Also remove code that deletes existing objects when final advice is updated
  - we have a new put method to update final level advice and there is no
    need to delete existing objects which was required for the previous
    method of update (where existing objects are deleted and new ones
    are created in their place).

Add tests for all of the above scenarios.

[LTD-3508](https://uktrade.atlassian.net/browse/LTD-3508)

[LTD-3508]: https://uktrade.atlassian.net/browse/LTD-3508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/migrations/0055_countersignadvice_valid.py
```diff
@@ -0,0 +1,23 @@
...
+        ("cases", "0054_countersignadvice"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="countersignadvice",
+            name="valid",
+            field=models.BooleanField(
+                blank=True,
+                default=True,
```
currycoder, line 18: My understanding is that all existing CountersignAdvice records will be set as `valid=True` from this.  Just checking but is that what we want?  Could it be possible that we will set some CountersignAdvice records as `valid=True` even though they have been superseded by newer ones?

api/cases/libraries/post_advice.py
```diff
@@ -156,6 +159,10 @@ def update_advice(request, case, level):
 
     serializer.save()
 
+    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
+    if settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED and request.user.govuser.team == lu_team:
+        mark_rejected_countersignatures_as_invalid(case)
```
currycoder, line 164: So other teams' CountersignAdvice cannot be `valid=False` right now?

api/applications/views/helpers/advice.py
```diff
@@ -131,3 +146,59 @@ def ensure_lu_countersign_complete(application):
                 "additional_text": "Removing flags as required countersignatures present and approved",
             },
         )
+
+
+def mark_rejected_countersignatures_as_invalid(case):
```
currycoder, line 151: This is specifically for LU countersign, right?  Could I suggest reflecting that in the function name?

api/applications/views/helpers/advice.py
```diff
@@ -49,24 +49,13 @@ def lu_sr_mgr_countersigning_flags():
...
-
-    Certain cases need double countersigning so it checks for countersignatures
-    of both orders (FIRST_COUNTERSIGN, SECOND_COUNTERSIGN) where applicable.
-
-    Once the countersignatures are valid then it removes the flags that blocks
-    finalising the Case.
-    """
-    case = application.get_case()
-    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
+def get_required_countersign_orders(case):
```
currycoder, line 52: Again, isn't this about lu countersign specifically?  Should the function name reflect that?

api/applications/views/helpers/advice.py
```diff
@@ -131,3 +146,59 @@ def ensure_lu_countersign_complete(application):
...
+        return
+
+    # Mark countersignatures as invalid in the required orders
+    # Usually either first or second countersign is rejected but not both because
+    # routing engine would've moved the Case back when first countersign is rejected
+    # But in the case where second countersign is rejected, then we also need to
+    # invalidate first countersign hence take the max_order
+    max_order = max(countersign_orders_to_invalidate)
+    for order in range(1, max_order + 1):  # +1 because the end is not included
+        countersign_advice = CountersignAdvice.objects.filter(
```
currycoder, line 196: Rather than doing two repeated queries, how about keeping track of the querysets from the first loop?  Then you can make a separate call down here about which ones actually need a `.update(valid=False)`

currycoder: Hopefully this will make this function a bit easier to follow

currycoder (commented):
Initial comments/questions without having looked at the tests yet

api/applications/views/helpers/advice.py
saruniitr: :+1:

api/cases/libraries/post_advice.py
saruniitr: Yeah other teams cannot invalidate hence limiting to LU only.
This gets called when LU caseworker edits their recommendation after a Case is sent back when countersign is rejected.

api/cases/migrations/0055_countersignadvice_valid.py
saruniitr: yeah the idea is they are all valid when they are created (when licensing manager or senior licensing manager first gives their comments whether they are approved or rejected).
If it is rejection then they go back and when caseworker edits, then we mark them as invalid and that will be end for these objects.
When the case is sent again for countersigning (after editing recommendation etc) then new `CountersignAdvice` objects gets created, does this clarify your understanding?

api/applications/views/helpers/advice.py
saruniitr: The query is slightly different here, I will see if this can be combined

saruniitr: Ok I think we have to repeat these queries.
In the first instance the query is different, looking for objects that have `outcome_accepted=False` because in the case where Senior manager countersigning is also required then the licensing manager would've approved but the Senior manager could reject - in this case we want to invalidate licensing manager countersignatures as well hence in the first I am collecting the order where it is rejected and in the second pass actually invalidating them.

api/applications/views/helpers/advice.py
```diff
@@ -131,3 +146,59 @@ def ensure_lu_countersign_complete(application):
...
+    come for the first time and possibly by a different officer so we can mark them as invalid.
+    They are still available on the case for the officer to check previous history but to prevent
+    using the same (eg by editing) these are marked as invalid.
+
+    Countersigning officers will provide new countersignatures to move it forward.
+
+    This is only done if there is a rejected countersignature. If all are approved it will go
+    back to caseworker for finalising.
+    """
+    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
```
hnryjmes, line 164: there is now a `LICENSING_UNIT` enum you can use for this in teams enums.py

api/applications/views/helpers/advice.py
```diff
@@ -75,16 +64,41 @@ def ensure_lu_countersign_complete(application):
...
+    are present and approved.
+
+    Certain cases need double countersigning so it checks for countersignatures
+    of both orders (FIRST_COUNTERSIGN, SECOND_COUNTERSIGN) where applicable.
+
+    Once the countersignatures are valid then it removes the flags that blocks
+    finalising the Case.
+    """
+    case = application.get_case()
+    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
```
hnryjmes, line 95: there is now a LICENSING_UNIT enum you can use for this in teams enums.py

api/cases/libraries/post_advice.py
```diff
@@ -156,6 +159,10 @@ def update_advice(request, case, level):
 
     serializer.save()
 
+    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
```
hnryjmes, line 162: there is now a LICENSING_UNIT enum you can use for this in teams enums.py

api/cases/tests/test_edit_advice.py
```diff
@@ -216,3 +228,251 @@ def test_edit_final_level_advice_types_success(self, advice_type, advice_data):
...
+            if flag["level"] == FlagLevels.CASE:
+                self.application.flags.add(Flag.objects.get(id=flag["id"]))
+            if flag["level"] == FlagLevels.DESTINATION:
+                # We emit audit entry of removing flags only if countersigning flags are set
+                # on the Party and skip otherwise. To cover the case where we skip it, don't
+                # set flags on one party (in this case last item is selected)
+                for party_on_application in list(self.application.parties.all())[:-1]:
+                    party_on_application.party.flags.add(Flag.objects.get(id=flag["id"]))
+
+        self.gov_user.team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
```
hnryjmes, line 344: there is now a LICENSING_UNIT enum you can use for this in teams enums.py

api/cases/tests/test_edit_advice.py
```diff
@@ -216,3 +228,251 @@ def test_edit_final_level_advice_types_success(self, advice_type, advice_data):
...
```
hnryjmes, line 436: there is now a LICENSING_UNIT enum you can use for this in teams enums.py

api/applications/views/helpers/advice.py
saruniitr: correct, was waiting for the SHA update.
Once all comments are addressed then I wil rebase and replace this with that enum.

api/cases/migrations/0055_countersignadvice_valid.py
currycoder: That does, thanks.

But thinking specifically about how this migration will set all existing CountersignAdvice records as `valid=True` - is that accurate?  Or do we need a data migration to set some as invalid..?

api/cases/libraries/post_advice.py
currycoder: OK understood. So the idea is that when LU update their recommendation (at all) the countersignature advice for LU is marked as invalid.

Is there any possibility that LU will update their advice in such a way where we don't want to mark countersignatures as invalid?  e.g. LU spot a typo and update that..?

api/applications/views/helpers/advice.py
currycoder: Understood.  Thanks for looking in to it, I think I follow why it needs to be like this.

api/cases/migrations/0055_countersignadvice_valid.py
currycoder: Awesome, thanks for clarifying.

api/cases/libraries/post_advice.py
currycoder: Yep that makes sense.  Just wanted to talk this possibility through.  I guess the point is that this will likely only be changed at the point that it lands back in LU post-circ queue; changes while it is in countersign are unlikely as they've already pushed the case forwards.

saruniitr: Correct, they won't be able to edit as the Case won't be in their queue

api/applications/serializers/advice.py
```diff
@@ -233,6 +233,7 @@ class Meta:
 
 class CountersignDecisionAdviceViewSerializer(serializers.Serializer):
     id = serializers.UUIDField()
+    valid = serializers.BooleanField()
```
currycoder, line 236: Just checking, this is only used for read-only API endpoints right?  We aren't requiring `valid` to be set for write API endpoints (e.g. a backward compatibility break)

api/cases/tests/test_edit_advice.py
```diff
@@ -146,6 +155,8 @@ def test_edit_advice_invalid_data_returns_error(self, level):
             [
                 {"type": ['"invalid" is not a valid choice.']},
                 {"type": ['"invalid" is not a valid choice.']},
+                {"type": ['"invalid" is not a valid choice.']},
+                {"type": ['"invalid" is not a valid choice.']},
```
currycoder, line 159: I don't fully understand why these were added?  In general it's a bit of an odd error response - lots of repeated lines!

api/cases/tests/test_edit_advice.py
```diff
@@ -216,3 +228,251 @@ def test_edit_final_level_advice_types_success(self, advice_type, advice_data):
...
+        - Makes API call to edit caseworker's recommendation
+        - Verifies that countersignatures are invalidated and their count matches.
+        - `expected_countersign_status` is the expected CountersignAdvice status after editing recommendation
+
+        Scenarios:
+        1. Only licensing manager countersigning required and they reject the case
+        2. Senior licensing manager countersigning required and they reject the case. In this case
+           even if licensing manager accepts, that needs invalidating as the Case starts the journey
+           from the beginning when it comes back again
+        3. Both manager accepts - nothings needs invalidating in this case
```
currycoder, line 331: Would recommend adding these as comments next to the parameters as I bet this docstring will go out of sync with the scenarios in later changes.

api/applications/views/helpers/advice.py
```diff
@@ -131,3 +146,59 @@ def ensure_lu_countersign_complete(application):
...
+    Countersigning officers will provide new countersignatures to move it forward.
+
+    This is only done if there is a rejected countersignature. If all are approved it will go
+    back to caseworker for finalising.
+    """
+    lu_team = Team.objects.get(id="58e77e47-42c8-499f-a58d-94f94541f8c6")
+    countersign_orders = get_lu_required_countersign_orders(case)
+
+    if check_lu_refused_outcome_exists(case, lu_team, countersign_orders):
+        raise CountersignInvalidAdviceTypeError("Cannot invalidate countersignatures as the outcome is refused")
```
currycoder, line 168: How do we expect LU post circ to update their advice when the countersigners say this should be refused?  Is there some other alternate process?

api/applications/serializers/advice.py
saruniitr: Correct, this serializer is used in the endpoint where Case details are retrieved.
There will be few changes in the UI that use this field to ignore invalidated countersignatures.

FE won't be able to set this field directly which is set by API based on user action (when they edit their recommendation)

api/applications/views/helpers/advice.py
saruniitr: This is more of additional protection.
Refusal process is handled manually and I think it needs a bit of work also.

If the original outcome is refuse then it won't go to countersigning stage so countersigning comments would be mostly to edit their current outcome but not the type.

currycoder: Ok got it - yeah I do remember you've explained this to me before about refusals being manual.  Could I suggest just popping a small comment in (what you said ^) to help the future dev seeing this error in sentry and not having the context?

currycoder (approved):
Will leave the remaining comments with you now

api/applications/views/helpers/advice.py
saruniitr: :+1:

antroy-madetech (approved):

api/cases/libraries/post_advice.py
saruniitr: oh I thought this was in `lite_routing`, updated now in all places with the enum

hnryjmes (approved):

PR #1365, opened 2023-03-28
LTD-3481: Update CaseStatusEnum to evaluate status choices at runtime

saruniitr:
## Change description

The new statuses introduced for LU Countersigning are behind a feature flag so we need to evaluate these at runtime (when running tests) otherwise they don't get picked up and tests fail.

### DIFF
--- a/api/staticdata/statuses/tests/test_statuses.py
+++ b/api/staticdata/statuses/tests/test_statuses.py
@@ -19,14 +19,16 @@ class TestCaseStatus:
         editable_case_status = get_case_status_by_status(status)
         assert editable_case_status.is_read_only is False
 
-    @override_settings(FEATURE_COUNTERSIGN_ROUTING_ENABLED=True)
-    def test_read_only_status_contains_lu_countersign_statuses(self):
-        statuses = get_case_statuses(read_only=True)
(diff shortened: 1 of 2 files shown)


api/staticdata/statuses/tests/test_statuses.py
```diff
@@ -16,3 +18,15 @@ def test_read_only_case_statuses(self, status):
     def test_editable_case_statuses(self, status):
         editable_case_status = get_case_status_by_status(status)
         assert editable_case_status.is_read_only is False
+
+    @override_settings(FEATURE_COUNTERSIGN_ROUTING_ENABLED=True)
```
currycoder, line 22: Could we have some testcases for the opposite?  e.g. when the feature flag is not enabled, assert these extra statuses are not present

currycoder (approved):
One comment I will leave with you.  Otherwise looks like it is good to go once the suite goes green

api/staticdata/statuses/tests/test_statuses.py
saruniitr: :+1:

hnryjmes (approved):

PR #1368, opened 2023-03-28
wip

hnryjmes:
### Aim


api/applications/views/helpers/advice.py
```diff
@@ -144,7 +144,7 @@ def ensure_lu_countersign_complete(application):
             payload={
                 "removed_flags": [flag.name for flag in flags_to_remove],
                 "destination_name": party_on_application.party.name,
-                "additional_text": "Removing flags as required countersignatures present and approved",
+                "is_lu_countersigning": True,
```
saruniitr, line 147: @hnryjmes this is not going to completely fix the issue (may be you are going to update this as it is still draft).
Currently we go through each party and remove these flags as this flag is applied on each party. So there will be multiple entries.
They are asking to emit just one entry instead of for each destination. So first thing is it needs to come out of this loop.

The new entry also is a consolidated one. So there will only be one audit entry if only licensing manager countersign is required.
If Senior manager countersigning is also required there will be a second entry.

hnryjmes: yeah I am just using the draft PR as a mechanism for running the e2e tests, not ready for proper review yet

I think I will move this out of the for loop but not sure yet how to deal with the consolidation

api/applications/views/helpers/advice.py
```diff
@@ -132,19 +132,23 @@ def ensure_lu_countersign_complete(application):
...
     )
     flags_to_remove = countersign_flags.intersection(countersign_process_flags)
+    any_flags_removed = False
     for party_on_application in application.parties.all():
         if not flags_to_remove.intersection(party_on_application.party.flags.all()):
             continue
 
         party_on_application.party.flags.remove(*flags_to_remove)
+        any_flags_removed = True
+    if any_flags_removed:
```
saruniitr, line 142: You can use `flags_to_remove` itself instead of another variable.
If `flags_to_remove` is not empty then we would've removed flags, not necessarily from all parties but we would've definitely removed flags from those parties where these flags are present.

api/applications/views/helpers/advice.py
```diff
@@ -132,19 +132,23 @@ def ensure_lu_countersign_complete(application):
...
+        # Even though flags may be removed from multiple parties,
+        # we only ever want to show max one audit event message to the user,
+        # hence the following is done once outside the for loop.
```
saruniitr, line 145: Below test need to be updated that checks for this audit event,
https://github.com/uktrade/lite-api/blob/dev/api/applications/tests/test_finalise_application.py#L561

api/audit_trail/formatters.py
```diff
@@ -112,12 +114,21 @@ def destination_add_flags(**payload):
...
 def destination_remove_flags(**payload):
     flags = [flag.strip() for flag in payload["removed_flags"].split(",")]
-    destination_name = payload["destination_name"].title()
-    if len(flags) == 1:
-        return f"removed the flag '{flags[0]}' from the destination '{destination_name}'."
-    elif len(flags) >= 2:
-        formatted_flags = f"{str(flags[:-1])[1:-1]} and '{flags[-1]}'"
-        return f"removed the flags {formatted_flags} from the destination '{destination_name}'."
+    # short audit message for LU countersigning case finalising only
+    is_finalise_case = bool(payload.get("is_finalise_case", False))
```
saruniitr, line 118: No need to cast into `bool`

api/applications/views/helpers/advice.py
hnryjmes: updated

api/audit_trail/formatters.py
hnryjmes: updated

api/applications/views/helpers/advice.py
```diff
@@ -132,19 +132,23 @@ def ensure_lu_countersign_complete(application):
...
         audit_trail_service.create_system_user_audit(
             verb=AuditType.DESTINATION_REMOVE_FLAGS,
-            action_object=party_on_application.party,
             target=case,
             payload={
                 "removed_flags": [flag.name for flag in flags_to_remove],
-                "destination_name": party_on_application.party.name,
-                "additional_text": "Removing flags as required countersignatures present and approved",
+                "is_finalise_case": True,
```
saruniitr, line 151: We are doing this during Case finalising but these audit entries are actually related to countersigning so what do you think about renaming to `lu_countersign_check` or `is_lu_countersign` or something similar?

saruniitr (approved):
one comment I leave it with you.

api/audit_trail/formatters.py
```diff
@@ -112,12 +114,21 @@ def destination_add_flags(**payload):
...
-    elif len(flags) >= 2:
-        formatted_flags = f"{str(flags[:-1])[1:-1]} and '{flags[-1]}'"
-        return f"removed the flags {formatted_flags} from the destination '{destination_name}'."
+    # short audit message for LU countersigning case finalising only
+    is_finalise_case = payload.get("is_finalise_case", False)
+    if settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED and is_finalise_case:
+        if len(flags) == 1:
+            return f"removed the flag '{flags[0]}'."
+        elif len(flags) >= 2:
+            formatted_flags = f"{str(flags[:-1])[1:-1]} and '{flags[-1]}'"
```
antroy-madetech, line 123: I know this is existing code, but it is horrible to mentally decode. From the ticket this looks like it will only be the two countersignature flags. And reading it it seems that it will output the formatted flags as something like: Flag_one flag_two and 'flag_3'

Perhaps better extracted into a formatted_flags method that makes it clearer what is going on.

Seems like it should be doing something like:

```
def formatted_flags(flags):
    last_flag = flags[-1]
    rest = flags[:-1]
    quoted_csv = ", ".join(["'f{flag}'" for flag in rest])
    return f"{quoted_csv} and '{last_flag}'"
```

Or something

api/applications/tests/test_finalise_application.py
```diff
@@ -611,7 +612,14 @@ def test_finalise_application_success_with_countersigning(self, advice_type, req
         # Finally check for expected audit events
         audit_qs = Audit.objects.filter(verb=AuditType.DESTINATION_REMOVE_FLAGS, target_object_id=case.id)
         flag_names = sorted(list(Flag.objects.filter(id__in=expected_flags_to_remove).values_list("name", flat=True)))
+        expected_text = {
+            1: "removed the flag 'LU Countersign Required'.",
+            2: "removed the flags 'LU Countersign Required' and 'LU Senior Manager check required'.",
+        }
```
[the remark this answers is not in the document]
hnryjmes: yeah not sure why I didn't think of that, good suggestion!

api/applications/views/helpers/advice.py
hnryjmes: updated

api/audit_trail/formatters.py
hnryjmes: updated

api/audit_trail/formatters.py
```diff
@@ -102,22 +104,41 @@ def good_add_remove_flags(**payload):
...
+        formatted_flags = f"{all_but_last_flag} and '{flags[-1]}'"
+        message = f"{action} the flags {formatted_flags}"
+
+    if destination_name and action == "added":
+        message += f" to the destination '{destination_name}'"
+
+    if destination_name and action == "removed":
+        message += f" from the destination '{destination_name}'"
+
+    return message + "."
```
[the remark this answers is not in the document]
hnryjmes: done

antroy-madetech (approved):
LGTM

PR #1371, opened 2023-03-30
Lte 3485 countersign audit

antroy-madetech:
### Aim

Currently there is no audit for LU advice nor LU advice countersigning. This PR adds in that functionality. See the ticket for the designs for the 8 cases we cover in this PR.

This PR also covers LTD-3544 - Countersign proviso advice (license conditions).

[LTD-3485](https://uktrade.atlassian.net/browse/LTD-3485)
[LTD-3544](https://uktrade.atlassian.net/browse/LTD-3544)

[LTD-3485]: https://uktrade.atlassian.net/browse/LTD-3485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LTD-3544]: https://uktrade.atlassian.net/browse/LTD-3544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/audit_trail/formatters.py
```diff
@@ -261,3 +261,12 @@ def generate_decision_letter(**payload):
...
 
     return f"invalid decision {decision} for this event."
+
+
+def create_lu_advice(firstname, lastname, advice_type):  # /PS-IGNORE
+    return f"{firstname} {lastname} added a recommendation to {advice_type}."  # /PS-IGNORE
+
+
+def update_lu_advice(firstname, lastname, advice_type, **payload):  # /PS-IGNORE
+    advice_type_noun = {AdviceType.APPROVE: "recommendation", AdviceType.REFUSE: "refusal"}[advice_type]
```
saruniitr, line 271: instead of `recommendation` this should be `approval` correct?
Perhaps use it as below and add a default value in case if we get a different type?
```
{advice_type_noun.get(advice_type, "INVALID_ADVICE_TYPE")}
```

antroy-madetech: This is case 9 in the ticket, which has recommendation in the text and approval in the image. I'll confirm with Colin, but I suspect you are correct here.

api/cases/tests/test_create_advice.py
```diff
@@ -364,11 +458,22 @@ def test_countersign_advice_with_decision_success(self, department_name):
         ]
 
         response = self.client.post(self.url, **self.gov_headers, data=data)
-        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
-        self.assertEqual(CountersignAdvice.objects.count(), len(data))
+        assert response.status_code == status.HTTP_201_CREATED
+        assert CountersignAdvice.objects.count() == len(data)
```
saruniitr, line 462: Any particular reason to change this?

api/audit_trail/formatters.py
```diff
@@ -261,3 +261,19 @@ def generate_decision_letter(**payload):
...
+    return f"{firstname} {lastname} added a recommendation to {advice_type}."  # /PS-IGNORE
+
+
+def update_lu_advice(firstname, lastname, advice_type, **payload):  # /PS-IGNORE
+    advice_type_noun = {AdviceType.APPROVE: "approval", AdviceType.REFUSE: "refusal"}[advice_type]
+    return f"{firstname} {lastname} edited their {advice_type_noun} reason."  # /PS-IGNORE
+
+
+def countersign_advice(firstname, lastname, department, order, countersign_accepted, **payload):  # /PS-IGNORE
+    senior_text = "senior " if order == 2 else ""
```
saruniitr, line 276: We created enums for the order values,
https://github.com/uktrade/lite-api/blob/dev/api/cases/enums.py#L359

saruniitr: `payload` is not used

api/cases/libraries/post_advice.py
```diff
@@ -115,6 +115,15 @@ def post_advice(request, case, level, team=False):
             GeneratedCaseDocument.objects.filter(
                 case_id=case.id, advice_type__isnull=False, visible_to_exporter=False
             ).delete()
+            if is_in_lu_team(request.user):
+                audit_payload = {
+                    "firstname": request.user.first_name,  # /PS-IGNORE
+                    "lastname": request.user.last_name,  # /PS-IGNORE
+                    "advice_type": (lu_advice_type(request)),
```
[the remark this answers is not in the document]
antroy-madetech: That is essentially what the lu_advice_type method is doing. But because it gets it from the data it is just filtering to make sure it's the LU advice. Update method is getting the advice type from the Advice objects filtered on the ids of the updated advice which we don't have in the post case.
Unless the data will only ever have LU advice?

api/cases/views/views.py
```diff
@@ -1087,6 +1089,35 @@ def audit_countersign(self, request, case):
...
+    def audit_countersign(self, request, case):
+        govuser = request.user.govuser
+        department = govuser.team.department
+        data = request.data
+        if department is not None:
+            department = department.name
+        else:
+            department = "department"
+
+        case_countersignature = [cs for cs in data if UUID(cs["case"]) == case.id][0]
```
saruniitr, line 1101: `order` is one of the parameter in `data`, it can be taken directly instead of this.

api/cases/tests/test_create_advice.py
saruniitr: ok that's fine.

api/cases/views/views.py
saruniitr: only `payload` is different in both of these audit method, is it worth checking to see if they can be combined?

api/audit_trail/formatters.py
antroy-madetech: It isn't used directly, but the way the additional comment works in the audit is that it uses a "additional_text" entry in the payload, so adding in payload like this (could equally be **kwargs as like you say we don't use it) allows the method to accept params that aren't explicitly in the parameter list. Without kwargs or an explicit additional_text method the templating fails as it tries to expand payload into parameters for the method.

saruniitr: ah that is correct

api/cases/libraries/post_advice.py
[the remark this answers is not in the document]
antroy-madetech: I've tried to simplify it a bit in the latest commit.

api/audit_trail/formatters.py
antroy-madetech: Colin has confirmed and I've changed this

api/cases/tests/test_create_advice.py
```diff
@@ -222,6 +226,92 @@ def test_can_create_advice_with_footnote_required(self):
...
+
+        assert response.status_code == status.HTTP_200_OK
+
+        lu_advice_audit = Audit.objects.filter(verb=AuditType.LU_EDIT_ADVICE)
+        assert lu_advice_audit.exists()
+        audit_obj = lu_advice_audit.first()
+        assert audit_obj.payload["firstname"] == "John"  # /PS-IGNORE
+        assert audit_obj.payload["lastname"] == "Smith"  # /PS-IGNORE
+        assert audit_obj.payload["advice_type"] == advice_type
+        assert audit_obj.payload["additional_text"] == "Updated Text"
```
saruniitr, line 313: we could also use the serializer here to check if message is formatted correctly

depsiatwal (approved):
Lgtm

saruniitr (approved):

api/audit_trail/formatters.py
```diff
@@ -261,3 +261,19 @@ def generate_decision_letter(**payload):
...
+def create_lu_advice(firstname, lastname, advice_type):  # /PS-IGNORE
+    return f"{firstname} {lastname} added a recommendation to {advice_type}."  # /PS-IGNORE
+
+
+def update_lu_advice(firstname, lastname, advice_type, **payload):  # /PS-IGNORE
+    advice_type_noun = {AdviceType.APPROVE: "approval", AdviceType.REFUSE: "refusal"}[advice_type]
+    return f"{firstname} {lastname} edited their {advice_type_noun} reason."  # /PS-IGNORE
+
+
+def countersign_advice(firstname, lastname, department, order, countersign_accepted, **payload):  # /PS-IGNORE
```
currycoder, line 275: Could the function name express that this is for LU only?

api/audit_trail/tests/test_formatters.py
```diff
@@ -447,3 +447,43 @@ def test_create_final_recommendation(self, payload, expected_result):
...
+        ]
+    )
+    def test_create_lu_advice(self, first, last, advice_status, expected_text):
+        result = formatters.create_lu_advice(first, last, advice_status)
+        assert result == expected_text
+
+    @parameterized.expand(
+        [
+            ("Test", "User", AdviceType.APPROVE, "Test User edited their approval reason."),
+            ("Dark", "Knight", AdviceType.REFUSE, "Dark Knight edited their refusal reason."),
```
currycoder, line 464: dark_knight

api/cases/libraries/post_advice.py
```diff
@@ -115,6 +115,15 @@ def post_advice(request, case, level, team=False):
             GeneratedCaseDocument.objects.filter(
                 case_id=case.id, advice_type__isnull=False, visible_to_exporter=False
             ).delete()
+            if is_in_lu_team(request.user):
```
currycoder, line 118: Rather than baking this team check in here, I wonder if it could be better to have some sort of helper `record_final_advice_audit(request.user, data)` and bake that in to there instead..?

api/cases/libraries/post_advice.py
```diff
@@ -160,13 +170,27 @@ def update_advice(request, case, level):
...
+        new_text = data[0]["text"]
+        audit_payload = {
+            "firstname": request.user.first_name,  # /PS-IGNORE
+            "lastname": request.user.last_name,  # /PS-IGNORE
+            "advice_type": advice_type,
+            "additional_text": new_text,
+        }
+        audit_trail_service.create(
+            actor=request.user, verb=AuditType.LU_EDIT_ADVICE, target=case, payload=audit_payload
+        )
```
currycoder, line 185: Again just wondering if a separate helper might make this cleaner?

api/cases/tests/test_create_advice.py
```diff
@@ -331,11 +426,30 @@ def test_countersign_advice_with_decision_serializer_invalid_failure(self):
         response = self.client.put(self.url, **self.gov_headers, data=data)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
-    @parameterized.expand([["DIT"], [None]])
+    @parameterized.expand(
+        [
+            ["DIT", 1, True, "Accepted reason", "John Smith countersigned all DIT recommendations."],
```
currycoder, line 431: Rather than values 1/2 for order here, is there an enum we can use so that it's more understandable for future devs?

api/cases/libraries/post_advice.py
currycoder: In addition I see that we now have two ways of recording audit entries in this function... Do we want to combine them in some way?

api/cases/tests/test_create_advice.py
antroy-madetech: Yes, there is. I'll update

api/cases/libraries/post_advice.py
antroy-madetech: I did have this in two separate helper functions, Arun asked me to combine them :-D

api/cases/libraries/post_advice.py
saruniitr: Team check is an additional guard here because only LU can do the final advice (enforced in FE).
The suggestion of helper function makes sense.

currycoder: OK if only LU can call this, I would suggest chucking this right up top in this function and raising an exception (or responding 400?)  We can save some indentation that way and make this easier to understand.

I would still suggest keeping all the audit related logic together if that is possible?  Right now we capture some audit entries a few lines up, but "FINAL" audit entries here...

api/cases/libraries/post_advice.py
```diff
@@ -160,13 +170,27 @@ def update_advice(request, case, level):
 
     serializer.save()
 
-    lu_team = Team.objects.get(id=TeamIdEnum.LICENSING_UNIT)
-    if settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED and request.user.govuser.team == lu_team:
+    if settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED and is_in_lu_team(request.user):
```
currycoder, line 173: Is this another API that can only be called by LU?  Again I would suggest moving the team check up as per the other comment

api/cases/libraries/post_advice.py
currycoder: OK no worries.  Let's just leave this as is and revisit later.

currycoder (approved):

PR #1377, opened 2023-04-03
LTD-3549: Consider NLR advice type when checking for countersignatures

saruniitr:
### Aim

When LU first consolidates their advice before sending for countersigning they consolidate advice for NLR products also (if that Case has NLR products). However when checking for Countersignatures are complete or not at the time of finalising, we are not including advice type for NLR.

Update tests.

[LTD-3549](https://uktrade.atlassian.net/browse/LTD-3549)

### DIFF
--- a/api/applications/views/helpers/advice.py
+++ b/api/applications/views/helpers/advice.py
@@ -112,7 +112,11 @@ def ensure_lu_countersign_complete(application):
             case=case,
             advice__user__team=lu_team,
             advice__level=AdviceLevel.FINAL,
-            advice__type__in=[AdviceType.APPROVE, AdviceType.PROVISO],
+            advice__type__in=[
+                AdviceType.APPROVE,
+                AdviceType.PROVISO,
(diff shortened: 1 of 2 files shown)


api/applications/views/helpers/advice.py
```diff
@@ -202,6 +210,10 @@ def mark_lu_rejected_countersignatures_as_invalid(case):
             case=case,
             advice__user__team=lu_team,
             advice__level=AdviceLevel.FINAL,
-            advice__type__in=[AdviceType.APPROVE, AdviceType.PROVISO],
+            advice__type__in=[
+                AdviceType.APPROVE,
+                AdviceType.PROVISO,
+                AdviceType.NO_LICENCE_REQUIRED,
+            ],
```
antroy-madetech, line 217: This could probably be a constant at this point, as it's the same in all 3 cases and is getting quite large.

antroy-madetech (approved):

hnryjmes (approved):

currycoder (approved):

PR #1382, opened 2023-04-05
lu refusal notification emails sent from case post save signal

MarkThomasLite:
### Aim

Trgiger email notification from https://uktrade.atlassian.net/browse/LTD-3240 when a case is rejected, the case post save hook was used as similar to flag calculation you check the previous status and the current one to identify when to trigger 

[LTD-3388](https://uktrade.atlassian.net/browse/LTD-3388)

[LTD-3388]: https://uktrade.atlassian.net/browse/LTD-3388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/signals.py
```diff
@@ -20,8 +21,20 @@ def case_post_save_handler(sender, instance, raw=False, **kwargs):
...
+        _check_for_countersign_rejection(instance)
+    instance._previous_status = instance.status
+
+
+def _check_for_countersign_rejection(case):
+    if (
+        settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED
+        and case.status.status == CaseStatusEnum.UNDER_FINAL_REVIEW
+        and case._previous_status.status
+        in [CaseStatusEnum.FINAL_REVIEW_COUNTERSIGN, CaseStatusEnum.FINAL_REVIEW_SECOND_COUNTERSIGN]
```
saruniitr, line 37: This works but there is one thing I would like to clarify which is not obvious.

- If a Case requires both countersignatures and first countersign is rejected then it doesn't go to second countersign queue and immediately goes back to finalise queue to edit the recommendation but the case status does not change like that.
Case status transitions like this: `under_final_review` -> `final_review_countersign` -> `final_review_second_countersign` -> `under_final_review`
-

api/applications/notify.py
```diff
@@ -27,16 +32,31 @@ def notify_exporter_case_opened_for_editing(application):
...
 
-def notify_caseworker_countersign_return(user_email, application, countersign_advice):
-    relative_url = reverse("cases:countersign_decision_advice", kwargs={"pk": application.id})
+# def notify_caseworker_countersign_return(user_email, application, countersign_advice):
+def notify_caseworker_countersign_return(case):
+    countersign_order = (
+        CountersignOrder.FIRST_COUNTERSIGN
+        if case._previous_status.status == CaseStatusEnum.FINAL_REVIEW_COUNTERSIGN
+        else CountersignOrder.SECOND_COUNTERSIGN
+    )
```
saruniitr, line 41: Please see below comment.

Because of the reasons mentioned there you will always get the order as second countersign.

Instead I suggest the following.
There will only be one `CountersignAdvice` rejection either from first or second manager but not both so just filter like this.

```
CountersignAdvice.objects.filter(
    case=case,
    valid=True,
    outcome_accepted=False,
    advice__user__team=Team.objects.get(id=TeamIdEnum.LICENSING_UNIT),
    advice__level=AdviceLevel.FINAL,
    advice_type__in=[APPROVE, REFUSE, NLR]
)
```

and user the `order` from the queryset.

api/applications/notify.py
```diff
@@ -27,16 +32,31 @@ def notify_exporter_case_opened_for_editing(application):
...
+    countersign_advice = CountersignAdvice.objects.get(
+        case=case,
+        valid=True,
+        order=countersign_order,
+        outcome_accepted=False,
+        advice__user__team=Team.objects.get(id=TeamIdEnum.LICENSING_UNIT),
+        advice__level=AdviceLevel.FINAL,
+        advice__type=AdviceType.REFUSE,
```
saruniitr, line 49: this needs to include all types as mentioned above.
`outcome_accepted=False` represents countersign rejection.
Advice type here is the type of the advice which is being countersigned, it could be either approve/refuse/NLR.

saruniitr (approved):

antroy-madetech (approved):

api/applications/tests/test_notify.py
```diff
@@ -33,35 +37,78 @@ def test_notify_licence_issued(self, mock_send_email):
...
+        # frontend_url = (
+        #     f"https://internal.lite.service.localhost.uktrade.digital/cases/{self.case.id}/countersign-decision-advice/"
+        # )
+        # data = {
+        #     "case_reference": self.case.reference_code,
+        #     "countersigned_user_name": f"{countersign_advice.countersigned_user.first_name} {countersign_advice.countersigned_user.last_name}",
+        #     "countersign_reasons": countersign_advice.reasons,
+        #     "recommendation_section_url": frontend_url,
+        # }
+        # expected_payload = CaseWorkerCountersignCaseReturn(**data)
```
saruniitr, line 111: this was not there before, could you please remove this?

PR #1389, opened 2023-04-12
Proviso audit

antroy-madetech:
### Aim
Licence conditions (Proviso advice) should be audited. Audit messages should be as per the designs in the ticket and "additional text" to display in the timeline should be the proviso text.

[LTD-3485](https://uktrade.atlassian.net/browse/LTD-3485)

[LTD-3485]: https://uktrade.atlassian.net/browse/LTD-3485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/audit_trail/formatters.py
+++ b/api/audit_trail/formatters.py
@@ -286,11 +286,15 @@ def generate_decision_letter(**payload):
     return f"invalid decision {decision} for this event."
 
 
-def create_lu_advice(firstname, lastname, advice_type):  # /PS-IGNORE
+def create_lu_advice(firstname, lastname, advice_type, **payload):  # /PS-IGNORE
+    if advice_type == AdviceType.PROVISO:
+        return f"{firstname} {lastname} added a licence condition."  # /PS-IGNORE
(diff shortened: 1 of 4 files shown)


saruniitr (approved):

api/audit_trail/migrations/0014_alter_audit_verb.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("audit_trail", "0013_alter_audit_verb"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
hnryjmes, line 14: is it possible to write a quick test for this migration operation using `MigratorTestCase`? 

you may be able to just check the count is the same using an assertion like
```
self.assertCountEqual(
            updated_audit._meta.get_field('verb').choices,
            new_choices
        )
```

or however you might want to check the count.

hnryjmes (approved):

PR #1396, opened 2023-04-13
Needed a way of setting active to False in create_routing_rule test_helper

Tllew:
Required for the activation helper tests


hnryjmes (approved):

test_helpers/clients.py
```diff
@@ -1118,7 +1116,9 @@ def create_licence(
         licence.decisions.set(decisions)
         return licence
 
-    def create_routing_rule(self, team_id, queue_id, tier, status_id, additional_rules: list, is_python_criteria=False):
+    def create_routing_rule(
+        self, team_id, queue_id, tier, status_id, additional_rules: list, is_python_criteria=False, active=True
```
hnryjmes, line 1120: I've already approved but someone else might have an opinion on what we set as the default here, assuming that `active=True` is the more common case

saruniitr: Could you give some context behind this change?
When we add new routing rules through migration we add them in inactive state and the changes are behind a feature flag, once we are ready to release then we activate them.

saruniitr: with default as `False`, this is good.

Tllew: I've made some tests for the activation helper, this is required to create a routing rule that's turned off, as you can see below active=active was added with the default active=True because in the RoutingRule Object we have this:
```active = models.BooleanField(default=True)```

If I set the default to active I believe it will break every test that uses create_routing_rule, which is a test_helper client

saruniitr (approved):

PR #1398, opened 2023-04-14
Create NCSC feature switch in lite-api

hnryjmes:


api/conf/settings.py
```diff
@@ -455,6 +455,8 @@ def _build_redis_url(base_url, db_number, **query_args):
 
 FEATURE_COUNTERSIGN_ROUTING_ENABLED = env.bool("FEATURE_COUNTERSIGN_ROUTING_ENABLED", True)
 
+FEATURE_NCSC_C7_ROUTING_ENABLED = env.bool("FEATURE_NCSC_C7_ROUTING_ENABLED", False)
```
saruniitr, line 458: Are we on-boarding any other OGDs besides NCSC?
Can it be renamed to `FEATURE_C7_ROUTING_ENABLED`?
If we NCSC in the name then may be `FEATURE_C7_NCSC_ROUTING_ENABLED`?

hnryjmes: my guess is that there will be no more OGD teams included in C7. I don't know that for sure, but perhaps in that case it is better to make this generic, as we have done for other places.

depsiatwal (approved):
lgtm

saruniitr (approved):

PR #1401, opened 2023-04-14
LTD-3513-add-is-ncsc-military-security-product 

depsiatwal:
New data model change to allow TAU office to select via a UI that a good is  a ncsc military security product 
This would then tigger a flag to be added and allowing routing rules to route this case to NCSC.

https://uktrade.atlassian.net/browse/LTD-3513


Tllew (approved):
Looks good, will need to add tests to the helper once this is in dev

hnryjmes (approved):

PR #1402, opened 2023-04-14
Ltd 3543 audit query responses

antroy-madetech:
### Aim

Adds an audit of the ECJU query response from an exporter.

[LTD-3543](https://uktrade.atlassian.net/browse/LTD-3543)

[LTD-3543]: https://uktrade.atlassian.net/browse/LTD-3543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/tests/test_case_ecju_queries.py
```diff
@@ -281,6 +284,12 @@ def _test_exporter_responds_to_query(self, add_documents, query_type):
         response = response.json()["ecju_query"]
         self.assertEqual(response["response"], data["response"])
 
+        lu_advice_audit = Audit.objects.filter(verb=AuditType.ECJU_QUERY_RESPONSE)
```
saruniitr, line 287: this is not specific to LU

api/cases/views/views.py
```diff
@@ -547,7 +547,13 @@ def put(self, request, pk, ecju_pk):
...
             if "validate_only" not in request.data or not request.data["validate_only"]:
                 serializer.save()
-
+                audit_trail_service.create(
+                    actor=request.user,
+                    verb=AuditType.ECJU_QUERY_RESPONSE,
+                    action_object=serializer.instance,
+                    target=serializer.instance.case,
+                    payload={"ecju_response": data["response"]},
+                )
```
saruniitr, line 556: this looks good.
Could you check if this shows up in "Latest action" on All cases page as we are expecting that it gets picked up there once this audit is created.

api/cases/tests/test_case_ecju_queries.py
antroy-madetech: Copy paste error. I will change

saruniitr (approved):

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -281,6 +284,12 @@ def _test_exporter_responds_to_query(self, add_documents, query_type):
         response = response.json()["ecju_query"]
         self.assertEqual(response["response"], data["response"])
 
+        query_response_audit = Audit.objects.filter(verb=AuditType.ECJU_QUERY_RESPONSE)
+        self.assertTrue(query_response_audit.exists())
+        audit_obj = query_response_audit.first()
+        audit_text = AuditSerializer(audit_obj).data["text"]
+        self.assertEqual(audit_text, " responded to an ECJU Query: Attached the requested documents.")
```
currycoder, line 291: Is it worth also making sure that this audit object is attached to the case in question?

currycoder (approved):
Looks great, just one optional comment

PR #1403, opened 2023-04-17
-Downgrades weasyprint to 5.25 where it does not require 1.44+ Pango.

seijihg:
### Aim

There were issues with the newest Weasyprint where it requires Pango 1.44++ version to work.

It was not caught up since my local version is `pango-view (pango) 1.50.9`

https://doc.courtbouillon.org/weasyprint/stable/first_steps.html

`If WeasyPrint is not available on your distribution, or if you want to use a more recent version of WeasyPrint, you have to be sure that [Python](https://www.python.org/) (at least version 3.7.0) and [Pango](https://pango.gnome.org/) (at least version 1.44.0) are installed on your system.
If the version of Pango provided by your distribution is too old, you can use version 52.5 of WeasyPrint which does not need recent Pango features.
`


kevincarrogan (approved):

antroy-madetech (approved):
LGTM

PR #1406, opened 2023-04-18
Ltd 3563 optimise cases search

MarkThomasLite:
### Aim

Reduce queries and duration when loading cases search (1000 DB queries and ~6.5s duration on local).

work was focused on the following:
- populate_destinations
- populate_activity_updates

queries down to 125 and ~1.1s duration, again on local

[LTD-3563](https://uktrade.atlassian.net/browse/LTD-3563)

[LTD-3563]: https://uktrade.atlassian.net/browse/LTD-3563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/views/search/service.py
```diff
@@ -215,7 +215,8 @@ def populate_activity_updates(cases: List[Dict]):
     retrieve the last 2 activities per case for the provided list of cases
     """
     obj_type = ContentType.objects.get_for_model(Case)
-    case_ids = [case["id"] for case in cases]
+    case_map = {case["id"]: case for case in cases}
+    case_ids = list(case_map.keys())
```
kevincarrogan, line 222: You appear to be doing this in both this function and the `populate_destinations`. I wonder if we could produce this map and the ids list just once outside this function and pass it in to both of them instead.

kevincarrogan: You could also have a single loop that just iterates through the cases and produces both the keys list and the dictionary for even MOAR PERFORMANCE GAInS.

MarkThomasLite: will do, was trying to keep them all seperate as the next step would be a rewrite and reuse of objects between several of the populate methods

api/cases/views/search/service.py
```diff
@@ -195,28 +187,26 @@ def get_hmrc_sla_hours(cases: List[Dict]):
...
+        application__in=list(case_map.keys()), deleted_at=None
+    )
+    for poa in poas:
+        serializer = CountrySerializer(poa.party.country)
+        data = serializer.data
+        case = case_map[str(poa.application_id)]
+        if case.get("destinations"):
+            case["destinations"].append({"country": data})
+        else:
+            case["destinations"] = [{"country": data}]
```
kevincarrogan, line 201: I think you've instantiated this elsewhere now.

kevincarrogan (approved):

api/cases/views/search/service.py
MarkThomasLite: course as part of a subsequent bug fix

saruniitr (approved):

PR #1419, opened 2023-04-26
Remove LU post circ feature switch in lite api

hnryjmes:
This is part of the work to cleanup LU post-circ feature switches. It has now been updated to take into account the feature switch work merged into lite-routing. 

This work contains a sha update. If lite-routing is updated before this is merged, the sha will need to be updated again.

Next up: frontend


api/applications/views/helpers/advice.py
```diff
@@ -149,7 +148,7 @@ def mark_lu_rejected_countersignatures_as_invalid(case, user):
     back to caseworker for finalising.
     """
     lu_team = Team.objects.get(id=TeamIdEnum.LICENSING_UNIT)
-    if not settings.FEATURE_COUNTERSIGN_ROUTING_ENABLED or not (user.govuser.team == lu_team):
+    if not (user.govuser.team == lu_team):
```
saruniitr, line 151: As you are changing this, instead of double negation, best to check for `team != lu_team`

hnryjmes: updated

currycoder (approved):
Looks good.  Could the lite-routing sha be updated to latest commit prior to merge though?

PR #1428, opened 2023-05-02
Ltd 3563 optimise case search misc

MarkThomasLite:
### Aim

Reduce queries and duration when loading cases search (125 DB queries and 1s duration on local).

additional work on several small areas

queries down to 60 and 0.7s duration, again on local

[LTD-3563](https://uktrade.atlassian.net/browse/LTD-3563)

[LTD-3563]: https://uktrade.atlassian.net/browse/LTD-3563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/signals.py
```diff
@@ -8,21 +8,26 @@
...
 
 
+@receiver(pre_save, sender=Case)
+def case_pre_save_handler(sender, instance, raw=False, **kwargs):
+    try:
+        previous_record = Case.objects.get(pk=instance.id)
+        instance._previous_status = previous_record.status
+    except Case.DoesNotExist:
+        pass
+
```
currycoder, line 18: Instead of adding a new signal, you could just modify what we had with the `__init__` method and persist the status **id** (`self._previous_status_id`) instead to avoid the query.

api/cases/signals.py
```diff
@@ -8,21 +8,26 @@
...
+
 @receiver(post_save, sender=Case)
 def case_post_save_handler(sender, instance, raw=False, **kwargs):
     if raw:
         return
 
-    if not instance.id:
-        return
```
currycoder, line 17: Did you mean to remove this?  I believe this is important as it means the signal will only act on existing objects and not brand new ones

api/cases/signals.py
currycoder: Although TBH, it's nice to see the tightly related code in the same file..

MarkThomasLite: yeah that was my thought, made it a bit clearer

api/cases/signals.py
MarkThomasLite: ah my understanding was that it always had an id during post but not necessarily during pre

currycoder: My bad, you are correct!

currycoder (approved):

PR #1434, opened 2023-05-05
LTD-3638-api-mentions-save

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-3638

Adds the ability yo save mentions. 

The return value will need some rework along with the ability yo save a Team.


Tllew (approved):
Looks good to me

PR #1438, opened 2023-05-12
 LTD-3654-pull-mentions-data 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-3654

This API call has been linked with the notes and timelines screen. 

The mentions portion will pull the mentions on a screen and display as per requirements. We could have used the same data shape as the Audit Table but I believe this is fundamentally wrong and a bit of a hack. Hence a new API call to pull real data from the source table.

### DIFF
--- a/api/cases/serializers.py
+++ b/api/cases/serializers.py
@@ -306,6 +306,11 @@ class CaseNoteMentionsSerializer(serializers.ModelSerializer):
     user = PrimaryKeyRelatedSerializerField(
         queryset=GovUser.objects.all(), serializer=GovUserViewSerializer, required=False
     )
+    case_note_user = PrimaryKeyRelatedSerializerField(
+        queryset=BaseUser.objects.all(), source="case_note.user", serializer=BaseUserViewSerializer, required=False
+    )
+    is_urgent = serializers.BooleanField(source="case_note.is_urgent", required=False)
+    case_note_text = serializers.CharField(source="case_note.text", required=False)
 
     class Meta:
         model = CaseNoteMentions
@@ -314,6 +319,10 @@ class CaseNoteMentionsSerializer(serializers.ModelSerializer):
             "team",
             "is_accessed",
             "user",
+            "created_at",
+            "case_note_user",
+            "is_urgent",
+            "case_note_text",
         )
 
 
--- a/api/cases/tests/test_case_notes.py
+++ b/api/cases/tests/test_case_notes.py
@@ -20,7 +20,6 @@ class CaseNotesGovCreateTests(DataTestClient):
 
     def test_create_case_note_successful(self):
         response = self.client.post(self.url, data=self.data, **self.gov_headers)
-
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         self.assertEqual(CaseNote.objects.count(), 1)
         self.assertEqual(CaseNote.objects.get().text, self.data.get("text"))
@@ -167,3 +166,22 @@ class CaseNotesViewTests(DataTestClient):
 
         response = self.client.get(reverse("cases:activity", kwargs={"pk": case.id}), **self.gov_headers)
         self.assertIn(text, str(response.json()))
+
+
+class CaseNoteMentionsViewTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        self.case = self.create_clc_query("Query", self.organisation)
+        self.url = reverse("cases:case_note_mentions", kwargs={"pk": self.case.id})
+
+    def test_view_case_mentions_successful(self):
+
+        case_note = self.create_case_note(self.case, "Hairpin Turns", self.gov_user.baseuser_ptr)
+        self.other_user = self.create_gov_user("test@gmail.com", self.team)  # /PS-IGNORE
+        self.create_case_note_mention(case_note, self.other_user)
+        response = self.client.get(self.url, **self.gov_headers)
+        result = response.json()
+
+        self.assertEqual(result["count"], 1)
+        self.assertEqual(result["results"][0]["user"]["id"], str(self.other_user.baseuser_ptr.id))
+        self.assertEqual(result["results"][0]["case_note_user"]["id"], str(self.gov_user.baseuser_ptr.id))
--- a/api/cases/tests/test_case_notes.py
+++ b/api/cases/tests/test_case_notes.py
@@ -42,14 +42,14 @@ class CaseNotesGovCreateTests(DataTestClient):
         self.data["mentions"] = mentions
         response = self.client.post(self.url, data=self.data, **self.gov_headers)
 
-        assert response.status_code == status.HTTP_201_CREATED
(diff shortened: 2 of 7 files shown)


Tllew (approved):
Looks good

api/cases/serializers.py
```diff
@@ -361,6 +361,11 @@ class CaseNoteMentionsSerializer(serializers.ModelSerializer):
     user = PrimaryKeyRelatedSerializerField(
         queryset=GovUser.objects.all(), serializer=GovUserViewSerializer, required=False
     )
+    case_note_user = PrimaryKeyRelatedSerializerField(
+        queryset=BaseUser.objects.all(), source="case_note.user", serializer=BaseUserViewSerializer, required=False
```
saruniitr, line 365: best to limit this to `GovUser` only

api/cases/tests/test_case_notes.py
```diff
@@ -167,3 +166,22 @@ def test_view_pre_submitted_case_notes_post_submit(self):
...
+
+        case_note = self.create_case_note(self.case, "Hairpin Turns", self.gov_user.baseuser_ptr)
+        self.other_user = self.create_gov_user("test@gmail.com", self.team)  # /PS-IGNORE
+        self.create_case_note_mention(case_note, self.other_user)
+        response = self.client.get(self.url, **self.gov_headers)
+        result = response.json()
+
+        self.assertEqual(result["count"], 1)
+        self.assertEqual(result["results"][0]["user"]["id"], str(self.other_user.baseuser_ptr.id))
+        self.assertEqual(result["results"][0]["case_note_user"]["id"], str(self.gov_user.baseuser_ptr.id))
```
currycoder, line 187: Would suggest asserting the whole result e.g. `assert result == expected_result`.  Keeps the tests tight and exhaustive

api/cases/views/views.py
```diff
@@ -1162,3 +1164,11 @@ def get_queryset(self):
                 "control_list_entries",
             )
         )
+
+
+class CaseNoteMentionList(ListAPIView):
```
currycoder, line 1169: I wonder if this should all be in its own views file - this one is getting pretty chonky with a lot of unrelated stuff.

api/cases/serializers.py
currycoder: I don't really know what `PrimaryKeyRelatedSerializerField` gets us; I think replacing with the following gives us more options for selecting all the data in a single query;
```
    case_note_user = BaseUserViewSerializer(
        source="case_note.user", required=False
    )
```

api/cases/serializers.py
```diff
@@ -361,6 +361,11 @@ class CaseNoteMentionsSerializer(serializers.ModelSerializer):
     user = PrimaryKeyRelatedSerializerField(
```
currycoder, line 361: Same as below, maybe replace with;
```
    user = GovUserViewSerializer()
```

api/cases/views/views.py
```diff
@@ -1162,3 +1164,11 @@ def get_queryset(self):
...
+    authentication_classes = (GovAuthentication,)
+    serializer_class = CaseNoteMentionsSerializer
+
+    def get_queryset(self):
+        return CaseNoteMentions.objects.filter(case_note__case_id=self.kwargs["pk"]).order_by("-created_at")
```
currycoder, line 1174: ATM, I think the serializer will perform three extra queries per CaseNoteMention record to populate `case_note`, `user`, `case_note.user`.  I think this could be avoided with;
```
CaseNoteMentions.objects
    .select_related("case_note", "user", "case_note__user")    
    .filter(case_note__case_id=self.kwargs["pk"]).order_by("-created_at")
```

api/cases/views/views.py
```diff
@@ -1162,3 +1164,11 @@ def get_queryset(self):
                 "control_list_entries",
             )
         )
+
+
+class CaseNoteMentionList(ListAPIView):
+    authentication_classes = (GovAuthentication,)
+    serializer_class = CaseNoteMentionsSerializer
+
+    def get_queryset(self):
```
currycoder, line 1173: Functionality wise, it seems like this pulls all CaseNoteMention objects for the case - is there a need to get mentions just for a single user?

api/cases/serializers.py
depsiatwal: > best to limit this to `GovUser` only

I was forced to use baseUser as it seems to have issues serialising Team field. 
I think it's save for now since this is used for relationships as opposed to creating.

depsiatwal: > I don't really know what `PrimaryKeyRelatedSerializerField` gets us; I think replacing with the following gives us more options for selecting all the data in a single query;
> 
> ```
>     case_note_user = BaseUserViewSerializer(
>         source="case_note.user", required=False
>     )
> ```
PrimaryKeyRelatedField this allows a list of Id's to be passed to create a relationship between the tables. 
else it would expected the user object to be created and expects a dict.

api/cases/tests/test_case_notes.py
depsiatwal: yeah I also prefer assert result == expected_result
was just trying to keep it consistent  with the other tests 

anyway changed ...

api/cases/views/views.py
depsiatwal: ahh nice one changed

api/cases/views/views.py
depsiatwal: there will be that I think will require a very different view of the data 
will consolidate and refactor is need be

api/cases/tests/test_case_notes.py
currycoder: Sorry my point wasn't about `self.assertEqual` vs `assert` but about asserting that the *whole* result JSON was what we expected it to be rather than just individual fields.  IMO that gives us proper coverage of the format of the API.

currycoder (approved):
Nice one! Just one comment left regarding testing result formats which I will leave with you.

api/cases/tests/test_case_notes.py
```diff
@@ -30,21 +29,27 @@ def test_create_case_note_with_mentions_successful(self):
...
-        self.assertEqual(CaseNote.objects.get().is_urgent, True)
-        self.assertEqual(CaseNoteMentions.objects.count(), 1)
+        assert response.status_code == status.HTTP_201_CREATED
+        assert CaseNote.objects.count() == 1
+        assert CaseNote.objects.get().text == self.data.get("text")
+        assert CaseNote.objects.get().is_visible_to_exporter is False
+        assert CaseNote.objects.get().is_urgent is True
+        assert CaseNoteMentions.objects.count() == 2
         assert response.json()["case_note"]["mentions"][0]["user"]["id"] == mentions[0]["user"]
+        assert response.json()["case_note"]["mentions"][1]["user"]["id"] == mentions[1]["user"]
```
kevincarrogan, line 52: We use `self.assertX` for the API not `assert`.

PR #1449, opened 2023-05-18
LTD-3674-api-mentions-user-case-view 

depsiatwal:
This is the API to support this screen. 

https://uktrade.atlassian.net/browse/LTD-3622

I questioned if this should site with users/cases app. Since it's so closely related to case-notes for now it resides with case notes where we have all the tests/serializer and views for case-note mentions. 

During implementation one thing that’s come to lite is which queue is used when we link back to the notes and timelines section for the case when responding to a mention. 

 

Currently I’m checking the queues of the case and if any of the team queues match we use the matching queue. If none are matching i.e Case has moved on or it’s no longer with this team. We default to “All cases Queue” which will limit what actions the user can action with the case.


Tllew (approved):
Looks good to me

api/cases/views/case_notes.py
```diff
@@ -88,7 +88,31 @@ def get_queryset(self):
...
+            CaseNoteMentions.objects.select_related(
+                "case_note",
+                "user",
+                "case_note__user",
+                "case_note__user__govuser__team",
+                "case_note__case",
+            )
+            .filter(user_id=request.user.pk)
+            .order_by("-created_at")
+        )
```
currycoder, line 115: Do we want to do any limiting/pagination here?  Feels like this is going to be weighty over time

api/cases/tests/test_case_notes.py
```diff
@@ -194,3 +194,56 @@ def test_view_case_mentions_successful(self):
...
+        first_mention = result[0]
+
+        self.assertEqual(len(result), 2)
+        self.assertEqual(first_mention["user"]["id"], str(self.gov_user.baseuser_ptr.id))
+        self.assertEqual(first_mention["case_note_user"]["id"], str(self.other_user.baseuser_ptr.id))
+        self.assertEqual(first_mention["case_note_text"], self.case_note_other.text)
+        self.assertEqual(first_mention["case_note"], str(self.case_note_other.id))
+        self.assertEqual(first_mention["reference_code"], self.case_note_other.case.reference_code)
+        self.assertEqual(first_mention["case_id"], str(self.case_note_other.case.id))
+        self.assertEqual(first_mention["is_urgent"], self.case_note_other.is_urgent)
```
currycoder, line 229: Can we please just assert the *whole result* rather than individual parts of it?  This will force us to update tests when the serializers change in future - we will otherwise miss doing these updates for additions IMO

```
expected_result = {...}
self.assertEqual(result, expected_result)
```

api/cases/serializers.py
```diff
@@ -366,6 +367,17 @@ class CaseNoteMentionsSerializer(serializers.ModelSerializer):
     )
     is_urgent = serializers.BooleanField(source="case_note.is_urgent", required=False)
     case_note_text = serializers.CharField(source="case_note.text", required=False)
+    reference_code = serializers.CharField(source="case_note.case.reference_code", required=False, read_only=True)
+    case_id = serializers.UUIDField(source="case_note.case.id", required=False, read_only=True)
+    case_queue_id = serializers.SerializerMethodField()
+
+    def get_case_queue_id(self, instance):
+        # Check that the case's queue is still within the users team queue else fallback to all cases queue
+        if isinstance(instance, CaseNoteMentions):
```
currycoder, line 376: Why do we need to assert this?  Do we try to serialize records that aren't `CaseNoteMentions` with this serializer?

api/cases/serializers.py
```diff
@@ -366,6 +367,17 @@ class CaseNoteMentionsSerializer(serializers.ModelSerializer):
...
+            for case_queue in instance.case_note.case.queues.all():
+                if case_queue in instance.user.team.queue_set.all():
+                    return case_queue.id
```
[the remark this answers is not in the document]
depsiatwal: I did a write up about it here 
https://uktrade.atlassian.net/browse/LTD-3622

basically it's because the case can move on from the queue and user may not have access to the original queue.

api/cases/views/case_notes.py
depsiatwal: I think one for the future atm this is early MVP and we not even sure how mentions will be used. I didn't want to build too much around this ATM until we know more.

currycoder: Ok sounds reasonable, could I suggest that we have a ticket for that though so that it's not something we lose track of?

api/cases/serializers.py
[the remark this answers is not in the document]
currycoder: OK fair enough, maybe something to come back to later.  Could I suggest that this approach is sort of written up in a comment here for people coming along and having no reference back to the jira?

api/cases/views/case_notes.py
depsiatwal: ahh yeah of course added https://uktrade.atlassian.net/browse/LTD-3695

api/cases/serializers.py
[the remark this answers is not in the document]
depsiatwal: added to main description

api/cases/serializers.py
depsiatwal: ahh it turns out this is because of the relationship serialisers on the CaseNote turns out we don't need that for now. And hence I have removed this check.

PR #1454, opened 2023-05-23
 LTD-3699-is-new-reset 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-3699

This adds a new end point which allows bulk update of case note mentions. 
Predominantly used to update the is_accessed flag. This is extending a new view ideally the get and patch could be in a single view. However the get is making use of the DRF ListAPIView and the new can't be mixed. 

I will add in a jira to collapse these into a single view.

### DIFF
--- a/api/cases/tests/test_case_notes.py
+++ b/api/cases/tests/test_case_notes.py
@@ -178,7 +178,7 @@ class CaseNoteMentionsViewTests(DataTestClient):
     def setUp(self):
         super().setUp()
         self.case = self.create_clc_query("Query", self.organisation)
-        self.url = reverse("cases:case_note_mentions", kwargs={"pk": self.case.id})
+        self.url = reverse("cases:case_note_mentions_list", kwargs={"pk": self.case.id})
 
     def test_view_case_mentions_successful(self):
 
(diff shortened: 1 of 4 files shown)


Tllew (approved):

api/cases/views/case_notes.py
```diff
@@ -96,6 +96,31 @@ def get_queryset(self):
...
+            data=update_data,
+            partial=True,
+            many=True,
+        )
+
+        if serializer.is_valid():
+            serializer.save()
+            return JsonResponse(
+                data={"mentions": serializer.data},
+                status=status.HTTP_201_CREATED,
```
saruniitr, line 119: this is an update so cannot be 201

api/cases/tests/test_case_notes.py
```diff
@@ -261,3 +261,36 @@ def test_view_user_case_mentions_case_queue_match(self):
...
+
+        case_note_mention = self.create_case_note_mention(self.case_note, self.gov_user)
+        case_note_mention_2 = self.create_case_note_mention(self.case_note, self.gov_user)
+        self.assertEqual(case_note_mention.is_accessed, False)
+        self.assertEqual(case_note_mention_2.is_accessed, False)
+
+        url = reverse("cases:case_note_mentions")
+        update_data = [
+            {"id": case_note_mention.pk, "is_accessed": True},
+            {"id": case_note_mention_2.pk, "is_accessed": True},
```
saruniitr, line 275: can one of this to be set to `False` so we check updating of both?

api/cases/views/case_notes.py
depsiatwal: update put and patch can be 200 or 201 since no other codes are available. 
I think for most updates we use 201

saruniitr (approved):

PR #1460, opened 2023-05-31
 LTD-3741-api-call-new-mentions 

depsiatwal:
A count of new mentions for a user is required in the UI menu.
Hence a light weight API call just to get a count.  I did think about making this more generic or extend one of the other review for filtering. But figured this is used so often and is imperative to the application functioning it's worth having it's own API call.

### DIFF
--- a/api/cases/views/case_notes.py
+++ b/api/cases/views/case_notes.py
@@ -153,3 +153,23 @@ class UserCaseNoteMention(APIView):
         serializer = self.serializer_class(qs, many=True)
 
         return JsonResponse(data={"mentions": serializer.data})
+
+
+class UserCaseNoteMentionsNewCount(APIView):
+    authentication_classes = (GovAuthentication,)
+
+    def get(self, request):
+        """Gets count of all new mentions for user."""
+        qs = (
+            CaseNoteMentions.objects.select_related(
+                "case_note",
+                "user",
+                "case_note__user",
+                "case_note__user__govuser__team",
+                "case_note__case",
+            )
+            .filter(user_id=request.user.pk)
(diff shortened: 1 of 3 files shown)


api/cases/views/case_notes.py
```diff
@@ -153,3 +153,23 @@ def get(self, request):
...
         return JsonResponse(data={"mentions": serializer.data})
+
+
+class UserCaseNoteMentionsNewCount(APIView):
+    authentication_classes = (GovAuthentication,)
+
+    def get(self, request):
+        """Gets count of all new mentions for user."""
+        qs = (
+            CaseNoteMentions.objects.select_related(
```
kevincarrogan, line 164: I don’t think you need the select related stuff here seeing as you’re just getting the count.

api/cases/views/case_notes.py
```diff
@@ -153,3 +153,23 @@ def get(self, request):
...
+                "case_note",
+                "user",
+                "case_note__user",
+                "case_note__user__govuser__team",
+                "case_note__case",
+            )
+            .filter(user_id=request.user.pk)
```
kevincarrogan, line 171: You can combine these two filter calls into one here.

api/cases/views/case_notes.py
depsiatwal: yeah good point done

api/cases/views/case_notes.py
depsiatwal: done

kevincarrogan (approved):

Tllew (approved):

PR #1483, opened 2023-06-20
Updated control list entry filter to accept multiple values

antroy-madetech:
### Aim

API changes to support multi-select of the CLE filter.

[LTD-3828](https://uktrade.atlassian.net/browse/LTD-3828)

[LTD-3828]: https://uktrade.atlassian.net/browse/LTD-3828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/views/search/views.py
+++ b/api/cases/views/search/views.py
@@ -143,7 +143,7 @@ class CasesSearchView(generics.ListAPIView):
             filters[selected_tab] = True
 
         filters["flags"] = request.GET.getlist("flags", [])
-        filters["control_list_entry"] = request.GET.getlist("control_list_entry", [])
+        filters["control_list_entry"] = [cle for cle in request.GET.getlist("control_list_entry", []) if cle]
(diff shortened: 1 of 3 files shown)


api/cases/views/search/views.py
```diff
@@ -143,6 +143,7 @@ def get_filters(self, request):
             filters[selected_tab] = True
 
         filters["flags"] = request.GET.getlist("flags", [])
+        filters["control_list_entry"] = request.GET.getlist("control_list_entry", [])
```
currycoder, line 146: Just to talk this through, would this still be backwards compatible in the case where a user searched CLE where the API has been updated and the frontend is yet to be released?

antroy-madetech: Tested this branch on api against dev on frontend and all seems fine (applying a filter with a CLE and clicking/saving/edit/delete of bookmarks from that filter)

PR #1484, opened 2023-06-21
Regime multi select

antroy-madetech:
### Aim

Support Multi-selection in the case filter for regime.

[LTD-3829](https://uktrade.atlassian.net/browse/LTD-3829)

[LTD-3829]: https://uktrade.atlassian.net/browse/LTD-3829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/views/search/views.py
+++ b/api/cases/views/search/views.py
@@ -143,7 +143,7 @@ class CasesSearchView(generics.ListAPIView):
             filters[selected_tab] = True
 
         filters["flags"] = request.GET.getlist("flags", [])
-        filters["regime_entry"] = request.GET.getlist("regime_entry", [])
+        filters["regime_entry"] = [regime for regime in request.GET.getlist("regime_entry", []) if regime]
         filters["assigned_queues"] = request.GET.getlist("assigned_queues", [])
         filters["submitted_from"] = make_date_from_params("submitted_from", filters)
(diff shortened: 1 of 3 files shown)


currycoder (approved):
Approved subject to comment

api/cases/views/search/views.py
```diff
@@ -143,6 +143,7 @@ def get_filters(self, request):
             filters[selected_tab] = True
 
         filters["flags"] = request.GET.getlist("flags", [])
+        filters["regime_entry"] = request.GET.getlist("regime_entry", [])
```
currycoder, line 146: Ah I just found a bug with this which may occur during the release.  We talked about something similar on the CLE PR and I think it's ok for that but doesn't work due to the frontend component choice here..

If you checkout this API branch and have frontend on `dev` branch, you will get a 500 error from the API when applying filters when the `regime_entry` field is unpopulated.  Maybe there's something we easy we can put in here to prevent that issue?

antroy-madetech: Is it sending an empty string then rather than just not populating it? I'll test it.

antroy-madetech:
Merged with CLE work in a different branch

PR #1487, opened 2023-06-26
Regime and CLE multi-select for Product filter tab

antroy-madetech:
### Aim
Add multi-select functionality to regime and CLE filters.

[LTD-3828](https://uktrade.atlassian.net/browse/LTD-3828)
[LTD-3829](https://uktrade.atlassian.net/browse/LTD-3829)

[LTD-3828]: https://uktrade.atlassian.net/browse/LTD-3828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-3829]: https://uktrade.atlassian.net/browse/LTD-3829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -110,8 +110,8 @@ class CaseQuerySet(models.QuerySet):
             | Q(baseapplication__goods_type__control_list_entries__rating__in=[control_list_entry])
         )
 
-    def with_regime_entry(self, regime_entry):
-        return self.filter(baseapplication__goods__regime_entries__id=regime_entry)
+    def with_regime_entries(self, regime_entries):
+        return self.filter(baseapplication__goods__regime_entries__id__in=regime_entries)
 
     def with_flags(self, flags):
         case_flag_ids = self.filter(flags__id__in=flags).values_list("id", flat=True)
@@ -357,7 +357,7 @@ class CaseManager(models.Manager):
             case_qs = case_qs.with_control_list_entry(control_list_entry)
 
         if regime_entry:
-            case_qs = case_qs.with_regime_entry(regime_entry)
+            case_qs = case_qs.with_regime_entries(regime_entry)
 
         if flags:
             case_qs = case_qs.with_flags(flags)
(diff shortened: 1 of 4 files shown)


api/cases/managers.py
```diff
@@ -104,14 +104,14 @@ def with_exporter_site_address(self, exporter_site_address):
             | Q(baseapplication__application_sites__site__address__address__icontains=exporter_site_address)
         )
 
-    def with_control_list_entry(self, control_list_entry):
+    def with_control_list_entries(self, control_list_entries):
         return self.filter(
-            Q(baseapplication__goods__good__control_list_entries__rating__in=[control_list_entry])
-            | Q(baseapplication__goods_type__control_list_entries__rating__in=[control_list_entry])
+            Q(baseapplication__goods__good__control_list_entries__rating__in=control_list_entries)
+            | Q(baseapplication__goods_type__control_list_entries__rating__in=control_list_entries)
```
saruniitr, line 110: `baseapplication_goods` are the instances of `GoodOnApplication` where as
`baseapplication__goods__good` are instances of `Good`.

I think we should search only in `GoodOnApplication` as that is where the assessment is recorded.
We update in `Good` also but that might change when we address this properly.

Also we can ignore `goods_type` as they are related to a different licence type other than SIELs

currycoder (approved):
Approved pending Arun's comment

api/cases/managers.py
antroy-madetech: This was the existing filter behaviour. (I just removed the list wrapping the value) but I can look at this. So change to just "baseapplication__goods__control_list_entries__rating__in" you mean?

antroy-madetech: I've replaced these two Q() | Q() filters with just the baseapplication__goods__control_list_entries__rating__in and I get considerably different results. setting the CLEs searched for to MEND2 and ML11a (UAT data but a few weeks old) I get 27 results for the original filter, but only 19 with the new filter.

antroy-madetech: Note that removing the goods_type part of the query has no effect on the results (with or without the __good__ in the application query), so that seems OK to remove on a very small sample set.

saruniitr (approved):

PR #1508, opened 2023-07-14
LTD-3959-filtering

seijihg:
### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-3959

### Reviewing help

### Merging

### DIFF
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -229,8 +229,8 @@ class CaseQuerySet(models.QuerySet):
     def exclude_sanction_matches(self):
         return self.exclude(baseapplication__parties__sanction_matches__is_revoked=False)
 
-    def includes_refusal_recommendation(self):
-        return self.filter(advice__type=AdviceType.REFUSE, status__status=CaseStatusEnum.OGD_ADVICE)
+    def includes_refusal_recommendation_from_ogd(self):
+        return self.filter(advice__type=AdviceType.REFUSE, advice__team__is_ogd=True)
 
 
 class CaseManager(models.Manager):
@@ -290,7 +290,7 @@ class CaseManager(models.Manager):
         exclude_sanction_matches=None,
         max_total_value=None,
         report_summary=None,
-        includes_refusal_recommendation=None,
+        includes_refusal_recommendation_from_ogd=None,
         **kwargs,
     ):
         """
@@ -465,8 +465,8 @@ class CaseManager(models.Manager):
         if max_total_value:
             case_qs = case_qs.with_max_total_value(max_total_value)
 
-        if includes_refusal_recommendation:
-            case_qs = case_qs.includes_refusal_recommendation()
+        if includes_refusal_recommendation_from_ogd:
+            case_qs = case_qs.includes_refusal_recommendation_from_ogd()
 
         return case_qs.distinct()
 
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -230,7 +230,7 @@ class CaseQuerySet(models.QuerySet):
         return self.exclude(baseapplication__parties__sanction_matches__is_revoked=False)
 
     def includes_refusal_recommendation_from_ogd(self):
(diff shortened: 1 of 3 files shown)


api/cases/managers.py
```diff
@@ -228,6 +229,9 @@ def exclude_denial_matches(self):
     def exclude_sanction_matches(self):
         return self.exclude(baseapplication__parties__sanction_matches__is_revoked=False)
 
+    def includes_refusal_recommendation(self):
```
depsiatwal, line 232: do we need a new function can we not use 
https://github.com/uktrade/lite-api/blob/118387ce120042e534238afb384764ed20c692a4/api/cases/managers.py#L140C9-L140C20
with_advice

seijihg: ha nice did not see that :D

api/cases/managers.py
```diff
@@ -228,6 +229,9 @@ def exclude_denial_matches(self):
...
+        return self.filter(advice__type=AdviceType.REFUSE, status__status=CaseStatusEnum.OGD_ADVICE)
+
```
saruniitr, line 234: Problem with this is that this filter only works when the case is with OGDs, once it has come back to LU then it won't work because the status would have changed, is this the requirement?
From the ticket it doesn't look like that is the case.

I think instead of status you need to check the team of the user who gave the advice whether that is one of the OGD.

seijihg: ah ur correct on this.

api/cases/tests/test_case_search.py
```diff
@@ -744,6 +744,20 @@ def test_get_cases_filter_by_country_no_match(self, countries_on_application, fi
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(len(response_data), 0)
 
+    def test_get_cases_filter_by_includes_refusal_recommendation(self):
+        case = self.create_standard_application_case(self.organisation)
```
saruniitr, line 748: Can you also add a negative test where we don't get anything if the user team is not ogd.

saruniitr (approved):

api/cases/tests/test_case_search.py
```diff
@@ -744,6 +744,31 @@ def test_get_cases_filter_by_country_no_match(self, countries_on_application, fi
...
+        good = self.create_good("A good", self.organisation)
+        self.gov_user.team = team_ogd
+        self.gov_user.save()
+
+        factories.UserAdviceFactory(user=self.gov_user, case=case, good=good, type=AdviceType.REFUSE)
+
+        url = f'{reverse("cases:search")}?includes_refusal_recommendation_from_ogd=on'
+        response = self.client.get(url, **self.gov_headers)
+        response_data = response.json()["results"]["cases"]
+        self.assertTrue(response_data)
```
kevincarrogan, line 759: Could we check that we get back the results we expect here as opposed to just checking that there are results?

api/cases/tests/test_case_search.py
```diff
@@ -744,6 +744,31 @@ def test_get_cases_filter_by_country_no_match(self, countries_on_application, fi
...
+    def test_get_cases_filter_by_includes_refusal_recommendation_not_met(self):
+        case = self.create_standard_application_case(self.organisation)
+        good = self.create_good("A good", self.organisation)
+
+        factories.UserAdviceFactory(user=self.gov_user, case=case, good=good, type=AdviceType.REFUSE)
+
+        url = f'{reverse("cases:search")}?includes_refusal_recommendation_from_ogd=on'
+        response = self.client.get(url, **self.gov_headers)
+        response_data = response.json()["results"]["cases"]
+        self.assertFalse(response_data)
```
kevincarrogan, line 770: I think we're really saying here that there's no cases right? I think this test would be better if it showed that instead of just checking that `response_data` was falsey.

kevincarrogan (approved):

PR #1512, opened 2023-07-18
LTD-3938-add-case-advice-column

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-3958

This enriches the data to show case recommendation  on the default queue view page.


api/cases/views/search/service.py
```diff
@@ -240,6 +242,37 @@ def populate_denials(case_map):
...
 
+def populate_advice(case_map):
+    for case in case_map.values():
+        case["advice"] = []
+
+    case_advices = (
+        Advice.objects.select_related(
+            "user",
+        )
+        .prefetch_related()
```
kevincarrogan, line 253: This `prefetch_related` isn't doing anything here.

api/applications/serializers/advice.py
```diff
@@ -65,6 +66,19 @@ class AdviceViewSerializer(serializers.Serializer):
...
 
+class AdviceSearchViewSerializer(serializers.Serializer):
+    id = serializers.UUIDField()
+    text = serializers.CharField()
+    type = KeyValueChoiceField(choices=AdviceType.choices)
+    level = serializers.CharField()
+    proviso = serializers.CharField()
+    denial_reasons = PrimaryKeyRelatedSerializerField(
+        queryset=DenialReason.objects.all(), many=True, serializer=DenialReasonSerializer
+    )
```
kevincarrogan, line 77: Are you displaying the denial reasons on the page? If so then this might want to be part of a `select_related` or `prefetch_related` (not sure which) or you've just introduced a bunch of extra queries.

Tllew (approved):
Looks good

api/cases/views/search/service.py
```diff
@@ -240,6 +242,37 @@ def populate_denials(case_map):
...
+        result = defaultdict(list)
+        for advice in case_advice:
+            # Group cases by team and type
+            advice_key = advice["user"]["team"]["id"] + advice["type"]["key"]
+            # Filter duplicate advice records, which is duplicated per good
+            if not result[advice_key]:
+                result[advice_key].append(advice)
+        case["advice"] = result
+
+    return list(case_map.values())
```
kevincarrogan, line 273: You don't do anything with this returned value so we don't need it.

api/applications/serializers/advice.py
```diff
@@ -65,6 +66,19 @@ class AdviceViewSerializer(serializers.Serializer):
...
+    user = PrimaryKeyRelatedSerializerField(queryset=GovUser.objects.all(), serializer=GovUserListSerializer)
```
kevincarrogan, line 78: This seems weird to be using the User*List* serializer as that sounds like it's specifically for the list page.

api/cases/views/search/service.py
depsiatwal: removed seem to be the same pattern used on a lot of these populate

api/applications/serializers/advice.py
depsiatwal: add denial reason to prefetch

api/cases/views/search/service.py
```diff
@@ -240,6 +242,37 @@ def populate_denials(case_map):
...
+    for case in case_map.values():
+        case_advice = case["advice"]
+        result = defaultdict(list)
+        for advice in case_advice:
+            # Group cases by team and type
+            advice_key = advice["user"]["team"]["id"] + advice["type"]["key"]
+            # Filter duplicate advice records, which is duplicated per good
+            if not result[advice_key]:
+                result[advice_key].append(advice)
+        case["advice"] = result
```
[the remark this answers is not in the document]
depsiatwal: Resolved the re-looping i'm hoping this is more optimised now and clearer

api/cases/tests/test_case_search.py
```diff
@@ -1067,6 +1071,8 @@ def test_api_success(self):
         self.assertEqual(case_api_result["sla_days"], 0)
         self.assertEqual(case_api_result["sla_remaining_days"], None)
         self.assertEqual(case_api_result["status"]["key"], self.case.status.status)
+        self.assertEqual(len(case_api_result["advice"][str(self.advice.user.team.id) + self.advice.type]), 1)
```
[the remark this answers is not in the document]
depsiatwal: added a test for groups where duplicate should be removed and a non group where I check the keys and a couple of values, The payload for the whole of denial reasons might be too much to check

api/cases/views/search/service.py
```diff
@@ -240,6 +242,31 @@ def populate_denials(case_map):
...
+
+    for case in case_map.values():
+        case_advice = case_advice_map[case["id"]]
+        case_advice_result = defaultdict(list)
+        for advice in case_advice:
+            serializer = AdviceSearchViewSerializer(advice)
+            # Filter duplicate advice records, which is duplicated per good
+            advice_key = str(advice.user.team.id) + advice.type
+            if not case_advice_result[advice_key]:
+                case_advice_result[advice_key].append(serializer.data)
```
kevincarrogan, line 266: Does the frontend actually need to care about this `advice_key` or is this more so we can filter out duplicates?

kevincarrogan: Also. It appears that this thing is a list that we're appending to but we'll only ever append to it once because once we've appended to it we'll never get in here again. So does this need to be a list?

kevincarrogan (approved):

PR #1513, opened 2023-07-19
Added end-users / ult-end-users to the case payload

antroy-madetech:
### Aim

Added end-users / ult-end-users to the case payload

This is to enable the frontend to display the end-users in a configurable column in the queue display.

[LTD-3960](https://uktrade.atlassian.net/browse/LTD-3960)

[LTD-3960]: https://uktrade.atlassian.net/browse/LTD-3960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -318,7 +318,7 @@ class CaseManager(models.Manager):
                 "queues__team",
                 Prefetch(
                     "baseapplication__parties",
-                    to_attr="enduser_parties",
+                    to_attr="end_user_parties",
                     queryset=PartyOnApplication.objects.select_related("party").filter(
                         deleted_at__isnull=True,
                         party__type__in=["end_user", "ultimate_end_user"],
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -307,6 +307,7 @@ class CaseManager(models.Manager):
                 "case_assignments__queue",
                 "queues",
                 "queues__team",
+                "baseapplication__parties__party",
             )
         )
 
--- a/api/cases/managers.py
+++ b/api/cases/managers.py
@@ -3,7 +3,14 @@ from typing import List
 
 from compat import get_model
 from django.db import models, transaction
-from django.db.models import Q, Case, When, BinaryField, Sum
+from django.db.models import (
+    BinaryField,
+    Case,
+    Prefetch,
+    Q,
+    Sum,
+    When,
+)
 from django.utils import timezone
 
 from api.cases.enums import AdviceLevel, CaseTypeEnum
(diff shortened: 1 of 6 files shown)


api/cases/managers.py
```diff
@@ -302,7 +311,14 @@ def search(  # noqa
                 "case_assignments__queue",
                 "queues",
                 "queues__team",
-                "baseapplication__parties__party",
+                Prefetch(
+                    "baseapplication__parties",
+                    to_attr="enduser_parties",
+                    queryset=PartyOnApplication.objects.select_related("party").filter(
+                        deleted_at__isnull=True,
```
saruniitr, line 318: shouldn't this be `False`?
otherwise you only get deleted parties. This is set only when a party is deleted.

antroy-madetech: I think this is correct, as it checks if deleted_at is null.

saruniitr: actually you are correct

api/cases/tests/test_case_search.py
```diff
@@ -97,6 +99,27 @@ def test_get_cases_no_filter_returns_all_cases(self):
             response_data["filters"]["gov_users"],
         )
 
+    def test_get_cases_returns_all_cases_with_enduser_and_ultimate_enduser_data(self):
+        """
+        Given multiple Cases exist with different statuses and case-types
+        When a user requests to view all Cases with no filter
+        Then all Cases are returned
+        """
```
kevincarrogan, line 107: I don't think this comment matches this test.

kevincarrogan (approved):

api/cases/serializers.py
```diff
@@ -128,6 +128,7 @@ class CaseListSerializer(serializers.Serializer):
     has_open_queries = serializers.BooleanField()
     case_officer = serializers.SerializerMethodField()
     intended_end_use = serializers.SerializerMethodField()
+    endusers = serializers.SerializerMethodField()
```
saruniitr, line 131: may be `end_users` is better to keep it consistent

api/cases/serializers.py
```diff
@@ -177,6 +178,14 @@ def get_intended_end_use(self, instance):
         except BaseApplication.DoesNotExist:
             return ""
 
+    def get_endusers(self, instance):
+        if hasattr(instance, "baseapplication") and hasattr(instance.baseapplication, "enduser_parties"):
+            return [
+                {"name": party_on_app.party.name, "type": party_on_app.party.type}
+                for party_on_app in instance.baseapplication.enduser_parties
+            ]
+        return []
```
saruniitr, line 187: I think it is more readable if we flip the `if` condition such that we return empty list there (-ve case) and if it doesn't satisfy then return list of parties.

saruniitr (approved):
few minor comments

PR #1514, opened 2023-07-21
Remove GoodsQuery

kevincarrogan:
This is also sometimes referred to as CLC query

### Aim


api/cases/tests/test_case_search.py
```diff
@@ -220,7 +154,7 @@ def test_get_cases_filter_by_case_officer_not_assigned(self):
         When a user requests to view All Cases with no assigned case officer
         Then only cases without an assigned case officer are returned
         """
-        all_cases = self.application_cases + self.clc_cases
```
stuaxo, line 223: [not a blocker]

(same as above) - can probably remove this and replace the one usage of all_cases with `self.application_cases` itself.

- same may apply to all these places where `all_cases` is just one thing below.

api/cases/tests/test_sla.py
```diff
@@ -192,7 +191,7 @@ def test_sla_update_gifting_mod(
         self.assertEqual(case.sla_days, 1)
         self.assertEqual(case.sla_remaining_days, target - 1)
 
-    @parameterized.expand([(CaseTypeSubTypeEnum.GOODS,), (CaseTypeSubTypeEnum.EUA,)])
+    @parameterized.expand([(CaseTypeSubTypeEnum.EUA,)])
     def test_sla_doesnt_update_queries(self, query_type):
```
stuaxo, line 195: [not a blocker]

Now this only has one query, the `parameterized.expand` could probably go and just use the query directly in the test.

stuaxo:
Nice removal of complexity :)

Just a couple of non-blocker comments from me, LGTM - pre-emptively approving.

stuaxo (approved):
Pre-emptively approving, LGTM - A couple of small changes recommended.

PR #1517, opened 2023-07-25
LTD-3994-new-variables 

seijihg:
### Aim

New vars added appeal_deadline and date_application_submitted in the requested format. 
Also fixed a bug were variables were not working before.

Exporter reference (pulled from the LITE application) is {{ details.user_reference }} which already exists.

Added exporter_reference since details.user_reference  is not clear.

https://uktrade.atlassian.net/browse/LTD-3994

### DIFF
--- a/api/letter_templates/helpers.py
+++ b/api/letter_templates/helpers.py
@@ -1,5 +1,4 @@
 import os
-import datetime
 from django.template import Context, Template
 
 from django.conf import settings
@@ -67,33 +66,9 @@ def generate_preview(
 
     if case:
         context.update(get_document_context(case, additional_contact))
-        context.update(additional_context(case))
         context["user_content"] = format_user_text(convert_var_to_text(text, context))
 
     return render_to_string(template_name, context)
-
-
-def additional_context(case):
-    # The reason for not using get_document_context is because the file is not all covered by tests and codecov will fail
-    base_application = case.baseapplication if getattr(case, "baseapplication", "") else None
-
-    appeal_deadline = datetime.date.today() + datetime.timedelta(days=28)
-    exporter_reference = ""
-    date_application_submitted = ""
-
-    if base_application:
-        if base_application.name:
-            exporter_reference = base_application.name
-
-        if base_application.submitted_at:
-            date_application_submitted = base_application.submitted_at.strftime("%d %B %Y")
-
-    data = {
-        "appeal_deadline": appeal_deadline.strftime("%d %B %Y"),
-        "date_application_submitted": date_application_submitted,
-        "exporter_reference": exporter_reference,
-    }
-    return data
 
 
 def convert_var_to_text(text, data):
--- a/api/letter_templates/helpers.py
+++ b/api/letter_templates/helpers.py
@@ -1,6 +1,6 @@
 import os
 import datetime
-from jinja2 import Template
+from django.template import Context, Template
 
 from django.conf import settings
 from django.template.loader import render_to_string
@@ -9,7 +9,6 @@ from django.utils.html import (
     mark_safe,
 )
 from markdown import markdown
-from api.applications.models import BaseApplication
 
 from api.letter_templates.context_generator import get_document_context
 
@@ -69,7 +68,7 @@ def generate_preview(
 
     if case:
         context = {**context, **get_document_context(case, additional_contact)}
-        context.update(get_additional_var_data(case))
+        context.update(get_additional_var_data_for_template(case))
 
     context["user_content"] = convert_var_to_text(text, context)
 >>>>>>> acf4f0f2 (Adds unit tests)
@@ -91,21 +90,26 @@ def generate_preview(
     return render_to_string(template_name, context)
 
 
-def get_additional_var_data(case):
-    application = BaseApplication.objects.get(id=case.id)
+def get_additional_var_data_for_template(case):
+    # In the future if more variables are needed for Edit Text to generate PDFs. It can be added here
     today = datetime.date.today()
 
     appeal_deadline = today + datetime.timedelta(days=28)
-    date_application_submitted = application.submitted_at
+    exporter_reference = case.baseapplication.name
+    date_application_submitted = ""
+    if case.baseapplication.submitted_at:
+        date_application_submitted = case.baseapplication.submitted_at.strftime("%d %B %Y")
 
     data = {
         "appeal_deadline": appeal_deadline.strftime("%d %B %Y"),
-        "date_application_submitted": date_application_submitted.strftime("%d %B %Y"),
+        "date_application_submitted": date_application_submitted,
+        "exporter_reference": exporter_reference,
     }
     return data
 
 
 def convert_var_to_text(text, data):
     template = Template(text)
+    context = Context(data)
 
-    return template.render(data)
+    return template.render(context)
--- a/api/letter_templates/helpers.py
+++ b/api/letter_templates/helpers.py
@@ -1,4 +1,6 @@
 import os
+import datetime
+from jinja2 import Template
 
 from django.conf import settings
 from django.template.loader import render_to_string
@@ -7,6 +9,7 @@ from django.utils.html import (
     mark_safe,
 )
 from markdown import markdown
+from api.applications.models import BaseApplication
 
 from api.letter_templates.context_generator import get_document_context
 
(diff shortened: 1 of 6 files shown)


api/letter_templates/helpers.py
```diff
@@ -81,3 +87,23 @@ def generate_preview(
     context["css"] = css_string
 
     return render_to_string(template_name, context)
+
+
+def get_additional_var_data(case):
```
depsiatwal, line 92: can the naming be more specific here as it's not generic data

api/letter_templates/helpers.py
```diff
@@ -81,3 +87,23 @@ def generate_preview(
...
+    application = BaseApplication.objects.get(id=case.id)
```
depsiatwal, line 93: is this the most optimised way to get the application does case not have a related field name ?

api/letter_templates/tests/test_helpers.py
```diff
@@ -23,3 +25,34 @@ def test_document_layouts(self):
...
             # check it renders with no errors
             generate_preview(layout=layout, case=case, text="")
+
+    def test_document_variables(self):
+        case = self.create_standard_application_case(self.organisation, user=self.exporter_user)
+
+        generate_preview(
+            layout="refusal",
+            case=case,
+            text="Having carefully considered your application, {{ appeal_deadline }} {{ date_application_submitted }}",
```
depsiatwal, line 35: Is this being added to every preview the paragraphs for letters are held in the database  I suspect we only need to add if the users has asked when entering the text

api/letter_templates/helpers.py
```diff
@@ -1,9 +1,12 @@
 import bleach
 import os
+import datetime
+from jinja2 import Template
```
antroy-madetech, line 4: How come Jinja2 rather than django templates?

kevincarrogan: I'm not particularly keen on adding a different templating engine so it would be good to know why we can't use Django templates.

seijihg: ha just looked at it now and yea Django Template works. was too fixated on the Spike fix proposal which suggested Jinja :)

api/letter_templates/helpers.py
seijihg: I think the idea is whenever in the future they request more variables you can always add it here in this function

api/letter_templates/tests/test_helpers.py
seijihg: I am just testing if there will be any errors if there are variables.

api/letter_templates/helpers.py
```diff
@@ -69,8 +70,12 @@ def generate_preview(
         title = TemplateTitles.SIEL
 
     context = {"include_digital_signature": include_digital_signature, "user_content": text, "title": title}
+
     if case:
         context = {**context, **get_document_context(case, additional_contact)}
+        context.update(get_additional_var_data_for_template(case))
```
kevincarrogan, line 76: Why have we got a new function to get the vars when it seems like that's exactly what `get_document_context` was already doing?

api/letter_templates/helpers.py
```diff
@@ -81,3 +86,28 @@ def generate_preview(
     context["css"] = css_string
 
     return render_to_string(template_name, context)
+
+
+def get_additional_var_data_for_template(case):
+    # In the future if more variables are needed for Edit Text to generate PDFs. It can be added here
+    today = datetime.date.today()
+
+    appeal_deadline = today + datetime.timedelta(days=28)
```
kevincarrogan, line 95: This date is an interesting one. This will change depending on when they generate or potentially regenerate the licence. I wonder if this actually needs to be statically set somewhere else.

api/letter_templates/tests/test_helpers.py
```diff
@@ -23,3 +24,38 @@ def test_document_layouts(self):
...
+        )
```
kevincarrogan, line 35: There are no assertions in this test so what is this checking for?

api/letter_templates/tests/test_helpers.py
```diff
@@ -23,3 +24,38 @@ def test_document_layouts(self):
...
+        appeal_deadline = today + datetime.timedelta(days=28)
+        date_application_submitted = case.baseapplication.submitted_at
+        exporter_reference = case.baseapplication.name
+
+        data = get_additional_var_data_for_template(case)
+
+        assert data["appeal_deadline"] == appeal_deadline.strftime("%d %B %Y")
+        assert data["date_application_submitted"] == date_application_submitted.strftime("%d %B %Y")
+        assert data["exporter_reference"] == exporter_reference
+        assert len(data) == 3
```
kevincarrogan, line 52: If you want to check that the dictionary is exactly these three items then just check the whole dictionary is equal instead of doing three separate assertions and then a `len` check.

api/letter_templates/helpers.py
seijihg: ah interesting did not realise :O

seijihg:
I did not put it in the get_document_context() since the file is not covered by codecov like 1/3 of it. It will become a tech dept if I was to try to cover all the untested lines.

kevincarrogan (approved):

api/letter_templates/helpers.py
```diff
@@ -64,5 +66,37 @@ def generate_preview(
     }
     if case:
         context.update(get_document_context(case, additional_contact))
+        context.update(additional_context(case))
+        context["user_content"] = convert_var_to_text(text, context)
```
kevincarrogan, line 70: This overwrites the `user_content` from above which will lose the formatting provided previously.

api/letter_templates/helpers.py
```diff
@@ -64,10 +64,11 @@ def generate_preview(
         "user_content": format_user_text(text),
         "css": css_string,
     }
+
     if case:
         context.update(get_document_context(case, additional_contact))
         context.update(additional_context(case))
-        context["user_content"] = convert_var_to_text(text, context)
+        context["user_content"] = format_user_text(convert_var_to_text(text, context))
```
kevincarrogan: We should have tests covering this. The fact nothing failed is a sign that there's missing tests.

It would be good to see some tests that show the interaction between markdown formatted text and variables too.

kevincarrogan (approved):

api/letter_templates/helpers.py
```diff
@@ -62,7 +64,40 @@ def generate_preview(
...
     if case:
         context.update(get_document_context(case, additional_contact))
+        context.update(additional_context(case))
+        context["user_content"] = format_user_text(convert_var_to_text(text, context))
 
     return render_to_string(template_name, context)
+
+
+def additional_context(case):
+    # The reason for not using get_document_context is because the file is not all covered by tests and codecov will fail
```
currycoder, line 77: Did you see codecov fail for this?  I would expect if you were adding new lines to a function and those new lines were covered, it would be fine.

As a general point, I feel like codecov complaints should not inform how we structure the code - sometimes we are going to have to bite the bullet and improve coverage.

seijihg: yes I have seen it it was complaining about the lines I did not touch. it was like 1/3 file were the red lines :O

kevincarrogan: In theory it should only check against the diff and not check any lines that were untouched, but you could possibly inadvertently make the diff bigger depending on where you put the new code because it's shifting other lines around.

seijihg: maybe the black was reformatting things lemme try to put it back again there.

depsiatwal (approved):
lgtm

api/letter_templates/tests/test_helpers.py
```diff
@@ -71,3 +77,36 @@ def test_document_layouts(self):
...
+        case = self.create_standard_application_case(self.organisation, user=self.exporter_user)
+        generated_preview = generate_preview(
+            case=case,
+            layout="user_content",
+            text="""**{{ exporter_reference }}**
+*{{ exporter_reference }}*
+[{{ exporter_reference }}](http://example.com), {{ exporter_reference }}""",
+        )
+        with open(TEST_DATA_PATH / "markdown-variables.html") as expected_output_file:
+            expected_output = expected_output_file.read()
```
stuaxo, line 110: Not a blocker, but it may be worth using pathlib read_text here to avoid the need for `expected_output_file`,  along the lines of:
```python
expected_output = (TEST_DATA_PATH / "markdown-variables.html").read_text()
```

seijihg: more modern and concise approach i see :D

PR #1519, opened 2023-07-26
Remove some code related to unsupported application types

saruniitr:
## Change description

Currently we only support SIEL types but some of the helper functions account for supporting other licence types OIEL, F680, Exhibition clearance etc so if we change those function we need to maintain these additional types as well. This is unnecessary and can be removed.

In future if we add support for these licence types then we probably do it in a different way so removing the code where possible and associated tests as well.

Looking to do this in multiple stages instead of one big PR. All changes included in one commit so it is easy to revert in case if we need to do that.


api/applications/libraries/get_applications.py
```diff
@@ -17,77 +17,62 @@ def get_application(pk, organisation_id=None):
         kwargs["organisation_id"] = str(organisation_id)
 
     application_type = _get_application_type(pk)
+    if application_type != CaseTypeSubTypeEnum.STANDARD:
+        raise NotImplementedError(f"Unsupported application type: {application_type}")
```
saruniitr, line 21: main change is in this file, rest of the changes are removing associated tests

api/applications/tests/test_adding_countries.py
kevincarrogan, line 1: Would these tests still be applicable if we had just changed `create_draft_open_application` to `create_draft_standard_application`?

I ask this because when I did a similar thing with the `GoodsQuery` I found there were tests I had to change because they still related but they just happened to be using `create_goods_query` instead of `StandardApplication`.

api/applications/tests/test_consignee.py
```diff
@@ -127,36 +127,6 @@ def test_consignee_deleted_when_new_one_added(self, delete_s3_function):
         self.assertEqual(deleted_consignees.count(), 0)
         delete_s3_function.assert_not_called()
 
-    def test_set_consignee_on_open_draft_application_failure(self):
```
kevincarrogan, line 130: Could this still be applicable to `standard`?

api/applications/tests/test_contract_types_on_countries.py
kevincarrogan, line 1: Could this be applicable to `standard`?

api/applications/tests/test_copy_application.py
```diff
@@ -50,36 +43,6 @@ def test_copy_draft_standard_application_successful(self):
...
-        self.response = self.client.post(self.url, self.data, **self.exporter_headers)
-        self.response_data = self.response.json()["data"]
-
-        self.assertEqual(self.response.status_code, status.HTTP_201_CREATED)
-        self.assertNotEqual(self.response_data, self.original_application.id)
-
-        self.copied_application = StandardApplication.objects.get(id=self.response_data)
-
-        self._validate_standard_application()
-
```
kevincarrogan, line 82: This appears to relate to `StandardApplication` so why is this being removed?

api/cases/tests/test_get_case.py
```diff
@@ -129,28 +124,6 @@ def test_case_assignments(self):
...
-            "Small Arms",
-            "UK Military/Para: Sch 2",
-        ]
-        actual_flags_on_case = [flag["name"] for flag in response_data["case"]["all_flags"]]
-        actual_flags_on_goods_type = [flag["name"] for flag in response_data["case"]["data"]["goods_types"][0]["flags"]]
-
-        self.assertIn(actual_flags_on_case[0], expected_flags)
-        for expected_flag in expected_flags:
-            self.assertIn(expected_flag, actual_flags_on_goods_type)
-
```
kevincarrogan, line 153: Could this be applicable to `StandardApplication`?

saruniitr: No this is specific to Open type applications

api/applications/tests/test_copy_application.py
saruniitr: I will double check, `TradeControlProductCategory` is not related to standard type afaik

api/applications/tests/test_consignee.py
saruniitr: we should have a separate test for these as most of these are parameterized for various types, will check again

api/applications/tests/test_contract_types_on_countries.py
saruniitr: No, contract types are not applicable for standard

kevincarrogan (approved):

saruniitr:
Will come back to this later

PR #1520, opened 2023-07-26
Retain denial reasons after advice has been countersigned

kevincarrogan:
### Aim

Stop denial reasons from being deleted.

[LTD-4030](https://uktrade.atlassian.net/browse/LTD-4030)

[LTD-4030]: https://uktrade.atlassian.net/browse/LTD-4030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -486,6 +486,14 @@ class Advice(TimestampableModel):
 
         super(Advice, self).save(*args, **kwargs)
 
+        # We retain the denial reasons from the previous object because we're effectively really doing an edit here by
+        # removing the previous advice and create a new one in its place, however we lose the many-to-many relationship
+        # the old object had so we reinstate it here.
+        # This may look like it would overwrite the denial reasons that were set currently if you were editing this
+        # object but if you were editing the denial reasons you'd first have to save this object and _then_ set the
+        # denial reasons afterwards, so in that case this just gets overwritten anyway.
+        # We do this after the save method as this object needs to have its id in place so we can create these
+        # many-to-many objects.
         if denial_reasons:
             self.denial_reasons.set(denial_reasons)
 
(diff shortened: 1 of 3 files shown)


api/cases/models.py
```diff
@@ -486,21 +479,16 @@ def save(self, *args, **kwargs):
...
-                    )
+                denial_reasons = [denial_reason for denial_reason in old_advice.denial_reasons.all()]
                 old_advice.delete()
         except Advice.DoesNotExist:
             pass
 
         super(Advice, self).save(*args, **kwargs)
 
+        if denial_reasons:
+            self.denial_reasons.set(denial_reasons)
```
saruniitr, line 490: If the new instance has `denial_reasons` then they are overwritten by the old one.

kevincarrogan: Is there a real use case where this happens so I can write a test for this?

The problem here is that if I combine them there might be the case where someone is purposefully overwriting the denial reasons and combining them is wrong i.e. when someone is editing.

The biggest issue we've got at the moment is that we're just deleting them, which is what we need to resolve to stop this from happening.

saruniitr: Users can edit their recommendation and change the criteria.
I think we need tests covering both cases.
- FCDO user gives refusal recommendation with some criteria and submits. He then decides to edit their recommendation and changes criteria so in this case we only want to save the new criteria

api/cases/tests/test_create_advice.py
```diff
@@ -422,6 +427,35 @@ def test_countersign_advice_success(self):
         self.assertEqual(audit_qs.count(), 1)
         self.assertEqual(audit_qs.first().actor, self.gov_user)
 
+    def test_countersigning_retains_denial_reasons(self):
+        advice = Advice.objects.create(
+            user=self.gov_user,
+            case=self.case,
+            note="Advice",
```
saruniitr, line 434: this should actually be set to `text` field.
`note` is where additional notes to exporter field

saruniitr (approved):

depsiatwal (approved):
lgtm

PR #1522, opened 2023-07-27
LTD-4029-refusal-note

seijihg:
### Aim

Adds new is_refusal_note as boolean so we can distinct it between refusal reasons or note.

https://uktrade.atlassian.net/browse/LTD-4029

Related:
https://github.com/uktrade/lite-frontend/pull/1411


api/applications/serializers/advice.py
```diff
@@ -68,6 +68,7 @@ class AdviceViewSerializer(serializers.Serializer):
         queryset=GovUser.objects.all(), serializer=GovUserListSerializer
     )
     countersign_comments = serializers.CharField()
+    is_refusal_note = serializers.BooleanField()
```
saruniitr, line 71: Could you add some help text that explains the purpose of this field?
I am not able to understand it completely, what does it mean when this is `True` or `False`?

saruniitr: As explained internally I think it is best to use a separate text field to save refusal meeting notes instead of combining with an existing field (`text` in this case) and then use this boolean to determine whether the value in that field is either 'reasons for refusal' or 'meeting notes'

seijihg: I am a bit confused since reasons for refusal is being saved to `text` field in advice model. So essentially thats what we are doing since meeting note is being saved in the same place as `reasons for refusal` and `is_refusal_note` is telling us it is a note.

depsiatwal (approved):
lgtm

api/cases/migrations/0058_advice_is_refusal_note.py
```diff
@@ -0,0 +1,18 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cases', '0057_auto_20230505_1209'),
+    ]
+
+    operations = [
```
Tllew, line 12: do we need migration tests?

looks good otherwise

api/applications/serializers/advice.py
currycoder: So here's a thing..  We are using this field to characterise whether the `text` field is a refusal reason or a refusal meeting note, right?

To me, that is a little odd.  This is an `Advice` record which could be of `type` "approve" or "refuse", and then we have this  boolean value saved to the DB which will only be true when `type=refuse`, `team=lu`.  I feel like we are de-normalising the data here in a way that makes the overall DB record more messy; this flag only makes sense for a specific subset of the records and not all advice the records.

So, I wonder if this field could instead be a `@property` e.g.
```
    @property
    def is_refusal_note(self):
        return self.team == "lu" and self.type == "refuse"
```

This way, the value becomes a calculated field which will *only* be True under this circumstance.  Otherwise we are letting the frontend call the shots on when to set `is_refusal_note` and that may lead to data inconsistencies - e.g. setting `is_refusal_note=True` for approvals/for teams other than LU.

seijihg: @currycoder by doing the property we cant differentiate the old Advice Model being refusal_reasons or refusal_note. Colin still need to know which ones were refusal_reasons before this change is implemented. By adding the property I am not sure we have a record of what is a note and what is the old reasons.

api/cases/migrations/0058_advice_is_refusal_note.py
seijihg: this kind of test might be relatively slow, given that it involves applying and unapplying migrations. It is generally not necessary to test every migration, particularly simple ones IMO but I have added one this time 👍

api/cases/libraries/post_advice.py
```diff
@@ -149,6 +150,10 @@ def update_advice(request, case, level):
     data = request.data
     advice_ids = [item["id"] for item in data]
     advice_to_update = Advice.objects.filter(id__in=advice_ids)
+    # Preserving order as in advice_ids since the serializer AdviceUpdateListSerializer is based on the order to get things updated
+    preserved_order = Case(*[When(pk=pk, then=pos) for pos, pk in enumerate(advice_ids)], output_field=IntegerField())
```
kevincarrogan, line 154: Why is this a `Case` `When`?

saruniitr: `AdviceUpdateListSerializer` takes list of ids so order should not matter

seijihg: thats the thing it does not, just updated with the unit-test for this exact sceniario. Going to try to fix it in serializer instead inside the `View`

api/cases/migrations/0058_advice_is_refusal_note.py
stuaxo: For migrating forwards - I guess that our CI will just fail if the migration is really broken.

Missing backwards migrations is a problem it's easy to have, we it might be worth us having an integration test that enforces this (assuming we can already migrate all the way backwards and none are missing)

kevincarrogan (approved):

PR #1531, opened 2023-08-03
[LTD-4056] Endpoint to send generated case documents to exporters 

currycoder:
### Aim

Add API endpoint for sending generated case documents to exporters; the endpoint marks a document as `visible_to_exporter=True` and sends a notification if a notification would be applicable for the given template.  For now, only generated docs from the "inform_letter" template will result in a notification - but we can extend this as we use this endpoint for more documents in the future.

[LTD-4056](https://uktrade.atlassian.net/browse/LTD-4056)

[LTD-4056]: https://uktrade.atlassian.net/browse/LTD-4056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/generated_documents/urls.py
+++ b/api/cases/generated_documents/urls.py
@@ -7,5 +7,6 @@ app_name = "generated_documents"
 urlpatterns = [
     path("", views.GeneratedDocuments.as_view(), name="generated_documents"),
     path("<uuid:dpk>/", views.GeneratedDocument.as_view(), name="generated_document"),
+    path("<uuid:document_pk>/send/", views.GeneratedDocumentSend.as_view(), name="send_generated_document"),
     path("preview/", views.GeneratedDocumentPreview.as_view(), name="preview"),
 ]
--- a/api/cases/generated_documents/views.py
+++ b/api/cases/generated_documents/views.py
@@ -1,5 +1,6 @@
 from django.db import transaction
 from django.http import JsonResponse
+from django.shortcuts import get_object_or_404
 from django.utils import timezone
 
 from rest_framework import status, generics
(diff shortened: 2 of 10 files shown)


api/cases/generated_documents/urls.py
```diff
@@ -7,5 +7,6 @@
 urlpatterns = [
     path("", views.GeneratedDocuments.as_view(), name="generated_documents"),
     path("<uuid:dpk>/", views.GeneratedDocument.as_view(), name="generated_document"),
+    path("<uuid:document_pk>/send/", views.GeneratedDocumentSend.as_view(), name="send_generated_document"),
```
currycoder, line 10: Note that this is a kind of "action" endpoint rather than a PATCH endpoint (in order to change individual fields).  That is because there is a possible side effect (sending a notification).

seijihg (approved):

api/cases/generated_documents/urls.py
seijihg: still very confusing to use POST instead of PATCH though. tbh I dont understand why we cant use PATCH. What kind of side effect we are talking about? maybe would be good to have comments here for the future reference

depsiatwal (approved):
look spot on to me :-)

api/cases/generated_documents/urls.py
currycoder: This sort of explains what I'm getting at here; https://restful-api-design.readthedocs.io/en/latest/methods.html#actions

Broadly, it's not a PATCH - PATCH expresses that we are able to change one or more fields on a particular resource.  This is an action; we are doing a specific thing (sending to exporter) which will adjust the data AND result in a side effect (potentially sending them a notification).

api/cases/generated_documents/views.py
```diff
@@ -145,3 +147,25 @@ def get(self, request, pk):
...
+        document.visible_to_exporter = True
+        document.save()
+
+        layout_name = document.template.layout.filename
+
+        if NOTIFICATION_FUNCTIONS.get(layout_name):
+            NOTIFICATION_FUNCTIONS[layout_name](document.case)
+            return Response(data={"notification_sent": True})
+
+        return Response(data={"notification_sent": False})
```
saruniitr, line 171: minor - could you consider inverting the condition returning early with the response?

At this point is it is fine but if the body of `if` expands it may be easier to read.

saruniitr (approved):

PR #1534, opened 2023-08-09
[LTD-3975] Ensure countersign flags are removed when LU refuse a case

currycoder:
### Aim

TODO

[LTD-3975](https://uktrade.atlassian.net/browse/LTD-3975)


api/applications/views/helpers/advice.py
```diff
@@ -134,6 +123,28 @@ def ensure_lu_countersign_complete(application):
...
+    )
+
+
+def remove_countersign_process_flags(application):
+    """
+    Delete all flags from parties on a case which govern routing in the countersign process.
+    """
+    countersign_process_flags = get_countersign_process_flags()
+    for party_on_application in application.parties.all():
+        party_on_application.party.flags.remove(*countersign_process_flags)
```
saruniitr, line 145: we need an audit event here for the flag removal similar to the helper function

Tllew (approved):
Looks good to me

stuaxo (approved):

PR #1537, opened 2023-08-15
LTD-4018-Adds-note-timeline

seijihg:
### Aim

Adding additional audit functions for edit and create note.

https://uktrade.atlassian.net/browse/LTD-4018


api/audit_trail/enums.py
```diff
@@ -134,6 +134,8 @@ class AuditType(LiteEnum):
     LU_ADVICE = autostr()
     LU_EDIT_ADVICE = autostr()
     LU_COUNTERSIGN = autostr()
+    LU_EDIT_NOTE = autostr()
+    LU_CREATE_NOTE = autostr()
```
kevincarrogan, line 138: Could we name this more specifically so that it's obvious this is about a meeting note.

api/cases/libraries/post_advice.py
```diff
@@ -114,7 +114,12 @@ def post_advice(request, case, level, team=False):
             GeneratedCaseDocument.objects.filter(
                 case_id=case.id, advice_type__isnull=False, visible_to_exporter=False
             ).delete()
-            audit_lu_countersigning(AuditType.LU_ADVICE, data[0]["type"], data, case, request)
+
+            if data[0].get("is_refusal_note", False):
+                audit_lu_countersigning(AuditType.LU_CREATE_NOTE, data[0]["type"], data, case, request)
+            else:
+                audit_lu_countersigning(AuditType.LU_ADVICE, data[0]["type"], data, case, request)
```
kevincarrogan, line 121: What did `LU_ADVICE` used to say in the audit log? I feel that this isn't an either or thing maybe. Advice is a separate thing from the note creation isn't it? I feel we would still want to log that LU have given advice even if they've set a refusal meeting note.

saruniitr: Agree, I think we need to log refusal note in additional to refusal advice.
Please check with UCD if that is not the case.

api/audit_trail/enums.py
saruniitr: +1

api/audit_trail/formatters.py
```diff
@@ -303,3 +303,11 @@ def lu_countersign_advice(department, order, countersign_accepted, **payload):
...
         return f" {senior_text}countersigned all {department} recommendations."
     return f" declined to {senior_text}countersign {department} recommendations."
+
+
+def update_lu_note(advice_type, **payload):
+    return " edited their refusal meeting note."
+
+
+def create_lu_note(advice_type, **payload):
+    return " added a refusal meeting note."
```
saruniitr, line 313: similarly make the function names clear that these are related to refusal note

api/audit_trail/migrations/0017_alter_audit_verb.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.20 on 2023-08-15 08:51
+
+import api.audit_trail.enums
```
currycoder, line 3: Just a note but this will need re-generating (deleting and re-creating) after Kevin's comment above RE: renaming the audit type.

api/cases/libraries/post_advice.py
```diff
@@ -163,12 +168,22 @@ def update_advice(request, case, level):
     serializer.save()
 
     mark_lu_rejected_countersignatures_as_invalid(case, request.user)
-    audit_lu_countersigning(AuditType.LU_EDIT_ADVICE, advice_to_update.first().type, data, case, request)
+
+    if advice_to_update.first().is_refusal_note:
```
currycoder, line 172: How come we only deal with the first record?

api/cases/libraries/post_advice.py
seijihg: ah yea ur right lemme go through things again.

seijihg: Colin confirmed after abit of thoughts there will be two logs at the time of refusal meething note creation:
```
Licensing Unit: Le Ngo added a recommendation to refuse.
Licensing Unit: Le Ngo added a refusal meeting note.
```

api/audit_trail/formatters.py
seijihg: done :D

api/cases/libraries/post_advice.py
seijihg: Hi Advice is attached to the good as well so all goods will have the same advice.

api/cases/libraries/post_advice.py
```diff
@@ -173,6 +179,12 @@ def update_advice(request, case, level):
     serializer.save()
 
     mark_lu_rejected_countersignatures_as_invalid(case, request.user)
+
+    if advice_to_update.first().is_refusal_note:
+        audit_lu_countersigning(AuditType.LU_EDIT_MEETING_NOTE, advice_to_update.first().type, data, case, request)
+    else:
+        audit_lu_countersigning(AuditType.LU_EDIT_ADVICE, advice_to_update.first().type, data, case, request)
```
kevincarrogan, line 186: Isn't this the same as before where we actually want to say that they are editing both their meeting note _and_ advice and it's not that it's one or the other?

seijihg: no, this one is correct since both cant be happening at the same time.

kevincarrogan (approved):

Tllew (approved):
Looks good

PR #1540, opened 2023-08-17
Add application appeal endpoint

kevincarrogan:
### Aim

Adds an endpoint to make an appeal object.

[LTD-4116](https://uktrade.atlassian.net/browse/LTD-4116)


depsiatwal (approved):
finally good to see django API being used for it's purpose

api/applications/models.py
```diff
@@ -195,6 +196,8 @@ class BaseApplication(ApplicationPartyMixin, Case):
     agreed_to_foi = models.BooleanField(blank=True, default=None, null=True)
     foi_reason = models.TextField(blank=True, default="")
 
+    appeal = models.OneToOneField(Appeal, blank=True, null=True, on_delete=models.SET_NULL)
```
currycoder, line 199: So a single application can have at most one appeal.  Just out of interest, but has it been decided how a subsequent appeal might be dealt with?

api/appeals/models.py
```diff
@@ -0,0 +1,10 @@
+import uuid
+
+from django.db import models
+
+from api.common.models import TimestampableModel
+
+
+class Appeal(TimestampableModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    grounds_for_appeal = models.TextField()
```
currycoder, line 10: Do we need to record the exporter user that initiated the appeal?

api/applications/models.py
currycoder: And also to talk this through...  Why do the relation this way round, rather than having Appeal refer back to Application as a ForeignKey?

api/appeals/models.py
saruniitr: is it worth saving the Organisation/Exporter that started appeal?
and appeal result?

api/applications/models.py
kevincarrogan: In this case we're are making the decision that there will only be a single appeal. We won't be building any functionality around multiple appeals so I decided to build the data model to reflect that. If I link it with a foreign key the other way round then we're saying that there's a many-to-one relationship and there's not, it's one-to-one.

As for the relationship I went backwards and forwards on this.

The end rationale was the difference in accessing this property. This way round it was this:

```
appeal = application.appeal
print(appeal)
None
```

vs

```
appeal = application.appeal

Raises a Django exception
```

The former felt cleaner.

api/appeals/models.py
kevincarrogan: There's nothing in the designs that show this anywhere. We can add this as an additional ticket if it looks like it's valuable to capture. There will be additional fields added as we work through appeals, this is just to facilitate the bare minimum to create an appeal.

api/applications/models.py
currycoder: Thanks for explaining the rationale, that makes sense to me.

currycoder (approved):

PR #1541, opened 2023-08-18
Ltd 4117 appeal documents

kevincarrogan:
### Aim

Adds an endpoint to create appeal documents.

[LTD-4117](https://uktrade.atlassian.net/browse/LTD-4117)


api/appeals/views.py
```diff
@@ -0,0 +1,25 @@
...
+
+from django.http import Http404
+
+from api.core.authentication import ExporterAuthentication
+
+from .models import Appeal
+from .serializers import AppealDocumentSerializer
+
+
+class AppealDocuments(CreateAPIView):
```
currycoder, line 11: So just to understand, exporters upload some set of supporting documents for their appeal.  Is that right?

currycoder, line 11: Might we also want to list the documents in the future?  I guess for now this is keeping the scope limited.

seijihg (approved):

currycoder (approved):

api/appeals/views.py
kevincarrogan: That is correct. They will upload zero or more supporting documents for their appeal.

kevincarrogan: So far I've found I don't need to create a separate list endpoint but adding the appeal object to an existing endpoint seems to do the job.

PR #1544, opened 2023-08-21
LTD-4115: Add appeal_deadline field

saruniitr:
### Aim

Exporters can appeal on a refusal within a certain date. Add field to capture this deadline.
This is used to show/hide appeal options on Exporter dashboard

[LTD-4115](https://uktrade.atlassian.net/browse/LTD-4115)


api/letter_templates/context_generator.py
```diff
@@ -899,7 +901,7 @@ def get_document_context(case, addressee=None):
     if not addressee and case.submitted_by:
         addressee = case.submitted_by
 
-    appeal_deadline = datetime.date.today() + datetime.timedelta(days=28)
+    appeal_deadline = make_aware(datetime.now() + timedelta(days=APPEAL_DAYS))
```
kevincarrogan, line 904: Django has `timezone.now` saves you having to do the make aware stuff.

api/letter_templates/context_generator.py
```diff
@@ -910,6 +912,11 @@ def get_document_context(case, addressee=None):
         if base_application.submitted_at:
             date_application_submitted = base_application.submitted_at.strftime("%d %B %Y")
 
+        # Update appeal_deadline for Refusal Cases
+        if AdviceType.REFUSE in final_advice.values_list("type", flat=True):
+            base_application.appeal_deadline = appeal_deadline
+            base_application.save()
```
kevincarrogan, line 918: I'm not keen on this function having a side effect as it's not obvious to the caller that something like this is going to happen. Is there possibly another way of achieving this?

api/letter_templates/tests/test_helpers.py
```diff
@@ -110,3 +112,26 @@ def test_markdown_and_variables(self):
...
+        preview_output = generate_preview(layout=layout, case=case, text="")
+
+        application.refresh_from_db()
+        if layout == "refusal":
+            self.assertIsNotNone(application.appeal_deadline)
+        else:
+            self.assertIsNone(application.appeal_deadline)
+
+        # Additional check to make sure that we actually get some output even if there are no errors
+        assert preview_output
```
kevincarrogan, line 137: You can check for the actual value here if you wanted to the same as the above tests do.

api/letter_templates/tests/test_helpers.py
```diff
@@ -110,3 +112,26 @@ def test_markdown_and_variables(self):
...
+        for good_on_application in application.goods.all():
+            FinalAdviceFactory(user=self.gov_user, case=case, good=good_on_application.good, type=advice_type)
+
+        preview_output = generate_preview(layout=layout, case=case, text="")
+
+        application.refresh_from_db()
+        if layout == "refusal":
+            self.assertIsNotNone(application.appeal_deadline)
+        else:
+            self.assertIsNone(application.appeal_deadline)
```
kevincarrogan, line 134: Instead of an `if` statement in the tests this could be another parameter as part of the parametrized set.

api/letter_templates/context_generator.py
saruniitr: Moved to "publish to exporter" flow which is where Case status is set to finalised and various notifications sent to Exporter.
Appeal deadline is populated before we notify exporter

api/letter_templates/tests/test_helpers.py
saruniitr: removed this test as this is no longer required and updated existing tests to check for `appeal_deadline`

currycoder (approved):

seijihg (approved):

api/letter_templates/context_generator.py
```diff
@@ -899,7 +901,7 @@ def get_document_context(case, addressee=None):
     if not addressee and case.submitted_by:
         addressee = case.submitted_by
 
-    appeal_deadline = datetime.date.today() + datetime.timedelta(days=28)
+    appeal_deadline = timezone.localtime() + timedelta(days=APPEAL_DAYS)
```
kevincarrogan, line 904: If there's a gap between this letter being generated and the letter going out then this deadline date is going to be different to the one we're setting.

UCD have also said that the deadline should reset when the letter is generated, which we aren't doing.

This should be using the date that we set on the object itself but I don't think the is the function to do that saving.

saruniitr: actually this variable is not used in the template. It is passed in `context` but we don't display that in the letter.
I was thinking of removing this altogether but left as is.

Once letter is generated then user is going to publish to exporter, at which point we are resetting the deadline.
If we reset the deadline as soon as letter is generated (and not published) it means exporter is not notified and the timer already started but they are not aware which doesn't look right hence I moved to publishing flow where we finalise the case, set deadline and also notify exporter.

saruniitr: @kevincarrogan I have moved this to the view where the letter is created.
I think below is not the right place to reset deadline
https://github.com/uktrade/lite-api/blob/dev/api/letter_templates/context_generator.py#L902

This is part of the preview which is a multiform wizard in FE, when that is submitted we make another call to create the document.
I am currently resetting the deadline in that view and it is the same in case of regeneration also.

Added tests to check for deadline once refusal letter is generated.

seijihg: its passing in the context because the guys wanted to have is an option for them to put it in as a custom text in edit. so the 28days is from the date PDF is being generated? probably need to ask Jenna

PR #1545, opened 2023-08-21
 LTD-4129-inform-letter-status-decision-docs

depsiatwal:
Advice documents drive many of the document generation. 

Inform has been added to advice_type as it's required by login in the FE. 
With out this too many hacks are required for figure out if a document or template is of type inform


api/cases/libraries/finalise.py
```diff
@@ -19,6 +19,10 @@ def get_required_decision_document_types(case):
         required_decisions.add(AdviceType.APPROVE)
         required_decisions.remove(AdviceType.PROVISO)
 
+    # Ensure that REFUSE advice requires an inform document
+    if AdviceType.REFUSE in required_decisions:
+        required_decisions.add(AdviceType.INFORM)
```
currycoder, line 24: Do we want to put this behind the feature flag too?

currycoder (commented):
Looks good, just one comment

api/cases/libraries/finalise.py
depsiatwal: We can put we have in FF in FE for decision documents
also we will still need to block inform letter from the finalisation screen anyway so don't see it adding value. 

But having said that lets do it to be safer

seijihg (approved):

api/cases/libraries/finalise.py
depsiatwal: add with tests

currycoder (approved):

PR #1547, opened 2023-08-21
Ltd 4126 appeal details

kevincarrogan:
### Aim

Adds an endpoint to get the details of an appeal document.

Adds appeal to the standard application serializer.

[LTD-4126](https://uktrade.atlassian.net/browse/LTD-4126)

### DIFF
--- a/api/appeals/factories.py
+++ b/api/appeals/factories.py
@@ -1,8 +1,21 @@
 import factory
 
-from .models import Appeal
+from .models import (
+    Appeal,
+    AppealDocument,
+)
 
 
 class AppealFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = Appeal
+
+
+class AppealDocumentFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = AppealDocument
+
+    name = factory.Faker("file_name")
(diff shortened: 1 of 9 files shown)


api/appeals/factories.py
```diff
@@ -1,8 +1,21 @@
 import factory
 
-from .models import Appeal
+from .models import (
```
saruniitr, line 3: Can this file be moved to `tests` directory as it is only used in tests?

kevincarrogan: Done.

api/appeals/tests/factories.py
```diff
@@ -0,0 +1,21 @@
+import factory
+
+from ..models import (
```
currycoder, line 3: Not to say that this is invalid, but generally we tend to use absolute import paths.  Do we want to keep that consistent?

currycoder (approved):

api/appeals/tests/factories.py
kevincarrogan: We should use absolute imports when it's across Djangos apps but we should use relative imports when it's within a Django app is my personal feeling.

PR #1548, opened 2023-08-21
LTD-4050-Criterion-1

seijihg:
### Aim

Migration for criterion 1 which already exists in the database

https://uktrade.atlassian.net/browse/LTD-4050


api/staticdata/denial_reasons/migrations/tests/test_0003_criterion_1_to_true.py
```diff
@@ -0,0 +1,17 @@
...
+    migrate_to = ("denial_reasons", "0003_criterion_1_to_true")
+
+    def prepare(self):
+        DenialReason = self.old_state.apps.get_model("denial_reasons", "DenialReason")
+        DenialReason.objects.create(id=1, deprecated=True, display_value="1")
+
+    def test_criterion_1_to_true(self):
+        DenialReason = self.new_state.apps.get_model("denial_reasons", "DenialReason")
+        denial_reason = DenialReason.objects.get(id=1)
+        self.assertFalse(denial_reason.deprecated)
```
kevincarrogan, line 17: This is called `_to_true` and yet it checks that the value is `False`.

seijihg: ah sorry naming is wrong :P should be to_false haha

api/staticdata/denial_reasons/migrations/tests/test_0003_criterion_1_to_false.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+@pytest.mark.django_db()
+class TestCriterion1ToTrue(MigratorTestCase):
+    migrate_from = ("denial_reasons", "0002_denialreason_display_value")
+    migrate_to = ("denial_reasons", "0003_criterion_1_to_false")
+
+    def prepare(self):
+        DenialReason = self.old_state.apps.get_model("denial_reasons", "DenialReason")
+        DenialReason.objects.create(id=1, deprecated=True, display_value="1")
```
Tllew, line 12: Should deprecated here be false?

seijihg: no its correct because in the DB is True right now we want to switch to False since they are re-enabling it

kevincarrogan (approved):

PR #1551, opened 2023-08-24
LTD-4017-4151-refusal-criteria-audit

seijihg:
https://uktrade.atlassian.net/browse/LTD-4017

also combining this bug ticket into here:
https://uktrade.atlassian.net/browse/LTD-4151

Some duplicates on timeline for create/edit meeting refusal note.


api/audit_trail/enums.py
```diff
@@ -136,6 +136,7 @@ class AuditType(LiteEnum):
     LU_COUNTERSIGN = autostr()
     LU_EDIT_MEETING_NOTE = autostr()
     LU_CREATE_MEETING_NOTE = autostr()
+    CREATE_REFUSAL_CRITERIA = autostr()
```
saruniitr, line 139: We already add refusal advice so what is the need for a new event or how is this different?

seijihg: I think ur right :D could just reuse the other one

seijihg: no actually there is a need for a new one, since we need:

" added a recommendation to refusal." AND
a seperate one " added refusal criteria."

Tllew (approved):
Looks good to me

api/audit_trail/enums.py
saruniitr: When users give refusal recommendation then they also need to specify the criteria so " added a recommendation to refusal." implies that already.

We also have a separate event for refusal meeting note.

seijihg: ah ur right! I could just re-use. ok updated

seijihg: no this is actually no good:

You need a new event since its separated:

if its OGD refusals:

- 1 added a recommendation to
- 2 added refusal criteria

if its FINAL:

- 1 added a recommendation to
- 2 added refusal criteria
- 3 added a refusal meeting note.

we cant reuse event LU_ADVICE since then there is no way for formatter to know which message to use since both TYPES are REFUSE unless we combine it into one message:

like this:
Le Ngo added a recommendation to Refusal and added refusal criteria.

api/cases/tests/test_create_advice.py
```diff
@@ -653,6 +653,41 @@ def test_create_lu_refusal_note_has_audit(self):
...
+        }
+
+        response = self.client.post(self.final_case_url, **self.gov_headers, data=[data])
+
+        assert response.status_code == status.HTTP_201_CREATED
+
+        criteria_advice_audit = Audit.objects.filter(verb=AuditType.CREATE_REFUSAL_CRITERIA)
+
+        assert criteria_advice_audit.exists()
+        criteria_audit_obj = criteria_advice_audit.first()
```
saruniitr, line 685: Are we expecting more than one audit entries of this type?
I suggest check for exact count instead of taking the first one.

seijihg: no, we are expecting only one yes going to assert it to count 1

saruniitr (approved):
With the last comment addressed, it looks good to me

PR #1552, opened 2023-08-25
LTD-4142-update-test-to-confirm-finalisation-no-inform

depsiatwal:
This Updates the tests during refusal finalisation to ensure we can proceed with just a refusal letter

https://uktrade.atlassian.net/browse/LTD-4142

### DIFF
--- a/api/cases/tests/test_refuse_licence.py
+++ b/api/cases/tests/test_refuse_licence.py
@@ -41,7 +41,9 @@ class RefuseCaseTests(DataTestClient):
             self.assertTrue(document.visible_to_exporter)
 
         # Confirm finalisation can succeed with only a refuse doc
-        self.assertEqual(GeneratedCaseDocument.objects.filter(case=self.application.id)[0].advice_type, "refuse")
+        case_documents_query = GeneratedCaseDocument.objects.filter(case=self.application.id)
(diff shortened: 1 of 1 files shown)


api/cases/tests/test_refuse_licence.py
```diff
@@ -40,6 +40,8 @@ def test_refuse_standard_application_success(self, send_exporter_notifications_f
         for document in GeneratedCaseDocument.objects.filter(advice_type__isnull=False):
             self.assertTrue(document.visible_to_exporter)
 
+        # Confirm finalisation can succeed with only a refuse doc
+        self.assertEqual(GeneratedCaseDocument.objects.filter(case=self.application.id)[0].advice_type, "refuse")
```
kevincarrogan, line 44: This says "with only a refuse doc" but then the assertion doesn't check that there's _only_ a refuse doc, it just checks that the _one_ of the documents is a refusal.

depsiatwal: good point need to make sure only one exists

depsiatwal: done

kevincarrogan (approved):

Tllew (approved):
LGTM

PR #1560, opened 2023-08-31
Allow cases queues to be set by both gov and exporter users

kevincarrogan:
### Aim

Allow cases queues to be set by both gov and exporter users. **Note** this doesn't mean that an exporter can change this value themselves just that we allow calls on behalf of the exporter to make this change.

[LTD-4120](https://uktrade.atlassian.net/browse/LTD-4120)

[LTD-4120]: https://uktrade.atlassian.net/browse/LTD-4120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/views/views.py
+++ b/api/cases/views/views.py
@@ -160,7 +160,7 @@ class CaseDetailBasic(RetrieveAPIView):
 
 
 class SetQueues(APIView):
-    authentication_classes = (GovAuthentication,)
+    authentication_classes = (SharedAuthentication,)
 
     @transaction.atomic
     def put(self, request, pk):
--- a/api/cases/tests/test_case_queues.py
+++ b/api/cases/tests/test_case_queues.py
@@ -1,6 +1,7 @@
 import uuid
 
 from django.urls import reverse
+from parameterized import parameterized
 from rest_framework import status
 
 from api.audit_trail.models import Audit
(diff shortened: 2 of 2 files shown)


api/cases/views/views.py
```diff
@@ -160,7 +160,7 @@ class CaseDetailBasic(RetrieveAPIView):
 
 
 class SetQueues(APIView):
-    authentication_classes = (GovAuthentication,)
+    authentication_classes = (SharedAuthentication,)
```
currycoder, line 164: So really the net effect here is that the exporter **app** is able to call the "set queues" API.  It's up to that app to use this ability responsibly and sanitise inputs etc to avoid misuse.

I suppose that is OK, but this does feel a little conceptually odd that the exporter app (concerned with creating export applications) is now going to know a bit more about the workflow guts of lite.  It does feel like a bit of a details leak - though I understand the reasoning we arrived at for this.

depsiatwal: I also agree that some of the core UI type validation should be in FE. 

But surely API Authentication calls (a common rest pattern)  should still protect it self.

kevincarrogan: I think that works if we think this is an API that is going to be used by multiple consumers and we want to make sure that all consumers abide by the business logic that's encapsulated in our API. The truth is that's just not the case. The API exists for our frontend to function and this separation is a byproduct of completely separate decisions that aren't "this API will have multiple consumers".

This change doesn't say "this is a free for all and we can do anything", it's just saying "there are two trusted users of this API, the exporter and the caseworker application, and we trust that both of them will do the right thing". Why do I think that's ok? Because we are writing both of these applications as a singular distributed monolith, adding additional checks here doesn't give us anything.

kevincarrogan: Let's put it this way, if we had a single monolithic Django application I guarantee this view would have done:

```
case.queues.add(appeal_queue)
```

and absolutely no one would have batted an eyelid.

currycoder (approved):

PR #1561, opened 2023-08-31
 LTD-4014-n-and-t-letter-sent 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4014

Entry in N&T for when an inform letter is sent. 

Note that this currently only works for inform letter when the other letters come into play and we get the formal text then we can update the audit verbs. 

image


api/cases/generated_documents/views.py
```diff
@@ -169,6 +169,13 @@ def post(self, request, pk, document_pk):
         document.visible_to_exporter = True
         document.save()
 
+        audit_trail_service.create(
+            actor=request.user,
+            verb=AuditType.DECISION_LETTER_SENT,
+            target=document.case,
+            payload={"case_reference": document.case.reference_code, "decision": document.advice_type},
```
currycoder, line 176: Double checking, will `document.advice_type` always be something other than `None`?

api/audit_trail/formatters.py
```diff
@@ -285,6 +285,15 @@ def generate_decision_letter(**payload):
     return f"invalid decision {decision} for this event."
 
 
+def decision_letter_sent(**payload):
+    decision = payload["decision"]
+
+    if decision == AdviceType.INFORM:
+        return "sent an inform letter."
+
+    return f"invalid decision {decision} for this event."
```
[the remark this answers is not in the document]
currycoder: I agree with Kevin, this is not right.

I would prefer if we made the fallback `"sent a document for decision {decision}"` or just made it so that all of the advice types were covered + a sensible fallback.

api/cases/generated_documents/views.py
depsiatwal: covered all feasible possibilities and fall back to is raise an error.

depsiatwal: yeah with this DECISION_LETTER_SENT you would need to have a support advice type

kevincarrogan (approved):

currycoder (approved):

PR #1562, opened 2023-08-31
LTD-4156: Fix resetting of appeal deadline when documents are regenerated

saruniitr:
### Change description

The requirement is to reset appeal deadline every time refusal letter is generated.
When the document is created we use `advice_type` to identify a refusal and accordingly reset the deadline however this is not provided in case of regeneration hence it is not getting reset.
FE cannot provide this during regeneration hence updated API to check for template decisions and derive this.
It is possible to enquire template details from FE and provide `advice_type` but we will have to expose more details (such as decision types etc).

Refactor endpoint code to helper functions.

[LTD-4156](https://uktrade.atlassian.net/browse/LTD-4156)

[LTD-4156]: https://uktrade.atlassian.net/browse/LTD-4156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/generated_documents/helpers.py
```diff
@@ -82,3 +84,40 @@ def get_generated_document_data(request_params, pk, include_css=True):
...
+
+    # When generating documents during finalise this is provided
+    if advice_type:
+        return advice_type
+
+    decisions = template.decisions.values_list("name", flat=True)
+    if {AdviceType.APPROVE, AdviceType.PROVISO}.intersection(decisions):
+        return AdviceType.APPROVE
+
+    if {AdviceType.REFUSE}.intersection(decisions):
```
kevincarrogan, line 100: `if AdviceType.REFUSE in decision` might be more readable here as you're only looking for a single elements as opposed to above where you look for multiple.

api/cases/generated_documents/helpers.py
```diff
@@ -82,3 +84,40 @@ def get_generated_document_data(request_params, pk, include_css=True):
...
 
     return GeneratedDocumentPayload(case=case, template=template, document_html=document_html, text=text)
+
+
+def get_advice_type(request_data, template):
+    advice_type = request_data.get("advice_type")
+
+    # When generating documents during finalise this is provided
+    if advice_type:
+        return advice_type
```
kevincarrogan, line 94: I think this function should accept `advice_type` instead of `request_data` which would make it more flexible.

saruniitr: ok, as it is returning `advice_type` I thought it looks odd if we provide `advice_type`, I will rename the function to `get_decision_type()`

kevincarrogan (approved):

seijihg (approved):

PR #1565, opened 2023-08-31
LTD-3978-refusal-letter-format

seijihg:
This is formatting without changes to db description and display value for criterions

https://uktrade.atlassian.net/browse/LTD-3978

Criterions updates will be in the below ticket:

https://uktrade.atlassian.net/browse/LTD-3995

This ticket has been a bit confusing, need to update the automatic generated text in edit text as well since they have changed paragraph. <--- this part actually is being done by LITE team. We put base in.


depsiatwal (approved):
small comment else LGTM

assets/css/refusal.css
```diff
@@ -1,53 +1,90 @@
...
+
+.document .header .header-1 div.addressee {
+  margin-top: 50px;
+}
+
+.document .header .header-subtext * {
+  font-size: 3mm;
+}
+
+.document .header .header-subtext p {
```
currycoder, line 84: OK this looks like a stack of duplication from inform_letter.css.  Have you tried loading in the stylesheets from HTML?
e.g.
```
<link href="/media/examples/link-element-example.css" rel="stylesheet" />
```

If we could do that, we wouldn't need to pass in the CSS and need to have this duplication.

api/letter_templates/templates/letter_templates/refusal.html
```diff
@@ -1,22 +1,23 @@
-{% extends "letter_templates/ecju_base.html" %}
+{% extends "letter_templates/inform_letter_base.html" %}
```
currycoder, line 1: I don't think the refusal letter should inherit from something called "inform_letter_base".  In fact, the "inform_letter_base.html" template isn't specific to inform letter - maybe it can be renamed to be e.g. "dbt_letter_base.html"

currycoder (commented):
In addition to the comments inline, I think there's a big thing missing here from what is specified in the Jira ticket; the default editable text needs changing.  The text starts "Dear {{ addressee.name }}" and ends "Export Control Joint Unit" as per the attached .docx file on the Jira.

Here's an example of how we have done this before by creating a migration to update the appropriate PicklistItem; https://github.com/uktrade/lite-api/blob/dev/api/letter_templates/migrations/0007_inform_letter_template_update_paragraphs.py

assets/css/refusal.css
seijihg: resolved good idea did not think of this, but an empty refusal.css (`css_string = load_css(layout)`) is still needed (not needed but will be used in the future)

api/letter_templates/templates/letter_templates/refusal.html
seijihg: done changed

seijihg:
> In addition to the comments inline, I think there's a big thing missing here from what is specified in the Jira ticket; the default editable text needs changing. The text starts "Dear {{ addressee.name }}" and ends "Export Control Joint Unit" as per the attached .docx file on the Jira.
> 
> Here's an example of how we have done this before by creating a migration to update the appropriate PicklistItem; https://github.com/uktrade/lite-api/blob/dev/api/letter_templates/migrations/0007_inform_letter_template_update_paragraphs.py

like in the Slack chat I dont think we should update this with migration. This is something LITE team can edit/create by themselves.

by adding a paragraph in here:
https://internal.lite.service.uat.uktrade.digital/team/picklists/?type=letter_paragraph

then create/edit a NEW refusal letter here with a just added paragraph above:
https://internal.lite.service.uat.uktrade.digital/document-templates/

api/letter_templates/migrations/0008_refusal_letter_update.py
```diff
@@ -0,0 +1,29 @@
...
+
+    pick_list_item = PicklistItem.objects.filter(name="Refusal letter content").first()
+    admin = Team.objects.get(name="Admin")
+
+    with open("lite_content/lite_api/letter_paragraphs/refusal_letter.txt", "r", encoding="utf_8") as f:
+        text = f.read()
+
+        pick_list_item, created = PicklistItem.objects.get_or_create(
+            team=admin, name="Refusal letter content", text=text, type=PicklistType.LETTER_PARAGRAPH
+        )
```
kevincarrogan, line 17: I think you could use `update_or_create` here.

api/letter_templates/migrations/0008_refusal_letter_update.py
```diff
@@ -0,0 +1,29 @@
...
+
+        if not created:
+            pick_list_item.text = text
+            pick_list_item.save()
```
kevincarrogan, line 21: I don't see any tests around this specific condition here.

kevincarrogan (approved):

PR #1569, opened 2023-09-04
LTD-4124: Add management command to remove Refusal advice given by LU

saruniitr:
## Change description

In the situation where an Appeal is overturned we need to prepare the Case for LU to issue a licence. This can only be done after removing the initial refusal advice given by LU.

Add management command that removes refusal advice. Once this command is run then LU can finalise the Case again and issue licence.

It also moves the case to 'LU Post-Circulation Cases to finalise' queue and removes any countersigning flags.

A dry run option is included to test before we can commit changes.

[LTD-4124](https://uktrade.atlassian.net/browse/LTD-4124)

[LTD-4124]: https://uktrade.atlassian.net/browse/LTD-4124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/support/management/commands/remove_refusal_advice.py
```diff
@@ -0,0 +1,74 @@
...
+        parser.add_argument(
+            "--case_reference",
+            type=str,
+            nargs="?",
+            help="Reference code of the Case (eg GBSIEL/2023/0000001/P)",
+        )
+        parser.add_argument("--dry_run", help="Is it a test run?", action="store_true")
+
+    def handle(self, *args, **options):
+        breakpoint()
```
seijihg, line 40: breakpoint()

seijihg (approved):

api/support/management/commands/remove_refusal_advice.py
```diff
@@ -0,0 +1,73 @@
...
+            # Ensure final advice given by LU is refusal
+            final_advice = case.advice.filter(level=AdviceLevel.FINAL, type=AdviceType.REFUSE)
+            if not final_advice.exists():
+                logger.error("Invalid Advice data, LU not refused the Case")
+                return
+
+            if not dry_run:
+                # Move the Case to 'LU Post-Circulation Cases to Finalise' queue
+                # as it needs to be in this queue to finalise and issue
+                case.queues.set(Queue.objects.filter(id=QueuesEnum.LU_POST_CIRC))
```
hnryjmes, line 59: for some reason I thought that we would be using the LU Appeals queue for the whole of the appeals journey, is that not right?

hnryjmes: or is LU Appeals queue set in addition to the LU Post-Circulation Cases to Finalise queue maybe?

saruniitr: I think initially when appeal is submitted it will be on the Appeals queue and Appeals team adds relevant OGD queues etc.
They all look at the Case again and move it forward (which removes their queue) and we are left with Appeals queue at the end.
Currently to finalise it needs to be on the post-circ queue. We need to make code changes if we were to allow finalise from Appeals queue. If this a requirement then we will have to update this. This will be answered once the steps I am documenting are reviewed.

hnryjmes (approved):

PR #1574, opened 2023-09-07
LTD-4178: Stop marking Licence as exhausted when products are completely used

saruniitr:
### Change description

HMRC sends usage data when Exporter ships products that are licenced. LITE keeps track of the usage and when it reaches the allowed quantity we currently mark the licence as exhausted and inform HMRC to cancel the licence.

Because of an issue in how we update usage we sometimes double count and cancel the licence prematurely. Also HMRC is source of truth and they won't allow Exporters to ship more than the quantity they are licenced for so LITE should not really be doing this.

[LTD-4178](https://uktrade.atlassian.net/browse/LTD-4178)

[LTD-4178]: https://uktrade.atlassian.net/browse/LTD-4178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/licences/libraries/hmrc_integration_operations.py
```diff
@@ -209,15 +208,6 @@ def _update_licence(validated_data: dict) -> str:
...
 
-    if (
-        action != HMRCIntegrationActionEnum.EXHAUST
-        and licence.case.case_type_id in CaseTypeEnum.STANDARD_LICENCE_IDS + CaseTypeEnum.MOD_LICENCE_IDS
-    ):
-        # If all Goods have been Exhausted; Exhaust the Licence
-        if not licence.goods.filter(usage__lt=F("quantity")).exists():
-            send_status_change_to_hmrc = action == HMRCIntegrationActionEnum.OPEN
-            change_status = licence.exhaust
-
```
kevincarrogan, line 220: I see other code paths other than this that could still mean that we send the exhausted message to HMRC and I'm keen to make sure we completely stop that from happening.

e.g. there's still

```
    if action == HMRCIntegrationActionEnum.EXHAUST:
        change_status = licence.exhaust
```

above.

Is there any way to change this more fundamentally so that we just never send HMRC anything if the licence is set to expired on our side not just in this particular case?

api/licences/tests/test_hmrc_integration_to_api.py
```diff
@@ -315,13 +315,6 @@ def test_update_usages_all_goods_exhausted_on_licence(self, create_licence):
         self.assertEqual(response.json()["licences"]["rejected"], [])
         self.assertEqual(licence.goods.first().usage, original_usage + usage_data)
         self.assertTrue(HMRCIntegrationUsageData.objects.filter(id=usage_data_id, licences=licence).exists())
-        self.assertEqual(licence.status, LicenceStatus.EXHAUSTED)
```
currycoder, line 318: Do we want to still assert that the status is non-exhausted?

api/licences/tests/test_hmrc_integration_to_api.py
```diff
@@ -431,16 +424,6 @@ def test_update_usages_all_goods_exhausted_when_action_is_open_does_inform_hmrc_
...
```
currycoder, line 434: Do we want to still assert that the status is non-exhausted?

currycoder (approved):

api/licences/libraries/hmrc_integration_operations.py
saruniitr: I wanted to remove this as well but this is coming from HMRC itself when they send usage data so retained.

We could just remove this path and nothing gets sent as `change_status` will remain `None`.
is that ok?

api/licences/tests/test_hmrc_integration_to_api.py
saruniitr: hmm, good point, I will update

api/licences/libraries/hmrc_integration_operations.py
kevincarrogan: But wouldn't that mean we'd get an exhaust value from HMRC and then we'd go and send a cancel response straight back?

kevincarrogan: I guess the real bit that's making me nervous is that there's still a codepath lurking around that means that we set a licence to exhausted and then it will still go and cancel the licence on HMRC's side.

https://github.com/uktrade/lite-api/blob/dev/api/licences/enums.py#L51-L60

It would be good to break that connection completely.

saruniitr: ok, I will look for all places where can potentially cancel/exhaust and remove them completely

PR #1578, opened 2023-09-08
Ltd 4170 sub status UI

kevincarrogan:
### Aim

Add an endpoint to allow sub-status to be set on an application.

[LTD-4170](https://uktrade.atlassian.net/browse/LTD-4170)

[LTD-4170]: https://uktrade.atlassian.net/browse/LTD-4170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/applications/tests/test_application_sub_status.py
```diff
@@ -0,0 +1,142 @@
...
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIsNone(self.standard_application.sub_status)
+
+    def test_set_sub_status_of_different_parent_status(self):
+        self.assertIsNone(self.standard_application.sub_status)
+
+        CaseSubStatus.objects.create(
+            name="test_sub_status",
+            parent_status=self.standard_application.status,
```
currycoder, line 72: Could it be worth making this an explicit status for readability rather than whatever `DataTestClient.standard_application` chooses?

currycoder (approved):

api/applications/tests/test_application_sub_status.py
kevincarrogan: Can do, we know what status it is because the call to `self.submit_application(self.standard_application)` will set it explicitly to `CaseStatusEnum.SUBMITTED`, which is kind of what I'm relying on here.

saruniitr (approved):

PR #1588, opened 2023-09-11
 LTD-4174-add-substatus-filter

depsiatwal:
This adds the filter options on the UI and goes together with 

https://github.com/uktrade/lite-frontend/pull/1502

UI changes are here :
https://github.com/uktrade/lite-frontend/pull/1502


api/cases/views/search/service.py
```diff
@@ -27,6 +27,11 @@ def get_case_status_list() -> List[Dict]:
     return CaseStatusEnum.as_list()
 
 
+def get_case_sub_status_list() -> List[Dict]:
+    # Once data is populated this should return sub_statues
+    return []
+
+
```
kevincarrogan, line 34: We already have the model to get this so we don't need to stub it out we can already implement it.

kevincarrogan (approved):

api/cases/managers.py
```diff
@@ -307,7 +311,9 @@ def search(  # noqa
 
         case_qs = (
             self.submitted()
-            .select_related("status", "case_type", "case_officer", "case_officer__baseuser_ptr", "baseapplication")
+            .select_related(
+                "sub_status", "status", "case_type", "case_officer", "case_officer__baseuser_ptr", "baseapplication"
```
currycoder, line 315: :ok_hand:

api/cases/views/search/service.py
```diff
@@ -21,12 +21,17 @@
...
 from api.users.models import BaseUser, GovUser
+from api.staticdata.statuses.models import CaseSubStatus
 
 
 def get_case_status_list() -> List[Dict]:
     return CaseStatusEnum.as_list()
 
 
+def get_case_sub_status_list() -> List[Dict]:
+    return [{"value": key, "key": key} for key in CaseSubStatus.objects.all().values_list("name", flat=True)]
```
currycoder, line 32: Would suggest that the key for the choices is the ID, rather than the name

api/cases/managers.py
```diff
@@ -81,6 +81,9 @@ def not_terminal(self):
     def has_status(self, status):
         return self.filter(status__status=status)
 
+    def has_sub_status(self, sub_status):
+        return self.filter(sub_status__name=sub_status)
```
currycoder, line 85: As per the other comment, if the choice key is the ID this can be changed to filtering on ID instead of name string.

api/cases/views/search/service.py
depsiatwal: It was ID originally but then looked odd as status is also based on 'name' or test rather then the id. 
so it was changed to name as the url looks cleaner (IMO) and consistent with  the status and other searches.

currycoder (approved):

PR #1590, opened 2023-09-12
[LTD-4172] Set refusals sub statuses

currycoder:
### Aim

This change ensures that case sub status is set as follows:
- "Inform letter sent" - when an inform letter is sent and the case status is "Under final review".
- "Approved" - when the case is finalised: approved.
- "Refused" - when the case is finalised: refused.

These sub statuses are set as the result of calling specific API endpoints

[LTD-4172](https://uktrade.atlassian.net/browse/LTD-4172)

[LTD-4172]: https://uktrade.atlassian.net/browse/LTD-4172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/generated_documents/views.py
```diff
@@ -179,6 +181,12 @@ def post(self, request, pk, document_pk):
 
         layout_name = document.template.layout.filename
 
+        if layout_name == "inform_letter":
+            try:
+                document.case.set_sub_status(CaseSubStatusIdEnum.UNDER_FINAL_REVIEW__INFORM_LETTER_SENT)
+            except BadSubStatus:
+                pass  # Sub-status cannot be set.. This is only a side effect so do not raise the error
```
kevincarrogan, line 188: Would a logger be useful here just in case this does happen we could at least then track it back somehow?

api/cases/tests/test_models.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class CaseTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        application = self.create_draft_standard_application(self.organisation)
+        self.case = self.submit_application(application)
+
+    def test_set_sub_status_invalid(self):
+        self.assertRaises(BadSubStatus, self.case.set_sub_status, CaseSubStatusIdEnum.FINALISED__APPROVED)
```
kevincarrogan, line 16: I didn't know you could do assertRaises like this. I tend to use the `with` variant.

kevincarrogan (approved):

depsiatwal (approved):
lgtm

api/cases/generated_documents/views.py
kevincarrogan: Ok. If we're considering this normal then don't add the logger.

PR #1598, opened 2023-09-15
LTD-4213: Allow LU users to set finalised status

saruniitr:
### Change description

To support Appeals process we need LU users to be able to set the status to 'Finalised' at the end of an appeal. Currently this cannot be done so update endpoint to allow only LU users with necessary permissions can change status to 'Finalised'

Using this opportunity to refactor some of the checks into a helper function.

Add test that checks only LU can set status to finalised and fails for other internal users.

[LTD-4213](https://uktrade.atlassian.net/browse/LTD-4213)

### DIFF
--- a/api/applications/libraries/application_helpers.py
+++ b/api/applications/libraries/application_helpers.py
@@ -1,11 +1,20 @@
+from django.http import JsonResponse
+
+from rest_framework import status
+from rest_framework.request import Request
+from rest_framework.exceptions import PermissionDenied
+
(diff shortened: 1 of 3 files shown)


Tllew (approved):
Looks good to me, code cov isn't happy though

hnryjmes:
I changed the ticket to reference LTD-4203 as this looks like the correct ticket for this work

hnryjmes (approved):

depsiatwal (commented):
LGTM just lacking code cov

PR #1614, opened 2023-09-25
Ltd-4222-emit-substatus

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4222

N&T update for sub status update. This includes system update and user update.


kevincarrogan:
There are three places that sub-statuses are set and I think this only covers two:

  - When a user explicitly sets the sub-status (covered)
  - When the sub-status is reset on a status change (covered)
  - When the sub-status is set by the system on other changes e.g. on refusal (not covered)

api/applications/tests/test_application_sub_status.py
```diff
@@ -9,15 +12,15 @@
...
+from api.staticdata.statuses.models import CaseStatus, CaseSubStatus
 
 
 class ApplicationManageStatusTests(DataTestClient):
     def setUp(self):
         super().setUp()
         self.standard_application = self.create_draft_standard_application(self.organisation)
         self.submit_application(self.standard_application)
-        self.status = CaseStatusFactory(status="test_status")
+        self.status = CaseStatus.objects.get(status="under_final_review")
```
kevincarrogan, line 23: Is there a specific reason why we can't use the factory here? It's done like this so that it's less coupled to any specific status that might change in the future.

api/applications/tests/test_application_sub_status.py
```diff
@@ -41,6 +44,37 @@ def test_gov_set_application_sub_status(self):
...
+        self.assertEqual(Audit.objects.all().count(), 3)
+
+        audit = Audit.objects.all().first()
+        self.assertEqual(AuditType(audit.verb), AuditType.UPDATED_SUB_STATUS)
+        self.assertEqual(
+            audit.payload,
+            {"status": "Under final review", "sub_status": "test_sub_status"},
+        )
+        audit_text = AuditSerializer(audit).data["text"]
+        self.assertEqual(audit_text, "updated the status to Under final review - test_sub_status")
```
kevincarrogan, line 57: Checks for three audit events but we then only check the output of one?

api/cases/models.py
```diff
@@ -133,16 +133,22 @@ def save(self, *args, **kwargs):
         super(Case, self).save(*args, **kwargs)
 
     def _reset_sub_status_on_status_change(self):
+        from api.audit_trail import service as audit_trail_service
```
kevincarrogan, line 136: Does this need to be imported in this method?

depsiatwal:
> There are three places that sub-statuses are set and I think this only covers two:
> 
> * When a user explicitly sets the sub-status (covered)
> * When the sub-status is reset on a status change (covered)
> * When the sub-status is set by the system on other changes e.g. on refusal (not covered)

wasn't sure if it existed  found it

api/cases/models.py
depsiatwal: Unfortunately this is done though out the whole model file.  This is because of a circular dependency between the classes. Hence the why the import is being done at the time of use. 

I think it may be neater to wrap this in a helper function which does the import. perhaps I should create a follow up ticket  under support/bau?

api/applications/tests/test_application_sub_status.py
depsiatwal: Not ideal but since we get the CaseStatus text using

kevincarrogan (commented):
Why have so many tests changed in regards to the number of audit events being sent? I wouldn't have presumed that these sub-status audit events would be being sent that frequently as it's only when a sub-status changes and I can't see where a sub-status is changing in some of these tests.

All these additional audit changes are caused by status changes on the underlying case. i.e Draft to final for example. 

The culprit is this change 
https://github.com/uktrade/lite-api/blob/ee2edb423ad09d07762427e9179dc072b9bfd7fa/api/cases/models.py#L145

Where the sub_status is set to None  when the status changes. I've noticed in the majority of the cases sub_status is None anyway so an additional check here would help where we only update the sub_status if it's non None.

api/queries/goods_query/tests/test_goods_queries.py
```diff
@@ -455,7 +455,7 @@ def test_user_cannot_respond_to_pv_grading_with_case_in_terminal_state(self):
 
         response = self.client.put(self.url, self.data, **self.gov_headers)
 
-        self.assertEqual(Audit.objects.count(), 0)
+        self.assertEqual(Audit.objects.count(), 1)
```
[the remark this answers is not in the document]
depsiatwal: These have been minimised where we can. 
We now only change the status to None when required i.e it's not None

api/applications/tests/test_application_sub_status.py
kevincarrogan: Ok. It seems like we're doing this because we later put this through the `AuditSerializer` which I'm not certain is really that necessary. We should just be able to check that the audit object has been created.

api/cases/models.py
```diff
@@ -133,16 +133,22 @@ def save(self, *args, **kwargs):
...
         try:
             case = Case.objects.get(id=self.pk)
         except Case.DoesNotExist:
             return  # If our case record does not yet exist in the DB, return early
         old_status = case.status
         status_changed = old_status != self.status
-
-        if status_changed:
+        if status_changed and self.sub_status:
             self.sub_status = None  # Reset sub-status on any status change
```
kevincarrogan, line 146: The comment on this line isn't valid anymore.

api/cases/models.py
```diff
@@ -133,16 +133,22 @@ def save(self, *args, **kwargs):
...
+            audit_trail_service.create_system_user_audit(
+                verb=AuditType.UPDATED_SUB_STATUS,
+                target=case,
+                payload={"sub_status": "", "status": CaseStatusEnum.get_text(self.status.status)},
```
kevincarrogan, line 150: I feel `"sub_status": None` might be more appropriate.

api/applications/tests/test_application_sub_status.py
```diff
@@ -41,6 +44,31 @@ def test_gov_set_application_sub_status(self):
...
+        self.standard_application.refresh_from_db()
+
+        # Check add audit
+        audit = Audit.objects.filter(verb=AuditType.UPDATED_SUB_STATUS).order_by("-created_at")[0]
+        self.assertEqual(
+            audit.payload,
+            {"status": "Under final review", "sub_status": "none"},
+        )
+        audit_text = AuditSerializer(audit).data["text"]
+        self.assertEqual(audit_text, "updated the status to Under final review - none")
```
kevincarrogan, line 70: This is checking for the string of `"none"` but I don't think that's really what we should be doing, it's no different from the test above, did you mean to set this to an empty string or maybe the `None` object? I don't need that we're doing anything different or special with the string `"none"`.

api/applications/tests/test_application_sub_status.py
depsiatwal: yeah last time there was some comments around using AuditSerializer to be used in tests to check the text. i've removed for now so we just check audit.

api/audit_trail/formatters.py
```diff
@@ -190,6 +190,13 @@ def get_updated_status(**payload):
     return format_text(strings.Audit.UPDATED_STATUS, **payload)
 
 
+def get_updated_sub_status(**payload):
+    sub_status = payload.get("sub_status", "")
```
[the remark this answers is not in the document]
depsiatwal: updated

api/cases/models.py
depsiatwal: agree removed

api/cases/models.py
depsiatwal: updated

Tllew (approved):
Looks good to me

api/audit_trail/formatters.py
```diff
@@ -190,6 +190,13 @@ def get_updated_status(**payload):
     return format_text(strings.Audit.UPDATED_STATUS, **payload)
 
 
+def get_updated_sub_status(**payload):
+    sub_status = payload.get("sub_status")
+    sub_status_text = "" if sub_status == "" else f" - {sub_status}"
+    status = payload.get("status")
```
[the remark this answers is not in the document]
depsiatwal: removed the get 
the blank is set by UI when we reset the status to 'None' hense why we have this condition.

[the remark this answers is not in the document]
kevincarrogan: It's not set to a blank string at any point.

```
        if not sub_status:
            sub_status = "none"
        else:
            sub_status = CaseSubStatus.objects.get(id=sub_status).name
        # Update the model
        audit_trail_service.create(
            actor=request.user,
            verb=AuditType.UPDATED_SUB_STATUS,
            target=case,
            payload={"sub_status": sub_status, "status": CaseStatusEnum.get_text(case.status.status)},
        )
```

Here `sub_status` is always a string `"none"` if it doesn't have a value and `status` is always set to a string.

```
            audit_trail_service.create_system_user_audit(
                verb=AuditType.UPDATED_SUB_STATUS,
                target=case,
                payload={"sub_status": None, "status": CaseStatusEnum.get_text(self.status.status)},
            )
```

Here `sub_status` is set to `None` but not a blank string and again `status` is always a string.

```
        audit_trail_service.create_system_user_audit(
            verb=AuditType.UPDATED_SUB_STATUS,
            target=case,
            payload={"sub_status": self.sub_status.name, "status": CaseStatusEnum.get_text(self.status.status)},
        )
```

Here `sub_status is set to a string and `status` is always set to a string.

There's no case where `sub_status` is ever set to an empty string.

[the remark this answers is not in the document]
kevincarrogan: Additonally,`status` is always in the payload so theres' no need to do `status = payload.get("status")`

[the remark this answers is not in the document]
depsiatwal: updated

api/applications/tests/test_application_sub_status.py
saruniitr: I think checking with `AuditSerializer()` is helpful in some cases as we can check how it looks in N&T which UCD checks in PR.

api/applications/tests/test_application_sub_status.py
saruniitr: From the ticket, I think if sub-status is removed then we just show the status without appending anything.

api/letter_templates/tests/test_view.py
```diff
@@ -50,6 +50,22 @@ def test_get_letter_templates_success_inform_letter_feature_enabled(self):
...
+    def test_get_letter_templates_success_inform_letter_feature_disabled(self):
+        url = reverse("letter_templates:letter_templates")
+        letter_template = LetterTemplate.objects.create(
+            name="Inform letter",
+            layout=LetterLayout.objects.first(),
+            visible_to_exporter=False,
+            include_digital_signature=False,
+        )
+        letter_template.decisions.set([AdviceType.ids[AdviceType.REFUSE]])
+        letter_template.case_types.set([CaseTypeEnum.SICL.id])
```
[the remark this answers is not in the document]
depsiatwal: updated

api/applications/tests/test_application_sub_status.py
kevincarrogan: I think we should have tests for `AuditSerializer` I'm just not certain that this is the right place for it.

kevincarrogan (approved):

PR #1619, opened 2023-09-26
LTD-4232-add-draft-rejection-substatus 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4231
Add a new sub status to under appeal status


Tllew (approved):

api/staticdata/statuses/migrations/0010_add_draft_rejection_sub_status.py
```diff
@@ -0,0 +1,26 @@
...
+        id=SUB_STATUS__UNDER_APPEAL__DRAFT_REJECTION_LETTER,
+        name="Draft rejection letter",
+        parent_status_id=CASE_STATUS_UNDER_APPEAL,
+    )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("statuses", "0009_update_request_received_name"),
+        ("lite_routing", "0001_initial"),
```
currycoder, line 21: Assume this was copied from a previous migration?  This one won't need to depend on lite_routing as a previous statuses migration already depends on lite_routing; https://github.com/uktrade/lite-api/blob/dev/api/staticdata/statuses/migrations/0007_add_case_sub_status_records.py#L93

kevincarrogan: If this is copied then we should really stop that as a thing we do as that's going to mean we end up copying the wrong thing at some point.

`./manage.py makemigrations --empty ...` is what we want to be doing.

kevincarrogan (approved):

api/staticdata/statuses/migrations/0010_add_draft_rejection_sub_status.py
depsiatwal: fixed

PR #1629, opened 2023-10-04
Use circleci parallelism for CI test suites

currycoder:
### Aim

Improve speed of test runs by using circleci's parallelism features.

This is achieved by adding `parallelism: n` to various test jobs in the circleci config file.  The API test suite is run with 10 workers, routing BDD/routing test suites are run with 5 workers each.

This means that for non-e2e test jobs, the time to run our suite has decreased from ~60min to ~12min.

Various additional fixes have been made to ensure that our suites can be run across parallel workers - mainly due to changes in test ordering and fixing bad test isolation.  In addition, most django migration tests have been removed as these were extremely slow and provide little value (as the migrations have already run in production).

**Note:** Depends on https://github.com/uktrade/lite-routing/pull/228 merging first (and lite-routing sha update)


api/common/dates.py
```diff
@@ -89,13 +90,15 @@ def get_bank_holidays(call_api=True, data=[]):  # noqa
         return get_backup_bank_holidays()
 
     r = requests.get(BANK_HOLIDAY_API)
+    print(r.status_code)
```
saruniitr, line 93: leftover debug statements? one more below

saruniitr (approved):

api/common/dates.py
currycoder: whoops - great spot Arun, thanks

Pipfile
```diff
@@ -71,6 +71,7 @@ redis = "~=4.4.4"
 psycopg2-binary = "~=2.9.3"
 django-test-migrations = "~=1.2.0"
 django-silk = "~=5.0.3"
+pytest-circleci-parallelized = "~=0.1.0"
```
kevincarrogan, line 74: This should probably be in the `dev-packages` set of requirements as opposed to here. It seems to have moved `pytest` with it to the distributed packages in the lockfile too.

kevincarrogan (approved):

PR #1631, opened 2023-10-05
Fix licence document quantity value not showing

hnryjmes:
### Aim

This is a fix for the production issue where licence quantities and values were not showing on the generated licence document. This affected multiple exporters who raised the issue to lite support.

After initial investigation it looked as if the issue was appearing only for PROVISO licences. Due to an earlier change which fixed the variable output in the generated document from `applied_for_quantity`  and `applied_for_value` to `quantity` and `value` taken from the good on licence, the quantity and value was not showing on the generated licence document.

The problem is in the way PROVISO good data are combined with APPROVE good data in the context generator.

This PR fixes the issue in a way that makes sure we are still using the correct data taken from the good on licence, but combines the APPROVE and PROVISO data in a way that the quantity and value fields are copied from the APPROVE data to the PROVISO data and this then gets put back into APPROVE later in the context function (the other parts of the function could use simplifying, but is out of scope for the time being...) so that, finally, all fields show on the generated licence document as expected.

[LTD-4240](https://uktrade.atlassian.net/browse/LTD-4240)

[LTD-4240]: https://uktrade.atlassian.net/browse/LTD-4240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/letter_templates/context_generator.py
+++ b/api/letter_templates/context_generator.py
@@ -1048,7 +1048,7 @@ def _get_goods_context(application, final_advice, licence=None):
     # we need to make sure only to keep approved goods that are not in proviso goods
     # otherwise goods are duplicated in the licence document. Also we need to copy quantity
     # and value data from the approve good object to the proviso good object.
-    if AdviceType.PROVISO in goods_context:
+    if goods_context[AdviceType.PROVISO] is not []:
         approve_goods = goods_context[AdviceType.APPROVE]
         proviso_goods = goods_context[AdviceType.PROVISO]
         # Copy quantity and value data from approve good to proviso good
--- a/api/letter_templates/context_generator.py
+++ b/api/letter_templates/context_generator.py
@@ -1048,7 +1048,7 @@ def _get_goods_context(application, final_advice, licence=None):
     # we need to make sure only to keep approved goods that are not in proviso goods
     # otherwise goods are duplicated in the licence document. Also we need to copy quantity
     # and value data from the approve good object to the proviso good object.
-    if goods_context[AdviceType.PROVISO] is not []:
+    if goods_context[AdviceType.PROVISO] != []:
         approve_goods = goods_context[AdviceType.APPROVE]
         proviso_goods = goods_context[AdviceType.PROVISO]
         # Copy quantity and value data from approve good to proviso good
--- a/api/letter_templates/context_generator.py
+++ b/api/letter_templates/context_generator.py
@@ -1044,14 +1044,16 @@ def _get_goods_context(application, final_advice, licence=None):
             good_on_application = goods_on_application_dict[advice.good_id]
             goods_context[advice.type].append(_get_good_on_application_context_with_advice(good_on_application, advice))
 
-    # Because we append goods that are approved with proviso to the approved goods below
-    # we need to remove them from the approved goods list otherwise they are duplicated
-    # in the licence document.
-    proviso_good_ids = [item["id"] for item in goods_context[AdviceType.PROVISO]]
-    if proviso_good_ids:
-        current_approved_goods = goods_context.pop(AdviceType.APPROVE)
-        goods_context[AdviceType.APPROVE] = [
-            item for item in current_approved_goods if item["id"] not in proviso_good_ids
+    # Because we append goods that are approved with proviso to the approved goods below,
+    # only add proviso goods if they are not already in approved goods. If the lists are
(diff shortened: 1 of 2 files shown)


api/letter_templates/context_generator.py
```diff
@@ -1045,14 +1045,25 @@ def _get_goods_context(application, final_advice, licence=None):
...
+        approve_goods = goods_context[AdviceType.APPROVE]
+        proviso_goods = goods_context[AdviceType.PROVISO]
+        # Copy quantity and value data from approve good to proviso good
+        for proviso_good in proviso_goods:
+            corresponding_approve_good = next(
+                (x for x in approve_goods if x["id"] == proviso_good["id"]), None
+            )  # get the corresponding approve good or None
+            if corresponding_approve_good is not None:
+                proviso_good.update(quantity=corresponding_approve_good["quantity"])
+                proviso_good.update(value=corresponding_approve_good["value"])
```
saruniitr, line 1061: My understanding is we need to copy `proviso_reason` etc rather than `value` and `quantity`?

hnryjmes: I sneaked it back to the old way, where we keep the proviso goods. that means we then need to add in the missing fields from the approve goods

this line
`approve_goods = [item for item in approve_goods if item["id"] not in [item["id"] for item in proviso_goods]]`
does much the same thing as this did before:
`goods_context[AdviceType.APPROVE] = [item for item in current_approved_goods if item["id"] not in proviso_good_ids]`

so because we are keeping the proviso good, it has the 4 extra fields already:
```
reason
note
denial_reasons
proviso_reason
```

api/letter_templates/context_generator.py
```diff
@@ -1045,14 +1045,25 @@ def _get_goods_context(application, final_advice, licence=None):
...
-    proviso_good_ids = [item["id"] for item in goods_context[AdviceType.PROVISO]]
-    if proviso_good_ids:
-        current_approved_goods = goods_context.pop(AdviceType.APPROVE)
-        goods_context[AdviceType.APPROVE] = [
-            item for item in current_approved_goods if item["id"] not in proviso_good_ids
-        ]
+    # we need to make sure only to keep approved goods that are not in proviso goods
+    # otherwise goods are duplicated in the licence document. Also we need to copy quantity
+    # and value data from the approve good object to the proviso good object.
+    if goods_context[AdviceType.PROVISO] is not []:
```
hnryjmes, line 1051: this line was changed as the old condition actually didn't work, so codecov found something useful

the key `AdviceType.PROVISO` will always be in the dictionary because the dictionary `goods_context` is made to have all the choices of keys in it, so checking that the list is not empty is better

api/letter_templates/tests/test_context_generation.py
```diff
@@ -575,6 +575,61 @@ def test_generate_context_with_proviso_advice_on_goods(self):
...
+            AdviceLevel.FINAL,
+            advice_text="some advice text",
+            good=good,
+        )
+
+        licence = self.create_licence(case, status=LicenceStatus.ISSUED, start_date=date(2023, 10, 5))
+        good_on_licence = GoodOnLicenceFactory(good=good_on_application, quantity=3, value=420, licence=licence)
+
+        context = get_document_context(case)
+        render_to_string(template_name="letter_templates/case_context_test.html", context=context)
```
kevincarrogan, line 596: Is this just to make sure that there's no error? We don't seem to be checking the output of this in anyway.

kevincarrogan (approved):

api/letter_templates/tests/test_context_generation.py
hnryjmes: I was just thinking, we probably don't need this

I think I will remove it and run the test again

api/letter_templates/context_generator.py
kevincarrogan, line 1051: What's weird about the code coverage is that it's saying that we've only tested one of the two possible branches for this `if` statement.

From the tests it looks like we do have both covered but it's making me wonder if there isn't a test that definitely has proviso with an empty list.

hnryjmes: I think you were right, this has now been updated to use `!=` instead which I think is what I should have used initially

PR #1636, opened 2023-10-13
LTD-4282-management-command-report-sla-numbers

depsiatwal:
This is to allow fine tuning of back calculating the SLA numbers produced in Lite. 
This report will be used as a tool to report on the numbers and compare it with the published numbers by ECJU. 
Once we are confident the numbers re correct we can update the SLA numbers in Lite. 

Work in progress as we still need to 

1) Push report to AWS to allow extraction 
2) fine tune the algorithm to match spire as close as possible

### DIFF
--- a/api/cases/management/commands/sla_processing_time_report.py
+++ b/api/cases/management/commands/sla_processing_time_report.py
@@ -1,0 +1,85 @@
+import logging
+from datetime import datetime, time, timedelta
+
+from background_task import background
+from django.conf import settings
+from django.db import transaction
+from django.db.models import F
+from django.db.models import Q
+from django.utils import timezone
+from pytz import timezone as tz
+
+from api.cases.enums import CaseTypeSubTypeEnum
+from api.cases.models import Case, CaseAssignmentSLA, CaseQueue, DepartmentSLA
+from api.cases.models import EcjuQuery
+from api.common.dates import is_weekend, is_bank_holiday
+from api.staticdata.statuses.enums import CaseStatusEnum
+from api.staticdata.statuses.models import CaseStatus
+
+from api.documents.libraries import s3_operations
+
+# DST safe version of midnight
+SLA_UPDATE_TASK_TIME = time(22, 30, 0)
+SLA_UPDATE_CUTOFF_TIME = time(18, 0, 0)
+LOG_PREFIX = "update_cases_sla background task:"
+import csv
+
+def get_all_case_sla():
+    case_sla_records = []
+    for case in Case.objects.all():
+        case_sla_records.append(get_case_sla(case.id))
+
+    with open("case_sla.csv", "w", newline="") as file:
+        writer = csv.writer(file)
+        writer.writerow(["id", "case_ref", "elapsed_days", "working_days", "rfi_queries", "elapsed__rfi_days", "rfi_working_days", "sla_days"])
+        for case_sla_record in case_sla_records:
+            writer.writerow(case_sla_record.values())
+
+
+def get_case_sla(case_id):
+
+    date = timezone.localtime()
+    case = Case.objects.get(id=case_id)
+    elapsed_days = 0
+    working_days = 0
+    sla_days = 0
+    rfi_working_days = 0
+    rfi_non_working_days = 0
+
+    for date in daterange(case.created_at, case.updated_at):
+        elapsed_days += 1
+        if not is_bank_holiday(date, call_api=True) and not is_weekend(date):
+            working_days += 1
+            if not is_active_ecju_queries(date, case_id):
+                sla_days += 1
+            else:
+                rfi_working_days += 1
+        elif is_active_ecju_queries(date, case_id):
+            rfi_non_working_days += 1
+    return {
+            "id": case.id,
+            "reference_code": case.reference_code,
+            "elapsed_days": elapsed_days,
+            "working_days": working_days,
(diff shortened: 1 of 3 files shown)


api/cases/management/commands/sla_processing_time_report.py
```diff
@@ -0,0 +1,97 @@
...
+        if update_audit_log.payload.get("status", {}).get("new") == "submitted":
+            return update_audit_log.created_at
+    return case.created_at
+
+def get_end_date(case):
+    update_audit_logs = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).order_by("-created_at")
+    for update_audit_log in update_audit_logs:
+        if update_audit_log.payload.get("status", {}).get("new") == "finalised":
+            return update_audit_log.created_at
+    return case.updated_at
```
kevincarrogan, line 80: If this case hasn't been finalised then the date we want is todays date, not the updated date (the case may not have been updated for multiple days but the SLA clock should keep ticking).

api/cases/management/commands/sla_processing_time_report.py
```diff
@@ -0,0 +1,97 @@
...
+            "rfi_working_days": rfi_working_days,
+            "sla_days": sla_days,
+        }
+
+def get_start_date(case):
+    update_audit_logs = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).order_by("created_at")
+    for update_audit_log in update_audit_logs:
+        if update_audit_log.payload.get("status", {}).get("new") == "submitted":
+            return update_audit_log.created_at
+    return case.created_at
```
kevincarrogan, line 73: If the case hasn't been submitted then the clock should just be 0, if we set this as `created_at` then we're starting the clock early on cases that haven't been submitted.

api/cases/management/commands/sla_processing_time_report.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def daterange(start_date, end_date):
+    for n in range(int((end_date - start_date).days + 2)):
+        yield start_date + timedelta(n)
+
+
+def is_active_ecju_queries(date, case_id):
+    compare_date_min = datetime.combine(date, time.min)
+    compare_date_max = datetime.combine(date, time.max)
```
saruniitr, line 90: What is the need to compare with this?

api/cases/management/commands/sla_processing_time_report.py
```diff
@@ -0,0 +1,97 @@
...
+
+    return EcjuQuery.objects.filter(
+        Q(responded_at__isnull=True, created_at__lte=compare_date_min)
```
saruniitr, line 93: There have been issues where even if this is null the query is being considered as resolved. Similarly with the `responded_by` field, so may be worth combining both to correctly identify open queries or introduce other means.

api/cases/management/commands/sla_processing_time_report.py
```diff
@@ -0,0 +1,97 @@
...
+
+def get_start_date(case):
+    update_audit_logs = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).order_by("created_at")
+    for update_audit_log in update_audit_logs:
+        if update_audit_log.payload.get("status", {}).get("new") == "submitted":
+            return update_audit_log.created_at
+    return case.created_at
+
+def get_end_date(case):
+    update_audit_logs = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).order_by("-created_at")
```
saruniitr, line 76: Relying on audit events can complicate things.
Eg currently if we have to re-issue a licence (following an Appeal or any other reason) then it will be finalised again so there will be two finalised events.
Another scenario is if they change their final decision, previously refused and now approved.

We also need to consider sub-statuses to filter some of those.

Current reports already rely on these and probably one of the reasons for the errors.

api/cases/management/commands/sla_processing_time_report.py
```diff
@@ -0,0 +1,97 @@
...
+            "elapsed_days": elapsed_days,
+            "working_days": working_days,
+            "rfi_queries": EcjuQuery.objects.filter(case_id=case_id).count(),
+            "elapsed__rfi_days": rfi_working_days + rfi_non_working_days,
+            "rfi_working_days": rfi_working_days,
+            "sla_days": sla_days,
+        }
+
+def get_start_date(case):
+    update_audit_logs = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).order_by("created_at")
```
saruniitr, line 69: Same as below

api/cases/management/commands/sla_processing_time_report.py
```diff
@@ -0,0 +1,97 @@
...
+
+def get_end_date(case):
+    update_audit_logs = Audit.objects.filter(target_object_id=case.id, verb=AuditType.UPDATED_STATUS).order_by("-created_at")
+    for update_audit_log in update_audit_logs:
+        if update_audit_log.payload.get("status", {}).get("new") == "finalised":
+            return update_audit_log.created_at
+    return case.updated_at
+
+
+def daterange(start_date, end_date):
```
saruniitr, line 83: Why +2?

Giving today as both start and end returning tomorrow also,
```
In [8]: for item in daterange(datetime.today(), datetime.today()):
   ...:     print(item)
   ...: 
2023-10-13 14:37:31.482981
2023-10-14 14:37:31.482981
```

PR #1638, opened 2023-10-16
LTD-4301:  Save details of the advisor who assessed a product on the application

saruniitr:
### Change description

Save the user details who assessed a particular product on application.
When users searching for historic product information they are interested in previous assessments as well as the officer who made that assessment. Having these fields in the data model helps injecting into Elasticsearch document easily and make it available in product search results.

[LTD-4301](https://uktrade.atlassian.net/browse/LTD-4301)

[LTD-4301]: https://uktrade.atlassian.net/browse/LTD-4301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/goods/views.py
```diff
@@ -173,6 +174,9 @@ def post(self, request, case_pk):
         for good in self.get_queryset():
             data = request.data.copy()
 
+            # record assessment date
+            good.assessment_date = timezone.now()
+
```
saruniitr, line 179: Rather than updating FE we have an option to add `assessed_by` here in the API using `request.user`.
As this is not really an input from the user, is this ok or update FE to send user id?

hnryjmes: `request.user` should be a reliable indicator of who assessed the product

api/applications/migrations/0073_auto_20231016_1508.py
```diff
@@ -0,0 +1,30 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="goodonapplication",
+            name="assessed_by",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.DO_NOTHING,
```
hnryjmes, line 20: you could also consider `SET_NULL` here instead of `DO_NOTHING`

api/goods/views.py
kevincarrogan: I think we should use `request.user` seems like a simpler change and doesn't rely on us needing to make changes in the frontend.

api/applications/models.py
```diff
@@ -540,6 +538,11 @@ class GoodOnApplication(AbstractGoodOnApplication):
         default=None, blank=True, null=True, help_text="trigger to NCSC for a recommendation"
     )
 
+    assessment_date = models.DateTimeField(blank=True, null=True)
+    assessed_by = models.ForeignKey(
+        GovUser, on_delete=models.DO_NOTHING, blank=True, null=True, related_name="assessed_by"
```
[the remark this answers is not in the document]
kevincarrogan: I would prefer `DO_NOTHING` here, so that in the weird case where we do delete this user, we will still retain a key that could uniquely identify that this did belong to something and was assessed by someone instead of losing all information by making this `null`.

api/goods/views.py
```diff
@@ -117,23 +116,13 @@ def get_queryset(self):
...
 
-    def get_serializer_class(self):
-        if self.application.case_type.sub_type in [CaseTypeSubTypeEnum.OPEN, CaseTypeSubTypeEnum.HMRC]:
-            return ClcControlGoodTypeSerializer
-        return ControlGoodOnApplicationSerializer
-
-    def get_serializer(self, *args, **kwargs):
-        serializer_class = self.get_serializer_class()
-        return serializer_class(*args, **kwargs, data=self.request.data)
-
```
saruniitr, line 136: Not related to this change but used this opportunity to remove this

api/applications/migrations/tests/test_0074_back_populate_assessor_details.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+from django_test_migrations.contrib.unittest_case import MigratorTestCase
+
+from api.audit_trail.enums import AuditType
+
+
+@pytest.mark.django_db()
+class TestBackPopulateProductAssessorDetails(MigratorTestCase):
```
saruniitr, line 8: I am going to add a ticket to remove this test after this is live.

api/applications/migrations/0074_back_populate_product_assessor_details.py
kevincarrogan, line 1: Do we have an idea of how long this may take to run?

kevincarrogan (approved):

hnryjmes (approved):

api/applications/migrations/0074_back_populate_product_assessor_details.py
saruniitr: locally it took under a min

kevincarrogan: Ok. If it's close to a minute then that _might_ cause an issue when we release this. Have you tried it on a dev environment?

The reason why is that we run `./manage migrate` in the Procfile, when we do a deployment there is a healthcheck that runs waiting for that Procfile to bring up the actual web server.

I _think_ that healthcheck has a timeout so it's possible that the migration takes too long, the healthcheck then thinks that the service is unhealthy and then terminates because the migration is still running.

saruniitr: once CI is done then I will merge and deploy to an env.

PR #1645, opened 2023-10-19
LTD-4296: Improve relevance of Product search results

saruniitr:
### Change description

Update the document defined to index product data (GoodOnApplication) to improve the relevance of search results.
Explicitly specify the fields to be searched.

[LTD-4296](https://uktrade.atlassian.net/browse/LTD-4296)

[LTD-4296]: https://uktrade.atlassian.net/browse/LTD-4296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/search/product/documents.py
+++ b/api/search/product/documents.py
@@ -101,7 +101,7 @@ class ProductDocumentType(Document):
     context = fields.Keyword()
 
     # used for grouping
-    canonical_name = fields.KeywordField(attr="good.description")  # is overwritten in prepare
+    canonical_name = fields.KeywordField(attr="good.name")  # is overwritten in prepare
 
     # base details. iteration 1
     id = fields.KeywordField()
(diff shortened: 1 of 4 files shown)


currycoder (commented):
Do we need to explicitly re-index these documents after we deploy this?  How do we do that, out of interest?

saruniitr:
> Do we need to explicitly re-index these documents after we deploy this? How do we do that, out of interest?
Yeah we need to re-index because we are updating document definition. There is a jenkins job for UAT that does after db refresh.
I think we need to do the same for prod as well. There is no job yet as this is the first time.
For Denials and Sanctions data they re-index when new data is added/updated.

api/search/product/documents.py
```diff
@@ -101,7 +101,7 @@ class ProductDocumentType(Document):
     context = fields.Keyword()
 
     # used for grouping
-    canonical_name = fields.KeywordField(attr="good.description")  # is overwritten in prepare
+    canonical_name = fields.KeywordField(attr="good.name")  # is overwritten in prepare
```
kevincarrogan, line 104: Will this automatically get backpopulated?

kevincarrogan (approved):

api/search/product/documents.py
saruniitr: yes as we will rebuild indexes

hnryjmes (approved):

PR #1646, opened 2023-10-20
LTD 4290 added original_submitted_at to case

Tllew:
Backend for the first_submitted at, gets the first updated status for a case.

Technically this might need to filter by payload where status has "new":"submitted", 

I can't find a case where this isn't the value though.

```
select distinct(target_object_id), first_value(payload) over (partition by target_object_id order by created_at) as payload from audit_trail_audit where verb = 'updated_status';
```
^ query to look for a value where the first updated_status isn't new: submitted

[LTD-4290](https://uktrade.atlassian.net/browse/LTD-4290)

[LTD-4290]: https://uktrade.atlassian.net/browse/LTD-4290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/views/views.py
```diff
@@ -139,6 +139,7 @@ def get(self, request, pk):
 
         if case.case_type.sub_type == CaseTypeSubTypeEnum.OPEN:
             data["data"]["destinations"] = get_destinations(case.id)  # noqa
+        data["data"]["first_submitted_at"] = get_first_submitted_at(case.id)
```
kevincarrogan, line 142: Why is this not in the `CaseDetailSerializer` itself?

api/cases/service.py
```diff
@@ -64,3 +64,8 @@ def retrieve_latest_activity(case):
         id=latest_activity.actor_object_id
     )
     return serialize_case_activity(latest_activity, actor)
+
+
+def get_first_submitted_at(case_id):
+    audit_history = Audit.objects.filter(target_object_id=case_id, verb="updated_status").order_by("created_at").first()
```
kevincarrogan, line 70: Is there a constant or enum that could be used for the `verb` here?

kevincarrogan: Also, I wonder if the case itself already has a link to the audit objects it's on so you could do `case.audits.filter`, which could also potentially mean optimising this a little bit.

api/cases/views/views.py
Tllew: You're absolutely right, not sure how I missed the serializer 2 lines above..

api/cases/service.py
saruniitr: there is an enum for this verb, https://github.com/uktrade/lite-api/blob/dev/api/audit_trail/enums.py#L44

api/cases/serializers.py
```diff
@@ -389,6 +394,11 @@ def get_audit_notification(self, instance):
         if queryset.exists():
             return {"audit_id": queryset.first().object_id}
 
+    def get_first_submitted_at(self, instance):
+        queryset = Audit.objects.filter(target_object_id=instance.id, verb=AuditType.UPDATED_STATUS).first()
```
depsiatwal, line 398: what if we have multiple  we may have to order by created desc

api/cases/serializers.py
```diff
@@ -281,6 +284,7 @@ class CaseDetailSerializer(serializers.ModelSerializer):
     latest_activity = serializers.SerializerMethodField()
     case_type = PrimaryKeyRelatedSerializerField(queryset=CaseType.objects.all(), serializer=CaseTypeSerializer)
     next_review_date = serializers.SerializerMethodField()
+    first_submitted_at = serializers.SerializerMethodField()
```
depsiatwal, line 287: I prefer orginal_submitted_at but personal pref :-)

depsiatwal (commented):
Also we will need to add tests

depsiatwal (commented):
lgtm is it worth running the performance test metrics to make sure no impact ?

PR #1651, opened 2023-10-24
LTD-4323 added processing_time to case model

Tllew:
added processing_time to case model


api/cases/models.py
```diff
@@ -110,6 +110,7 @@ class Case(TimestampableModel):
     last_closed_at = models.DateTimeField(null=True)
 
     sla_days = models.PositiveSmallIntegerField(null=False, blank=False, default=0)
+    processing_time = models.PositiveSmallIntegerField(null=False, blank=False, default=0)
```
saruniitr, line 113: Could you add `help_text` explaining the purpose of this field, difference wrt sla days?

PR #1653, opened 2023-10-25
 LTD-4337-fix-sla-report 

depsiatwal:
Refactor and fixes reporting errors


Tllew:
can we rename the file to be sla_processing_time

just so that it doesn't feel odd importing it in tasks.py

kevincarrogan (approved):

kevincarrogan:
The fact that these tests were wrong and we are kind of just switching their numbers suggests there should be better tests here. 

I look at the tests and don’t really know why those numbers are like they are. 

For this I think I’d prefer to see some smaller more obvious tests to actually describe what all of the calculations are doing e.g. a test of a case that is open and closed on the same day, a test that shows a case being open and closed on Monday and then Friday, a test that shows a case that spans a weekend. Keeping them simple means you could know exactly what the number means instead of it just being “33” which feels like we’ve just picked two dates and that happens to be the number.

depsiatwal:
> The fact that these tests were wrong and we are kind of just switching their numbers suggests there should be better tests here.
> 
> I look at the tests and don’t really know why those numbers are like they are.
> 
> For this I think I’d prefer to see some smaller more obvious tests to actually describe what all of the calculations are doing e.g. a test of a case that is open and closed on the same day, a test that shows a case being open and closed on Monday and then Friday, a test that shows a case that spans a weekend. Keeping them simple means you could know exactly what the number means instead of it just being “33” which feels like we’ve just picked two dates and that happens to be the number.

I'be broken the tests up a little more logical, 
I think the fact we try to flip the is_weekend doesn't help. 

so now we have two sets of tests 

- once checks with just bank holidays being flipped and the figures and another with the same data and RFI data includes 
this should test and validate every case and also prevents the flaky test issue.

Tllew (approved):

PR #1657, opened 2023-10-26
Match the docker linux version that's on cloudfoundry

kevincarrogan:
This is so that we can have a development environment that is more consistent with our production environments

### Aim

Match linux versions to what's on cloudfoundry


depsiatwal (approved):
lgtm 
this will be tested on dev right ? before merge

Tllew (approved):

PR #1661, opened 2023-10-30
Match cf linux in docker e2e

kevincarrogan:
### Aim

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,4 +1,4 @@
-FROM python:3.8.18-slim-bullseye
+FROM ubuntu:22.04
 
 ARG DOCKERIZE_VERSION=v0.6.1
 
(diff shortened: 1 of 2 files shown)


stuaxo:
Not a blocker:  
consider the deadsnakes ppa, it should keep build times reasonable if that's a thing for pyenv.

```
sudo add-apt-repository ppa:deadsnakes/ppa
sudo apt install python3.8 python3.9
```

https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa

The PPA should be OK, it's been the ubuntu place for different python versions for a number of years

hnryjmes:
I can see you've updated commands in a few places to use pipenv run, e.g. `./manage.py ingest_ui_test_sanctions` becomes `pipenv run ./manage.py ingest_ui_test_sanctions`

We also should do this for the docker commands here for completeness sake:

https://github.com/uktrade/lite-api/blob/dev/makefile#L16

I only know this because I pulled this branch and tried to run `make doc-migrate` locally and it didn't work, but it does work if you exec into the container and then do `pipenv shell` and `python manage.py migrate`. I guess changing those makefile commands to use `pipenv run` would have the same effect, because it would make sure that pipenv was happy first

hnryjmes (approved):
Approved, have also made a suggestion about the makefile in my other comment

hnryjmes:
@stuaxo I am not sure I agree with the deadsnakes suggestion just because the point of this exercise is to make our dev environments as similar to the cf environments as reasonably possible, so using pipenv here seems like it contributes to that aim

kevincarrogan:
> I can see you've updated commands in a few places to use pipenv run, e.g. `./manage.py ingest_ui_test_sanctions` becomes `pipenv run ./manage.py ingest_ui_test_sanctions`
> 
> We also should do this for the docker commands here for completeness sake:
> 
> https://github.com/uktrade/lite-api/blob/dev/makefile#L16
> 
> I only know this because I pulled this branch and tried to run `make doc-migrate` locally and it didn't work, but it does work if you exec into the container and then do `pipenv shell` and `python manage.py migrate`. I guess changing those makefile commands to use `pipenv run` would have the same effect, because it would make sure that pipenv was happy first

I'll make another PR to fix this.

seijihg (approved):

PR #1662, opened 2023-10-30
[LTD-4359] Ensure precedents API excludes products with is_good_controlled=None

currycoder:
### Aim

Ensure the precedents API excludes GoodOnApplication entries with `is_good_controlled=None` from results.  These are either unassessed GoodOnApplication records or pre-date `is_good_controlled` being set on assessment, so it is best to exclude them from precedents.

[LTD-4359](https://uktrade.atlassian.net/browse/LTD-4359)

[LTD-4359]: https://uktrade.atlassian.net/browse/LTD-4359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/views/views.py
+++ b/api/cases/views/views.py
@@ -1184,6 +1184,10 @@ class GoodOnPrecedentList(ListAPIView):
         return (
             GoodOnApplication.objects.filter(good__in=goods, good__status=GoodStatus.VERIFIED)
             .exclude(application=case)
+            # Ensure any precedents we return have a non-None value for is_good_controlled.
+            # GoodOnApplication records with is_good_controlled=None are either not yet assessed
+            # or are legacy records
+            .exclude(is_good_controlled=None)
             .select_related(
                 "report_summary_prefix",
                 "report_summary_subject",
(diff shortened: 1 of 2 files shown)


kevincarrogan (approved):

api/cases/views/views.py
```diff
@@ -1184,6 +1184,7 @@ def get_queryset(self):
         return (
             GoodOnApplication.objects.filter(good__in=goods, good__status=GoodStatus.VERIFIED)
             .exclude(application=case)
+            .exclude(is_good_controlled=None)
```
kevincarrogan, line 1187: Is there actually a different between this and `__isnull=True`?

currycoder: Good question.  Personally I find this a bit more readable as we are in ORM land rather than DB land, but I'm easy

kevincarrogan: Could we get a code comment around here as well? I feel like it's the kind of thing someone might find in future and choose to get rid of it if it gets raised as a bug.

saruniitr: If it is not assessed then `is_good_controlled` will be None but the status won't be verified as well. So the above won't exclude these instances.

saruniitr: Discussed internally and agreed that `exclude` is still required so marking this as resolved.

saruniitr (approved):

PR #1664, opened 2023-10-30
Add option to search by regime

hnryjmes:
### Aim

Update the product search API to support searching by regime. Both the regime `name` and `shortened_name` can be searched on.

[LTD-4330](https://uktrade.atlassian.net/browse/LTD-4330)

[LTD-4330]: https://uktrade.atlassian.net/browse/LTD-4330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/search/product/documents.py
+++ b/api/search/product/documents.py
@@ -56,6 +56,22 @@ email_analyzer = analysis.analyzer(
 )
 
 
+class Regime(InnerDoc):
+    # e.g. "W"
+    shortened_name = fields.KeywordField(
+        fields={
+            "raw": fields.KeywordField(normalizer=lowercase_normalizer),
+            "suggest": fields.CompletionField(),
+        },
+        copy_to="wildcard",
+    )
+    # e.g. "Wassenaar Arrangement"
+    name = fields.TextField(
+        copy_to="wildcard",
+        analyzer=descriptive_text_analyzer,
+    )
+
+
 class Rating(InnerDoc):
     rating = fields.KeywordField(
         fields={
@@ -169,7 +185,8 @@ class ProductDocumentType(Document):
         copy_to="wildcard",
     )
 
-    regime = fields.Keyword()
+    regime_entries = fields.NestedField(attr="regime_entries", doc_class=Regime)
+
     assessed_by = fields.NestedField(doc_class=GovUser, attr="assessed_by")
     assessment_date = fields.DateField(
         attr="assessment_date",
--- a/api/search/product/tests/test_views.py
+++ b/api/search/product/tests/test_views.py
@@ -25,6 +25,9 @@ class ProductSearchTests(DataTestClient):
         self.tau_user2 = GovUserFactory(
             baseuser_ptr=BaseUserFactory(first_name="TAU", last_name="Advisor2"), team=self.team
         )
+        self.tau_user3 = GovUserFactory(
+            baseuser_ptr=BaseUserFactory(first_name="TAU", last_name="Advisor3"), team=self.team
+        )
 
         # Create few products and add them to an application
         good = GoodFactory(
@@ -125,7 +128,7 @@ class ProductSearchTests(DataTestClient):
             is_good_controlled=True,
             report_summary="marine position fixing equipment",
             regime_entries=["Wassenaar Arrangement"],
-            assessed_by=self.tau_user2,
+            assessed_by=self.tau_user3,
             assessment_date=parse("2023-10-22T15:51:28.529110+00:00"),
         )
 
@@ -142,7 +145,7 @@ class ProductSearchTests(DataTestClient):
             is_good_controlled=True,
             report_summary="imaging cameras",
             regime_entries=["Wassenaar Arrangement Sensitive"],
-            assessed_by=self.tau_user2,
+            assessed_by=self.tau_user3,
             assessment_date=parse("2023-10-23T15:51:28.529110+00:00"),
         )
 
@@ -159,7 +162,7 @@ class ProductSearchTests(DataTestClient):
             is_good_controlled=True,
             report_summary="chemicals used for general laboratory work/scientific research",
             regime_entries=["CWC Schedule 3"],
-            assessed_by=self.tau_user2,
+            assessed_by=self.tau_user3,
             assessment_date=parse("2023-10-24T15:51:28.529110+00:00"),
         )
 
@@ -270,7 +273,7 @@ class ProductSearchTests(DataTestClient):
     @parameterized.expand(
         [
             ({"search": "Advisor1"}, 3),
-            ({"search": "Advisor2"}, 5),
+            ({"search": "Advisor2"}, 2),
         ]
     )
     def test_product_search_by_assessing_officer(self, query, expected_count):
--- a/api/search/product/tests/test_views.py
+++ b/api/search/product/tests/test_views.py
@@ -147,7 +147,7 @@ class ProductSearchTests(DataTestClient):
         )
 
         good = GoodFactory(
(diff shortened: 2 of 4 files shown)


api/search/product/documents.py
```diff
@@ -56,6 +56,22 @@
...
+        fields={
+            "raw": fields.KeywordField(normalizer=lowercase_normalizer),
+            "suggest": fields.CompletionField(),
+        },
+        copy_to="wildcard",
+    )
+    # e.g. "Wassenaar Arrangement"
+    name = fields.TextField(
+        copy_to="wildcard",
+        analyzer=descriptive_text_analyzer,
```
saruniitr, line 71: At this point we are not looking to expose suggestions etc but you can add `suggest` for this field as well.

api/search/product/tests/test_views.py
```diff
@@ -107,6 +107,57 @@ def setUp(self):
...
+            good=good,
+            is_good_controlled=True,
+            report_summary="imaging cameras",
+            regime_entries=["Wassenaar Arrangement Sensitive"],
+            assessed_by=self.tau_user2,
+            assessment_date=parse("2023-10-23T15:51:28.529110+00:00"),
+        )
+
+        good = GoodFactory(
+            name="mom's homemade lemonade",
```
saruniitr, line 145: Please rename this to something more sensible

api/search/product/tests/test_views.py
```diff
@@ -107,6 +107,57 @@ def setUp(self):
...
+            part_number="xyz123",
+            organisation=self.organisation,
+            is_good_controlled=True,
+            control_list_entries=["1C35016"],
+        )
+        GoodOnApplicationFactory(
+            application=self.case,
+            good=good,
```
saruniitr, line 153: This approach of creating products will not scale when we add full text search options, best to load a bunch of products using fixtures. I will create a separate task for that.

hnryjmes: should I wait for you to create that task or rewrite this test data now?

api/search/product/documents.py
hnryjmes: would it be like the above? i.e.
```
fields={
            "raw": fields.KeywordField(normalizer=lowercase_normalizer),
            "suggest": fields.CompletionField(),
        },
```

I would guess that lowercase_normalizer is appropriate in this context as the regime names aren't case sensitive

saruniitr: yes, same as above

api/search/product/tests/test_views.py
saruniitr: No need to rewrite, for this PR this is ok.

api/search/product/views.py
```diff
@@ -33,19 +33,13 @@ class ProductDocumentView(DocumentViewSet):
...
 
-    search_fields = [
-        "name",
-        "part_number",
-        "control_list_entries",
-        "report_summary",
-        "organisation",
-        "assessment_note",
-    ]
+    search_fields = ["name", "part_number", "control_list_entries", "report_summary", "regime_entries"]
```
saruniitr, line 36: This change is not required.
This is also removing `organisation, assessment_note` from the search fields.
I think it probably an issue with resolving merge conflicts.

We actually don't want `regime_entries` here as it is a nested field which is added below.

hnryjmes: yep, definitely a bad resolution to a merge conflict. thanks for spotting it

saruniitr (approved):

api/search/product/documents.py
```diff
@@ -56,6 +56,22 @@
...
 
 
+class Regime(InnerDoc):
+    # e.g. "W"
+    shortened_name = fields.KeywordField(
+        fields={
+            "raw": fields.KeywordField(normalizer=lowercase_normalizer),
+            "suggest": fields.CompletionField(),
+        },
+        copy_to="wildcard",
```
kevincarrogan, line 66: Won't some of these be so short that they won't actually be useful? I'm not completely convinced having the `shortened_name` really gives us much but I can't remember what they all are.

api/search/product/tests/test_views.py
```diff
@@ -174,12 +225,37 @@ def test_product_search_by_control_list_entries(self, query, expected_count, exp
...
+        response = response.json()
+        self.assertEqual(response["count"], expected_count)
+        self.assertIn(
+            expected_regime_shortened_name,
+            [entry["shortened_name"] for item in response["results"] for entry in item["regime_entries"]],
+        )
+        self.assertIn(
+            expected_regime_name,
+            [entry["name"] for item in response["results"] for entry in item["regime_entries"]],
+        )
```
kevincarrogan, line 251: Having these two asserts kind of makes me a bit confused as to what we're suggesting is being tested here. For some of these it just happens to equal that, not because it's actually what was searched for but just because that happens to be the other value.

api/search/product/tests/test_views.py
```diff
@@ -194,7 +270,7 @@ def test_product_search_by_report_summary(self, query, expected_count, expected_
     @parameterized.expand(
         [
             ({"search": "Advisor1"}, 3),
-            ({"search": "Advisor2"}, 2),
+            ({"search": "Advisor2"}, 5),
```
kevincarrogan, line 273: The kinds of tests where we just bump the number always seem off to me. It kind of shows that our tests aren't really very well isolated from each other or specific enough. I don't really know _why_ it's 5, I just know you've updated it but I have no understanding as to what's just happened to make it magically increase.

If this number goes up in the future do we know if something is breaking or are we just meant to increase it whenever the number goes up?

saruniitr: In the tests three new products are created and they all need `assessed_by` set.
Currently we define two users and in this case second user is used for all three products hence it increased by 3 so this particular test need updating.
Some what related to the discussion we were having internally about defining this test data better. I think this is more likely as we start adding more tests unless we create more users in the test but in some case we may want the same user assessing multiple goods.

api/search/product/documents.py
hnryjmes: a lot of this is based on assuming, on my part. the fact that we only show the shortened name (e.g. "W") in the tau tab in caseworker at the moment led me to believe that this is the most important form of the name for them

api/search/product/tests/test_views.py
hnryjmes: I thought this might come up, as I know kevin you're not a huge fan of "test this count is a specific number X" type tests. there are a huge number of these in this file, so if we do care about this, we should probably change all of them.

api/search/product/documents.py
saruniitr: In the product assessment tab also we only show short names

api/search/product/tests/test_views.py
hnryjmes: I suppose I am being a bit flexible with the test data, rather than exhaustively testing the variations of each regime I just picked some representative examples, and the assertions work for either the short name or the full name

api/search/product/tests/test_views.py
kevincarrogan: Well. It's less "test this count is a specific number X", it's more about it not being clear _why_ that number is that number and seeing diffs that just seemingly bump these numbers up and down give me a sense that we're not really actively thinking about why these numbers change we're just increasing and decreasing them when the test fails.

This isn't a blocker from my side though if we think this is extending this work unnecessarily.

kevincarrogan (approved):

PR #1666, opened 2023-11-01
LTD-4365 - Output CSV of recommendations to update historical report_summary prefix and suffixes.

stuaxo:
Output a CSV of recommendations to back populate the report_summary_prefix and report_summary_suffix on GoodOnApplication.

This normalises the report_summary, then looks up existing matches in ReportSummaryPrefix and ReportSummarySuffix

For suffix, the longest match is taken first for a better match.

Example run: `$ python manage.py suggest_summaries output.csv`


api/goods/management/commands/suggest_summaries.py
```diff
@@ -0,0 +1,104 @@
...
+    return filter_controlled_good_on_applications(
+        q.filter(report_summary__isnull=False, report_summary_prefix__isnull=True, report_summary_subject__isnull=True)
+    )
+
+
+def filter_populated_controlled_good_on_applications(q):
+    return filter_controlled_good_on_applications(q).filter(report_summary_subject__isnull=False)
+
+
+def annotate_matching_prefix(q):
```
currycoder, line 39: I'm kind of impressed that you managed to do this in the DB, but jeepers is this complex.  Why didn't you grab all the prefix strings/subject strings and do the matching in python?

api/goods/management/commands/suggest_summaries.py
```diff
@@ -0,0 +1,104 @@
...
+            "report_summary",
+            "suggested_prefix",
+            "suggested_prefix_id",
+            "suggested_subject",
+            "suggested_subject_id",
+        )
+        prefixes_by_id = {o.id: o for o in ReportSummaryPrefix.objects.all()}
+        subjects_by_name = {o.name: o for o in ReportSummarySubject.objects.all()}
+
+        with open(filename, "w") as infile:
```
currycoder, line 82: `outfile`, maybe?

api/goods/management/commands/suggest_summaries.py
```diff
@@ -0,0 +1,104 @@
...
+            ExpressionWrapper(
+                Func(F("report_summary"), Value(r"\s\(\d+\)$|\s\(x\)$"), Value(""), function="REGEXP_REPLACE"),
+                output_field=TextField(),
+            ),
+        ),
+    )
+
+
+def filter_controlled_good_on_applications(q):
+    return q.filter(Q(Q(is_good_controlled=True) | Q(is_good_controlled__isnull=True, good__is_good_controlled=True)))
```
kevincarrogan, line 26: Does the `Q` with the `Q` in it do anything special?

api/goods/management/commands/suggest_summaries.py
```diff
@@ -0,0 +1,104 @@
...
+
+def annotate_matching_prefix(q):
+    # Match the longest prefix first.
+    prefixes = ReportSummaryPrefix.objects.annotate(name_length=Length("name"))
+
+    best_matching_prefix = (
+        prefixes.annotate(prefix_position=StrIndex(OuterRef("report_summary"), "name"))
+        .filter(prefix_position=1)
+        .order_by("-name_length")
+        .values("pk")[:1]
```
kevincarrogan, line 47: When we say "best matching" why is there possibly more than one?

api/goods/management/commands/suggest_summaries.py
```diff
@@ -0,0 +1,104 @@
...
+            "suggested_subject",
+            "suggested_subject_id",
+        )
+        prefixes_by_id = {o.id: o for o in ReportSummaryPrefix.objects.all()}
+        subjects_by_name = {o.name: o for o in ReportSummarySubject.objects.all()}
+
+        with open(filename, "w") as infile:
+            writer = csv.DictWriter(infile, fieldnames=csv_headers, quoting=csv.QUOTE_ALL)
+            writer.writeheader()
+            for o in q:
```
kevincarrogan, line 85: I'm getting a bit lost with these super short variable names.

kevincarrogan (approved):

api/goods/management/commands/suggest_summaries.py
```diff
@@ -0,0 +1,104 @@
...
+        .values("pk")[:1]
+    )
+
+    q = q.annotate(suggested_prefix_id=Subquery(best_matching_prefix, output_field=TextField()))
+    return q
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument("filename", type=str, help="Path to the input CSV file")
```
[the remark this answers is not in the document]
kevincarrogan: Not to be changed but you can also do something like:

```
parser = argparse.ArgumentParser()
parser.add_argument('infile', nargs='?', type=argparse.FileType('r'))
parser.add_argument('outfile', nargs='?', type=argparse.FileType('w'))
```
https://docs.python.org/3/library/argparse.html

api/goods/management/commands/suggest_summaries.py
stuaxo: ISTR there are prefixes that like "abc" "abc efg" if this didn't have the `-name_length` you would find the shorter prefix first, so the most exact match is the longer one.    I'll rename this to `prefix`.

api/goods/management/commands/suggest_summaries.py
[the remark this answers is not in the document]
stuaxo: Ooh, FileType, that's nice - will file away for the future.

api/goods/management/commands/suggest_summaries.py
stuaxo: Good question... Quite possibly not, I can't remember if it worked without that or not.

api/goods/management/commands/suggest_summaries.py
stuaxo: I did start with that, it started to get complex and I had some bugs..   with prefix, for each report_summary you need to loop through all of the candidates and find the best match using startswith.  I guess if I wasn't trying to go fast I could have stopped and thought about it longer, but it seemed easier to punt it to the database.

stuaxo: Probably a habit of {other_project_in_this_org} where everything works that way.. that I will break / as I do more on this thing and become a master of doing things in basic types.

api/goods/management/commands/suggest_summaries.py
stuaxo: Fixed.

api/goods/management/commands/suggest_summaries.py
stuaxo: un-longified these, when when I did the commit to move the management to it's correct place in applications.

api/goods/management/commands/suggest_summaries.py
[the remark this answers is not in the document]
stuaxo: Resolving with AGREED-WONTFIX-THANKS-FOR-THE-INFO.

api/goods/management/commands/suggest_summaries.py
stuaxo: Went with csvfile since we only deal with one file here.

api/goods/management/commands/suggest_summaries.py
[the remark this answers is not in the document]
stuaxo: Noted - applied when I did the other small fixes just now.

hnryjmes (commented):
I understand that we're under time pressure here, but I still think a test for some of these functions (e.g. a test for the regex) would strengthen this PR

Tllew (approved):

Tllew (approved):
Looks good to me

hnryjmes (approved):
Thanks for adding the test

api/applications/tests/test_suggest_summaries.py
```diff
@@ -0,0 +1,124 @@
...
+        good = GoodFactory.create(
+            organisation=self.organisation,
+        )
+
+        # Input data, not including the ids for items that are created later:
+        report_data = [
+            # report_summary, suggested_prefix, suggested_subject
+            ("arts and crafts", None, "arts and crafts"),
+            ("training for arts and crafts", "training for", "arts and crafts"),
+            ("equipment for arts and crafts", "equipment for", "arts and crafts"),
```
currycoder, line 66: I think adding a row where suggested subject is None will cover some extra lines above

api/applications/management/commands/suggest_summaries.py
```diff
@@ -0,0 +1,122 @@
...
+def filter_unpopulated_controlled_good_on_applications(good_on_applications: QuerySet[GoodOnApplication]):
+    return filter_controlled_good_on_applications(
+        good_on_applications.filter(
+            report_summary__isnull=False, report_summary_prefix__isnull=True, report_summary_subject__isnull=True
+        )
+    )
+
+
+def filter_populated_controlled_good_on_applications(good_on_applications: QuerySet[GoodOnApplication]):
+    return filter_controlled_good_on_applications(good_on_applications).filter(report_summary_subject__isnull=False)
```
currycoder, line 40: Looks like this isn't called so can be dropped

currycoder (approved):
I think doing the fixes I suggest will make codecov happy.  Otherwise looking good

api/applications/tests/test_suggest_summaries.py
stuaxo: This is an unhappy path and shouldn't be output in the CSV.

I added the minimum test - in an ideal world I'd probably check the logging was called - let's see if it makes codecov happy.

PR #1668, opened 2023-11-01
Data migration to populate report_summary prefix and suffix from a cs…

stuaxo:
This uses the data generated by:  https://github.com/uktrade/lite-api/pull/1666

Once the data is placed in the lite-api, it will ingest it in migration 0075 of applications.


hnryjmes (commented):
I think that adding a data migration test here would strengthen this PR

api/applications/migrations/0075_back_populate_product_report_summary_prefix_and_suffix.py
```diff
@@ -0,0 +1,76 @@
...
+        for row in read_updates_csv(Migration.data_path):
+            good_on_application_pk = row["id"]
+            report_prefix_id = row["suggested_prefix_id"]
+            report_subject_id = row["suggested_subject_id"]
+            try:
+                good_on_application = GoodOnApplication.objects.get(id=good_on_application_pk)
+            except GoodOnApplication.DoesNotExist:
+                logger.warning("GoodOnApplication with id %s does not exist", good_on_application_pk)
+            else:
+                good_on_application.report_summary_prefix_id = report_prefix_id
```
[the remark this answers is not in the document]
stuaxo: Added + for prefixes.   Gets a little more if-blocky than I'd like; ordinarily I'd arrange things a little differently so this kind of error couldn't happen, log these all up front, but :shrug:

api/applications/migrations/0075_back_populate_product_report_summary_prefix_and_suffix.py
```diff
@@ -0,0 +1,76 @@
...
+    def populate_report_prefix_and_subject(apps, schema_editor):
+        GoodOnApplication = apps.get_model("applications", "GoodOnApplication")
+        for row in read_updates_csv(Migration.data_path):
+            good_on_application_pk = row["id"]
+            report_prefix_id = row["suggested_prefix_id"]
+            report_subject_id = row["suggested_subject_id"]
+            try:
+                good_on_application = GoodOnApplication.objects.get(id=good_on_application_pk)
+            except GoodOnApplication.DoesNotExist:
+                logger.warning("GoodOnApplication with id %s does not exist", good_on_application_pk)
```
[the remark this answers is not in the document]
stuaxo: Done

stuaxo:
> I think that adding a data migration test here would strengthen this PR

Added one that goes forward.

I'd have to do further refactoring to make this more test-friendly and able to be broken up (+ be more "unit" test and less "integration" test like).

Since this is an integration test, we end up having to make the file that this consumes ( - it has to contain references to things to existing GoodOnApplication instances to reference).

I can think of many ways of making this test better if I spent more time on it, TBF - it tests a happy path, and that the data we put in is actually created; which I'd consider a minimum.

The reverse test is a also not implemented here.

api/applications/migrations/tests/test_0075_back_populate_product_report_summary_prefix_and_suffix.py
```diff
@@ -0,0 +1,148 @@
...
+    )
+
+    GoodOnApplication = new_state.apps.get_model("applications", "GoodOnApplication")  # noqa N806
+    ReportSummaryPrefix = new_state.apps.get_model("report_summaries", "ReportSummaryPrefix")  # noqa N806
+    ReportSummarySubject = new_state.apps.get_model("report_summaries", "ReportSummarySubject")  # noqa N806
+
+    new_good_on_applications = GoodOnApplication.objects.filter(report_summary__in=report_summary_data)
+
+    assert new_good_on_applications.count() == len(report_summary_data)
+    for new_good_on_application in new_good_on_applications:
```
currycoder, line 129: This currently loops through data that was retrieved by matching `report_summary`.  Could it be better to loop through the CSV data itself in order to prove that the GoodOnApplication ids specified in that data were updated as expected?

stuaxo: I'll have a look... I think you're onto something that we should be matching ids and checking those are the items updated for this to be more watertight,

stuaxo: Checking the ids should be enough, since we just wrote the CSV a few lines up + we don't need to test CSVWriter itself succeeded.

The trap when writing this was being in a situation where the CSV path wasn't properly patched - so it read the real CSV (perfectly safe it's the test will fail).

I've made do the initial filter by primary keys, with that and the `.count()` check the length is the same it should be good.

hnryjmes (approved):
Approved. Tests look good. I will leave it to you to look at the issue with lite-routing-sha (I think there's something that complains if you do a sha update and it's not to the tip of the branch)

currycoder (approved):

PR #1673, opened 2023-11-02
LTD-4362: Add query string type full text search for Product search

saruniitr:
## Change description

Add "query string" type full text search capabilities to retrieve historic Product details.
This is achieved by adding a custom search filter backend which constructs `query_string` type queries that allows us to embed operators between the search terms. Read more about this [Opensearch query string query](https://opensearch.org/docs/latest/query-dsl/full-text/query-string/)

The plugins allow us to implement custom query backend and we need to implement `construct_search()` method which builds the query that is sent to opensearch for execution.

This enhances the capabilities of Product search and allows users to find relevant products efficiently.
Example searches they can perform,
 - product name with specific rating and regime
 - product name with rating that starts with some string (wildcard)
 - Products that match with exact phrase and certain details in assessment note etc
Please see tests for actual search terms.

**Follow up**
Query string queries are strict in the sense that if the search syntax is not valid then the engine throws an error. We need to discuss how these errors are to be conveyed to user hence doing that as a follow up so as to not hold up this for review.

Please see individual commit messages for additional context of some changes.

[LTD-4362](https://uktrade.atlassian.net/browse/LTD-4362)

[LTD-4362]: https://uktrade.atlassian.net/browse/LTD-4362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/search/product/tests/test_views.py
```diff
@@ -199,7 +249,7 @@ def test_product_search_by_name(self, query, expected_count, expected_name):
     @parameterized.expand(
         [
             ({"search": "ABC"}, 1, "ABC-123"),
-            ({"search": "HFL"}, 1, "HFL"),
+            ({"search": "H2SO4"}, 1, "H2SO4"),
```
saruniitr, line 252: Updated product part number hence this change

api/search/product/tests/test_views.py
```diff
@@ -286,11 +336,12 @@ def test_product_search_by_assessing_officer(self, query, expected_count):
     @pytest.mark.elasticsearch
     def test_product_search_by_applicant(self):
         query_params = {"search": f"{self.organisation.name}"}
+        expected_count = Good.objects.filter(organisation=self.organisation).count()
         response = self.client.get(self.product_search_url, query_params, **self.gov_headers)
         self.assertEqual(response.status_code, 200)
 
         response = response.json()
-        self.assertEqual(response["count"], 9)
+        self.assertEqual(response["count"], expected_count)
```
saruniitr, line 344: Created new products in this PR and we are using the same organisation for all products hence this change is required

api/search/product/tests/test_views.py
```diff
@@ -306,6 +357,27 @@ def test_product_search_by_assessment_note(self, query, expected_count):
...
+        [
+            ({"search": "rifle"}, 1),
+            ({"search": "shifter AND 6A004"}, 1),
+            ({"search": "sensor AND 6A006"}, 1),
+            ({"search": "sensor AND 6A*"}, 2),
+            ({"search": "sensor AND (thermal OR magnetic)"}, 2),
+            ({"search": "Wassenaar AND 6A001a1d"}, 1),
+            ({"search": "Chemicals AND (WS OR CWC)"}, 1),
+            ({"search": '"Thermal camera"'}, 1),
+            ({"search": "components NOT camera"}, 1),
```
saruniitr, line 371: tried to add one example for each possible combination.
Once users start using it then we will understand their common search scenarios (through GA) and we can update this with additional scenarios.

api/search/product/views.py
```diff
@@ -42,6 +43,17 @@ class ProductDocumentView(DocumentViewSet):
         "assessment_note",
     ]
 
+    query_string_search_fields = {
+        "name": None,
```
saruniitr, line 47: Instead of `None` we can specify a dict with `boost` parameter to boost the relevance score of that field.
For now we are not leaving this to default.

api/search/product/documents.py
```diff
@@ -73,11 +73,12 @@ class Regime(InnerDoc):
 
 
 class Rating(InnerDoc):
-    rating = fields.KeywordField(
+    rating = fields.TextField(
```
hnryjmes, line 76: I assume this is for ratings like `FR AI` which are multiple tokens

api/search/product/documents.py
```diff
@@ -73,11 +73,12 @@ class Regime(InnerDoc):
...
         fields={
             "raw": fields.KeywordField(normalizer=lowercase_normalizer),
```
hnryjmes, line 78: I'm wondering if we can use TextField everywhere? are there advantages to using KeywordField for this?

hnryjmes: I remember case sensitivity (e.g. "ML1a" vs "ml1a") is something to be aware of when making these choices

api/search/product/tests/test_views.py
hnryjmes: if control entry is now a TextField, and if that means input can be case insensitive (e.g. "6A001a1d" and "6a001a1d" would both do the same thing) - does that mean we can add some test cases for this?

api/search/product/documents.py
saruniitr: It needs to be a text field to be able to use in full text search as keyword fields are not analyzed.
https://opensearch.org/docs/latest/field-types/supported-field-types/keyword/

Although we don't use this nested field for full text search I have corrected it here.

api/search/product/documents.py
saruniitr: keywords are useful for exact matches, filtering etc

api/search/product/filter_backends.py
```diff
@@ -0,0 +1,13 @@
+from django_elasticsearch_dsl_drf.filter_backends import BaseSearchFilterBackend
+
+from api.search.product.query_backends import QueryStringQueryBackend
+
+
+class QueryStringSearchFilterBackend(BaseSearchFilterBackend):
+    """Query string search filter backend."""
```
depsiatwal, line 7: Possible to be more descriptive on what the purpose of this Class is ?

api/search/product/query_backends.py
```diff
@@ -0,0 +1,97 @@
+import copy
+
+from elasticsearch_dsl.query import Q
+from django_elasticsearch_dsl_drf.filter_backends.search.query_backends import BaseSearchQueryBackend
+
+
+class QueryStringQueryBackend(BaseSearchQueryBackend):
```
depsiatwal, line 7: again it would be useful to know what this class achieves

depsiatwal (commented):
also do we need more tests for this ?

api/search/product/tests/test_views.py
saruniitr: I can add a test for this. In the analyzers we are converting all text fields to lowercase so input is case insensitive already.

saruniitr:
> also do we need more tests for this ?

added tests for various combinations for search functionality but if there are other specific tests in mind please let me know I can add.
Will be adding few more tests to address codecov

api/search/product/filter_backends.py
saruniitr: this adds support for this new query type in elastic search plugin, will expand in the comment.

hnryjmes (approved):
I deployed this to demo and have tried many search input combinations. Everything is working as expected. There are still some known issues not related to this PR (e.g. searching "name AND" with a dangling "AND" causes the site to error) but maybe we can fix that in future with some frontend validation before it gets to the backend. 

Looks good.

saruniitr:
> I deployed this to demo and have tried many search input combinations. Everything is working as expected. There are still some known issues not related to this PR (e.g. searching "name AND" with a dangling "AND" causes the site to error) but maybe we can fix that in future with some frontend validation before it gets to the backend.
> 
> Looks good.

Yeah, there is a follow up ticket to fix this.

api/search/product/documents.py
```diff
@@ -186,8 +189,9 @@ class ProductDocumentType(Document):
     )
 
     regime_entries = fields.NestedField(attr="regime_entries", doc_class=Regime)
+    regimes = fields.TextField(attr="good.name", multi=True)  # is overwritten in prepare
 
-    assessed_by = fields.NestedField(doc_class=GovUser, attr="assessed_by")
+    assessed_by = fields.TextField(attr="good.name", multi=True)  # is overwritten in prepare
```
kevincarrogan, line 194: From the documentation it looks like `attr` can just be a `None` (or left unset) https://django-elasticsearch-dsl.readthedocs.io/en/latest/fields.html#available-fields, what happens if we do that instead of setting `attr` here to a field that we ultimately ignore?

kevincarrogan: I answered my own question, it would just set it to `assessed_by`. Maybe https://django-elasticsearch-dsl.readthedocs.io/en/latest/fields.html#using-prepare-field is the answer instead to not having to set `attr="good.name"`

kevincarrogan: Having looked at the code as to how `django_elasticsearch_dsl` works it seems that a `prepare_*` method takes precedence over `attr` so it looks like as long as you had a `prepare_assessed_by` method then you could remove `attr` and then just move the code from `prepare` to the relevant `prepare_*` method.

saruniitr: oh nice, didn't check internally. I will add `prepare_*` methods for each of those fields.

api/search/product/tests/test_querybackends.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class ProductSearchQueryStringQueryBackendTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        self.product_search_url = reverse("product_search-list")
+
+        # When these test are executed it is possible that index is not aware of Product documents
+        # so rebuild indexes to ensure successful query execution
+        call_command("search_index", models=["applications.GoodOnApplication"], action="rebuild", force=True)
```
kevincarrogan, line 17: What's the impact of this on the time to run tests?

saruniitr: negligible (varying in the range of 200-500 msec) as we are only adding 11 products.

api/search/product/tests/test_querybackends.py
```diff
@@ -0,0 +1,46 @@
...
+        Search fields are specified as dict but that is not provided then the
+        backend falls back to 'search_fields' list. In this test we delete the dict
+        to test whether it falls back to using the list.
+        We are always going to use dict but to address test coverage we need to check
+        with list as well.
+        """
+        delattr(mock_view.__class__, "query_string_search_fields")
+
+        response = self.client.get(self.product_search_url, {"search": "shifter AND 6A004"}, **self.gov_headers)
+        self.assertEqual(response.status_code, 200)
```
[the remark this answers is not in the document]
saruniitr: this is sufficient. Main thing I wanted to check is it is falling back to using `search_fields`

kevincarrogan (approved):

api/search/product/tests/test_querybackends.py
saruniitr: I have updated to re-index only once for all tests

PR #1675, opened 2023-11-03
[LTD-4352] Bulk distinct assessments endpoint

currycoder:
### Aim

The aim with this change is to create a new endpoint for creating bulk TAU assessments of goods.  The existing goods assessment endpoint has two major drawbacks;
- It does not accept bulk distinct assessments (just a single assessment applied to multiple GoodOnApplication records).
- It is slow.

The new endpoint created in this PR has aimed to port over all of the necessary functionality from the old endpoint while leaning more on django rest framework to make things much cleaner.

For previous assessment submissions (companion PR; https://github.com/uktrade/lite-frontend/pull/1606), response time for submitting 27 assessments has dropped from ~22seconds to under 3 seconds.
Further time savings to get this example under 1 second are possible, but rely on us improving search document indexing and so are separate Jira stories;
https://uktrade.atlassian.net/browse/LTD-4383
https://uktrade.atlassian.net/browse/LTD-4388

Ways the new endpoint diverges from the existing one (https://github.com/uktrade/lite-api/blob/dev/api/goods/views.py#L107);
- There is no permissions check on the endpoint - as we have a distributed monolith, we are relying on permissions checking to occur on the frontend.  Note that caseworker auth is of course required.
- The old endpoint would explicitly set a WASSENAAR system flag.  This flag is no longer used, so the new endpoint does not set it.
- The old endpoint would apply good flagging rules.  The flagging rules are applied when the case moves on from TAU, so this has not been reflected in the new endpoint.
- Various fields have been removed from the serializer; e.g. `is_precedent`, `is_wassenaar`, `end_use_control`, `assessed_by`

[LTD-4352](https://uktrade.atlassian.net/browse/LTD-4352)

[LTD-4352]: https://uktrade.atlassian.net/browse/LTD-4352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update --fix-missing
 RUN apt-get install -y libpq-dev gcc curl \
   build-essential python3-dev python3-pip python3-setuptools python3-wheel \
   python3-cffi libcairo2 libpango-1.0-0 libpangocairo-1.0-0 libgdk-pixbuf2.0-0 \
-  libffi-dev shared-mime-info swig git imagemagick poppler-utils openssl
+  libffi-dev shared-mime-info swig git imagemagick poppler-utils openssl libsqlite3-dev
 RUN curl https://pyenv.run | bash
 ENV HOME /root
 ENV PYENV_ROOT $HOME/.pyenv
--- a/api/assessments/serializers.py
+++ b/api/assessments/serializers.py
@@ -18,7 +18,11 @@ from lite_content.lite_api import strings
 
 class AssessmentUpdateListSerializer(serializers.ListSerializer):
     def update(self, instances, validated_data):
-        instance_data_pairs = zip(instances, validated_data)
+        instances_by_id = {str(instance.id): instance for instance in instances}
+        validated_data_by_id = {str(data["id"]): data for data in validated_data if data.get("id")}
+        instance_data_pairs = []
+        for instance_id, instance in instances_by_id.items():
+            instance_data_pairs.append((instance, validated_data_by_id[instance_id]))
         result = [
             self.child.update(instance, validated_update_data)
             for instance, validated_update_data in instance_data_pairs
@@ -37,6 +41,7 @@ class AssessmentUpdateListSerializer(serializers.ListSerializer):
 
 class AssessmentSerializer(GoodControlReviewSerializer):
 
+    id = serializers.UUIDField()
     regime_entries = PrimaryKeyRelatedField(
         many=True,
         queryset=RegimeEntry.objects.all(),
@@ -52,6 +57,7 @@ class AssessmentSerializer(GoodControlReviewSerializer):
     class Meta:
         model = GoodOnApplication
         fields = (
+            "id",
             "control_list_entries",
             "is_good_controlled",
             "comment",
--- a/api/assessments/serializers.py
+++ b/api/assessments/serializers.py
@@ -94,7 +94,7 @@ class AssessmentSerializer(GoodControlReviewSerializer):
         good.save()
 
     def emit_audit_entry(self, instance, validated_data, old_values):
-        case = get_case(instance.application_id)
+        case = get_case(instance.application_id, select_related=["status"])
         default_control = [strings.Goods.GOOD_NO_CONTROL_CODE]
         default_regimes = ["No regimes"]
         new_control_list_entries = [item.rating for item in validated_data["control_list_entries"]]
@@ -121,10 +121,10 @@ class AssessmentSerializer(GoodControlReviewSerializer):
 
     def get_old_values_for_audit(self, instance):
         return {
-            "control_list_entry": list(instance.control_list_entries.values_list("rating", flat=True)),
+            "control_list_entry": [cle.rating for cle in instance.control_list_entries.all()],
             "is_good_controlled": instance.is_good_controlled,
             "report_summary": instance.report_summary,
-            "regime_entries": list(instance.regime_entries.values_list("name", flat=True)),
+            "regime_entries": [regime_entry.name for regime_entry in instance.regime_entries.all()],
         }
 
     def update(self, instance, validated_data):
--- a/api/assessments/serializers.py
+++ b/api/assessments/serializers.py
@@ -1,0 +1,52 @@
+from rest_framework import serializers
+from rest_framework.relations import PrimaryKeyRelatedField
+
+from api.applications.models import GoodOnApplication
+from api.core.serializers import GoodControlReviewSerializer
+from api.staticdata.report_summaries.models import ReportSummarySubject, ReportSummaryPrefix
+from api.staticdata.regimes.models import RegimeEntry
+from api.users.enums import UserStatuses
+from api.users.models import GovUser
+
+
+class UpdateListSerializer(serializers.ListSerializer):
+    def update(self, instances, validated_data):
+
+        instance_hash = {index: instance for index, instance in enumerate(instances)}
+
+        result = [self.child.update(instance_hash[index], attrs) for index, attrs in enumerate(validated_data)]
+
+        return result
+
+
+class AssessmentSerializer(GoodControlReviewSerializer):
+
+    regime_entries = PrimaryKeyRelatedField(
+        many=True,
+        queryset=RegimeEntry.objects.all(),
+        required=False,  # Not required until we completely do away with the string report_summary field..
+    )
+    report_summary_prefix = PrimaryKeyRelatedField(
+        required=False, allow_null=True, queryset=ReportSummaryPrefix.objects.all()
+    )
+    report_summary_subject = PrimaryKeyRelatedField(
+        required=False, allow_null=True, queryset=ReportSummarySubject.objects.all()
+    )
+    assessed_by = PrimaryKeyRelatedField(
+        required=False, allow_null=True, queryset=GovUser.objects.filter(status=UserStatuses.ACTIVE)
+    )
+
+    class Meta:
+        model = GoodOnApplication
+        fields = (
+            "control_list_entries",
+            "is_good_controlled",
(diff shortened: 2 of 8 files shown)


api/assessments/serializers.py
```diff
@@ -0,0 +1,146 @@
...
+        return data
+
+
+class AssessmentSerializer(GoodControlReviewSerializer):
+
+    id = serializers.UUIDField()
+    regime_entries = PrimaryKeyRelatedField(
+        many=True,
+        queryset=RegimeEntry.objects.all(),
+        required=False,  # Not required until we completely do away with the string report_summary field..
```
saruniitr, line 48: assuming this is because of a copy-paste issue?

api/assessments/tests/test_views.py
```diff
@@ -0,0 +1,220 @@
...
+
+        good = good_on_application.good
+        assert good.status == GoodStatus.VERIFIED
+        assert [cle.rating for cle in good.control_list_entries.all()] == ["ML1"]
+        assert good_on_application.report_summary == f"{report_summary_prefix.name} {report_summary_subject.name}"
+        assert good.report_summary_prefix_id == report_summary_prefix.id
+        assert good.report_summary_subject_id == report_summary_subject.id
+
+        audit_entry = Audit.objects.order_by("-created_at").filter(verb=AuditType.PRODUCT_REVIEWED).first()
+        assert audit_entry.payload == {
```
saruniitr, line 85: In this case we are expecting only one event so could you also check for expected count?
We have `AuditSerializer`, may be comparing the serialized output simplifies this?

api/assessments/tests/test_views.py
```diff
@@ -0,0 +1,220 @@
...
+from api.staticdata.report_summaries.models import ReportSummarySubject, ReportSummaryPrefix
+from api.staticdata.statuses.models import CaseStatus
+
+
+class MakeAssessmentsViewTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        self.application = self.create_draft_standard_application(organisation=self.organisation)
+        self.case = self.submit_application(self.application)
+        self.good = GoodFactory(organisation=self.organisation)
```
saruniitr, line 24: I see what you have done here, reused the same for all :)

Tllew (approved):
Looks good, much cleaner

Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get update --fix-missing
 RUN apt-get install -y libpq-dev gcc curl \
   build-essential python3-dev python3-pip python3-setuptools python3-wheel \
   python3-cffi libcairo2 libpango-1.0-0 libpangocairo-1.0-0 libgdk-pixbuf2.0-0 \
-  libffi-dev shared-mime-info swig git imagemagick poppler-utils openssl
+  libffi-dev shared-mime-info swig git imagemagick poppler-utils openssl libsqlite3-dev
```
hnryjmes, line 9: 👀

api/assessments/serializers.py
```diff
@@ -0,0 +1,146 @@
...
+            "is_good_controlled": instance.is_good_controlled,
+            "report_summary": instance.report_summary,
+            "regime_entries": [regime_entry.name for regime_entry in instance.regime_entries.all()],
+        }
+
+    def update(self, instance, validated_data):
+        old_values = self.get_old_values_for_audit(instance)
+        super().update(instance, validated_data)
+        instance.assessed_by = validated_data["user"]
+        instance.assessment_date = timezone.now()
```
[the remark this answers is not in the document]
hnryjmes: @saruniitr could you explain this comment a bit more please?

api/assessments/tests/test_views.py
```diff
@@ -0,0 +1,220 @@
...
+            "new_regime_entries": [regime_entry.name],
+            "old_control_list_entry": ["No control code"],
+            "old_is_good_controlled": "No",
+            "old_regime_entries": ["No regimes"],
+            "old_report_summary": None,
+            "report_summary": good_on_application.report_summary,
+        }
+
+    @freeze_time("2023-11-03 12:00:00")
+    def test_valid_data_updates_multiple_records(self):
```
hnryjmes, line 100: is there a way to test the audit entries are as expected for this function too? as Arun said above, you might be able to use `AuditSerializer` - which also should mean the test assertion won't get too large when you're testing multiple audit entries

hnryjmes (commented):
looking good so far, will be happy to approve if and when you address the codecov patch check

api/assessments/serializers.py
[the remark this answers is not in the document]
saruniitr: `assessment_date` is set when user submit their assessment for a product. In this case a previously assessed good is used in another application.
User decided to accept those previous assessments and submitted, in this case we are updating date and user details which is correct. I was just wondering if there are no previous assessments what would happen in that case. I think that is currently being worked on where they see a combination of goods which have previous assessments and some doesn't.

api/assessments/serializers.py
currycoder: No, we still need this for now until `report_summary` has been removed

api/assessments/serializers.py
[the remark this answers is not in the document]
currycoder: The overall intention is for this API endpoint to replace the existing assessment endpoint also.  I'd like to remove the old one.  All that this does is set the `assessed_by` as the user that made the request; doesn't matter if they are confirming a previous assessment or (eventually) creating a new assessment.  They are the user that assessed this GoodOnApplication.

api/assessments/views.py
```diff
@@ -0,0 +1,62 @@
...
+        instances = self.get_queryset(ids)
+        line_numbers = self.get_application_line_numbers(instances)
+        serializer = self.get_serializer(instances, data=request.data, partial=False, many=True)
+        serializer.is_valid(raise_exception=True)
+        self.perform_update(serializer, user=request.user.govuser, line_numbers=line_numbers)
+
+        return JsonResponse(data={}, status=status.HTTP_200_OK)
+
+
+def validate_ids(data, unique=True):
```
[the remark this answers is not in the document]
currycoder: Yes, really this shouldn't be possible - as you say, the frontend should be built in such a way that only one assessment is sent per GoodOnApplication id.  But I've put this in as a safeguard in the case where we don't do that for some reason.

Yes, it should be possible to express duplicate IDs - will do that.

api/assessments/tests/test_views.py
currycoder: Maybe, will take a look.  What might be better is just collapsing this all in to one test with parameterized

api/assessments/serializers.py
saruniitr: ok but this is `regime_entries` field and the comment is about `report_summary`

api/assessments/serializers.py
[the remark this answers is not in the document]
saruniitr: agree, we need to replace the existing one with this.

api/assessments/serializers.py
currycoder: Aha!  Thank you Arun :D

api/assessments/tests/test_views.py
currycoder: Turns out that a parametrized testcase made this a bit of a nightmare so have just made a new testcase to cover this off.

saruniitr (approved):

api/assessments/serializers.py
```diff
@@ -0,0 +1,146 @@
...
+from api.goods.enums import GoodStatus
+from api.staticdata.report_summaries.models import ReportSummarySubject, ReportSummaryPrefix
+from api.staticdata.regimes.models import RegimeEntry
+from api.staticdata.statuses.enums import CaseStatusEnum
+
+from lite_content.lite_api import strings
+
+
+class AssessmentUpdateListSerializer(serializers.ListSerializer):
+    def update(self, instances, validated_data):
```
stuaxo, line 20: [Not a blocker]

Every time I come to one of these, I have no idea what structure the data coming in has by looking at just the signature.   It would be great if we included some more info (many options so won't solutionise right here).

api/assessments/serializers.py
```diff
@@ -0,0 +1,146 @@
...
+        if data.get("report_summary_subject"):
+            if data.get("report_summary_prefix"):
+                data["report_summary"] = f"{data['report_summary_prefix'].name} {data['report_summary_subject'].name}"
+            else:
+                data["report_summary"] = data["report_summary_subject"].name
+        return data
+
+    def update_good(self, instance, validated_data):
+        good = instance.good
+        # Add control entries to a verified good
```
stuaxo, line 84: ^ Possibly don't need this comment, it probably should be inside the `if` if kept.

api/assessments/serializers.py
```diff
@@ -0,0 +1,146 @@
...
+        if good.status == GoodStatus.VERIFIED:
+            good.control_list_entries.add(*validated_data["control_list_entries"])
+        # Overwrite control entries for a previously unverified good
```
stuaxo, line 87: This comment probably wants to be inside the `else` block.

api/assessments/serializers.py
```diff
@@ -0,0 +1,146 @@
...
+        else:
+            good.status = GoodStatus.VERIFIED
+            good.control_list_entries.set(validated_data["control_list_entries"])
+            good.flags.remove(SystemFlags.GOOD_NOT_YET_VERIFIED_ID)
+
+        # Set report summary fields on the good
```
stuaxo, line 93: [Not a blocker]  Not entirely sure this comment adds much.

stuaxo (approved):

PR #1678, opened 2023-11-08
LTD-4385 add stopwords to address denialdocument

Tllew:
updated address to use the address_analyzer and added suite to the stopwords

[LTD-4385](https://uktrade.atlassian.net/browse/LTD-4385)

[LTD-4385]: https://uktrade.atlassian.net/browse/LTD-4385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/external_data/documents.py
+++ b/api/external_data/documents.py
@@ -12,40 +12,6 @@ class DataField(fields.ObjectField):
         return instance.data
 
 
-class DenialDocumentType(Document):
-    id = fields.KeywordField()
-    name = fields.TextField()
-    address = fields.TextField()
-    reference = fields.KeywordField()
-    regime_reg_ref = fields.KeywordField()
-    notifying_government = fields.TextField()
-    country = fields.TextField(
-        fields={
-            "raw": fields.KeywordField(normalizer=lowercase_normalizer),
-        },
-    )
-    item_list_codes = fields.TextField()
-    item_description = fields.TextField()
-    consignee_name = fields.TextField()
-    end_use = fields.TextField()
-    data = DataField()
-    is_revoked = fields.BooleanField()
-
-    class Index:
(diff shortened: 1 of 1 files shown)


api/external_data/documents.py
```diff
@@ -12,40 +12,6 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
         return instance.data
 
 
-class DenialDocumentType(Document):
-    id = fields.KeywordField()
-    name = fields.TextField()
-    address = fields.TextField()
```
saruniitr, line 18: Is this move really required? My only concern is that we lose history in git blame.
Of course we can check how it was before it was moved but wondering if this is required for any other reasons.

depsiatwal (commented):
agree with comments the change looks good just need to make move back to the top.

api/external_data/documents.py
Tllew: the move was so that the address_stop_words_filter was in scope, seemed better than moving the stop_word_filter to the top of the file

saruniitr: in that case I think moving analyzers, filters above the document is better.

Tllew: This allows any of the analyzers to be used going forward, I don't think there's much gain in doing it the other way.

saruniitr: the end result is same in both cases, all the analyzers, filters will be at the top followed by document definition except that the diff will be with respect to filters.
I am easy on this but usually better not to move things unless necessary for the functionality.

saruniitr (approved):

depsiatwal (approved):
LGTM

PR #1681, opened 2023-11-09
 LTD-4393-add-denial-reason-entity-id 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4393

new fields being pulled from spire db

reason_for_denial - Requested by business
entity_id - is spire entity_id which may be used for delta updates in future imports.


api/external_data/models.py
```diff
@@ -32,6 +32,8 @@ class Denial(TimestampableModel):
     data = models.JSONField(default=dict)
     is_revoked = models.BooleanField(default=False, help_text="If true do not include in search results")
     is_revoked_comment = models.TextField(default="")
+    reason_for_refusal = models.TextField("The reason for denial'", blank=True, default="", null=True)
```
hnryjmes, line 35: it looks like you have `"The reason for denial'"` here instead of probably `"The reason for denial"`

hnryjmes (approved):
small comment but looks good otherwise

PR #1683, opened 2023-11-10
LTD-4391-denials-re-import-data 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4391

This bring in new fields in the data import from Spire reason_for_denial and entity_id. 

Existing records where we have marked as an application remain intact and we refresh all the other data.


api/external_data/models.py
```diff
@@ -32,6 +32,8 @@ class Denial(TimestampableModel):
     data = models.JSONField(default=dict)
     is_revoked = models.BooleanField(default=False, help_text="If true do not include in search results")
     is_revoked_comment = models.TextField(default="")
+    reason_for_refusal = models.TextField("The reason for denial'", blank=True, default="", null=True)
```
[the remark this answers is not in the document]
Tllew: @depsiatwal this does look a little odd, is it intentional to have the ' in the textfield?

`models.TextField("The reason for denial", blank=True, default="")`

should sort out all the comments on this

api/external_data/migrations/0018_auto_20231109_1425.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('external_data', '0017_denial_regime_reg_ref'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='denial',
+            name='reason_for_refusal',
+            field=models.TextField(blank=True, default='', null=True, verbose_name="The reason for denial'"),
```
Tllew, line 16: is this the same as below where it doesn't need null=True?

Tllew (approved):
Gonna put this as approved, there's a load of comments from the code review bot though. and I'm not sure about the text string

api/external_data/migrations/0018_auto_20231109_1425.py
depsiatwal: Yeah this is coming through with django doctor however we do need both null=true and default=true

this is because we have existing records we want to default and also import new records which are optional and may be null.

kevincarrogan (approved):

hnryjmes (approved):
all looks good to me

PR #1684, opened 2023-11-10
LTD-4402: Handle syntax errors in query string search terms

saruniitr:
## Change description

query_string queries are strict and does not tolerate syntax errors in
the search terms. eg a trailing AND, OR operators.

In this case we get an exception when the query is executed. To avoid
this and present a meaningful error response to the user, validate the
query before executing and bailout early if the query is not valid.

[LTD-4402](https://uktrade.atlassian.net/browse/LTD-4402)

[LTD-4402]: https://uktrade.atlassian.net/browse/LTD-4402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/search/product/views.py
```diff
@@ -85,6 +86,40 @@ def get_search_indexes(self):
...
+        # create a query with the given query params
+        query = {
+            "query": {
+                "query_string": {
+                    "fields": ["*"],
+                    "query": f"{search_term}",
+                }
+            }
+        }
+        return self.document._index.validate_query(body=query)
```
currycoder, line 106: Could this function just return True or False so that callers don't need to understand the validate response?

currycoder (approved):
Just a minor suggestion

stuaxo (approved):

api/search/product/views.py
saruniitr: We cannot change the response as it is coming from the plugin itself. It actually returns more data, I am just using `valid` field.
Complete response,
```
{'_shards': {'total': 1, 'successful': 1, 'failed': 0}, 'valid': True}
```

currycoder: I just mean that this helper function could return True/False and so `dispatch` didn't need to care about ES validate API responses.  I'm easy though

saruniitr: I will update

PR #1685, opened 2023-11-10
Augment search results with model instances

kevincarrogan:
## Change description

We define an elastic search document type based on a model which allows to index the fields we are interested in.
These are later useful to perform searches based on those fields and results contain the fields specified in the document.
However we have requirement where we want to show additional details from the model which are not necessarily index and made available for search.
To achieve this we need to augment the search results (hits) with model instances so that at the time of generating the response we can include additional fields using the model instance.

In this case we override the `list()` method that the plugin uses and hydrate the results with additional fields.

[LTD-4429](https://uktrade.atlassian.net/browse/LTD-4429)

[LTD-4429]: https://uktrade.atlassian.net/browse/LTD-4429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/search/product/views.py
+++ b/api/search/product/views.py
@@ -1,3 +1,4 @@
+from django_elasticsearch_dsl.search import Search
 from django_elasticsearch_dsl_drf import filter_backends
 from django_elasticsearch_dsl_drf.viewsets import DocumentViewSet
 from elasticsearch_dsl.query import Query
@@ -81,6 +82,11 @@ class ProductDocumentView(DocumentViewSet):
 
     highlight_fields = {"*": {"enabled": True, "options": {"pre_tags": ["<b>"], "post_tags": ["</b>"]}}}
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.search = Search(using=self.client, index=self.index, doc_type=self.document._doc_type.name)
+
     def get_search_indexes(self):
         if self.request.GET.get("database") in settings.ELASTICSEARCH_PRODUCT_INDEXES:
             return settings.ELASTICSEARCH_PRODUCT_INDEXES[self.request.GET["database"]]
@@ -148,7 +154,8 @@ class ProductDocumentView(DocumentViewSet):
                 }
             }
         )
-        return super().get_queryset()
+        queryset = super().get_queryset()
+        return queryset
 
 
 class ProductSuggestDocumentView(APIView):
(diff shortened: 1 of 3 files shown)


api/search/product/views.py
```diff
@@ -148,7 +151,35 @@ def get_queryset(self):
...
             }
         )
-        return super().get_queryset()
+        queryset = super().get_queryset()
+        return queryset
+
+    def get_model_queryset(self, document_queryset):
+        pks = [result.meta.id for result in document_queryset]
+        qs = self.document.Django.model.objects.filter(pk__in=pks)
+        preserved_order = Case(*[When(pk=pk, then=pos) for pos, pk in enumerate(pks)], output_field=IntegerField())
```
hnryjmes, line 160: what is the order if this isn't set here?

kevincarrogan: There isn't any order if this isn't set.

saruniitr: It will be `assessment_date` and `score`.
This is just to maintain the same order.

hnryjmes (commented):
I think another test that explicitly uses `augment_hits_with_instances` would make this better, and it might fix the codecov warning as well

saruniitr:
> I think another test that explicitly uses `augment_hits_with_instances` would make this better, and it might help to see how to fix the codecov warning as well

Every other test other than `test_search_results_serializer()` uses that `list()` method so that is covered.
`codecov` is complaining when pagination is not used, I will parameterize this test itself to cover that case.

api/search/product/views.py
stuaxo: This will definitely do that in a generic way, I guess both of these aren't accessible from the database, with `score` at least being an ES thing ?

stuaxo: The order preservation is useful.
If uses this and requires an `.order_by` on the query afterwards the work would be wasted - at that point it might be an idea for them to modify this so there is a way of getting these results without that step.

saruniitr:
> I think another test that explicitly uses `augment_hits_with_instances` would make this better, and it might help to see how to fix the codecov warning as well

My previous comment is not entirely correct, we cannot directly trigger `list()` and override pagination etc so I added a separate test to address codecov issue.

Tllew (approved):
Looks good to me

depsiatwal (approved):
Approved

PR #1686, opened 2023-11-10
Set "djangorestframework>=3.13.1,!=3.14.0" as djangorestframework>=3.…

stuaxo:
…13.1,!=3.14.0

As NullBoolean field is going away.


stuaxo:
Withdrawing as can't repro (which is good in this case).

PR #1687, opened 2023-11-10
LTD-4395 - Ensure previous assessment table only shows assessments from cases that went past TAU

stuaxo:
Ensure previous assessment table only shows assessments from cases that went past TAU

This adds filtering by status, so any statuses after TAU should be filtered out - it's acknowledged in the original ticket that to not be perfect - status may not be the only thing that decides this - but in practice, for the assessments here this gets us most of the way there.

### DIFF
--- a/api/staticdata/statuses/enums.py
+++ b/api/staticdata/statuses/enums.py
@@ -71,7 +71,7 @@ class CaseStatusEnum:
 
     compliance_visit_statuses = [OPEN, UNDER_INTERNAL_REVIEW, RETURN_TO_INSPECTOR, AWAITING_EXPORTER_RESPONSE, CLOSED]
 
-    post_tau_statuses = [
+    precedent_statuses = [
         UNDER_REVIEW,
         OGD_ADVICE,
         OGD_CONSOLIDATION,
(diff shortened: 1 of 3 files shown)


api/staticdata/statuses/enums.py
```diff
@@ -71,6 +71,16 @@ class CaseStatusEnum:
...
+    post_tau_statuses = [
+        UNDER_REVIEW,
+        OGD_ADVICE,
+        OGD_CONSOLIDATION,
+        UNDER_FINAL_REVIEW,
+        FINAL_REVIEW_COUNTERSIGN,
+        FINAL_REVIEW_SECOND_COUNTERSIGN,
+        FINALISED,
+    ]
+
```
stuaxo, line 83: TODO:  Refer back to statuses flow and put in the final list here.

currycoder: Is this TODO done?

currycoder (approved):
Nice work!

api/staticdata/statuses/enums.py
stuaxo: Had a look at this yesterday it seems about right, but I didn't come away from it with the certainty I wished for + some other eyes on this would be good - I'll revisit it myself as well.

Wasn't sure if I should add a comment to reflect the imperfect-ness of this ... it looks like status alone doesn't determine whether it has passed TAU, but again I think I need a bit more context to write something that's useful.

hnryjmes (approved):

PR #1688, opened 2023-11-13
LTD-4405: Update field name in serializer to match with document

saruniitr:
## Change description

Add a test to ensure fields in document match with serializer

[LTD-4405](https://uktrade.atlassian.net/browse/LTD-4405)

[LTD-4405]: https://uktrade.atlassian.net/browse/LTD-4405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/search/product/tests/test_views.py
+++ b/api/search/product/tests/test_views.py
@@ -4,6 +4,7 @@ from parameterized import parameterized
 from unittest.mock import patch
 
 from django.core.management import call_command
+from django.test import RequestFactory
 from django.urls import reverse
 from rest_framework import status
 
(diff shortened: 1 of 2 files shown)


api/search/product/tests/test_views.py
```diff
@@ -234,6 +238,24 @@ def tearDownClass(cls):
         cls.application.delete()
         cls.team.delete()
 
+    def test_search_results_serializer(self):
```
saruniitr, line 241: This test is included in this class because we need products to be created so that we get some search results, to avoid duplication in another class this is included as part of this class.

kevincarrogan (approved):

depsiatwal (commented):
looks like you need more tests other then that LGTM

saruniitr:
> looks like you need more tests other then that LGTM

I already added a test, could you clarify what other tests are required?

PR #1693, opened 2023-11-14
 fix-GeneratedDocumentSendTests 

depsiatwal:
Fixing tests where reference numbers don't match 
this could possibly be due to uncleaned up data

### DIFF
--- a/api/search/product/tests/test_views.py
+++ b/api/search/product/tests/test_views.py
@@ -234,6 +234,9 @@ class ProductSearchTests(DataTestClient):
 
     @classmethod
     def tearDownClass(cls):
+        # Clean up all the objects created for these tests
+        # Some other tests expecting certain reference codes and count of goods
+        # so they will fail if these are not cleanedup properly.
         Good.objects.filter(organisation=cls.organisation).delete()
         GovUser.objects.filter(team=cls.team).delete()
         cls.application.delete()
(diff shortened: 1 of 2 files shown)


currycoder (approved):

api/search/product/tests/test_views.py
```diff
@@ -213,23 +214,34 @@ def get_products_data(organisation, application, gov_users):
...
 
         # Rebuild indexes with the products created
         call_command("search_index", models=["applications.GoodOnApplication"], action="rebuild", force=True)
 
+    @classmethod
+    def tearDownClass(cls):
+        Good.objects.filter(organisation=cls.organisation).delete()
+        GovUser.objects.filter(team=cls.team).delete()
+        cls.application.delete()
+        CaseReferenceCode.objects.delete()
```
depsiatwal, line 240: why not just CaseReferenceCode.objects.all().delete()

depsiatwal (commented):
small comment

api/search/product/tests/test_views.py
```diff
@@ -236,6 +237,7 @@ def tearDownClass(cls):
         Good.objects.filter(organisation=cls.organisation).delete()
         GovUser.objects.filter(team=cls.team).delete()
         cls.application.delete()
+        CaseReferenceCode.objects.all().delete()
```
depsiatwal, line 240: Also possible to put a comment in so we know why this clean up is required. This is incase in the future we have same issue with other objects.

kevincarrogan (approved):

PR #1694, opened 2023-11-15
LTD-4392: Product search by destination

saruniitr:
## Change description

Add option to search by destination (consignee, end-user or ultimate end-users)

[LTD-4392](https://uktrade.atlassian.net/browse/LTD-4392)

[LTD-4392]: https://uktrade.atlassian.net/browse/LTD-4392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/search/product/documents.py
```diff
@@ -242,6 +246,9 @@ def prepare_assessed_by(self, instance):
             else []
         )
 
+    def prepare_ultimate_end_user_country(self, instance):
```
hnryjmes, line 249: how do these `prepare_` functions actually get used? I can see there are a few of them defined this way, but then they aren't referenced anywhere... what am I missing?

saruniitr: These are called when this object is prepared for indexing. We may want to do some work before it is indexed and these methods allow us to do that.
In case of multiple entries we are flattening them as a top level array in this prepare field.
These are to be defined as `prepare_<field_name>`

saruniitr: https://django-elasticsearch-dsl.readthedocs.io/en/latest/fields.html#using-prepare-field

hnryjmes (approved):
all looks good to me

Tllew (approved):
Looks good, I like the create and delete, it's a shame it can't be in setup (I know it created dupes)

PR #1702, opened 2023-11-21
Ltd 4433 populate report summaries part 3 tau feedback

stuaxo:
Outputs data for TAU to review to populate remaining report summaries.

This is in the same format as used in the PRs for LTD-4406, adding good_name and url as columns.

After the data is exported there is a manual process to add it to an excel workbook that is shared with TAU for review.

Once the data has been collected, the sheet can be input using a data migration, this will be in a separate PR but can copy+past from the data migration on 0075 or 0076.


hnryjmes (approved):
nice

stuaxo:
Updates today, to add the migration (copied from 0076 and dependent migration number bumped), the test was also moved from 0076 to 0077 and has no new code.

Tllew (approved):
Looks good, some interesting syntax in there too with the *()

PR #1711, opened 2023-11-26
LTD-4383-celery

seijihg:


test_helpers/clients.py
```diff
@@ -111,7 +111,7 @@
 class Static:
     seeded = False
 
-
+@override_settings(CELERY_TASK_ALWAYS_EAGER=True, CELERY_TASK_EAGER_PROPAGATES=True)
```
stuaxo, line 114: ^ You almost always want these when testing with celery

Here's what we did in TAP
https://github.com/uktrade/tamato/blob/c5fe1353a295553d64fba3d44a2e9a84f38bed30/conftest.py#L112

stuaxo:
I didn't see anything in here that starts celery beat or it's worker, but then when I looked for examples of "django_elasticsearch_dsl.signals.CelerySignalProcessor" the docs on elasticsearch DSL itself only say 
`. And celery needs to be pre-configured in the django project,`
https://django-elasticsearch-dsl.readthedocs.io/en/latest/settings.html?highlight=celery#elasticsearch-dsl-signal-processor

And point at the celery getting started page
https://docs.celeryq.dev/en/stable/django/first-steps-with-django.html

I think we are missing a celery.py to configure celery and a tasks.py to hold our tasks, instructions for starting celery beat, and then the worker.

The docs mention creating a CustomSignalProcessor where the which would manage the jobs, this is probably using celery signals https://docs.celeryq.dev/en/stable/userguide/signals.html to trigger a one of the tasks in tasks.py that actually does the indexing.

The docs in elasticsearch-dsl are so sparse, that if we got all this working it would be patching them upstream to make them work.

stuaxo:
There's a really good example https://bplociennik.com/post/django-elasticsearch-dsl-celery-signal/ that shows the setup of the crucial missing `task.py` and other bits that need adding to this branch to get this working.

PR #1713, opened 2023-11-28
LTD-4468: Add endpoint to provide suggestions for the search terms

saruniitr:
## Change description

To help users specify a data category when performing a search we are providing suggestions by checking which fields given search term matches with. This is done using multi match prefix query where prefix match is performed on each word. Highlight indicates where given search term is matched and these fields are returned as suggestions.

[LTD-4468](https://uktrade.atlassian.net/browse/LTD-4468)

[LTD-4468]: https://uktrade.atlassian.net/browse/LTD-4468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/search/product/urls.py
+++ b/api/search/product/urls.py
@@ -12,7 +12,7 @@ router.register(r"search", views.ProductDocumentView, basename="product_search")
 
 urlpatterns = [
     path("", include(router.urls)),
-    path("suggest/", views.ProductSuggestDocumentView.as_view(), name="product_suggest"),
+    path("suggest-poc/", views.ProductSuggestPoCDocumentView.as_view(), name="product_suggest_poc"),
     path("spire/<str:pk>/comment/", views.CommentView.as_view(), name="retrieve_spire_product_comment"),
     path("lite/<uuid:pk>/comment/", views.CommentView.as_view(), name="retrieve_lite_product_comment"),
     path("lite/<uuid:pk>/", views.RetrieveLiteProductView.as_view(), name="retrieve_lite_product"),
--- a/api/search/product/urls.py
+++ b/api/search/product/urls.py
@@ -12,6 +12,7 @@ router.register(r"search", views.ProductDocumentView, basename="product_search")
 
 urlpatterns = [
     path("", include(router.urls)),
+    path("suggest/", views.ProductSuggestDocumentView.as_view(), name="product_suggest"),
     path("suggest-poc/", views.ProductSuggestPoCDocumentView.as_view(), name="product_suggest_poc"),
     path("spire/<str:pk>/comment/", views.CommentView.as_view(), name="retrieve_spire_product_comment"),
     path("lite/<uuid:pk>/comment/", views.CommentView.as_view(), name="retrieve_lite_product_comment"),
--- a/api/search/product/views.py
+++ b/api/search/product/views.py
@@ -187,7 +187,7 @@ class ProductDocumentView(DocumentViewSet):
         return Response(serializer.data)
 
 
-class ProductSuggestDocumentView(APIView):
+class ProductSuggestPoCDocumentView(APIView):
     allowed_http_methods = ["get"]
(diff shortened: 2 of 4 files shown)


api/search/product/urls.py
```diff
@@ -13,6 +13,7 @@
 urlpatterns = [
     path("", include(router.urls)),
     path("suggest/", views.ProductSuggestDocumentView.as_view(), name="product_suggest"),
+    path("suggest-poc/", views.ProductSuggestPoCDocumentView.as_view(), name="product_suggest_poc"),
```
currycoder, line 16: "proof of concept"?  Do we expect this to stick around?

api/search/product/views.py
```diff
@@ -187,7 +187,70 @@ def list(self, request, *args, **kwargs):
...
 
-class ProductSuggestDocumentView(APIView):
+class ProductSuggestDocumentView(RetrieveAPIView):
+    authentication_classes = (GovAuthentication,)
+
+    def get(self, request):
+        query_str = self.request.GET.get("q", "")
+
+        # Use only most recent word to look for suggestions
+        query_str = query_str.split()[-1]
```
currycoder, line 197: If my querystring is "hi there arun" - this will use just "arun" right?

api/search/product/views.py
```diff
@@ -187,7 +187,70 @@ def list(self, request, *args, **kwargs):
...
+                    },
+                },
+            },
+            "_source": False,
+        }
+
+        search = ProductDocumentType.search().from_dict(query)
+        search._index = list(settings.ELASTICSEARCH_PRODUCT_INDEXES.values())
+        suggests = []
+        executed = search.execute()
```
currycoder, line 234: Maybe `response` could be better than `executed`?

api/search/product/views.py
```diff
@@ -187,7 +187,70 @@ def list(self, request, *args, **kwargs):
...
+
+        flat_suggestions = set()
+        for hit in executed.hits:
+            for field, value in hit.meta.highlight.to_dict().items():
+                value = value[0]
+                if value not in flat_suggestions:
+                    suggests.append(
```
currycoder, line 241: Is it worth creating a serializer to codify what the response format will be?

api/search/product/urls.py
saruniitr: No, I would like to remove this at the end as part of clean-up

api/search/product/views.py
saruniitr: yes that's correct.
I have more details about this in individual commit message,
https://github.com/uktrade/lite-api/pull/1713/commits/0b25031ff59bf36c4e65956a4ac5dd3fa95e029a

saruniitr: I will try to summarize here as well

currycoder (approved):

api/search/product/views.py
```diff
@@ -187,7 +187,77 @@ def list(self, request, *args, **kwargs):
...
+                        "ratings",
+                        "regimes",
+                        "report_summary",
+                        "organisation",
+                        "assessed_by",
+                        "consignee_country",
+                        "end_user_country",
+                        "ultimate_end_user_country",
+                        "assessment_note",
+                    ],
```
hnryjmes, line 224: these fields look similar to those defined here: https://github.com/uktrade/lite-api/blob/07e9ad24e2d1aa258fbd7b6ac310c10e8792f88d/api/search/product/views.py#L53-L65

is there a way we can re use this somehow, or create a new reusable data structure that can be used in both places?

I am just thinking about the added maintenance of having various lists of fields to update / amend as this functionality evolves

hnryjmes (approved):
approving as I'm satisfied all review feedback has been addressed

PR #1719, opened 2023-12-06
Multiple final advice debug

currycoder:
### Aim

TODO


api/letter_templates/context_generator.py
```diff
@@ -1024,9 +1025,9 @@ def _get_goods_context(application, final_advice, licence=None):
     final_advice = final_advice.filter(good_id__isnull=False)
     goods_context = {advice_type: [] for advice_type, _ in AdviceType.choices}
 
-    goods_on_application_dict = {
-        good_on_application.good_id: good_on_application for good_on_application in goods_on_application
-    }
+    goods_on_application_dict = defaultdict(lambda: [])
+    for good_on_application in goods_on_application:
+        goods_on_application_dict[good_on_application.good_id].append(good_on_application)
```
stuaxo, line 1030: If we only need the first one of these, perhaps we can `dict.setdefault`

```python
good_on_application_dict = {}
for good_on_application in goods_on_applications
    good_on_application_dict.setdefault(good_on_application.good_id, good_on_application)
```

And then you wouldn't need the `.pop(0)` later.

currycoder: Yes this is not obvious, but the whole point is to use `.pop(0)` to ensure that we make every GoodOnApplication present **once** in the end context.

I've added some comments that hopefully make that less subtle.

api/letter_templates/context_generator.py
```diff
@@ -1020,13 +1021,29 @@ def _get_good_on_licence_context(good_on_licence):
...
+
+    What would probably be better would be to start from a context datastructure from all the GoodOnApplication
+    objects that **we know** need to be present on the licence; e.g. application.goods.filter(is_good_controlled=True)
+
+    From there, it would probably be clearer to go through each of the Advice records, GoodOnLicence records etc
+    and hydrate those original GoodOnApplication objects.
+
+    Right now, we do things a little backwards and add records to the context datastructure/grab/rewrite/overwrite them.
+    That makes this quite hard to understand what is going on and very easy for bugs to manifest.
+    """
```
stuaxo, line 1036: ^ This is good context.

api/letter_templates/context_generator.py
```diff
@@ -1020,13 +1021,29 @@ def _get_good_on_licence_context(good_on_licence):
...
     final_advice = final_advice.filter(good_id__isnull=False)
     goods_context = {advice_type: [] for advice_type, _ in AdviceType.choices}
 
-    goods_on_application_dict = {
-        good_on_application.good_id: good_on_application for good_on_application in goods_on_application
-    }
+    # Create a mapping to get from Good.id to corresponding GoodOnApplication records
+    #  Note: we map to a list of GoodOnApplication records as there may be more than one
+    #  GoodOnApplication per Good for this application
+    good_ids_to_goods_on_application = defaultdict(lambda: [])
```
stuaxo, line 1044: Not a blocker - interesting, have never seen this style vs `defaultdict(list)`

api/letter_templates/context_generator.py
hnryjmes, line 1036: I had a lot of these same thoughts when working on this function. Anything else we can do as well as this (very helpful) comment?

api/letter_templates/context_generator.py
```diff
@@ -1039,9 +1056,13 @@ def _get_goods_context(application, final_advice, licence=None):
...
         final_advice = final_advice.exclude(type=AdviceType.APPROVE)
 
+    # Ensure that for each proviso final advice record, we add a record to the goods
+    #  context
     for advice in final_advice:
-        if advice.good_id in goods_on_application_dict:
-            good_on_application = goods_on_application_dict[advice.good_id]
+        if advice.good_id in good_ids_to_goods_on_application:
+            # Grab the next GoodOnApplication for this Good.id - this ensures that
+            #  each GoodOnApplication is present once on the end licence
```
hnryjmes, line 1064: this is maybe the only bit of this I'm a bit confused about... do you mean that each Good is present once on the end licence? I think the confusion comes from the meaning of having multiple GoAs for one Good, still not worked out what that means

hnryjmes: I mean the way it was done before now looks clearly wrong, so I am pretty much happy to approve it if you know this code solves the problem

api/letter_templates/context_generator.py
```diff
@@ -1020,13 +1021,29 @@ def _get_good_on_licence_context(good_on_licence):
...
+    """
     goods_on_application = application.goods.all().order_by("created_at")
     final_advice = final_advice.filter(good_id__isnull=False)
     goods_context = {advice_type: [] for advice_type, _ in AdviceType.choices}
 
-    goods_on_application_dict = {
-        good_on_application.good_id: good_on_application for good_on_application in goods_on_application
-    }
+    # Create a mapping to get from Good.id to corresponding GoodOnApplication records
+    #  Note: we map to a list of GoodOnApplication records as there may be more than one
```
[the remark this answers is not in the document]
currycoder: it won't help, the values are a list of unique items already

api/letter_templates/context_generator.py
currycoder: I'm hoping that this will be a signal next time round that it's time to really think about this and justify some time to rewrite.  We just don't currently have the room for that with various workstreams ending IMO

api/letter_templates/context_generator.py
currycoder: We'd need to fundamentally rewrite the function and be confident that it works across all the different documents that use it.  TBH, that is too much effort for what we have time for right now given the state of current workstreams/deadlines.

I'm absolutely open to any concrete examples, but this is the best that I can come up with.

stuaxo (approved):

hnryjmes (approved):
nice, this looks like a big improvement over what was there before

PR #1722, opened 2023-12-07
Remove unused product search code

hnryjmes:
Remove the unused proof of concept product search code from as much as 3 years ago. Although some of it was used for our current implementation of product search, most of it was not and can be removed.


api/search/product/views.py
```diff
@@ -243,119 +241,3 @@ def get(self, request):
...
-                            "value": option,
-                            "index": "spire" if "spire" in hit.meta.index else "lite",
-                        }
-                    )
-                    flat_suggestions.add(option)
-
-        return Response(suggests)
-
-
-class MoreLikeThisView(APIView):
```
saruniitr, line 304: Actually this is used/linked to product details tab in caseworker site.

PR #1724, opened 2023-12-08
improve  by adding  countersign_advice

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4502

There are timeout on this case due to unusual 299 products on inspection we found countersign_advice_ to be biggest culprit. These are the current metric improvements 

 22585 Q (202480 ms)
 countersign_advice__countersigned_user = 16594 Q (158557 ms) 
 countersign_advice__advice = 9099 q (11924)
 countersign_advice__advice__user = 7903 q (101582)


kevincarrogan (approved):

api/cases/views/views.py
```diff
@@ -133,10 +161,20 @@ def get(self, request, pk):
...
                 "case_officer",
                 "case_officer__team",
+                "organisation",
+                "submitted_by",
+                "status",
+                "sub_status",
+                "copy_of",
+                "baseapplication",
+                "compliancesitecase",
+                "compliancevisitcase",
```
saruniitr, line 171: these two and `opengenerallicencecase` below can be removed as we don't support those yet

saruniitr (approved):

api/cases/views/views.py
depsiatwal: removed and re-tested to ensure it has no impact

PR #1728, opened 2023-12-11
Cache queue objects on gov user serializer

kevincarrogan:
This caches the queue objects upfront so that this serializer doesn't need to do a queue lookup each time it's accessed

This is particularly useful when this serializer is called multiple times per page

### DIFF
--- a/api/gov_users/serializers.py
+++ b/api/gov_users/serializers.py
@@ -83,6 +83,21 @@ class GovUserViewSerializer(serializers.ModelSerializer):
     role = RoleListStatusesSerializer()
     default_queue = serializers.SerializerMethodField()
 
+    @classmethod
+    def _set_queue_cache(cls):
+        queue_cache = {}
+        for queue in Queue.objects.all():
+            queue_cache[str(queue.pk)] = queue
+        cls._queue_cache = queue_cache
+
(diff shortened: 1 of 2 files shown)


Tllew (approved):

depsiatwal (commented):
this has proven to work and tested on Dev2 just need test coverage. 
As discussed we can decide the urgency of retro of adding them depending on progress of this issue.

api/gov_users/serializers.py
```diff
@@ -83,6 +83,21 @@ class GovUserViewSerializer(serializers.ModelSerializer):
...
     default_queue = serializers.SerializerMethodField()
 
+    @classmethod
+    def _set_queue_cache(cls):
+        queue_cache = {}
+        for queue in Queue.objects.all():
+            queue_cache[str(queue.pk)] = queue
+        cls._queue_cache = queue_cache
+
+    @classmethod
```
stuaxo, line 93: You could use lru_cache and avoid having _set_queue_cache, it would look something like this:

```
@lru_cache(max_size=0)
@classmethod
def get_queue(cls, uuid):
    return cls._queue_cache[uuid]
```

hnryjmes (approved):

PR #1738, opened 2023-12-15
[LTD-4532] Remove warning log for 'CLE copied from Good'

currycoder:
### Aim

We have a warning log in our codebase which is triggered when a GoodOnApplication is saved with CLEs that appear to match the ones present in the parent Good.

The log appears to have been added to highlight some copying behaviour - but I can't see where in the API this "copy" may have been triggered.  I don't believe there is anywhere explicit in the API that this happens.

Possibly, it was supposed to highlight that a TAU officer used an exporter's suggestion for the CLEs.

In any case, the log is currently clogging up our alerts channel and is something we ignore as a team.  So best to be rid of it.

[LTD-4532](https://uktrade.atlassian.net/browse/LTD-4532)

[LTD-4532]: https://uktrade.atlassian.net/browse/LTD-4532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


saruniitr (approved):
one less alert

api/applications/models.py
```diff
@@ -563,23 +563,6 @@ def get_control_list_entries(self):
...
-    def save(self, *args, **kwargs):
-        """LTD-2541 - we want to flag when a GoodOnApplication object
-        is saved with properties that seem to be copied from the
-        associated Good.
-        """
-        super().save(*args, **kwargs)
-        cle = set(self.control_list_entries.all())
-        good_cle = set(self.good.control_list_entries.all())
-        if cle == good_cle and cle != set():
-            gona_copy_logger.warning(
```
kevincarrogan, line 575: I think this is the only place this was ever used so you can probably also remove the instantation of this logger on line 65 and then also get rid of the references to this in `settings.py`.

currycoder: Done

kevincarrogan (approved):

kevincarrogan (approved):

PR #1745, opened 2023-12-19
LTD-4546-enable-hawk-dev 

depsiatwal:
This PR is to switch on HAWK by default on API/Exporter/Caseworker 

The exporter keys are no longer extracted from Vault since their fault is the key name i,e 

LITE_EXPORTER_HAWK_KEY = LITE_EXPORTER_HAWK_KEY
LITE_INTERNAL_HAWK_KEY = LITE_INTERNAL_HAWK_KEY

These are the keys set on caseworker/exporter locally so setting these to default to keep things simple on local dev. 

Caseworker PR is here 
https://github.com/uktrade/lite-frontend/pull/1698

https://uktrade.atlassian.net/browse/LTD-4546


hnryjmes (approved):

local.env
```diff
@@ -52,10 +52,9 @@ UPLOAD_DOCUMENT_ENDPOINT_ENABLED=<<FROM_VAULT>>
 
 DJANGO_SECRET_KEY=<<FROM_VAULT>>
 
-HAWK_AUTHENTICATION_ENABLED=<<FROM_VAULT>>
 HAWK_LITE_DATA_WORKSPACE_KEY=<<FROM_VAULT>>
-LITE_EXPORTER_HAWK_KEY=<<FROM_VAULT>>
-LITE_INTERNAL_HAWK_KEY=<<FROM_VAULT>>
+LITE_EXPORTER_HAWK_KEY=LITE_EXPORTER_HAWK_KEY
+LITE_INTERNAL_HAWK_KEY=LITE_INTERNAL_HAWK_KEY
```
saruniitr, line 57: Besides this I think we need to set `HAWK_AUTHENTICATION_ENABLED` to `True` by default in settings file as well.

depsiatwal: no need since it's set here as default True 

https://github.com/uktrade/lite-api/blob/0b70b34423a0bbc7081d4924de1d5b5af6a29bd8/api/conf/settings.py#L21

kevincarrogan (approved):

PR #1747, opened 2023-12-20
When unsetting a report summary subject, the legacy report_summary fi…

stuaxo:
When unsetting a report summary subject, the legacy report_summary field was not being set.

On the frontend in TAU (and probably elsewhere) this manifests as the user seeing stale data, as the legacy field was used as fallback.

### DIFF
--- a/api/assessments/serializers.py
+++ b/api/assessments/serializers.py
@@ -70,15 +70,20 @@ class AssessmentSerializer(GoodControlReviewSerializer):
         list_serializer_class = AssessmentUpdateListSerializer
 
     def validate(self, data):
-        # If we have a report summary subject, overwrite whatever report_summary value
-        # we have with the string from the subject/prefix
-        if "report_summary_subject" in data:
+        if data.get("is_good_controlled") is False:
+            # Goods that are not controlled should have a blank report summary
+
+            data["report_summary"] = None
+            data["report_summary_prefix"] = None
+            data["report_summary_subject"] = None
+        elif "report_summary_subject" in data:
+            # If we have a report summary subject, overwrite whatever report_summary value
+            # we have with the string from the subject/prefix
             if data.get("report_summary_prefix") and data.get("report_summary_subject"):
                 data["report_summary"] = f"{data['report_summary_prefix'].name} {data['report_summary_subject'].name}"
             elif data.get("report_summary_subject"):
                 data["report_summary"] = data["report_summary_subject"].name
             else:
-                # Goods that are not controlled do not need a report summary
                 data["report_summary"] = None
         return data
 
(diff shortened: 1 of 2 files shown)


kevincarrogan:
I feel that this requires some tests.

api/assessments/serializers.py
```diff
@@ -72,11 +72,14 @@ class Meta:
...
-            if data.get("report_summary_prefix"):
+        if "report_summary_subject" in data:
+            if data.get("report_summary_prefix") and data.get("report_summary_subject"):
                 data["report_summary"] = f"{data['report_summary_prefix'].name} {data['report_summary_subject'].name}"
-            else:
+            elif data.get("report_summary_subject"):
                 data["report_summary"] = data["report_summary_subject"].name
+            else:
+                # Goods that are not controlled do not need a report summary
+                data["report_summary"] = None
```
kevincarrogan, line 82: Can we check more explicitly that this good is in fact not controlled instead of just relying on there not being any report summary values?

api/assessments/serializers.py
```diff
@@ -72,11 +72,14 @@ class Meta:
     def validate(self, data):
         # If we have a report summary subject, overwrite whatever report_summary value
         # we have with the string from the subject/prefix
-        if data.get("report_summary_subject"):
-            if data.get("report_summary_prefix"):
+        if "report_summary_subject" in data:
```
kevincarrogan, line 75: Is there a potential case where we may not send through the `report_summary_subject` and still want to reset the `report_summary` because it's NLR?

api/assessments/serializers.py
kevincarrogan: In fact, regardless of whether any report summary details are sent through should we just be setting this as blank if it's marked as NLR?

stuaxo: @kevincarrogan we should probably also clear the control list entries if someone unticks this the Licence Required, to be consistent.

stuaxo: Removed this last commit - this code is used in a lot of places + raising a validation error here might have other consequences.

api/assessments/serializers.py
stuaxo: Maybe ?   It's not possible in the TAU frontend, but I've added logic for this.

stuaxo: Explicit logic to clear these added for the NLA case, the existing test exercises this + needed a small update to pass.

hnryjmes (approved):

stuaxo:
To avoid needing a new string in lite-content, I've used `Picklists.REQUIRED_REPORT_SUMMARY` this is the same error, originally I had the new one under assessments for consistency, but this covers things without the need for a new string.

stuaxo:
> I feel that this requires some tests.

One of the existing tests was updated for this to pass, and a new one to cover the validation error we added.

kevincarrogan (approved):

stuaxo:
See https://github.com/uktrade/lite-routing/pull/233

PR #1748, opened 2023-12-20
LTD-4534: Update search signals to refresh Product document also

saruniitr:
## Change description

Updates the various models are intercepted in save() signal to refresh the indexes however it is only updating Application document.

When an application is first submitted by Exporter we run flagging rules which result in saving of underlying 'Good' instances. At this point the ProductDocument is not refreshed so if we search for a product in this application we see the reference_code as None as if it is still in draft state. Updating product document here pulls inner document for the Application with the updated reference code.

Without this change also eventually the reference code gets refreshed as the application progresses through various teams which result in updating of product attributes but if it is in submitted state then it doesn't get updated.

[LTD-4534](https://uktrade.atlassian.net/browse/LTD-4534)

[LTD-4534]: https://uktrade.atlassian.net/browse/LTD-4534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/search/tasks.py
+++ b/api/search/tasks.py
@@ -6,12 +6,12 @@ UPDATE_SEARCH_INDEX_QUEUE = "update_search_index_queue"
 
 
 @background(queue=UPDATE_SEARCH_INDEX_QUEUE, schedule=0)
-def update_search_index(model_name, *ids):
+def update_search_index(model_pk_pairs):
     """Update the search index with instances of a model as specified by
     the supplied model_name and ids.
     """
-    model = apps.get_model(model_name)
 
-    for id_ in ids:
-        instance = model.objects.get(pk=id_)
+    for model_name, pk in model_pk_pairs:
+        model = apps.get_model(model_name)
+        instance = model.objects.get(pk=str(pk))
         registry.update(instance)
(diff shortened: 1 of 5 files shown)


api/search/tests/test_signals.py
```diff
@@ -12,7 +12,7 @@ class UpdateApplicationDocumentTest(DataTestClient):
     def test_standard_application(self, mock_task):
         application = self.create_standard_application_case(self.organisation)
 
-        mock_task.assert_has_calls([call("applications.BaseApplication", str(application.pk))])
+        assert [call([("applications.BaseApplication", application.pk)])] in mock_task.call_args_list
```
saruniitr, line 15: there will be many other calls hence checking for the presence of expected one in each test

kevincarrogan (approved):

depsiatwal (commented):
looks sensible however tests are failing

api/search/tasks.py
```diff
@@ -6,12 +6,12 @@
 
 
 @background(queue=UPDATE_SEARCH_INDEX_QUEUE, schedule=0)
-def update_search_index(model_name, *ids):
+def update_search_index(model_pk_pairs):
```
hnryjmes, line 9: is this better than doing iterable unpacking?

saruniitr: Not necessarily.
Previously all arguments were for the same model but not it can be different hence I am passing them as pairs.

hnryjmes (approved):

PR #1749, opened 2023-12-20
LTD-4543-django-4-2-upgrade 

depsiatwal:
Upgrade of lite api Django to 4.2 LTE 

django-staff-sso-client  had to be upgrades as a result as previous version was incompatible. 

https://uktrade.atlassian.net/browse/LTD-4543


api/cases/managers.py
```diff
@@ -1,7 +1,7 @@
 from datetime import datetime
 from typing import List
 
-from compat import get_model
+from django.apps import apps
```
seijihg, line 4: this is interesting the django-compat never got removed before this upgrade. We dont even use Django 1.xx

seijihg (dismissed):

hnryjmes (approved):

saruniitr:
Kept in draft to avoid accidental merging.

currycoder (approved):

PR #1751, opened 2023-12-27
origin/LTD-4562-migrate-scan-document-for-viruses-celery

depsiatwal:
This is the first migration of the current django-backend to celery.

I thought about adding a config for some of the retries but then decided against it as each will probably be unique and I don't see we will need to override. This may change in the next conversion though. 

Also i've set following for tests
CELERY_ALWAYS_EAGER
CELERY_TASK_STORE_EAGER_RESULT

This normally goes against best practices since it move away real live scenario (i.e race conditions) but figured out tasks are relatively simple  and are shared_tasks so impact was low. 

https://uktrade.atlassian.net/browse/LTD-4562


api/documents/celery_tasks.py
```diff
@@ -0,0 +1,44 @@
...
+        document = Document.objects.select_for_update(nowait=True).get(id=document_id)
+        if document.virus_scanned_at:
+            logging.info(f"Document '{document_id}' has already been scanned; safe={document.safe}")
+            return
+
+        try:
+            document.scan_for_viruses()
+        except Exception as exc:  # noqa
+            logging.exception(exc, document)
+            raise Exception(exc)
```
kevincarrogan, line 35: You can just call `raise` here on its own instead of specifying an `Exception`.

api/documents/celery_tasks.py
```diff
@@ -0,0 +1,44 @@
...
+
+        document = Document.objects.select_for_update(nowait=True).get(id=document_id)
+        if document.virus_scanned_at:
+            logging.info(f"Document '{document_id}' has already been scanned; safe={document.safe}")
+            return
+
+        try:
+            document.scan_for_viruses()
+        except Exception as exc:  # noqa
+            logging.exception(exc, document)
```
kevincarrogan, line 34: I'm not sure this is the right parameters you should be passing to `logging.exception`.

If you're using `logging.exception` you'll already get the exception details so you don't need to pass it in. Also, the other arguments are used to be formatted into the log message, which at the moment is an exception so I think this might even break.

api/conftest.py
```diff
@@ -123,3 +127,14 @@ def apply(self, app, migrate_to):
...
+
+@pytest.fixture(autouse=True)
+def celery_app():
+    # Setup the celery worker to run in process for tests
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "api.conf.settings")
+    celeryapp = Celery("api")
+    celeryapp.autodiscover_tasks(related_name="celery_tasks")
+    celeryapp.conf.update(CELERY_ALWAYS_EAGER=True)
+    celeryapp.conf.update(CELERY_TASK_STORE_EAGER_RESULT=True)
+    return celeryapp
```
kevincarrogan, line 140: Now that we have this do we need to be mocking `api.documents.celery_tasks.scan_document_for_viruses.apply_async` everywhere? Surely we can just let that run, the bit we actually need to mock now is the call to the external service itself.

api/documents/tests/test_celery_tasks.py
kevincarrogan, line 1: I feel too much might be being mocked here. Can't we just let a lot of this run now and just mock out the call to the external service?

I also don't see a test that actually checks for the happy path where the file is scanned and comes back without any issues.

api/conftest.py
depsiatwal: Ealier I had this set to just  api.documents.celery_tasks.scan_document_for_viruses and the tests were happy however the issue with this is that the actual getting the S3 document was failing in CI and had the side effect of the attempt of the document delete.  

So either we remove the mocks or we should wait until we have the ability to mock the S3 bucket.

api/documents/tests/test_celery_tasks.py
depsiatwal: Same reason above happy path isn't fully implemented for getting document and AV service, 
We are testing test_document_scan_document_for_viruses_called which is the happy path for the celery task
I could check the functions in document are called but then feel i'm actually testing the document functions as opposed to the celery functions.

api/documents/celery_tasks.py
depsiatwal: updated

api/documents/celery_tasks.py
depsiatwal: corrected

api/conftest.py
depsiatwal: ohh your saying we should mock the external services.

depsiatwal: all api.documents.celery_tasks.scan_document_for_viruses.apply_async have been replaced to mock only the S3 and virus scan operations meaning we run the code deeper.

api/documents/tests/test_celery_tasks.py
depsiatwal: add happy path now

api/documents/celery_tasks.py
```diff
@@ -0,0 +1,44 @@
...
+    max_retries=MAX_ATTEMPTS,
+    retry_backoff=RETRY_BACKOFF,
+)
+def scan_document_for_viruses(self, document_id):
+    """
+    Scans documents for viruses
+    :param document_id:
+    """
+    with transaction.atomic():
+        logging.info(f"Fetching document '{document_id}'")
```
saruniitr, line 24: f-strings are readable but usually not advised in logging methods as these are calculated immediately and depending on the loglevel we may not even emit this event. Old style formatting is used because they get deferred till the time the log is actually emitted.

api/documents/tests/test_celery_tasks.py
```diff
@@ -0,0 +1,71 @@
...
+    @mock.patch("api.documents.libraries.av_operations.scan_file_for_viruses")
+    def test_document_scan_document_success(self, mock_virus_scan, mock_s3_operations_get_object):
+        mock_s3_operations_get_object.return_value = {"doc_data"}
+        mock_virus_scan.return_value = False
+        document = self.create_case_document(case=self.case, user=self.gov_user, name="jimmy")
+        self.assertIsNone(document.virus_scanned_at)
+
+        scan_document_for_viruses(str(document.id))
+        document.refresh_from_db()
+        self.assertIsNotNone(document.virus_scanned_at)
```
saruniitr, line 27: Can we also check `safe=True` here?

depsiatwal: done

api/documents/celery_tasks.py
depsiatwal: updated

saruniitr (approved):
With Kevin's comments addressed it looks good to me.

api/documents/tests/test_celery_tasks.py
```diff
@@ -0,0 +1,73 @@
...
+
+class DocumentVirusScan(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        self.case = self.create_standard_application_case(self.organisation)
+
+    @mock.patch("api.documents.libraries.s3_operations.get_object")
+    @mock.patch("api.documents.libraries.av_operations.scan_file_for_viruses")
+    def test_document_scan_document_success(self, mock_virus_scan, mock_s3_operations_get_object):
+        mock_s3_operations_get_object.return_value = {"doc_data"}
```
kevincarrogan, line 20: We're expecting this to return a set?

api/organisations/tests/test_documents.py
```diff
@@ -19,8 +19,11 @@ def create_document_on_organisation(self, name):
         }
         return self.client.post(url, data, **self.exporter_headers)
 
-    @mock.patch("api.documents.tasks.scan_document_for_viruses.now", mock.Mock)
-    def test_create_organisation_document(self):
+    @mock.patch("api.documents.libraries.s3_operations.get_object")
+    @mock.patch("api.documents.libraries.av_operations.scan_file_for_viruses")
+    def test_create_organisation_document(self, mock_virus_scan, mock_s3_operations_get_object):
+        mock_s3_operations_get_object.return_value = {"doc_data"}
+        mock_virus_scan.return_value = False
```
kevincarrogan, line 26: The fact that this is copied so frequently makes me feel that there's maybe a higher level fixture to be created here. I say this because I also don't think you mean to be returning a set in `mock_s3_operations_get_object.return_value` and you're potentially going to have to change this in multiple places.

kevincarrogan: Actually, we're not using pytest here so we can't build a higher level fixture, but it certainly feels like there might be something we could do instead of having to copy this over and over again.

api/documents/tests/test_celery_tasks.py
depsiatwal: for purpose of test it's fine as long as it's not null else code throws an exception

api/organisations/tests/test_documents.py
depsiatwal: yeah issue we have here is we are using Text Classes so higher level fixtures aren't allowed.  only mocking

We did have a higher level mock i.e One level up at document.scan_for_virus and felt it was too high level. 
hence we moved the mocks down to only external opps which is 
getting document from s3 bucket
scan document

kevincarrogan (approved):

PR #1755, opened 2024-01-02
LTD-4563-update-sanction-search-index-celery

depsiatwal:
Sanctions update task job migrating away from django-background tasks to Celery 

https://uktrade.atlassian.net/browse/LTD-4563


api/conf/settings.py
```diff
@@ -274,7 +274,9 @@ def _build_redis_url(base_url, db_number, **query_args):
     CELERY_BROKER_URL = _build_redis_url(REDIS_BASE_URL, REDIS_CELERY_DB, **url_args)
     CELERY_RESULT_BACKEND = CELERY_BROKER_URL
 
+CELERY_ALWAYS_EAGER = env.bool("CELERY_ALWAYS_EAGER", False)
```
currycoder, line 277: I thought the setting was `CELERY_TASK_ALWAYS_EAGER` - the line below.  Do we need this line?

docker-compose.e2e.yml
```diff
@@ -36,7 +36,7 @@ services:
       - elasticsearch
       - redis
     entrypoint: dockerize -wait tcp://db:5432 -wait tcp://api:8100
-    command: celery -A api.conf worker -l info
+    command: bash -c "celery -A api.conf worker -l info & celery -A api.conf beat"
```
currycoder, line 39: Maybe it would be better to have two containers instead of executing two scripts simultaneously in a single container.  I assume this will better match production.

docker-compose.yml
```diff
@@ -43,7 +43,7 @@ services:
       - redis
     depends_on:
       - api
-    command: pipenv run watchmedo auto-restart -d . -R -p '*.py' -- celery -A api.conf worker -l info
+    command:  bash -c "pipenv run watchmedo auto-restart -d . -R -p '*.py' -- celery -A api.conf worker -l info & pipenv run watchmedo auto-restart -d . -R -p '*.py' -- celery -A api.conf beat"
```
currycoder, line 46: As above; Maybe it would be better to have two containers instead of executing two scripts simultaneously in a single container.  I assume this will better match production.

currycoder (commented):
Have made some comments throughout.

One additional thing though, I don't think this scheduled task will actually run in production right now!  For that I think we need some changes to the `Procfile` - would suggest adding an extra worker on a new line for this.

api/conf/settings.py
depsiatwal: yeah I played for this for quite some time and e2e tests where failing when we didn't have CELERY_ALWAYS_EAGER=False Hence why we override it in FE api.config 

I was under the same conclusion CELERY_TASK_ALWAYS_EAGER replaces CELERY_ALWAYS_EAGER but when I was running e2e tasks it had no impact and they kept failing.

docker-compose.yml
depsiatwal: Agree this one should defo be shifted

Tllew (approved):

depsiatwal:
> Have made some comments throughout.
> 
> One additional thing though, I don't think this scheduled task will actually run in production right now! For that I think we need some changes to the `Procfile` - would suggest adding an extra worker on a new line for this.

Done

api/conf/settings.py
stuaxo: I guess the FE tasks were failing because the they expect to be able to access the data that has changed as a result of the celery task and it hasn't run yet (celery runs it at some time in the future, usually that seems instant to us).

It sounds like running the tasks eagerly is probably about right, but it may be worth coming back to the tests and checking things are as expected - in some places we may need to change code to do something different if the task they have started isn't complete.

depsiatwal: At the moment the document link ins't available until the document has successfully been scanned and then become available when scanned. We definitely have unit tests to cover both scenarios but doubt we have e2e tests to cover both. Don't think they're that granular atm and cover mainly the larger happy path scenarios.

currycoder (approved):

PR #1757, opened 2024-01-03
LTD-4577 Legacy GoodOnApplication instances may have a report_summary but not …

stuaxo:
Legacy GoodOnApplication instances can have only a report_summary, but no report_summary_prefix or report_summary_subject - so don't error out on these.

### DIFF
--- a/api/assessments/serializers.py
+++ b/api/assessments/serializers.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.utils import timezone
 from rest_framework import serializers
 from rest_framework.relations import PrimaryKeyRelatedField
@@ -14,6 +16,8 @@ from api.staticdata.regimes.models import RegimeEntry
 from api.staticdata.statuses.enums import CaseStatusEnum
 
 from lite_content.lite_api import strings
+
+logger = logging.getLogger(__name__)
 
 
 class AssessmentUpdateListSerializer(serializers.ListSerializer):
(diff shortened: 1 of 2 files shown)


stuaxo:
The two e2e failing look like they are not related to the code on here, would be nice to get confirmation.

api/assessments/serializers.py
```diff
@@ -83,7 +97,7 @@ def validate(self, data):
             # we have with the string from the subject/prefix
             if data.get("report_summary_prefix") and data.get("report_summary_subject"):
                 data["report_summary"] = f"{data['report_summary_prefix'].name} {data['report_summary_subject'].name}"
-            else:
+            elif data.get("report_summary_prefix"):
```
currycoder, line 100: I don't think I follow this change.  The point is that if we have both prefix and subject we construct the report_summary by concatenating them (above), otherwise we just use the subject.

stuaxo: Same with fresh eyes -

Separated out the legacy format so it's easier to remove this later.

Legacy = a report summary, no subject, no prefix.

Non legacy:

Must have a subject.

Use "prefix subject" if prefix is present, otherwise "subject"

stuaxo: (It's tempting to remove the exception handler -  thought I think we should be good now, in effect we are only catching non-legacy GOA where there is a prefix but no subject).

currycoder (approved):

saruniitr (approved):

stuaxo:
I missed that coverage wasn't happy.

stuaxo:
Had to re-run tests a few times before the flakey ones would pass..

PR #1758, opened 2024-01-03
Add management command to create test applications

saruniitr:
## Change description

This command can be used to seed local database with certain number of test applications.

It first creates a configurable number of product catalogue and organisations and creates applications with random number of products (selected from the catalogue) and submit these applications.

Note that it deletes all of existing data before it creates new data. For this reason it is included as part of new `testing` app which is only installed when `DEBUG=True`.

It also adds product assessments with random ARS subject and prefix values. It makes certain assumptions on the probablity of prefix vs subject values.

It is not perfect in terms of creating an perfect application but can quickly seed specified number of applications which can be handy for local env.


hnryjmes (approved):
This is good, I've been using an older version of this code which I copy-pasted from somewhere a while ago

I notice that this doesn't do anything with setting Flags on cases, and wonder if that's something that could be added in future

saruniitr:
> This is good, I've been using an older version of this code which I copy-pasted from somewhere a while ago
> 
> I notice that this doesn't do anything with setting Flags on cases, and wonder if that's something that could be added in future

correct, because the CLEs applied are random there may not be any flags.
We can definitely make it add random flags.

api/support/management/commands/create_test_data.py
```diff
@@ -0,0 +1,225 @@
...
+    "wave guide",
+    "Sulphuric acid",
+    "Hydrogen Fluoride",
+    "Motor assembly",
+    "Bolt action rifle",
+    "Ammonium hydrogen",
+    "piston",
+    "Imaging Camera",
+    "Sodium Chloride",
+    "Di-Hydrogen Oxide",
```
Tllew, line 77: Huge fan of some of these

Tllew (approved):
Looks good to me

api/support/management/commands/create_test_data.py
stuaxo: Stay hydrated

depsiatwal (commented):
not sure support is the best place for these 
as we commonly run support commands 

imagine somebody executing by accident in Prod 

Actually I wonder if we should only make this command visible or runnable only when Debug=True

saruniitr:
> not sure support is the best place for these as we commonly run support commands
> 
> imagine somebody executing by accident in Prod
> 
> Actually I wonder if we should only make this command visible or runnable only when Debug=True

Very good point.
I was not sure of the correct place for this command.

I will add conditions to only make it run when `Debug` is True. If possible I will check for env name and only allow it to run locally.

api/support/management/commands/create_test_data.py
```diff
@@ -0,0 +1,235 @@
...
+        # Create certain number of products for this organisation which
+        # will be used in various applications
+        for i in range(NUMBER_OF_PRODUCTS_IN_CATALOGUE):
+            organisation = random.choice(self.organisations)  # nosec
+            random.shuffle(product_names)  # nosec
+            name = random.choice(product_names)  # nosec
+            self.products.append(GoodFactory(name=name, organisation=organisation))
+
+    def handle(self, *args, **options):
+        env = settings.ENV
```
depsiatwal, line 159: will we need tests for these ?

saruniitr: No, we usually don't add tests for management commands

stuaxo:
> > not sure support is the best place for these as we commonly run support commands
> > imagine somebody executing by accident in Prod
> > Actually I wonder if we should only make this command visible or runnable only when Debug=True
> 
> Very good point. I was not sure of the correct place for this command.
> 

Firstly - thanks for this it's been really handy.

When I've been running my own fork of this, I've been running it in a Jupyter notebook, but that's just my own preference.

Your suggestion of checking for DEBUG is definitely a pointer in the right direction, I guess if this lived in it's own app then that could be only added when DEBUG is set.

depsiatwal:
> > > not sure support is the best place for these as we commonly run support commands
> > > imagine somebody executing by accident in Prod
> > > Actually I wonder if we should only make this command visible or runnable only when Debug=True
> > 
> > 
> > Very good point. I was not sure of the correct place for this command.
> 
> Firstly - thanks for this it's been really handy.
> 
> When I've been running my own fork of this, I've been running it in a Jupyter notebook, but that's just my own preference.
> 
> Your suggestion of checking for DEBUG is definitely a pointer in the right direction, I guess if this lived in it's own app then that could be only added when DEBUG is set.

I think that might be an overkill perhaps when we have more then a couple of management commands of this nature. 
I think at this stage just need to prevent accident execution in Production.

kevincarrogan (approved):

PR #1759, opened 2024-01-04
[LTD-4560] Move SLA update background task to celery

currycoder:
### Aim

Move background task for updating case SLAs to celery.

This has been tested on demo env (by temporarily forcing the schedule to another time and observing it run successfully).

[LTD-4560](https://uktrade.atlassian.net/browse/LTD-4560)

[LTD-4560]: https://uktrade.atlassian.net/browse/LTD-4560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal (approved):
LGTM
one to note but i'm wondering if we need is_weekend anymore since celery allows running mon-fri only

api/cases/celery_tasks.py
```diff
@@ -17,9 +17,8 @@
 from api.staticdata.statuses.models import CaseStatus
 
 # DST safe version of midnight
-SLA_UPDATE_TASK_TIME = time(22, 30, 0)
 SLA_UPDATE_CUTOFF_TIME = time(18, 0, 0)
-LOG_PREFIX = "update_cases_sla background task:"
+LOG_PREFIX = "update_cases_sla celery task:"
```
kevincarrogan, line 21: I know this is mostly just keeping things as is but it does feel like this is just trying to replicate what a specific logger made via `getLogger(__name__)` would do.

kevincarrogan (approved):

api/cases/celery_tasks.py
```diff
@@ -1,7 +1,7 @@
 import logging
```
stuaxo, line 1: This might want to use celery logging

https://docs.celeryq.dev/en/stable/reference/celery.app.log.html

currycoder: I'm fairly sure this works as-is.  What in particular should I adjust?

kevincarrogan: I've been doing some digging around to look at what Celery says in its documentation about logs.

https://docs.celeryq.dev/en/stable/userguide/tasks.html#logging i.e. using `logger = get_task_logger(__name__)` seems to be something that we might find useful.

currycoder: That looks good - how about I create a follow-up task where we can apply this pattern to all the celery tasks in lite-api?

PR #1763, opened 2024-01-05
LTD-4561: Migrate task that sends licence details to lite-hmrc to Celery

saruniitr:
## Change description

When a licence is issued/updated/re-issued we send these licence details to lite-hmrc by making an API call.

Previously we were using background_task to schedule a task for this. Migrate this task to now use Celery shared task.

[LTD-4561](https://uktrade.atlassian.net/browse/LTD-4561)

[LTD-4561]: https://uktrade.atlassian.net/browse/LTD-4561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/licences/celery_tasks.py
```diff
@@ -0,0 +1,51 @@
...
+    max_retries=MAX_ATTEMPTS,
+    retry_backoff=RETRY_BACKOFF,
+)
+def send_licence_details_to_lite_hmrc(licence_id, action):
+    """
+    Sends licence details to lite-hmrc
+
+    :param licence_id:
+    :param licence_reference:
+    """
```
hnryjmes, line 26: I think this is out of date, so would suggest either updating it or removing it completely if it's not adding value

hnryjmes (approved):

api/licences/apps.py
```diff
@@ -1,33 +1,5 @@
...
 
 class LicencesConfig(AppConfig):
     name = "api.licences"
-
-    def initialize_background_tasks(self, **kwargs):
-        if settings.LITE_HMRC_INTEGRATION_ENABLED:
-            self.schedule_not_sent_licences()
-
-    @staticmethod
-    def schedule_not_sent_licences():
```
currycoder, line 16: Just wondering what the rationale for removing this is?  I'm assuming that we now have this covered off elsewhere

api/licences/tests/test_api_to_hmrc_integration.py
```diff
@@ -431,68 +394,68 @@ def setUp(self):
...
-        schedule_licence_for_hmrc_integration.assert_not_called()
+        schedule_licence_details_to_lite_hmrc.assert_not_called()
 
     @override_settings(LITE_HMRC_INTEGRATION_ENABLED=True)
-    @mock.patch("api.licences.tasks.schedule_licence_for_hmrc_integration")
-    def test_licence_surrender_calls_schedule_licence_for_hmrc_integration(self, schedule_licence_for_hmrc_integration):
-        schedule_licence_for_hmrc_integration.return_value = None
+    @mock.patch("api.licences.celery_tasks.schedule_licence_details_to_lite_hmrc")
+    def test_licence_surrender_calls_schedule_licence_details_to_lite_hmrc(self, schedule_licence_details_to_lite_hmrc):
+        schedule_licence_details_to_lite_hmrc.return_value = None
```
currycoder, line 421: I believe what we agreed in dev meet this morning was to allow tests like this to call all the way through to the celery task - and only mock out external dependencies for the celery task itself; e.g. the only mocking the API call to send the licence in this case.

api/licences/celery_tasks.py
```diff
@@ -0,0 +1,51 @@
...
+    if (
+        licence.status == LicenceStatus.ISSUED
+        and action == HMRCIntegrationActionEnum.INSERT
+        and licence.goods.count() == 0
+    ):
+        logging.info(
+            "Licence %s contains no licenceable goods, skipping sending details to lite-hmrc",
+            licence.reference_code,
+        )
+        return
```
currycoder, line 48: Could you point out the test that proves that the schedule function will avoid queueing the task under these conditions?  I'm sure it exists but I'm having some trouble finding it!

api/licences/celery_tasks.py
```diff
@@ -0,0 +1,51 @@
...
+MAX_ATTEMPTS = 5
+RETRY_BACKOFF = 1200
+
+
+@shared_task(
+    autoretry_for=(Exception,),
+    max_retries=MAX_ATTEMPTS,
+    retry_backoff=RETRY_BACKOFF,
+)
+def send_licence_details_to_lite_hmrc(licence_id, action):
```
currycoder, line 20: This is all so much neater without background tasks - nice one!

currycoder (commented):
A few comments/questions inline

api/licences/celery_tasks.py
stuaxo: ^ It's good to have a narrower field than Exception.

If the error is because of a bug in our code we shouldn't retry.  

If an exception happens because we can't connect to a remote server / some database, then these are fair game to retry.

api/licences/tests/test_api_to_hmrc_integration.py
saruniitr: ok, these are the tests that only test whether the schedule function is called or not.
Actually we have other set of tests which only mock the API call. I thought of removing these but all of these are existing tests so retained them.
https://github.com/uktrade/lite-api/blob/dev/api/licences/tests/test_api_to_hmrc_integration.py#L677

api/licences/celery_tasks.py
saruniitr: Yes relevant test is here,
https://github.com/uktrade/lite-api/blob/dev/api/licences/tests/test_api_to_hmrc_integration.py#L699

This condition was added long back I think more as a precautionary because during normal process this situation won't arise.
We only create `Licence` if there are atleast one licenceable goods. If all are NLR or refusal then we don't create `Licence`. It can also happen if case moved back and then assessment is changed to NLR etc.

It is not possible to create this situation in the test because if we give NLR assessment we get an error when we try to create a draft licence which is why in the test it just deletes the goods on the licence.

api/licences/celery_tasks.py
saruniitr: updated to use `HMRCIntegrationException` only

api/licences/apps.py
saruniitr: My understanding was this is only required for background_tasks but realised that it is required even in celery case also now.

reverted these changes.

stuaxo (approved):

currycoder (approved):

PR #1764, opened 2024-01-08
Ltd 4564 update search index celery

Tllew:
### Aim

Updated search index to use celery

[LTD-4564](https://uktrade.atlassian.net/browse/LTD-4564)


api/search/celery_tasks.py
```diff
@@ -1,11 +1,19 @@
...
+from celery import shared_task
 from django.apps import apps
 from django_elasticsearch_dsl.registries import registry
 
 UPDATE_SEARCH_INDEX_QUEUE = "update_search_index_queue"
 
 
-@background(queue=UPDATE_SEARCH_INDEX_QUEUE, schedule=0)
+MAX_ATTEMPTS = 3
+RETRY_BACKOFF = 180
```
hnryjmes, line 9: can you say more about how you arrived at this number

hnryjmes: I might have misunderstood what this is doing but I know that the search index can take ~5 minutes to run sometimes, maybe that matters I'm not sure

api/search/tests/test_signals.py
```diff
@@ -74,7 +74,7 @@ def test_party(self, mock_task):
         party.save()
 
         self.assertCallsAreJSONEncodable(mock_task)
-        mock_task.assert_any_call(
+        mock_task.delay.assert_any_call(
```
currycoder, line 77: I think according to what we agreed in dev meet today, we actually want to let this call through without mocking the task `delay()` function.  We agreed instead to let these types of tests call all the way through and mock things that the task itself calls appropriately - in this case that would probably be the elasticsearch registry function

currycoder (commented):
Looks good - just one comment on tests which is hot off the press from dev meet today.

api/search/celery_tasks.py
```diff
@@ -1,11 +1,19 @@
...
+
+
+@shared_task(
+    autoretry_for=(Exception,),
```
stuaxo, line 13: Exception is a bit wide, it's good to retry on exceptions we expect at runtime - e.g. due to connectivity issues, and let errors in our coding raise exceptions.

api/search/celery_tasks.py
Tllew: This is just the value we've set in other celery_tasks to be honest.

api/search/celery_tasks.py
Tllew: Any recommendations on this?

We've used Exception elsewhere, not sure why the exception would be if it was strictly due to celery

api/search/celery_tasks.py
saruniitr: This task is not for rebuilding all indexes, this is only to update a document in an index whenever a model instance is updated in the database.
Eg when an instance of `GoodOnApplication` is updated then we get this signal and we schedule a task to update relevant document in the index. So it will just take couple of secs.

api/search/celery_tasks.py
stuaxo: My fault for not being clear in the comment.

I've given the same feedback on those other PRs.

Don't catch things due to celery, but let's say that sometimes you get errors because elasticsearch is down and they raise some particular Exception then I'd add those (I don't know this code well enough to name the exact exceptions).

In a different situation, if you were calling some remote API with the Requests module the ones to retry on would be stuff like timeouts, or connection error (to the remote API).

I guess the rule-of-thumb is:   retry on exceptions that may not occur the during one of the retries.

stuaxo: It's could be worth stopping ES running the task and see if any Connection errors surface.

If it's not obvious what the intermittent failure modes are, it may be best to have retries disabled + we can add them when if they occur.

Tllew: I used ConnectionError and ConnectionTimeout for ES but think that the default retry Exception might be more fitting for now incase we miss something unexpected within celery or ES

api/search/tests/test_signals.py
```diff
@@ -9,90 +9,58 @@
 
 
 class UpdateApplicationDocumentTest(DataTestClient):
-    def assertCallsAreJSONEncodable(self, mock_func):
-        """Check that the args and kwargs for a mock function are all JSON
```
saruniitr, line 13: this was recently added to fix an issue, is this not required in case of celery?

saruniitr (approved):

depsiatwal (commented):
small comment more of a question else LGTM

api/search/signals.py
```diff
@@ -58,11 +58,5 @@ def update_search_documents(sender, **kwargs):
...
     if to_update:
-        update_task = update_search_index
-
-        if not settings.BACKGROUND_TASK_ENABLED:
-            update_task = update_search_index.now
-
         to_update = [(model_name, str(pk)) for model_name, pk in to_update]
-
-        update_task(to_update)
+        update_search_index.delay(to_update)
```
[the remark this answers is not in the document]
saruniitr: it is same as calling `apply_async()` and allow us to specify arguments as a regular function call instead of as a list (in case of using `apply_async()` directly)

currycoder (approved):

PR #1765, opened 2024-01-08
[LTD-4560] Experiment with running celery worker in tests

currycoder:
### Aim

Experiment with running celery worker during tests.

[LTD-4560](https://uktrade.atlassian.net/browse/LTD-4560)

[LTD-4560]: https://uktrade.atlassian.net/browse/LTD-4560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/tests/test_sla.py
+++ b/api/cases/tests/test_sla.py
@@ -1,3 +1,5 @@
+import pytest
+
 from datetime import time, datetime
 from unittest import mock
 from unittest.mock import patch
@@ -38,7 +40,7 @@ def _set_submitted_at(case, time, date=timezone.localtime()):
 
 
 def run_update_cases_sla_task():
-    return update_cases_sla.apply().get()
+    return update_cases_sla.apply_async().get()
 
 
 class SlaCaseTests(DataTestClient):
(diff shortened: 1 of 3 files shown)


api/cases/tests/test_sla.py
```diff
@@ -38,7 +40,7 @@ def _set_submitted_at(case, time, date=timezone.localtime()):
 
 
 def run_update_cases_sla_task():
-    return update_cases_sla.apply().get()
+    return update_cases_sla.apply_async().get()
```
stuaxo, line 43: (Obviously not a blocker - probably not even a suggestion) -
Interesting, I normally spell this `.delay` since it makes passing arguments feel more natural, but it doesn't matter here as there aren't any :)

stuaxo:
Re: using redis.  We certainly can do this to make things close to E2E, a lot of the time tests use one of the other backends that is temporary, like :memory: if running things in process, or one of the others if running actual celery workers.

currycoder:
Closing as the spike is now concluded.

PR #1766, opened 2024-01-08
LTD-4600-update-api-model-changes-rfi 

depsiatwal:
This PR satisfies 2 PRs

https://uktrade.atlassian.net/browse/LTD-4600

Update the model on responded_by on ECJU queries since both exporter/govuser can now update ECJU queries we are changing the model to allow both. 

From Spike
https://uktrade.atlassian.net/browse/LTD-4590

Also 
https://uktrade.atlassian.net/browse/LTD-4575
Update Notes and timelines when a Caseworker closes a query. 

This PR can be merged independently since the model changes have no impact on current features and the N&T will only get executed when a caseworker is able to close a query in the UI.


stuaxo:
TIL about BaseUser..

api/cases/views/search/service.py
```diff
@@ -271,7 +271,8 @@ def populate_ecju_queries(case_map):
             "raised_by_user",
             "responded_by_user",
             "raised_by_user__baseuser_ptr",
-            "responded_by_user__baseuser_ptr",
+            "responded_by_user__govuser",
+            "responded_by_user__exporteruser",
```
hnryjmes, line 275: ah, this is the thing I couldn't figure out, nice one

hnryjmes (approved):
approving - I think the idea you said was to keep this in draft and then test together with my frontend changes on an environment whenever they're ready?

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -302,6 +302,25 @@ def test_exporter_responds_to_query(self, query_type):
         self._test_exporter_responds_to_query(False, query_type)
         self._test_exporter_responds_to_query(True, query_type)
 
+    def test_caseworker_manually_closes_query(self):
```
saruniitr, line 305: Once caseworker responds, I guess the option to respond for Exporter won't be available?

depsiatwal: ahh yes added another test for that so it can't happen

saruniitr (approved):

PR #1771, opened 2024-01-11
 LTD-4600-update-api-model-changes-rfi

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4605

Add a new endpoint to use in the UI. This is used in the UI to get a count of open queries. 
It will be used to block finalisation or display red dot to exporter/caseworker. 

We could have used the existing RFI Get API method but this returns all open queries and data so this is more streamlined. 

This PR can be merged since it's a new endpoint and has no impact on current functionality.


api/cases/urls.py
```diff
@@ -85,6 +85,11 @@
         views.EcjuQueryDocumentDetail.as_view(),
         name="case_ecju_query_document_detail",
     ),
+    path(
+        "<uuid:pk>/ecju-queries-open-count/",
+        views.ECJUQueriesOpenCount.as_view(),
+        name="case_ecju_query_open_count",
+    ),
```
saruniitr, line 92: Do we really need a separate endpoint for this?

We can get query count for case using `case.case_ecju_query.count()`

depsiatwal: Where are you proposing that ?
Just to add further context here we trying to avoid making case payload any heavier then it needs to be. Hence trying to make things a little more atomic and hence a separate call from FE.

depsiatwal:
Just to add further context here there will be a UI change that allows caseworkers to respond to a query. 

This API change allows that without impacting current functionality

Tllew (approved):
Looks good to me

api/cases/urls.py
saruniitr: we discussed offline and agreed on this.

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -167,6 +167,55 @@ def test_gov_user_can_get_an_individual_ecju_query(self):
...
+        case = self.create_standard_application_case(self.organisation)
+        ecju_query = EcjuQuery(
+            question="open",
+            case=case,
+            raised_by_user=self.gov_user,
+            responded_by_user=self.exporter_user,
+            response="I have a response",
+            responded_at=timezone.now(),
+        )
+        ecju_query.save()
```
saruniitr, line 206: opportunity to use factory here and previous test

api/cases/views/views.py
```diff
@@ -583,6 +583,18 @@ def post(self, request, pk):
...
 
 
+class ECJUQueriesOpenCount(APIView):
+    authentication_classes = (SharedAuthentication,)
+
+    def get(self, request, pk):
+        """Gets count of all open queries."""
+        qs = EcjuQuery.objects.filter(
+            case__pk=pk,
+            responded_at__isnull=True,
```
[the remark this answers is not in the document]
depsiatwal: UI and all the others use only responded_at so need to keep it consistent I guess at some point this might change as we progress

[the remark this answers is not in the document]
depsiatwal: Just for the record the endpoint is required as we don't pull ECJU queries on the FE until required. 
Current ECJU tab pulls all queries and all data. This is a light weight endpoint to pull only the number required on finialization screen atm.

api/cases/tests/test_case_ecju_queries.py
depsiatwal: using factory within this class now with few changes on the factory

api/cases/urls.py
depsiatwal: Just for the record the endpoint is required as we don't pull ECJU queries on the FE until required.
Current ECJU tab pulls all queries and all data. This is a light weight endpoint to pull only the number required on finialization screen atm.

depsiatwal: Just for the record the endpoint is required as we don't pull ECJU queries on the FE until required.
Current ECJU tab pulls all queries and all data. This is a light weight endpoint to pull only the number required on finialization screen atm.

test_helpers/clients.py
```diff
@@ -1092,8 +1092,26 @@ def create_letter_template(
...
+        self,
+        case,
+        question="ECJU question",
+        query_type=None,
+        created_at=None,
+        responded_at=None,
+        response=None,
+        responded_by_user=None,
+        raised_by_user=None,
+    ):
```
saruniitr, line 1105: Sorry I was not clear in the comment.
I don't mean using a helper function to create query, I mean using Django model factories.
eg https://github.com/uktrade/lite-api/blob/dev/api/cases/tests/factories.py#L110

We can have factories for eg, OpenEcjuQueryFactory(), ResolvedEcjuQueryFactory() or similar that creates instances with relevant fields populated.

depsiatwal: Done

api/cases/views/views.py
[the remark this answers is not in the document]
saruniitr: Actually in my previous comment I meant to check by `response__isnull=True` besides `responded_at__isnull=True`

I think the right thing is to check for both and even update the same in other places as well as it is better not to carry the same issue in other places.
`responded_at` can be set even if they submit empty response but I believe that raises an error on the form.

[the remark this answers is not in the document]
depsiatwal: check for both now.  This end point isn't going to be used except in finalisation. 
The other checks will change and possibly this one in P2 where we are thing of bring in a is_closed flag instead so we will change them all at this point.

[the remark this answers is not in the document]
saruniitr: ok

saruniitr (approved):

PR #1774, opened 2024-01-15
LTD 4599 - Red dot should persist as long as query is open.

stuaxo:
Only delete ECJU query notification when the user replies to a query in the ECJU query tab.

This PR only concerns this part of the change, everything else has been split out.

### DIFF
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -626,7 +626,6 @@ class EcjuQuery(TimestampableModel):
 
     @queryable_property
     def is_manually_closed(self):
-
         if self.responded_by_user and self.responded_by_user.type == UserType.INTERNAL:
             return True
         else:
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -640,17 +640,12 @@ class EcjuQuery(TimestampableModel):
 
     notifications = GenericRelation(ExporterNotification, related_query_name="ecju_query")
 
-    def save(self, *args, **kwargs):
-        existing_instance_count = EcjuQuery.objects.filter(id=self.id).count()
+    def send_notifications(self):
+        for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
+            user_relationship.send_notification(content_object=self, case=self.case)
 
-        # Only create a notification when saving a ECJU query for the first time
-        if existing_instance_count == 0:
-            super(EcjuQuery, self).save(*args, **kwargs)
-            for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
-                user_relationship.send_notification(content_object=self, case=self.case)
-        else:
-            self.responded_at = timezone.now()
-            super(EcjuQuery, self).save(*args, **kwargs)
+    def delete_notifications(self):
+        ExporterNotification.objects.get_notifications_for_object(self).delete()
 
 
 class EcjuQueryDocument(Document):
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -5,8 +5,11 @@ from typing import Optional
 
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
+from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.db.models import Q
+from django.utils import timezone
+
 from api.users.enums import UserType
 
 from rest_framework.exceptions import ValidationError
@@ -50,6 +53,7 @@ from api.users.models import (
     GovUser,
     UserOrganisationRelationship,
     ExporterNotification,
+    BaseNotification,
 )
 from lite_content.lite_api import strings
 
@@ -641,6 +645,21 @@ class EcjuQuery(TimestampableModel):
     def send_notifications(self):
         for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
             user_relationship.send_notification(content_object=self, case=self.case)
+
+    def save(self, *args, **kwargs):
+        is_new_query = not self.pk
+        super().save(*args, **kwargs)
+        if is_new_query:
+            self.send_notifications()
+        else:
+            self.responded_at = timezone.now()
+            if self.is_query_closed:
+                # When the query is responded to by user, delete notifications
+                # for the exporter and any internal users.
+                # Filter BaseNotification by GenericRelation to self
+                BaseNotification.objects.filter(
+                    object_id=self.id, content_type=ContentType.objects.get_for_model(EcjuQuery)
+                ).delete()
 
 
 class EcjuQueryDocument(Document):
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -644,9 +644,6 @@ class EcjuQuery(TimestampableModel):
         for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
             user_relationship.send_notification(content_object=self, case=self.case)
 
-    def delete_notifications(self):
-        self.notifications.all().delete()
-
 
 class EcjuQueryDocument(Document):
     query = models.ForeignKey(EcjuQuery, on_delete=models.CASCADE, related_name="ecjuquery_document")
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -7,7 +7,6 @@ from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
 from django.db.models import Q
-from django.utils import timezone
 from api.users.enums import UserType
 
 from rest_framework.exceptions import ValidationError
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -645,7 +645,7 @@ class EcjuQuery(TimestampableModel):
             user_relationship.send_notification(content_object=self, case=self.case)
 
     def delete_notifications(self):
-        ExporterNotification.objects.get_notifications_for_object(self).delete()
+        self.notifications.all().delete()
 
 
 class EcjuQueryDocument(Document):
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -647,7 +647,7 @@ class EcjuQuery(TimestampableModel):
             user_relationship.send_notification(content_object=self, case=self.case)
 
     def save(self, *args, **kwargs):
-        is_new_query = not self.pk
+        is_new_query = self._state.adding
         super().save(*args, **kwargs)
         if is_new_query:
             self.send_notifications()
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -8,7 +8,6 @@ from django.contrib.contenttypes.fields import GenericRelation
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.db.models import Q
-from django.utils import timezone
 
 from api.users.enums import UserType
 
@@ -646,20 +645,11 @@ class EcjuQuery(TimestampableModel):
         for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
             user_relationship.send_notification(content_object=self, case=self.case)
 
-    def save(self, *args, **kwargs):
-        is_new_query = self._state.adding
-        super().save(*args, **kwargs)
-        if is_new_query:
-            self.send_notifications()
-        else:
-            self.responded_at = timezone.now()
-            if self.is_query_closed:
-                # When the query is responded to by user, delete notifications
-                # for the exporter and any internal users.
-                # Filter BaseNotification by GenericRelation to self
-                BaseNotification.objects.filter(
-                    object_id=self.id, content_type=ContentType.objects.get_for_model(EcjuQuery)
-                ).delete()
+    def delete_notifications(self):
+        # Delete both internal and external notifications on this ECJUQuery
+        BaseNotification.objects.filter(
+            object_id=self.id, content_type=ContentType.objects.get_for_model(EcjuQuery)
+        ).delete()
 
 
 class EcjuQueryDocument(Document):
--- a/api/cases/serializers.py
+++ b/api/cases/serializers.py
@@ -678,7 +678,6 @@ class EcjuQueryCreateSerializer(serializers.ModelSerializer):
             "question",
             "case",
             "raised_by_user",
-            "responded_by_user",
             "query_type",
             "team",
         )
--- a/api/cases/serializers.py
+++ b/api/cases/serializers.py
@@ -678,6 +678,7 @@ class EcjuQueryCreateSerializer(serializers.ModelSerializer):
             "question",
             "case",
             "raised_by_user",
+            "responded_by_user",
             "query_type",
             "team",
         )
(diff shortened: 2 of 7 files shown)


api/cases/tests/test_case_ecju_queries.py
```diff
@@ -263,6 +264,28 @@ def test_submit_invalid_data_failure(self, data):
...
+    #     self.client.post(url, data, **self.gov_headers)
+    #     ecju_query = EcjuQuery.objects.get(case=case)
+    #     exporter_notification_count_after_creation = ExporterNotification.objects.get_notifications_for_object(ecju_query)
+    #
+    #     self.client.post(url, data, **self.gov_headers)
+    #
+    #     self.assertEqual(initial_exporter_notification_count, 0)
+    #     self.assertEqual(exporter_notification_count_after_creation.count(), 1)
+    #     import ipdb; ipdb.set_trace()
+
```
hnryjmes, line 288: do you plan to uncomment this later or is this to be deleted?

api/users/models.py
```diff
@@ -1,5 +1,7 @@
 import uuid
 from abc import abstractmethod
+
+from django.db.models import QuerySet
```
hnryjmes, line 4: is this needed

api/cases/tests/test_case_ecju_queries.py
stuaxo: Yep :)

api/cases/tests/test_sla.py
```diff
@@ -398,9 +398,9 @@ def test_answered_ecju_queries_today_after_cutoff(
         mock_is_weekend.return_value = False
         mock_is_bank_holiday.return_value = False
         case = self.create_standard_application_case(self.organisation)
-        _set_submitted_at(case, HOUR_BEFORE_CUTOFF)
+        _set_submitted_at(case, HOUR_AFTER_CUTOFF)
         query = self.create_ecju_query(case)
-        query.created_at = datetime.combine(TODAY, time=HOUR_BEFORE_CUTOFF, tzinfo=timezone.utc)
+        query.created_at = datetime.combine(TODAY, time=HOUR_AFTER_CUTOFF, tzinfo=timezone.utc)
```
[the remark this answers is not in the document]
saruniitr: yeah it is not clear what is the purpose of renaming this, probably discussed within the squad?

api/cases/tests/test_sla.py
```diff
@@ -421,7 +421,7 @@ def test_answered_ecju_queries_yesterday_before_cutoff(
         mock_is_weekend,
         mock_is_bank_holiday,
         responded_at=datetime.combine(YESTERDAY, time=HOUR_BEFORE_CUTOFF, tzinfo=timezone.utc),
-        expected_results=0,
+        expected_results=1,
```
saruniitr, line 424: there is no change in the test, any particular reason for updating expected value?

api/users/models.py
stuaxo: Good catch, removed.

api/cases/tests/test_sla.py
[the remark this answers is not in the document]
stuaxo: Moving this to it's own branch and issue.

The issue was that the test starts failing at a certain time of day, looking at the other tests it looked like this might have been the intention, but I can't be sure - let's pick this up in the other PR.

If this is still an issue, we may start seeing this test fail on this here again, but we can ignore it here.

api/cases/tests/test_sla.py
stuaxo: Have moved this to another PR, will try and re-familiarise in the meantime, and revisit it there.

stuaxo:
There is an expected fail on one of the FE tests, this has been split into another issue and PR.

api/cases/views/views.py
```diff
@@ -569,7 +565,15 @@ def post(self, request, pk):
         serializer = EcjuQueryCreateSerializer(data=data)
 
         if serializer.is_valid(raise_exception=True):
-            serializer.save()
+            is_new_query = not get_case(pk).case_ecju_query.exists()
```
hnryjmes, line 568: I don't think this will work if a case has multiple queries, because your predicate will only return True if there are no existing query objects on that case. 

from what I can tell, this variable `is_new_query` is used to send notifications, so it is important it works if it is e.g. the second or third new query opened on a case

api/cases/views/views.py
```diff
@@ -615,6 +619,7 @@ def put(self, request, pk, ecju_pk):
         If validate only, this will return if the data is acceptable or not.
         """
         ecju_query = get_ecju_query(ecju_pk)
+        is_new_query = not ecju_query.case.case_ecju_query.exists()
```
hnryjmes, line 622: as above, I don't think this is the right predicate to use

api/cases/serializers.py
```diff
@@ -678,6 +678,7 @@ class Meta:
             "question",
             "case",
             "raised_by_user",
+            "responded_by_user",
```
hnryjmes, line 681: is this just to make the test work? or does it have another use?

api/cases/tests/test_sla.py
hnryjmes, line 424: you said that this has been moved out to another PR but if you did that I would expect this to go back to `expected_results=0`

stuaxo: Good catch, I missed this.

api/cases/models.py
```diff
@@ -640,17 +638,9 @@ def is_query_closed(self, lookup, value):
...
-        if existing_instance_count == 0:
-            super(EcjuQuery, self).save(*args, **kwargs)
-            for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
-                user_relationship.send_notification(content_object=self, case=self.case)
-        else:
-            self.responded_at = timezone.now()
-            super(EcjuQuery, self).save(*args, **kwargs)
+    def send_notifications(self):
+        for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
+            user_relationship.send_notification(content_object=self, case=self.case)
```
saruniitr, line 643: Is this intentional that we are not updating `responded_at` now?

api/cases/serializers.py
saruniitr: yeah could you elaborate on the need for this when creating a query?

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -7,17 +7,17 @@
...
 from rest_framework import status
 
-from api.applications.models import BaseApplication
 from api.audit_trail.enums import AuditType
 from api.audit_trail.models import Audit
 from api.audit_trail.serializers import AuditSerializer
 from api.cases.enums import ECJUQueryType
 from api.cases.models import EcjuQuery
-from api.compliance.tests.factories import ComplianceSiteCaseFactory, ComplianceVisitCaseFactory
+from api.compliance.tests.factories import ComplianceSiteCaseFactory
```
saruniitr, line 15: We don't support these Case types yet so it is not clear why this needs changing. You can actually remove this line completely.

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -266,6 +266,45 @@ def test_submit_invalid_data_failure(self, data):
...
+        self.client.post(url, data, **self.gov_headers)
+
+        ecju_query = EcjuQuery.objects.get(case=case)
+        exporter_notification_count_after_creation = ecju_query.notifications.count()
+
+        self.client.post(
+            url,
+            {"pk": ecju_query.pk, **data},
+            **self.gov_headers,
+        )
```
saruniitr, line 288: As this is using gov_headers I am assuming this is the caseworker closing the query manually?
It is not clear because the same `data` is used, in this case `data` should contain `response` isn't it?

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -266,6 +266,45 @@ def test_submit_invalid_data_failure(self, data):
...
+        exporter_notification_count_after_update = ecju_query.notifications.count()
+
+        self.client.post(
+            url,
+            {"pk": ecju_query.pk, **data, "responded_by_user": export_user.baseuser_ptr.pk},
+            **self.gov_headers,
+        )
```
saruniitr, line 295: Same as above, using gov_headers but `data` is the same as query creation.
`responded_by_user` is actually populated in the view, it should not be sent as data.
It is actually sent as exporter user.

test_helpers/clients.py
```diff
@@ -1098,6 +1098,9 @@ def create_ecju_query(self, case, question="ECJU question", gov_user=None, creat
             ecju_query.created_at = created_at
         if responded_at:
             ecju_query.responded_at = responded_at
+        else:
+            # New ECJUQuery, send notification
+            ecju_query.send_notifications()
```
saruniitr, line 1103: I am trying to understand the need for manually calling this function here, this should happen as part of the view itself.

api/cases/tests/test_case_ecju_queries.py
stuaxo: This was pycharm highlighting unused imports.   I wonder where the first one is used.

test_helpers/clients.py
stuaxo: This was because we moved the code out of the model.save() and up into the serializer.

Since I've moved it back though, I'm revert this too.

api/cases/tests/test_case_ecju_queries.py
depsiatwal: It's worth checking if the is_query_closed property instead

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -266,6 +266,35 @@ def test_submit_invalid_data_failure(self, data):
...
+
+        case = self.create_standard_application_case(self.organisation, user=export_user)
+        url = reverse("cases:case_ecju_queries", kwargs={"pk": case.id})
+        data = {"question": "Test ECJU Query question?", "query_type": ECJUQueryType.ECJU}
+
+        initial_exporter_notification_count = ExporterNotification.objects.count()
+        self.client.post(url, data, **self.gov_headers)
+
+        ecju_query = EcjuQuery.objects.get(case=case)
+        exporter_notification_count_after_creation = EcjuQuery.objects.filter(case=case).count()
```
depsiatwal, line 282: These aren't exporter notifications but EcjuQuery count. 

You should be checking the ExporterNotification objects or the relationship it's self

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -266,6 +266,35 @@ def test_submit_invalid_data_failure(self, data):
...
+
+        self.client.post(
+            url,
+            {"pk": ecju_query.pk, **data, "responded_by_user": export_user.baseuser_ptr.pk},
+            **self.gov_headers,
+        )
+        exporter_notification_count_after_user_response = EcjuQuery.objects.filter(case=case).count()
```
depsiatwal, line 289: again this is ECJUQuery and not the notifications

test_helpers/clients.py
depsiatwal: These change hasn't been reverted back ?

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -266,6 +266,35 @@ def test_submit_invalid_data_failure(self, data):
         self.assertFalse(EcjuQuery.objects.exists())
 
 
+class ECJUQueriesNotificationsTest(DataTestClient):
+    @mock.patch("api.cases.views.views.notify.notify_exporter_ecju_query")
+    def test_exporter_notification(self, mock_notify):
```
depsiatwal, line 271: This is testing the post creates a notification ?

We will need to ensure a PUT to the detail deletes the notification for that particular ECJUQuery

api/cases/tests/test_case_ecju_queries.py
stuaxo: Good catch, changed this when I hadn't had enough coffee yet, and missed the notification count.

api/cases/models.py
```diff
@@ -640,17 +642,24 @@ def is_query_closed(self, lookup, value):
...
 
-        # Only create a notification when saving a ECJU query for the first time
-        if existing_instance_count == 0:
-            super(EcjuQuery, self).save(*args, **kwargs)
-            for user_relationship in UserOrganisationRelationship.objects.filter(organisation=self.case.organisation):
-                user_relationship.send_notification(content_object=self, case=self.case)
+    def save(self, *args, **kwargs):
+        is_new_query = not self.pk
```
saruniitr, line 650: I think the previous check using `existing_instance_count` is more clear than this.

I wonder if all of this can be done in the view itself rather than overriding save.
Because only caseworker can create a query so in the POST method we can just create notification objects (no need to check if this is a new query or not).

Either of them can resolve so in PUT method we can determine if it is closed and delete notifications etc.

api/cases/models.py
```diff
@@ -640,17 +642,24 @@ def is_query_closed(self, lookup, value):
...
+        super().save(*args, **kwargs)
+        if is_new_query:
```
depsiatwal, line 652: agree with the above comments 
All the logic can be self contained in the view. 

When we 

POST = Create objects and hence notification 

PUT on the ECJUQuery 
if closing query then Delete notification 

Job Done :-)

api/cases/serializers.py
stuaxo: Good catch, reverted.

api/cases/tests/test_case_ecju_queries.py
stuaxo: Cheers, that clarified how the headers on these work.

api/cases/models.py
stuaxo: Yup, it can revert back to that - especially as `is_new_query = not self.pk` didn't work because we use uuid's as PKs.

I prefered that approach as then we didn't need a custom save.

api/cases/models.py
stuaxo: Using .pk doesn't incur an extra query and is fairly standard... except it can't work here.

The second part of your comment really clarified things for me, around which bits of the lifecycle use POST vs PUT, and I've implemented that - it means we don't actually need the check for whether it's a new query or not.

Re: existing_instance_count, if we were going to do it, it would have been better to have has_existing_instance and use `.exists()` which was what I did in earlier iteration.

api/cases/models.py
stuaxo: Moved this out to the view, as per your other comment.

api/cases/tests/test_case_ecju_queries.py
stuaxo: Changed this to be more in line with the other tests.

api/cases/tests/test_case_ecju_queries.py
stuaxo: I'm not sure how to apply it to this test ?

test_helpers/clients.py
stuaxo: OK, so now.. everything is back in the views.

That particular change was there because moving some of the notification out of the .save() meant that the test didn't sent notifications, that's not a problem for this PR now, since the test doesn't use this method.

depsiatwal:
Close this as new PR supersedes this one

PR #1776, opened 2024-01-16
LTD-4616-add-is-query-closed-property 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4616

A new property is added to ECJUQuery object currently when a query is being tested for closure it's inconsistent 
as the code checked response is null , response_at is null. This property decouples these checks so we can test within the model. 

The are further designs coming into scope where an exporter can close a query without having a response. So this will allow the model to support this change.

### DIFF
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -9,6 +9,8 @@ from django.db import models
 from django.utils import timezone
 from rest_framework.exceptions import ValidationError
 
+from queryable_properties.properties import queryable_property
+
 from api.audit_trail.enums import AuditType
 from api.cases.enums import (
     AdviceType,
@@ -611,6 +613,10 @@ class EcjuQuery(TimestampableModel):
     query_type = models.CharField(
         choices=ECJUQueryType.choices, max_length=50, default=ECJUQueryType.ECJU, null=False, blank=False
     )
+
+    @queryable_property
+    def is_query_closed(self):
+        return self.responded_at is not None
 
     notifications = GenericRelation(ExporterNotification, related_query_name="ecju_query")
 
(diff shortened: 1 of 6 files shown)


api/cases/serializers.py
```diff
@@ -241,6 +242,9 @@ def get_raised_by_user(self, instance):
     def get_responded_by_user(self, instance):
         return self._user_name(instance.responded_by_user)
 
+    def get_is_query_closed(self, instance):
+        return self._user_name(instance.is_query_closed)
```
kevincarrogan, line 246: I can see that this is a copy-paste of some other code but calling _user_name is a super confusing thing to do. I don’t really understand how this user name thing is a function.

depsiatwal: not sure what I was thinking here. 
was trying to copy cat the code but just need a normal boolean field.

api/cases/models.py
```diff
@@ -612,6 +612,10 @@ class EcjuQuery(TimestampableModel):
         choices=ECJUQueryType.choices, max_length=50, default=ECJUQueryType.ECJU, null=False, blank=False
     )
 
+    @property
+    def is_query_closed(self):
+        return self.responded_at is not None
+
```
hnryjmes, line 618: nice, couldn't have done it better myself

hnryjmes (approved):
looking good

api/cases/models.py
depsiatwal: loll you did it was copied from your PR

kevincarrogan (approved):

depsiatwal:
This PR has been changed to use a new library to allow filtering and queries based on properties.  This has now been broken down into two parts

This PR - add the library and a queryable property  

Part 2 to follow : this will allow filtering and update the API to use the new filters.

PR #1777, opened 2024-01-16
LTD-4561: Wait till the lock is available before updating Licence table

saruniitr:
# Change description

The task that sends licence details to lite-hmrc is scheduled when licence is issued which is already part of a transaction.
Licence row would've already been locked during that transaction so if continue without waiting it will result in `OperationalError`.

[LTD-4561](https://uktrade.atlassian.net/browse/LTD-4561)

[LTD-4561]: https://uktrade.atlassian.net/browse/LTD-4561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/views/views.py
+++ b/api/cases/views/views.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.core.exceptions import PermissionDenied
 from django.db import transaction
 from django.http.response import JsonResponse, HttpResponse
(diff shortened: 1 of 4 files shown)


kevincarrogan (approved):

currycoder (approved):
Nice one - docstrings here are super helpful too

PR #1779, opened 2024-01-17
LTD-4565: Remove background_task and related env variables

saruniitr:
## Change description

We migrated all tasks to use celery so remove background_task app and related env variables

[LTD-4565](https://uktrade.atlassian.net/browse/LTD-4565)

[LTD-4565]: https://uktrade.atlassian.net/browse/LTD-4565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal (approved):
lgtm

api/search/signals.py
```diff
@@ -11,7 +10,7 @@
 
 @receiver(post_save)
 def update_search_documents(sender, **kwargs):
-    if not settings.LITE_API_ENABLE_ES or issubclass(sender, Task):
```
kevincarrogan, line 14: This is kind of interesting. We've now slightly changed this because of this Task thing, I don't really understand what it was there for before and don't really know how removing it might affect things.

saruniitr: Looking at git blame it is added when index refreshing is updated to happen asynchronously using background task. Previously registry update happening within this function.

As we are completely removing background tasks this won't affect us now.

kevincarrogan (approved):

PR #1780, opened 2024-01-17
LTD-4617-use-new-is-query-closed-property 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4617

This is the 2nd part of adding the property is_query_closed

This PR allows filtering and we update all API calls to use this new property..

### DIFF
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -7,7 +7,14 @@ from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
 from django.utils import timezone
+
+from queryable_properties.managers import QueryablePropertiesManager
+
+from django.db.models import Model, Q
+
 from rest_framework.exceptions import ValidationError
+
+from queryable_properties.properties import queryable_property
 
 from api.audit_trail.enums import AuditType
 from api.cases.enums import (
@@ -588,6 +595,9 @@ class EcjuQuery(TimestampableModel):
     Query from ECJU to exporters
     """
 
+    # Allows the properies to be queryable, allowing filters
+    objects = QueryablePropertiesManager()
+
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     question = models.CharField(null=False, blank=False, max_length=5000)
     response = models.CharField(null=True, blank=False, max_length=2200)
@@ -611,6 +621,14 @@ class EcjuQuery(TimestampableModel):
     query_type = models.CharField(
         choices=ECJUQueryType.choices, max_length=50, default=ECJUQueryType.ECJU, null=False, blank=False
     )
+
+    @queryable_property
+    def is_query_closed(self):
+        return self.responded_at is not None
+
+    @is_query_closed.filter(lookups=("exact",))
+    def is_query_closed(self, lookup, value):
+        return Q(responded_at__isnull=not value)
 
     notifications = GenericRelation(ExporterNotification, related_query_name="ecju_query")
 
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -6,15 +6,13 @@ from typing import Optional
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.db import models
+from django.db.models import Q
 from django.utils import timezone
 
-from queryable_properties.managers import QueryablePropertiesManager
-
-from django.db.models import Model, Q
-
 from rest_framework.exceptions import ValidationError
-
+from queryable_properties.managers import QueryablePropertiesManager
 from queryable_properties.properties import queryable_property
+
 
 from api.audit_trail.enums import AuditType
 from api.cases.enums import (
@@ -626,9 +624,10 @@ class EcjuQuery(TimestampableModel):
     def is_query_closed(self):
         return self.responded_at is not None
 
+    # This method allows the above propery to be used in filtering objects. Similar to db fields.
     @is_query_closed.filter(lookups=("exact",))
     def is_query_closed(self, lookup, value):
-        return Q(responded_at__isnull=not value)
+        return ~Q(responded_at__isnull=value)
 
     notifications = GenericRelation(ExporterNotification, related_query_name="ecju_query")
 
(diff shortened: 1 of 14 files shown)


api/cases/models.py
```diff
@@ -618,6 +624,10 @@ class EcjuQuery(TimestampableModel):
     def is_query_closed(self):
         return self.responded_at is not None
 
+    @is_query_closed.filter(lookups=("exact",))
+    def is_query_closed(self, lookup, value):
+        return Q(responded_at__isnull=not value)
+
```
kevincarrogan, line 630: There appears to be a boolean filter already https://django-queryable-properties.readthedocs.io/en/stable/filters.html#boolean-filters.

This is a minor syntactical thing but I'd prefer `~Q(responded_at__isnull=value)` as it kind of feels more Django like than the `not value` but that is possibly just my personal preference.

api/support/management/commands/mark_queries_as_resolved.py
```diff
@@ -31,26 +31,28 @@ def add_arguments(self, parser):
...
+        logging.info("Number of unresolved queries for Case %s:%s", case_reference, unresolved_queries.count())
 
         for query in unresolved_queries:
             query.response = "Marked as resolved by LITE System"
             if not dry_run:
                 query.save()
 
         logging.info(
-            f"Number of unresolved queries after update for Case {case_reference}: {case.case_ecju_query.filter(response=None).count()}"
+            "Number of unresolved queries after update for Case %s:%s ",
```
hnryjmes, line 55: could you change to put a space after the colon like this: `%s: %s`

api/cases/models.py
depsiatwal: knew this didn;t feel right couldn't think what

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -181,60 +181,41 @@ def test_ecju_query_open_query_count(self):
...
-
-        ecju_query = EcjuQueryFactory(
-            question="open",
-            case=case,
-            responded_by_user=self.exporter_user.baseuser_ptr,
-            responded_at=timezone.now(),
-        )
 
         EcjuQueryFactory(
             question="open",
```
hnryjmes, line 211: while you're updating this test, could you update this string `"open"` to be more descriptive? I was confused for a moment until I realised this is just an arbitrary string and it's closed because of the responded_at datetime added below it

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -181,60 +181,41 @@ def test_ecju_query_open_query_count(self):
...
         response = self.client.get(url, **self.gov_headers)
 
-        # Assert
         response_data = response.json()
 
         self.assertEqual(status.HTTP_200_OK, response.status_code)
         self.assertEqual(1, response_data["count"])
 
+        EcjuQueryFactory(
+            question="open",
```
hnryjmes, line 192: same as below, could you make it more obvious that this is just some arbitrary string and not related to the open/close status

hnryjmes (approved):
looks good, just some minor comments.

api/cases/tests/test_case_ecju_queries.py
depsiatwal: done

api/cases/models.py
```diff
@@ -618,6 +624,10 @@ class EcjuQuery(TimestampableModel):
     def is_query_closed(self):
         return self.responded_at is not None
 
+    @is_query_closed.filter(lookups=("exact",))
+    def is_query_closed(self, lookup, value):
+        return ~Q(responded_at__isnull=value)
+
```
saruniitr, line 630: Do we still need to above queryable_property?

depsiatwal: yes this allows us to use is_query_close='xyz' and the former is for use in the serializer to use in FE.

saruniitr: This really feels like syntactic sugar and not adding much value.
We are querying one value whose value depends on the value of another field so better to use Q expression directly based on original field.

If we remove above @queryable_property, does this still work?

saruniitr: Both of these named as `is_query_closed()` which is confusing as well.

depsiatwal: i'll add some context. 

So we have added a new property called is_query_closed to keep all the logic in a single place. This value is used in the frontend to check if a query is closed ,  or any call that can gets the serialised request. 

However when we do filtering i.e is_query_closed = True then we are unable to use this  since it's a property. So we would have to revert back to checking db fields negating the fact we want to keep all the logic in a single place. 

The first one is used for object serialisation the 2nd check is for filtering equality check. 

Alternatively we could use a manager class to do the same , but this is neater and encapsulates all the logic in the model class. 

i'll add a comment to the 2nd filtering method so it's clearer

saruniitr: ok but the logic is still not in one place.
Condition used in these two functions is different.

Are these two functions totally independent or the first one is required for the second one to work?

`@is_query_closed.filter(lookups=("exact",))` is this just to help use this as `is_query_close=True`?


Is below valid?

```
    @queryable_property
    def is_query_closed(self):
        return self.responded_at is not None

    # This method allows the above propery to be used in filtering objects. Similar to db fields.
    @is_query_closed.filter(lookups=("exact",))
    def is_query_closed(self, lookup, value):
        return self.is_query_closed()
```

depsiatwal: yes you need both first define the property it self the later 
is_query_closed.filter 

is the filtering functionality

Logic isn't in a single method but it's not scattered all over the code base. Frontend / backend views etc etc

kevincarrogan (approved):

api/cases/models.py
kevincarrogan: @saruniitr it's probably worth taking a look at the documentation for this library https://django-queryable-properties.readthedocs.io/en/stable/.

https://django-queryable-properties.readthedocs.io/en/stable/filters.html#filtering-querysets is the most relevant to what has been implemented here.

PR #1782, opened 2024-01-18
LTD-4622-ensure-ecjuquery-put-respone-field-optional 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4622

This change is to allow exporters/caseworkers to close a query without providing a response. 

Since we look at responded_at this end point is being used to close a query without a response.


api/cases/tests/test_case_ecju_queries.py
```diff
@@ -378,6 +378,23 @@ def test_caseworker_manually_closes_query(self):
...
+        case = self.create_standard_application_case(self.organisation)
+        ecju_query = self.create_ecju_query(case, question="provide details please", gov_user=self.gov_user)
+
+        self.assertIsNone(ecju_query.responded_at)
+
+        query_response_url = reverse("cases:case_ecju_query", kwargs={"pk": case.id, "ecju_pk": ecju_query.id})
+
+        data = {"response": None}
+
+        response = self.client.put(query_response_url, data, **self.gov_headers)
```
kevincarrogan, line 391: Can an exporter user also close a query without adding a response?

depsiatwal: Yes either can it will be optional 
want me to add another test ?

saruniitr: If exporter can close without response then that defeats the purpose raising query, is this expected behaviour?

kevincarrogan: Yeah. If it's to be expected that both gov users and exporters can close then it would be good to see a test for each.

depsiatwal: test added

hnryjmes (approved):

PR #1795, opened 2024-01-30
LTD-4619-exporter-new-query-respond-optional 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4619

This change is required so we no longer need to send the response in the payload. 
Since this field is now optional.


Tllew (approved):

stuaxo (approved):
LGTM

PR #1811, opened 2024-02-06
LTD-4621-rfi-email-reminder 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4621

A new schedule to send an email to the exporter if the case has an open query after 15 working days have elapsed.

The issue with these notifications is that sending of emails is stateless so  if there's an and another rerun would cause the emails to be potentially resent . We could resolve using a few solutions 

1) Add a status of type  ENUMS notification_state to indicate that the ecju_query chaser notification has been sent
2) Add an audit log to say an ECJU notification was sent. We can then check this before resending however we need to confirm if we want a log in N&T but I think this is quite useful. One potential fallback is that we are using the audit log  for yet another purpose. 
3) Add a NotificationLog table which logs when we send an email we could also record the data dict sent to record what info has been sent, user, case
NotificationLog
user
Enums - NotifyEnumsType
Data - payload sent
case
4) Ephemeral storage in cache (Redis) where we store a pattern key for the emails sent for that template for the day. Therefore if we rerun then we ignore if it's stored in cache. 

Above was solved by adding a flag to ECJU Query object which records the date/time email was sent on. 
We check this flag prior to sending a reminder. In addition we add if a case has been terminated prior to sending an email.

The check for ECJUQuery to be open between 15-20 days is an edge case in theory if nothing goes wrong we would send a reminder on day 15. The extra 5 days is a caution on the basis something goes wring the system would still send an email. 

This has been tested on Dev data and I successfully received a query  reminder on day 15.


api/cases/celery_tasks.py
```diff
@@ -156,3 +157,77 @@ def update_cases_sla():
...
+
+
+WORKING_DAYS_ECJU_QUERY_CHASER_REMINDER = 15
+WORKING_DAYS_APPLICATION = 20
+
+
+@shared_task(
+    autoretry_for=(Exception,),
+    max_retries=MAX_ATTEMPTS,
+    retry_backoff=RETRY_BACKOFF,
```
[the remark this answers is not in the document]
depsiatwal: good point and we have ability to rerun now removed

hnryjmes (approved):

api/cases/notify.py
```diff
@@ -104,11 +105,33 @@ def notify_exporter_ecju_query(case_pk):
     )
 
 
+def notify_exporter_ecju_query_chaser(ecju_query_id, callback):
+    APPLICATION_WORKING_DAYS = 20
```
saruniitr, line 109: Can this be imported from the constant defined in `celery_tasks.py` file so that we only need to update in one place or send as an argument to this.

api/cases/celery_tasks.py
```diff
@@ -156,3 +157,69 @@ def update_cases_sla():
...
+            ):
+                ecju_query_reminders.append(ecju_query.id)
+
+        for ecju_query_id in ecju_query_reminders:
+            # Now lets loop round and send the notifications
+            send_ecju_query_chaser_email.delay(ecju_query_id)
+
+        logger.info("Sending all ECJU query chaser emails started finished")
+
+    except Exception as e:  # noqa
```
saruniitr, line 195: why `# noqa` here and below as well?

we have `logger.exception()`.

api/cases/celery_tasks.py
```diff
@@ -156,3 +157,69 @@ def update_cases_sla():
...
+    logger.info("Sending all ECJU query chaser emails started")
+
+    try:
+
+        ecju_query_reminders = []
+        ecju_queries = EcjuQuery.objects.filter(
+            is_query_closed=False, chaser_email_sent_on__isnull=True, case__status__is_terminal=False
+        )
+
+        for ecju_query in ecju_queries:
```
saruniitr, line 182: Worth logging the count above this line.

depsiatwal: add an extra log to indicate we have added a scheduled notify for each id

api/cases/celery_tasks.py
depsiatwal: removed

api/cases/notify.py
depsiatwal: can't add to celete_tasks as this would create a circular reference 
infact this isn't really configurable as the actual email template is fixed hence i've just removed the variable completely to indicate that

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -678,6 +689,213 @@ def test_exporter_cannot_delete_documents_of_closed_query(self):
         response = response.json()
         self.assertIsNotNone(response["document"]["id"])
 
+
+class ECJUQueriesChaserNotificationTests(DataTestClient):
+    @freeze_time("2024-02-06 12:00:00")
+    def setUp(self):
+        super().setUp()
+        settings.GOV_NOTIFY_ENABLED = True
```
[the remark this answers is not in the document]
depsiatwal: that would only work on tests so I would have to enable on all tests that require the override. 

For this change it makes more sense it switch it on for all tests, also ensure we have the required level of mocking.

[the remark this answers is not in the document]
saruniitr: if it is required on all tests does adding this decorator to the `setUp()` works?

saruniitr (approved):
Please see if adding `override_settings()` to setup works otherwise we can go with the current approach.

currycoder (approved):

PR #1830, opened 2024-02-15
LTD-4700-working-add-open-days-template 

depsiatwal:
Add open days as a parameter to email

### DIFF
--- a/api/cases/notify.py
+++ b/api/cases/notify.py
@@ -116,6 +116,7 @@ def notify_exporter_ecju_query_chaser(ecju_query_id, callback):
             "case_reference": ecju_query.case.reference_code,
             "exporter_frontend_ecju_queries_url": exporter_frontend_ecju_queries_url,
             "remaining_days": 20 - ecju_query.open_working_days,
+            "open_working_days": ecju_query.open_working_days,
         },
         callback,
     )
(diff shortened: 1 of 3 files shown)


saruniitr (approved):

api/cases/notify.py
```diff
@@ -116,6 +116,7 @@ def notify_exporter_ecju_query_chaser(ecju_query_id, callback):
             "case_reference": ecju_query.case.reference_code,
             "exporter_frontend_ecju_queries_url": exporter_frontend_ecju_queries_url,
             "remaining_days": 20 - ecju_query.open_working_days,
```
hnryjmes, line 118: I know this isn't something you're changing in this PR but the `20` on its own on this line seems a bit random, maybe at some point it could be changed to be a uppercase constant variable that we stick somewhere, or maybe this already exists 

I can see in celery tasks there is `STANDARD_APPLICATION_TARGET_DAYS` so something like that but placed somewhere obvious and common

hnryjmes (approved):

api/cases/notify.py
depsiatwal: Yeah I did have that but then changed it back to show it wan't configurable

PR #1832, opened 2024-02-16
Add a healtcheck for document data backups

kevincarrogan:
### Aim

Add a healthcheck to be alerted about failures in the document data backups.

[LTD-4703](https://uktrade.atlassian.net/browse/LTD-4703)

[LTD-4703]: https://uktrade.atlassian.net/browse/LTD-4703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevincarrogan:
There's a debate to be had about whether this healtcheck is more complicated than it needs to be and whether this could be just an "outcome based" check i.e. just check whether the number of backed up files equals the number of documents just after we think the last backup ran.

The reason why this ended up as it is was for the following reasons:

  - Knowing when the last backup ran would be a bit of a guessing game, we could just hardcode this to something like 3am but I felt that this hardcoded time wasn't specific enough and could cause problems if/when this backup ever overruns
  - This healthcheck will be run constantly all hours of the day so you'd actually have to then build in a time period this shouldn't be checked e.g. 2am-3am (as files will be in the middle of being backed up so the number will be out), which is the window we think the backup will run, again it felt a bit like I was just picking numbers in the hope I'd get it right but also felt like if someone updated the time it ran they might not know that the healthcheck was harcoded to this time interval
  - Working on the premise of the two things above being a bit janky I thought about whether I could use the schedule definition itself so that the healthcheck knew exactly when this thing should be running
  - From that point I found that I could actually know very precisely when this task was going to start and when it exactly finished so the healthcheck could take all of that into account
  - On top of that I noticed that I could do extra checks that, if this were to fail, would give me precise information as to exactly why it failed, instead of just getting a high level "outcome based" error that just tells me that things weren't backed up

There may be additional concerns that this healthcheck knows too much about the inner workings of how the backup works, but considering this is all within the `document_data` app itself I think that's fine and I don't think that we'll be switching out the mechanisms for how this works any time soon.

I'm not against just swapping this out for a simpler check that ignores checks between 2-3am but I think that this healthcheck actually provides greater value and is less liekly to give false positives.

hnryjmes (approved):
no comments to make, you've explained everything very clearly

depsiatwal (approved):
I would have also argued that its an overkill and a bit too couples between the two tasks. 
But given your comments I think it make a decent case for it. 

Has this been tested on a dev environment 

approving for now,

PR #1834, opened 2024-02-19
[LTD-4710] Add capability to dump an anonymised DB sql file

currycoder:
### Aim

The overall goal for this work is to replace the mechanism for generating anonymised DB dumps for lite.  Previously, DB dumps were generated on an airflow instance using the following scripts; https://github.com/uktrade/lite-airflow-dags/blob/master/dags/export_lite_db.py

At a high level, this PR adds a new management command `dump_and_anonymise` to lite-api.

The command itself is provided by a companion PR in the django-db-anonymiser submodule; https://github.com/uktrade/django-db-anoynmiser/pull/1

### How it works

- A YAML configuration has been added which aims to match how anonymisation currently works (in lite-airflow-dags).  The YAML specifies DB column names along with the function to use to anonymise their values.
- The `dump_and_anonymise` django management command does the following;
    - Runs `pg_dump` against the django app's default DB.
    - Anonymises the resulting dump according to our YAML configuration.
    - Writes the anonymised dump to a temporary file location.
    - Uploads the dump to an S3 bucket.
- The `dump_and_anonymise` command will run nightly using the `task-runner` job on Jenkins.  This will result in a dump file in S3.
- Our existing refresh DB jenkins jobs can then use the resulting DB dumps.  We will just need to point these jobs to look at the new S3 bucket.

### Points of note
- Changes have been made to `settings.py` to support multiple S3 buckets provided in `VCAP_SERVICES`.  At the moment, different buckets are identified by tags - meaning that we will need to add/tag buckets correctly before deploying this.
- This has been deployed to devdata (with steps noted in the associated Jira story) and the resulting SQL dump verified.
- Integration tests have been added which add data to various DB tables, run the `dump_and_anonymise` command and verify that the resulting dump does not contain any data that escaped anonymisation.

[LTD-4710](https://uktrade.atlassian.net/browse/LTD-4710)

[LTD-4710]: https://uktrade.atlassian.net/browse/LTD-4710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/conf/anonymise_model_config.yaml
```diff
@@ -0,0 +1,46 @@
+config:
+  addons:
+    - django_db_anonymiser.db_anonymiser
+  extra_parameters:
+    pg_dump:
+        - "--exclude-table-data=document_data_documentdata"
+
+strategy:
+  users_baseuser:
+    first_name: faker.first_name
```
depsiatwal, line 10: just thinking about this 
but I presume if we added a new column in the database which requires animalization
somebody would need to remember to update the yaml config. 

Wondering if there's anyway to alert the user to remind then to potentially add a text column here or explicitly exclude it.  Also questioning if this is an overkill on the basis we don't add that many columns to the data which contain sensitive information.

depsiatwal (commented):
Did you manage to test this works as expected on a dev env ?

.gitmodules
```diff
@@ -6,3 +6,6 @@
 	path = lite_routing
 	url = git@github.com:uktrade/lite-routing.git
 	branch = main
+[submodule "django_db_anonymiser"]
+	path = django_db_anonymiser
+	url = git@github.com:uktrade/django-db-anoynmiser.git
```
hnryjmes, line 11: see my comment on the other PR, your repo should probably be named `django-db-anonymiser` not `django-db-anoynmiser`

Dockerfile
```diff
@@ -8,6 +8,10 @@ RUN apt-get install -y libpq-dev gcc curl \
   python3-cffi libcairo2 libpango-1.0-0 libpangocairo-1.0-0 libgdk-pixbuf2.0-0 \
   libffi-dev shared-mime-info swig git imagemagick poppler-utils openssl libsqlite3-dev
 RUN curl https://pyenv.run | bash
+RUN curl --fail https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add
+RUN sh -c 'echo "deb https://apt.postgresql.org/pub/repos/apt jammy-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
+RUN apt-get update
+RUN apt-get install -y postgresql-client-12
```
hnryjmes, line 14: I can see that we already have `apt-get install ...` and `apt-get update ...` earlier in this file, is it necessary to have these lines added separately or can they all be combined somehow?

api/conf/anonymise_model_config.yaml
hnryjmes: I like this suggestion ^ 

a fully automated way of de-risking this could be to have a test that checks that the fields defined in this file matches a list of fields generated by some script that checks what fields we have in lite-api programmatically. there are probably also manual ways we could achieve the same thing if we thought this was too much effort

api/anonymised_db_dumps/tests/test_anonymised_dumps.py
```diff
@@ -0,0 +1,146 @@
...
+from api.users.tests.factories import BaseUserFactory
+
+
+class TestAnonymiseDumps(TransactionTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.create_test_data()
+        try:
+            os.remove("/tmp/anonymised.sql")
```
hnryjmes, line 23: would it be better if this filename was `DB_ANONYMISER_DUMP_FILE_NAME` in case this test is ever run locally with this override variable set in the env file?

Dockerfile
currycoder: The issue is that the first `apt-get install` grabs curl, and we need curl to grab the key ahead of this `apt-get` install.  No matter how we slice it we need two `apt-get install` calls so I have preferred to keep the postgres one in a separate block.

api/conf/anonymise_model_config.yaml
currycoder: I can't think of a clean way to do this right now that isn't super annoying - 90% of schema changes will not need anonymisation.  I think for now I would prefer to keep this as a manual/review responsibility and keep the scope of this particular PR to "replace the airflow anonymiser".  If we can think of simple ways to layer checks on later then I'm totally open to that.

currycoder:
@depsiatwal 
> Did you manage to test this works as expected on a dev env ?

Yes, I deployed this to devdata and created SQL dumps successfully which I verified locally.

api/conf/anonymise_model_config.yaml
```diff
@@ -0,0 +1,46 @@
+config:
```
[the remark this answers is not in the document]
currycoder: Right now, this just lifts the same anonymisation rules that we have from our airflow solution (https://github.com/uktrade/lite-airflow-dags/blob/master/dags/anonymiser.py).  I have a follow-up piece for going through all the columns in lite-api and ensuring that we have everything anonymised that needs to be.

hnryjmes (approved):

api/queries/end_user_advisories/tests/factories.py
```diff
@@ -0,0 +1,10 @@
+import factory
+from api.queries.end_user_advisories.models import EndUserAdvisoryQuery
+from api.cases.tests.factories import CaseFactory
+
+
+class EndUserAdvisoryQueryFactory(CaseFactory):
+    end_user = factory.SubFactory("api.parties.tests.factories.PartyFactory")
```
saruniitr, line 7: My understanding is we don't support these queries yet so this can be removed.

saruniitr (approved):

depsiatwal (approved):
lgtm

PR #1836, opened 2024-02-21
LTD-4706-mock-virus-scan 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4706

Adding a mock virus scan if local and testing. 

The URLS and mock service is only available when the environment variable is set.


api/mock_virusscan/views.py
```diff
@@ -0,0 +1,24 @@
...
+    parser_classes = (MultiPartParser,)
+    EICAR_TEST = b"X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*"
+
+    def post(self, request):
+        if len(request.FILES) != 1:
+            return HttpResponseBadRequest("Provide a single file")
+
+        uploaded_file = request.FILES.get("file")
+
+        if uploaded_file.file.read() == self.EICAR_TEST:
```
saruniitr, line 21: this is comparing with the content of whole file, we need to check for the presence of EICAR_TEXT in the file correct? as it can be anywhere in the file.

api/conf/settings.py
```diff
@@ -121,6 +122,14 @@
     "api.document_data",
 ]
 
+MOCK_VIRUSSCAN_ACTIVATE_ENDPOINTS = env("MOCK_VIRUSSCAN_ACTIVATE_ENDPOINTS")
+
+if "MOCK_VIRUSSCAN_ACTIVATE_ENDPOINTS":
+    INSTALLED_APPS += [
+        "api.mock_virusscan",
```
saruniitr, line 129: could this be renamed to `mock_virus_scan`?

api/conftest.py
```diff
@@ -13,6 +14,19 @@
...
 
+DEFAULT_ENVFILE = ".env"
+
+
+def pytest_configure(config):
+    """
+    Load  env variables automagically
+    """
+    if not os.environ.get("PIPENV_DOTENV_LOCATION"):
+        load_dotenv(dotenv_path=DEFAULT_ENVFILE, override=True)
```
kevincarrogan, line 25: Is this bit necessary?

api/mock_virusscan/views.py
```diff
@@ -0,0 +1,23 @@
...
+@method_decorator(csrf_exempt, name="dispatch")
+class Scan(APIView):
+    parser_classes = (MultiPartParser,)
+    EICAR_TEST = b"X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*"
+
+    def post(self, request):
+        if len(request.FILES) != 1:
+            return HttpResponseBadRequest("Provide a single file")
+
+        uploaded_file = request.FILES.get("file")
```
kevincarrogan, line 18: `uploaded_file` here could be `None` because you're using `get`.

README.md
```diff
@@ -32,6 +32,11 @@ Service for handling backend calls in LITE.
     - Check LITE_INTERNAL_HAWK_KEY ENV Keys match on API and caseworker
     - Check LITE_EXPORTER_HAWK_KEY ENV Keys match on  API and expoter
 
+  - Mock Virus Scan
+    For file uploading for local, testing and development can use the mock scan set the following env variables
+    AV_SERVICE_URL=http://localhost:8100/mock_virusscan/scan
+    MOCK_VIRUSSCAN_ACTIVATE_ENDPOINTS=True
```
kevincarrogan, line 38: If this is what we want as the default for development we should just set this in `local.env` as the default.

README.md
```diff
@@ -100,13 +105,15 @@ EXPORTER_USERS='[{"email"=>"foo@bar.com"}]' ./manage.py seedexporterusers
...
-- django-background-tasks: `pipenv run ./manage.py process_tasks` will run all background tasks
+We currently use celery for async tasks and scheduling in LITE;
 - celery: a celery container is running by default when using docker-compose.  If a working copy
     "on the metal" without docker, run celery with `watchmedo auto-restart -d . -R -p '*.py' -- celery -A api.conf worker -l info`
 - celery-scheduler: a celery container is running by default when using docker-compose. This is to monitor any scheduled tasks  If a working copy
     "on the metal" without docker, run celery with `watchmedo auto-restart -d . -R -p '*.py' -- celery -A api.conf beat`
+- Running celery without non async for development set the following in .env
+  CELERY_ALWAYS_EAGER=True
+  CELERY_TASK_ALWAYS_EAGER=True
+  CELERY_TASK_STORE_EAGER_RESULT=True
```
kevincarrogan, line 116: If this is what we want as the default for development we should just set this in `local.env` as the default.

kevincarrogan: Actually. Why do we want to suggest running celery in non-async mode for development? It would be better to leave it as is.

api/mock_virusscan/views.py
depsiatwal: Well we just need to ensure the ECIR files return True which they do as these are what we should be using in the e2e tests. 

This a common pattern which return true if these are the file contents

api/conf/settings.py
depsiatwal: yeah done also moved out of API

api/conftest.py
depsiatwal: apparently not removed and move mock service to outside api

api/mock_virusscan/views.py
depsiatwal: I think we can expect file to be present here as we check length 
else it will return error anyway which means it failed virus scanning anyway. Changed to index rather then GET

README.md
depsiatwal: good point removed and added to local.env

README.md
depsiatwal: good point removed and added to local.env

mock_virus_scan/tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+
+
+def pytest_configure(config):
+    """
+    Load  env variables automagically
+    """
```
kevincarrogan, line 7: This comment isn't valid anymore.

mock_virus_scan/tests/test_mock_virus_scan.py
kevincarrogan, line 1: In API we don't tend to use pytest type tests and instead use unittest style tests.

mock_virus_scan/views.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework.views import APIView
+from rest_framework.parsers import MultiPartParser
+
+from django.http import JsonResponse, HttpResponseBadRequest
+from django.utils.decorators import method_decorator
+from django.views.decorators.csrf import csrf_exempt
+
+
+@method_decorator(csrf_exempt, name="dispatch")
```
kevincarrogan, line 9: I don't think you need to do `csrf_exempt` for an `APIView`.

README.md
```diff
@@ -100,9 +100,7 @@ EXPORTER_USERS='[{"email"=>"foo@bar.com"}]' ./manage.py seedexporterusers
 ```
 
 ## Running background tasks
-
-We currently have two mechanisms for background tasks in LITE;
-- django-background-tasks: `pipenv run ./manage.py process_tasks` will run all background tasks
+We currently use celery for async tasks and scheduling in LITE;
```
hnryjmes, line 103: nice

hnryjmes (commented):
confused as to why you have `mock_virusscan` in some places but `mock_virus_scan` in others
e.g. in `local.env` you have
```
MOCK_VIRUSSCAN_ACTIVATE_ENDPOINTS=True
```

in `ci.env` you have
```
MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS = True
```

are they actually different things?

mock_virus_scan/views.py
```diff
@@ -0,0 +1,23 @@
...
+class Scan(APIView):
+    parser_classes = (MultiPartParser,)
+    EICAR_TEST = b"X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*"  # noqa
```
hnryjmes, line 12: I'm wondering if there will be weird consequences of us having this string in our public codebase

mock_virus_scan/tests/test_mock_virus_scan.py
depsiatwal: redone

mock_virus_scan/views.py
depsiatwal: yeah manage to remove and it works

README.md
depsiatwal: This was an overhand from a rewording fixed now

mock_virus_scan/views.py
depsiatwal: don't know but it's a common pattern used for virus scan, 
one of the DIT repositories have the actual ECIR file

local.env
```diff
@@ -39,9 +39,10 @@ AWS_STORAGE_BUCKET_NAME=uploads
 AWS_REGION=eu-west-2
 
 # AV
-AV_SERVICE_URL=<<FROM_VAULT>>
-AV_SERVICE_USERNAME=<<FROM_VAULT>>
-AV_SERVICE_PASSWORD=<<FROM_VAULT>>
+AV_SERVICE_URL=http://localhost:8100/mock_virusscan/scan
```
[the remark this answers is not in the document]
depsiatwal: corrected

ci.env
```diff
@@ -15,7 +15,8 @@ AWS_SECRET_ACCESS_KEY=AWS_SECRET_ACCESS_KEY
 AWS_STORAGE_BUCKET_NAME=AWS_STORAGE_BUCKET_NAME
 AWS_REGION=eu-west-2
 # AV
-AV_SERVICE_URL=AV_SERVICE_URL
+MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS = True
+AV_SERVICE_URL=http://localhost:8100/mock_virusscan/scan
```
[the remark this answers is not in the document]
depsiatwal: corrected

kevincarrogan (approved):

hnryjmes (approved):

PR #1838, opened 2024-02-22
Adds ruff to pre-commit

seijihg:
https://uktrade.atlassian.net/browse/LTD-4705

 I have updated linting to be used with `Ruff` instead of `Prospector` which what we were using. I have added to `circleCI` and `pre-commit`. If this is ok I will add to `lite_routing` and `frontend`


kevincarrogan:
> I was playing with circleCI for ruff but there are like thousand of files need to be changed to be compatible with ruff so I added just to pre-commit

The intention would be to match our current rules using ruff so that we can replace our existing tools without having to change all of our files. I don't think just adding ruff as well as the other tools is useful here.

.circleci/config.yml
```diff
@@ -267,11 +267,8 @@ jobs:
     steps:
       - setup
       - run:
-          name: Prospector
-          command: pipenv run prospector -W pylint -W pep257 -W mccabe
-      - run:
-          name: Black
-          command: pipenv run black . --check --diff
+          name: Ruff and Black
+          command: pipenv run ruff . && pipenv run black . --check --diff
```
kevincarrogan, line 271: We should split these back out to be two separate `run` commands.

Also, I was looking through the docs for `ruff` and it says to use `ruff check .` so what does `ruff .` do?

.pre-commit-config.yaml
```diff
@@ -46,3 +46,13 @@ repos:
...
+
+    - repo: https://github.com/astral-sh/ruff-pre-commit
+      # Ruff version.
+      rev: v0.2.2
+      # Ruff can be configured through a pyproject.toml, ruff.toml, or .ruff.toml file
+      hooks:
+          - id: ruff
+            args: [--fix]
+          # Run the formatter.
+          - id: ruff-format
```
kevincarrogan, line 58: We now have two formatters running here, both ruff and black. We should one or the other but not both.

Also, this formatter isn't being run in ci.

pyproject.toml
```diff
@@ -23,3 +23,19 @@ exclude = '''
   | migrations
 )/
 '''
+[tool.ruff.lint]
+ignore = [
```
kevincarrogan, line 27: I can see we've ignored some but I would also expect us to have needed to have explicitly added some rules to have matched what our previously linter was doing.

.circleci/config.yml
```diff
@@ -267,11 +267,8 @@ jobs:
...
       - run:
           name: Bandit
```
kevincarrogan, line 273: Can we replace `bandit` as well with `ruff`?

.circleci/config.yml
```diff
@@ -267,11 +267,8 @@ jobs:
     steps:
       - setup
       - run:
-          name: Prospector
-          command: pipenv run prospector -W pylint -W pep257 -W mccabe
```
kevincarrogan, line 271: I think we've not fully covered all of these `pylint`, `pep257` and `mccabe` rules that we used to run with `ruff` based on the config I can see below.

.circleci/config.yml
seijihg: I think I just forgot to type `check`  :D but just tested see what it was doing and `ruff . `seems to do the same thing

PR #1839, opened 2024-02-22
Ensure that organisation data being provided to data workspace has a …

kevincarrogan:
### Aim

Ensure that organisation data being provided to data workspace has a stable sort across pages

[LTD-4768](https://uktrade.atlassian.net/browse/LTD-4768)

[LTD-4768]: https://uktrade.atlassian.net/browse/LTD-4768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/data_workspace/views.py
+++ b/api/data_workspace/views.py
@@ -17,7 +17,13 @@ class OrganisationListView(viewsets.ReadOnlyModelViewSet):
     authentication_classes = (DataWorkspaceOnlyAuthentication,)
     serializer_class = OrganisationDetailSerializer
     pagination_class = LimitOffsetPagination
-    queryset = Organisation.objects.all()
+    # We ensure that organisations with the same name always come out
+    # in the same order by additionally ordering by created_at
+    # and by id. If just name is used, the order is non-deterministic
+    # when multiple organisations have the same name. This can mean the
+    # same organisation being emitted on different pages, causing
+    # primary key violations in Data Workspace.
+    queryset = Organisation.objects.order_by("name", "created_at", "id")
 
 
 class PartyListView(viewsets.ReadOnlyModelViewSet):


api/data_workspace/views.py
```diff
@@ -17,7 +17,13 @@ class OrganisationListView(viewsets.ReadOnlyModelViewSet):
...
     serializer_class = OrganisationDetailSerializer
     pagination_class = LimitOffsetPagination
-    queryset = Organisation.objects.all()
+    # We ensure that organisations with the same name always come out
+    # in the same order by additionally ordering by created_at
+    # and by id. If just name is used, the order is non-deterministic
+    # when multiple organisations have the same name. This can mean the
+    # same organisation being emitted on different pages, causing
+    # primary key violations in Data Workspace.
+    queryset = Organisation.objects.order_by("name", "created_at", "id")
```
saruniitr, line 26: Is it worth changing the ordering in the model also?

saruniitr (approved):

api/data_workspace/views.py
kevincarrogan: I chose not to do that because I wasn't sure if anything else was relying on it and wanted to make the surface area of this change as small as possible.

currycoder: > This can mean the same organisation being emitted on different pages

What happens if duplicated records are emitted on either side of the page boundry, e.g. last record of page 1 and first record of page 2 are the same orgs.

kevincarrogan: Data workspace throws an error because it's seen a duplicate row.

currycoder: Ok so my point being, are we ok with that potentially still happening?  Are we expecting that this will vastly reduce the chance of the error?

kevincarrogan: Yes. I believe it will reduce it to the point we probably won't see this happen again.

kevincarrogan:
> I understand the reasoning to keep this change as small as possible, but, now I'm aware of there being a non-deterministic sort somewhere else in LITE, I think it would also be good to have it fixed at the model level

This is breaking us pulling in data to data workspace and needs to be rectified quicly. If I make a change that has a broader impact this requires us to do more testing to ensure that this hasn't broken something else.

api/data_workspace/views.py
kevincarrogan: This fix has been applied in other places already and we don't see this issue elsewhere.

currycoder (approved):

PR #1840, opened 2024-02-22
Add options to start the API server and populate the database with test data

Tllew:
Add options to start the API server and populate the database with test data

### Aim


makefile
```diff
@@ -23,6 +23,13 @@ doc-seed:
 	docker exec -it api pipenv run ./manage.py seedinternalusers
 	docker exec -it api pipenv run ./manage.py seedexporterusers
 
+first-run:
+	docker exec -it api pipenv run ./manage.py migrate
+	docker exec -it api pipenv run ./manage.py seedrolepermissions
```
depsiatwal, line 28: can we not just run seedall ?

depsiatwal (commented):
lgtm small comment

makefile
Tllew: done

seijihg (approved):

makefile
```diff
@@ -56,9 +57,7 @@ secrets:
 	cp local.env .env
 
 seed:
-	./manage.py seedrolepermissions
-	./manage.py seedinternalusers
-	./manage.py seedexporterusers
+	./manage.py seedall
```
hnryjmes, line 60: I see why we've done this and I think one command is cleaner than three

there might have been a reason that we had three separate commands in the README rather than just seedall but it's definitely been that way since I joined in 2022

can't really speak for the whole team but I think I'd prefer that we adapt seedall to work for our purposes rather than have individual commands listed in the README and this is a good first step towards doing that

hnryjmes (approved):

PR #1852, opened 2024-02-29
Remove exhibition clearance related code

saruniitr:
## Change description

These types are not supported yet.
Implementation can be different if we need to bring this back again so we can start fresh.


depsiatwal (approved):
Nicely done 
Do you think we could delete models/enums at any point to really tidy this up ?

saruniitr:
> Nicely done Do you think we could delete models/enums at any point to really tidy this up ?

Yes, I am actually deleting the models also in the next commit. There are test failures so fixing them up first.

api/applications/views/applications.py
```diff
@@ -255,13 +252,6 @@ def put(self, request, pk):
...
-        # Prevent minor edits of the f680 clearance types
-        if not application.is_major_editable() and request.data.get("types"):
-            return JsonResponse(
-                data={"errors": {"types": [strings.Applications.Generic.NOT_POSSIBLE_ON_MINOR_EDIT]}},
-                status=status.HTTP_400_BAD_REQUEST,
-            )
-
         # Prevent minor edits of additional_information
         if not application.is_major_editable() and any(
             [request.data.get(field) for field in constants.F680.ADDITIONAL_INFORMATION_FIELDS]
```
hnryjmes, line 257: this `any()` statement can probably be updated if F680 applications don't exist:
`any([request.data.get(field) for field in constants.F680.ADDITIONAL_INFORMATION_FIELDS])`

api/cases/managers.py
```diff
@@ -464,15 +457,7 @@ def search(  # noqa
             case_qs = case_qs.with_report_summary_subject_or_prefix(report_summary)
 
         if is_work_queue:
-            case_qs = case_qs.annotate(
-                case_order=Case(
-                    When(baseapplication__hmrcquery__have_goods_departed=False, then=0),
-                    default=1,
-                    output_field=BinaryField(),
-                )
-            )
```
hnryjmes, line 473: I'm not sure what this does but I'm assuming it's being deleted because HmrcQuery isn't a thing we care about

hnryjmes (approved):
for every file I've looked at the changes seem sensible and I can see the reason why it's being deleted

made a comment about something else that could be deleted but happy to approve

api/applications/views/applications.py
saruniitr: yeah this whole if condition can go now, will update.

api/cases/managers.py
saruniitr: correct. Depending on whether `have_goods_departed` then `case_order` will either have 0 or 1 which in-turn is used for ordering.
HMRCQuery code is already removed so this whole thing is removed now along with model.

PR #1854, opened 2024-03-01
LTD-4644-nt-close-query-format-v2

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4644

Various options were discussed here including migrations and just payload changes. 

This is the best solution adding a new additional_text field allows us to keep payloads as they are and add fields to additional text just for ui presentation.


saruniitr (approved):

api/audit_trail/serializers.py
```diff
@@ -66,4 +66,8 @@ def get_text(self, instance):
         return format_payload(verb, payload)
 
     def get_additional_text(self, instance):
-        return instance.payload.get("additional_text", "")
+        verb = AuditType(instance.verb)
+        return ADDITIONAL_TEXT_FORMATTERS.get(
+            verb,
+            lambda payload: payload.get("additional_text", ""),
+        )(instance.payload)
```
hnryjmes, line 73: can you explain to me how this works please? the parentheses don't look right to me so I'm finding it hard to follow

saruniitr: from documentation,
`itemgetter("ecju_response")`

Return a callable object that fetches item from its operand

here item in `ecju_response` and operand is `instance_payload`

this is the same as `instance.payload.get("ecju_response")`
If verb is not defined then it picks up `additional_text" from the payload (default lambda function)

hnryjmes: but here it looks like the function call `(instance.payload)` is on the outside of this expression, which is why I'm not following. 

if the verb is undefined, the lambda returns some string, and then `(instance.payload)` is called on the string? 

I don't see how it's guaranteed that `(instance.payload)` is always applied to a callable i.e. not a string

saruniitr: `(instance.payload)` is the argument to the function call.

if verb is defined then,

```
itemgetter("ecju_response")(instance.payload) which is the same as instance.payload.get("ecju_response")
```

if verb is not defined then the output is the value of lambda function.

@depsiatwal may be a test when the verb is not defined helps here.

depsiatwal: exactly as Arun said was gonna pretty much write-up the same thing i've added additional check 
test_exporter_responding_to_query_creates_case_note_mention_for_caseworker
where it checks additional_text this item wouldn't be in our dict so therefore fallback to lamda function.

Tllew (approved):
Looks good to me

api/audit_trail/serializers.py
hnryjmes: I just re-read this again and it makes sense to me now, I wasn't getting that the lambda function stays as a function and isn't immediately called, so just me being stupid

PR #1864, opened 2024-03-08
LTD: Remove Open application types code as they are not supported yet

saruniitr:
## Change description

Remove `OpenApplication` licence type related code and related models as we don't support this yet.


api/cases/tests/test_team_advice.py
```diff
@@ -1,414 +0,0 @@
...
-from api.goods.enums import PvGrading
-from api.staticdata.statuses.enums import CaseStatusEnum
-from api.staticdata.statuses.libraries.get_case_status import get_case_status_by_status
-from api.teams.tests.factories import TeamFactory
-from api.users.tests.factories import GovUserFactory
-from test_helpers.clients import DataTestClient
-from api.users.models import GovUser, Role
-
-
-class CreateCaseTeamAdviceTests(DataTestClient):
```
hnryjmes, line 20: is team advice from advice 1.0 ?

saruniitr: Yes. We still have team advice as one of the advice level but the process is different.

saruniitr:
> I've been looking at the 2 test failures on this and reading through the tests https://github.com/uktrade/lite-api/blob/dev/api/audit_trail/tests/test_streams.py#L26 https://github.com/uktrade/lite-api/blob/dev/api/audit_trail/tests/test_streams.py#L77
> 
> I think I just don't understand what this streams endpoint is used for. Does it need updating or can it all be removed completely?

@hnryjmes I think it is related to Activity stream and can be removed. I need to look at these failures.
It looks like it is an MVP and not fully extended,
https://github.com/uktrade/lite-api/blob/dev/api/audit_trail/streams/service.py#L2

depsiatwal:
Looks like api/applications/libraries/get_applications.py  has dropped in code cov. Is that a result of deleting or is it because the file was touched and now expects more coverage ?

depsiatwal:
> > Details
> 
> It is just application not found path. As this file is touched now it is complaining about coverage.

Yeah thought so will need to be forced merged. Would be good to raise a jira to increase the coverage.

depsiatwal (approved):
lgtm

Tllew (approved):

PR #1877, opened 2024-03-25
LTD: Add validator to validate an application for submission

saruniitr:
### Aim


depsiatwal (commented):
looks good but any reason why we don't do this in the serializer ?

saruniitr:
> looks good but any reason why we don't do this in the serializer ?

This is for the overall application covering various cases when they are submitting. We can probably create a generic serializer for this validation.

api/applications/creators.py
```diff
@@ -257,23 +256,6 @@ def _validate_goods(draft, errors, is_mandatory):
...
-    errors = _validate_siel_locations(draft, errors)
-    errors = _validate_end_user(draft, errors, is_mandatory=True)
-    errors = _validate_security_approvals(draft, errors, is_mandatory=True)
-    errors = _validate_consignee(draft, errors, is_mandatory=True)
-    errors = _validate_third_parties(draft, errors, is_mandatory=False)
-    errors = _validate_goods(draft, errors, is_mandatory=True)
-    errors = _validate_ultimate_end_users(draft, errors, is_mandatory=True)
-    errors = _validate_end_use_details(draft, errors, draft.case_type.sub_type)
-    errors = _validate_route_of_goods(draft, errors)
-    errors = _validate_temporary_export_details(draft, errors)
```
currycoder, line 272: Can all of these helper functions be deleted too?

api/applications/validators.py
```diff
@@ -0,0 +1,199 @@
...
+    party = application.end_user.party
+    documents_qs = PartyDocument.objects.filter(party=party).values_list("safe", flat=True)
+    if documents_qs.exists():
+        if None in documents_qs:
+            error = "We're still processing the end user document. Please submit again"
+        elif False in documents_qs:
+            error = "To submit the application, attach a document that does not contain a virus to the end user"
+    else:
+        if not party.end_user_document_available and not party.end_user_document_missing_reason:
+            error = "To submit the application, attach a document to the end user"
```
currycoder, line 45: Something to consider, it seems like we only return a single error at a time.  Is this desirable?  Just thinking through that from a user perspective it may be helpful to see all the errors at once?

api/applications/models.py
```diff
@@ -302,6 +304,10 @@ class StandardApplication(BaseApplication):
     f1686_approval_date = models.DateField(blank=False, null=True)
     other_security_approval_details = models.TextField(default=None, blank=True, null=True)
 
+    def validate(self):
+        validator = self.validator_class(self)
+        return validator.validate()
```
currycoder, line 309: Love this interface 🙌

api/applications/creators.py
saruniitr: yeah they can be deleted, will clean up in the PR

api/applications/tests/test_standard_application_submit.py
```diff
@@ -112,6 +112,7 @@ def test_submit_standard_application_without_end_user_document_success(self):
         PartyDocument.objects.filter(party=self.draft.end_user.party).delete()
```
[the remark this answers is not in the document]
saruniitr: yeah, will be adding more tests when this actually comes for review.

api/applications/validators.py
saruniitr: yeah initially just trying to keep the current behaviour as is but certainly possible to return all errors at the same time.

api/applications/tests/test_standard_application_submit.py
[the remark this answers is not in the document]
saruniitr: Checked again and we already have tests covering these in https://github.com/uktrade/lite-api/blob/dev/api/applications/tests/test_standard_application_submit.py

In this PR no new changes introduced to change behaviour.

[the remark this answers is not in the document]
saruniitr: added/updated tests to fix coverage issues

api/applications/validators.py
hnryjmes: "attach a document to the end user" sounds wrong to me, I think it should be "attach a document to the end user section". 

There are other error messages like this for consignee and third party, I think they should also end with "consignee section" and "third party section"

saruniitr: that was the existing error message so kept the same, will correct in this PR.

hnryjmes (commented):
is this PR still being worked on?

saruniitr:
> is this PR still being worked on?

Not actively though, just need to update and add tests.
I guess it can be closed and reopened again, I will close it for now.

PR #1879, opened 2024-03-25
LTD-4840-flags-spike

seijihg:
https://uktrade.atlassian.net/browse/LTD-4840

Current idea that we’re investigating:
- new table that contains the flags to remove
- new method to remove the flag
- new method to remove the audit trail
- new method that runs on finalisation to run the remove flag and audit trail notes


api/cases/tests/test_finalise_advice.py
```diff
@@ -121,6 +128,27 @@ class ApproveAdviceTests(DataTestClient):
     def setUp(self):
         super().setUp()
         self.application = self.create_standard_application_case(self.organisation)
+
+        # Add Flag to test with removal after Finalise
+        self.test_flag = Flag.objects.all().first()
+        self.test_flag.remove_on_finalised = True
```
Tllew, line 134: should this be on setup or can we add it to the course for datatestclient?

maybe in system_flags.csv?

api/cases/tests/test_finalise_advice.py
```diff
@@ -151,3 +179,55 @@ def test_approve_standard_application_success(
...
+        self.application.refresh_from_db()
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(self.application.status, CaseStatus.objects.get(status=CaseStatusEnum.FINALISED))
+        for document in GeneratedCaseDocument.objects.filter(advice_type__isnull=False):
+            self.assertTrue(document.visible_to_exporter)
+
+        # Make sure Flag was removed and Audit trail was removed
+        case = get_case(self.application.id)
+        flag = Flag.objects.filter(id="00000000-0000-0000-0000-000000000007").first()
```
Tllew, line 205: can we store self.applications.flags.id in a variable and search by the ID, this ID could change

Tllew (commented):
couple of comments, tests all look good though, let me know if you think the comments aren't doable

api/cases/tests/test_finalise_advice.py
seijihg: not sure this is just a random flag for testing this one specific case and system_flags is/may affect other things.

api/cases/tests/test_finalise_advice.py
seijihg: yes updated this.

Tllew (approved):
Looks good to me, happy to approve if all the tests pass

hnryjmes (commented):
are there any plans to update existing flags to use the new `remove_on_finalised` boolean? possibly you could write a data migration and test to do this for some of our existing flags. I think LU countersigning flag is one example which gets removed on finalisation currently.

seijihg:
> are there any plans to update existing flags to use the new `remove_on_finalised` boolean? possibly you could write a data migration and test to do this for some of our existing flags. I think LU countersigning flag is one example which gets removed on finalisation currently.

there is a test in lite_routing for this.

api/cases/libraries/finalise.py
```diff
@@ -40,3 +42,19 @@ def get_required_decision_document_types(case):
...
+    for flag in case_flags:
+        if flag.remove_on_finalised:
+            case.flags.remove(flag)
+    case.save()
+
+
+def remove_flags_from_audit_trail(case):
+    flags_to_remove = Flag.objects.filter(remove_on_finalised=True)
+    for flag in flags_to_remove:
+        audit_logs = Audit.objects.filter(target_object_id=case.id, payload__icontains=flag.name)
```
kevincarrogan, line 58: Do we actually know if there were any audit objects that this would actually delete at the moment?

I'm not fully convinced that just using `icontains` here on the payload is the right thing to do.

api/applications/views/applications.py
```diff
@@ -415,6 +417,13 @@ def put(self, request, pk):
 
         case_status = get_case_status_by_status(data["status"])
         data["status"] = str(case_status.pk)
+
+        # Remove needed flags when case is Withdrawn
+        withdrawn_status = CaseStatus.objects.filter(status=CaseStatusEnum.WITHDRAWN).first()
```
kevincarrogan, line 422: Why use `.first` here? There should only be one status of this type so you could use `.get`.

api/applications/views/applications.py
```diff
@@ -415,6 +417,13 @@ def put(self, request, pk):
...
+        if case_status == withdrawn_status:
```
kevincarrogan, line 423: Have we thought about any other statuses that might be relevant? Could a user set this to finalised here?

api/cases/libraries/finalise.py
kevincarrogan: Just to elaborate on my concern here. Right now the flags we are searching on have fairly unique names that shouldn't be in other audit logs but it's not impossible for us to have a future flag that has a more common name that could then go and delete audit logs that it shouldn't do.

seijihg: Your assessment seems accurate, using icontains might not be the most suitable choice. Refactoring with more case-sensitive checks ensures that the name must match exactly. This should make us confident enough IMO since there is no flags with the same name since unique is True:

`name = models.CharField(default="Untitled Flag", unique=True, max_length=100)`

api/applications/views/applications.py
seijihg: not sure, just feeling using `filter` is safer than `get` for some reason =))

Tllew (commented):
This looks closed and completed here :
https://github.com/uktrade/lite-api/pull/1925
@seijihg could you close this please?

PR #1880, opened 2024-03-26
LTD-4843-new-denials-model 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4843

This started as a spike but as we've firmed up the model I think it's worth PRing this. 

The new model moves Denial related properties as part of the Denial model and a new entry for each Denial Entry this normalises the models and stops duplication of denial records which is currently about 30%. 

The loader has been updated to load importing into the new fields. These don't currently impact any live code. The next steps would be to figure out how we transition and migrate across Denials search , Denials Matching and  denials view page.


api/external_data/models.py
```diff
@@ -38,6 +38,34 @@ class Denial(TimestampableModel):
     spire_entity_id = models.IntegerField(help_text="Entity_id from spire for matching data", null=True)
 
 
+class NewDenial(TimestampableModel):
```
depsiatwal, line 41: We calling it New for now eventually it would replace the current Denial Model

PR #1885, opened 2024-03-27
LTD-4879 Update api to check is_onward_incorporated on application creation

hnryjmes:
### Aim

Recently we updated the logic that shows "Ultimate end-user" information in the "Check your answers" page to also look at the `is_onward_incorporated` field of a good on application. This check was missing from the application creation logic in `creators.py` so it makes sense to update this also to use that field in the same way.

[LTD-4879](https://uktrade.atlassian.net/browse/LTD-4879)

[LTD-4879]: https://uktrade.atlassian.net/browse/LTD-4879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/creators.py
+++ b/api/applications/creators.py
@@ -141,7 +141,7 @@ def _validate_security_approvals(draft, errors, is_mandatory):
     return errors
 
 
-def _validate_ultimate_end_users(draft, errors, is_mandatory, open_application=False):
+def _validate_ultimate_end_users(draft, errors, is_mandatory):
     """
     Checks all ultimate end users have documents if is_mandatory is True.
     Also checks that at least one ultimate_end_user is present if there is an incorporated good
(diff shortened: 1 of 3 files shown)


api/applications/creators.py
```diff
@@ -152,20 +152,16 @@ def _validate_ultimate_end_users(draft, errors, is_mandatory, open_application=F
...
-            ultimate_end_user_required = True in [
-                goodstype.is_good_incorporated for goodstype in list(draft.goods_type.all())
-            ]
-        else:
-            ultimate_end_user_required = GoodOnApplication.objects.filter(
-                application=draft, is_good_incorporated=True
-            ).exists()
+        ultimate_end_user_required = GoodOnApplication.objects.filter(
+            application=draft, is_good_incorporated=True
+        ).exists() | GoodOnApplication.objects.filter(
```
hnryjmes, line 157: you could also use `Q()` syntax here but I think this is equivalent

hnryjmes: I probably also should write a test for this

api/applications/creators.py
```diff
@@ -152,20 +152,15 @@ def _validate_ultimate_end_users(draft, errors, is_mandatory, open_application=F
...
     if is_mandatory:
-        if open_application:
-            ultimate_end_user_required = True in [
-                goodstype.is_good_incorporated for goodstype in list(draft.goods_type.all())
-            ]
-        else:
-            ultimate_end_user_required = GoodOnApplication.objects.filter(
-                application=draft, is_good_incorporated=True
-            ).exists()
+        ultimate_end_user_required = (
```
depsiatwal, line 155: why removed the top portion
 if open_application: etc etc 

And add just additional query to the check Q(is_onward_incorporated = True | is_onward_incorporated=True)

saruniitr: we don't support open applications so good that it is removed.
I think with a test for this it looks good.

depsiatwal: ok that makes sense is it possible to change the query to a Q operator instead ?

hnryjmes: if you would prefer that, I can change it. I think what I have is equivalent. but now I have a test added I can refactor this to use the Q() syntax

hnryjmes: updated

saruniitr (approved):

PR #1887, opened 2024-03-28
 LTD-4874--new-denial-table 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4874

New model to support Denials , currently not in use.


api/external_data/models.py
```diff
@@ -7,6 +7,26 @@
 from api.users.models import GovUser
 
 
+class Denial(TimestampableModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False, db_index=True)
+    created_by_user = models.ForeignKey(
+        GovUser, related_name="denial_created_by_user", on_delete=models.DO_NOTHING, null=True, blank=True
+    )
+    reference = models.TextField(help_text="The reference assigned by the notifying government")
+    regime_reg_ref = models.TextField(unique=True, help_text="The unique reference assigned by the issuing regime")
```
hnryjmes, line 16: in the diagram you have `regime_reg_ref` showing as nullable, does this mean you have to set `null=True` here? or is the diagram not right?

I would think that there's not much point having a Denial record if there's not a regime reg ref to go with it

depsiatwal: it shouldn't be nullable diagram needs updating this has to be unique field.

kevincarrogan (approved):

PR #1888, opened 2024-04-02
LTD-4875-denial-id-DenialEntity 

depsiatwal:
Add a denail_id to link Denial and Entity 

These are blank/nullable for now, once populate we will introduce constraints


api/external_data/models.py
```diff
@@ -29,6 +29,9 @@ class Denial(TimestampableModel):
 
 class DenialEntity(TimestampableModel):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False, db_index=True)
+
+    denial_id = models.ForeignKey(Denial, related_name="denial_entity", on_delete=models.CASCADE, blank=True, null=True)
```
saruniitr, line 33: Can it just be `denial`? as django makes an implicit field with `_id` suffix that stores pk of `Denial`

depsiatwal: done

api/external_data/models.py
```diff
@@ -29,6 +29,9 @@ class Denial(TimestampableModel):
 
 class DenialEntity(TimestampableModel):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False, db_index=True)
+
+    denial = models.ForeignKey(Denial, related_name="denial_entity", on_delete=models.CASCADE, blank=True, null=True)
```
saruniitr, line 33: just want to double check on the cascade - is there a situation where we delete `Denial`?

depsiatwal: This is so if the Denial record is deleted then the Entity records also gets deleted this makes sense so we don't end up with orphaned records. 
 
https://docs.djangoproject.com/en/5.0/ref/models/fields/#arguments

saruniitr: actually I am asking if there is any situation where we need to delete `Denial` records

depsiatwal: not using current requirements but quite possible to the future in which case we should deleted any entities related to that as well.

saruniitr (approved):

kevincarrogan (approved):

PR #1890, opened 2024-04-03
LTD-4839-duplicates

seijihg:
https://uktrade.atlassian.net/browse/LTD-4839

This is probably a lot more simpler than I thought:

added a validation in serializer instead of relying on model uniqueness only so if in the future we can validate against api company houses we can always do it there.


api/addresses/tests/test_postcode_validation.py
```diff
@@ -77,6 +81,7 @@ def test_valid_postcodes(self):
 
         for postcode in valid_postcodes:
             data["site"]["address"]["postcode"] = postcode
+            data["registration_number"] = "".join([str(randint(0, 9)) for _ in range(8)])
```
Tllew, line 84: I'm not sure we want as many registration numbers as post codes?

api/addresses/tests/test_postcode_validation.py
```diff
@@ -35,14 +37,16 @@ def test_invalid_postcodes(self):
 
         for postcode in invalid_postcodes:
             data["site"]["address"]["postcode"] = postcode
+            data["registration_number"] = "".join([str(randint(0, 9)) for _ in range(8)])
```
Tllew, line 40: I'm not sure we want as many registration numbers as post codes?

Tllew: that way we could add a non unique registration_number and see it fail

Tllew (commented):
Looks good, added a couple of comments so that we can prove registration_number's uniqueness is required

api/addresses/tests/test_postcode_validation.py
seijihg: this is the old tests for poscode :D

api/addresses/tests/test_postcode_validation.py
seijihg: I have added a test for this

api/organisations/tests/test_serializers.py
```diff
@@ -54,10 +55,31 @@ def test_phone_number_validation_success(self, phone, exp_number):
...
+        "reg_number, expected",
+        [("12345678", "12345678"), ("GB123456", "GB123456")],
+    )
+    @pytest.mark.django_db()
+    def test_registration_number_validation_success(self, reg_number, expected):
+        data = {"registration_number": reg_number}
+
+        subj = OrganisationRegistrationNumberSerializer(data=data)
+        assert subj.is_valid()
+        assert subj._validated_data["registration_number"] == expected
```
kevincarrogan, line 79: We use the `self.assert*` assertions on the api side.

api/organisations/tests/test_serializers.py
```diff
@@ -54,10 +55,31 @@ def test_phone_number_validation_success(self, phone, exp_number):
             "01234@567921",
         ],
     )
+    @pytest.mark.django_db()
```
kevincarrogan, line 58: Is this actually required?

api/organisations/tests/test_serializers.py
```diff
@@ -54,10 +55,31 @@ def test_phone_number_validation_success(self, phone, exp_number):
...
         data = self.non_phone_data
         data["phone_number"] = phone
 
         subj = SimpleOrganisationCreateUpdateSerializer(data=data)
 
         assert not subj.is_valid()
+
+
+class TestOrganisationRegistrationNumberSerializer:
+    @pytest.mark.parametrize(
```
kevincarrogan, line 69: We tend to use `parametrized` instead of the pytest version on the api side.

api/organisations/views/organisations.py
```diff
@@ -268,3 +269,14 @@ def update(self, request, *args, **kwargs):
             return JsonResponse(data=serializer.data, status=status.HTTP_200_OK)
 
         return JsonResponse(data={"errors": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+
+class ValidateRegistrationNumber(generics.RetrieveAPIView):
+    # authentication_classes = (SharedAuthentication,)
```
kevincarrogan, line 275: I think we should have some authentication here.

saruniitr (approved):

kevincarrogan (approved):

PR #1909, opened 2024-04-08
LTD-4890-fix-broken-timestamp

depsiatwal:
This test started breaking after we moved to BST. 

From what I can tell we pass back next_time stamp 
https://github.com/uktrade/lite-api/blob/dev/api/audit_trail/streams/service.py#L192C63-L192C78

This appears to be sequence of events 

1) created_at in  UTC 
2) time.mktime(last_created_at.timetuple()) which was doing created_at -1 to make BST 
3) timezone.make_aware(datetime.fromtimestamp(timestamp) which was doing another -1 to make timezone suitable resulting in -2 hours. 

from what I can tell datetime.fromtimestamp was alway timezone aware since it depends on django settings however  time.mktime is timezone aware on circle CI but not in docker since it depends on TZ settings and this being picked up correctly. 

time.daylight = 1 on circle-ci and  time.daylight = 0 on docker both TZ is are set to Europe/London but I think due to OS files one can understand it as BST/GMT and other can't. Probably a way to fix that but didn't bother with that as UTC is a better option here. 

This wouldn't have been an issue since before we were in UTC time and all worked ok. 

I think a better option is to make everything UTC based since this function doesn't need to be timezone aware as long as we send back the timestamp of the next item to search for.

### DIFF
--- a/api/audit_trail/streams/service.py
+++ b/api/audit_trail/streams/service.py
@@ -1,9 +1,7 @@
 """
 MVP activity stream. To be extended appropriately as requirements are drawn up.
 """
-import time
 from datetime import datetime
-
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.utils import timezone
@@ -143,7 +141,7 @@ def get_stream(timestamp):
         verb__in=STREAMED_AUDITS,
     ).order_by("created_at")
     if timestamp > 0:
-        audit_qs = audit_qs.filter(created_at__gte=timezone.make_aware(datetime.fromtimestamp(timestamp)))
+        audit_qs = audit_qs.filter(created_at__gte=datetime.fromtimestamp(timestamp, timezone.utc))
     audit_qs = audit_qs[: settings.STREAM_PAGE_SIZE]
 
     if not audit_qs:
@@ -189,4 +187,4 @@ def get_stream(timestamp):
             countries = []
             stream.append(case_record_json(audit.target_object_id, audit.created_at, countries))
 
-    return {"data": stream, "next_timestamp": int(time.mktime(last_created_at.timetuple())) + 1}
+    return {"data": stream, "next_timestamp": int(last_created_at.timestamp()) + 1}
--- a/api/audit_trail/tests/test_streams.py
+++ b/api/audit_trail/tests/test_streams.py
@@ -1,5 +1,4 @@
-from datetime import timedelta
-
+from datetime import timedelta, datetime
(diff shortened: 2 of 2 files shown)


api/audit_trail/streams/service.py
```diff
@@ -188,5 +196,10 @@ def get_stream(timestamp):
...
             countries = []
             stream.append(case_record_json(audit.target_object_id, audit.created_at, countries))
-
+    print('---------------------------')
+    temp_ts = int(time.mktime(last_created_at.timetuple())) + 1
+    ts = datetime.fromtimestamp(temp_ts, timezone.utc)
+    print(f'timestamp timetuple {last_created_at.timetuple()}')
+    print(f'timestamp {temp_ts}' )
+    print(f'timestamp time obj {ts}' )
     return {"data": stream, "next_timestamp": int(time.mktime(last_created_at.timetuple())) + 1}
```
depsiatwal, line 205: It turns out when constructing the time.mktime(created_at) when creating a timestamp is different in the enviroments. 0 locally and 1 in ciricleci both ENV have TZ='Europe/London' set at OS level. 
one is talking day light saving into account and the other isn't
time.daylight = 1 (manually setting this makes no different)
time.tzname = (GMT,BST) on local this is set to Europe/Europe
os.environ['TZ'] = 'Europe/London'

The time tuple has tm_isdst=-1 which means it doesn't know and makes a guess but it's unclear why it's 1 hour behind

time.struct_time(tm_year=2024, tm_mon=4, tm_mday=9, tm_hour=12, tm_min=50, tm_sec=21, tm_wday=1, tm_yday=100, tm_isdst=-1)

api/audit_trail/streams/service.py
```diff
@@ -189,4 +187,4 @@ def get_stream(timestamp):
             countries = []
             stream.append(case_record_json(audit.target_object_id, audit.created_at, countries))
 
-    return {"data": stream, "next_timestamp": int(time.mktime(last_created_at.timetuple())) + 1}
+    return {"data": stream, "next_timestamp": int(last_created_at.timestamp()) + 1}
```
depsiatwal, line 190: time.mktime was removed since it's env specific which was incorrectly picking up bad timezone and marking daylight = 0 locally and daylight = 1 on the server. 

Changed to use timestamp() to move away from making this timezone aware and shift to UTC.

kevincarrogan (approved):

seijihg (approved):

PR #1910, opened 2024-04-10
Simplified timestamp and make it more accurate

seijihg:
Timestamp was not passed very accurate when it was int-ed so it made the result not very correct.


depsiatwal (commented):
that has many issues
What about old records ?
this could have a snowball effect across the app
stay away from TZ aware fields in DB use UTC where possible as good practice

PR #1911, opened 2024-04-10
[LTD-4855] Fix healthcheck deadlocks

currycoder:
### Aim

We have noticed that when calling the healthcheck endpoint every 30 seconds via pingdom, the production lite-api instance can get in to a bad state.  The bad state can surface quite quickly or sometimes takes days to surface; the likelihood of it surfacing appears to be tied to the weight of traffic that the application is dealing with.

The bad state manifests as a locked gunicorn worker which is unable to respond to incoming requests.  This results in errors being returned in production caseworker/exporter requests and general application instability.

### Fix
The fix was to upgrade the psycopg library to 3.1.18.

Our previous version - 2.9.3 - was not fully compatible with gevent workers.  It's likely that the bad state occurred because of a deadlock; it appears to be much more likely to occur when there is some interaction with postgres at the same time as some interaction with redis for a given gunicorn worker.

### Details

To diagnose this issue, it was necessary to find a reliable way to reproduce it.  The following two scripts were repeatedly run against a lite environment;

```python
import requests
import time

while True:
    ping_result = requests.get("https://lite-api-healthcheck-proxy-{{env}}/healthcheck/").status_code
    if ping_result == 200:
        print(f"{time.asctime()} PING")
    else:
        print(f"{time.asctime()} FAIL: {ping_result}")
```
This script repeatedly hits lite-api's healthcheck endpoint.

```python
import random
import requests
import time

while True:
    ping_result = requests.get("https://internal.lite.{{env}}/", cookies={"exporter": "{{secret}}"}).status_code
    if ping_result == 200:
        print(f"{time.asctime()} PING")
    else:
        print(f"{time.asctime()} FAIL: {ping_result}")
```
This script repeatedly hits lite caseworker's "all queues" page.

By running the two scripts in parallel simultaneously, the bad state could be reached within around 1H maximum.

From there, various different explorations/fixes were tried until the problem was narrowed down to being manifested by the celery/redis health checks.

[LTD-4855](https://uktrade.atlassian.net/browse/LTD-4855)

[LTD-4855]: https://uktrade.atlassian.net/browse/LTD-4855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/document_data/tests/test_celery_tasks.py
```diff
@@ -45,7 +45,7 @@ def test_backup_new_document_data(self):
             "thisisakey",
         )
         self.assertEqual(
-            document_data.data.tobytes(),
+            document_data.data,
```
currycoder, line 48: This is necessary as django defers the contract for `BinaryField` values to the database/driver package; https://forum.djangoproject.com/t/binaryfield-value-type-differs-across-databases/23785/4

This was a little bit surprising as usually the point of django's ORM is to keep code DB-agnostic, but still the changes to accommodate it here aren't far-reaching.

kevincarrogan (approved):

saruniitr (approved):

PR #1912, opened 2024-04-10
Ltd 4894 save team on advice objects

kevincarrogan:
### Aim

This ensures that a team is saved against all advice objects even when not explicitly.

There will be a follow up to migrate all existing objects to have a team.

[LTD-4894](https://uktrade.atlassian.net/browse/LTD-4894)

[LTD-4894]: https://uktrade.atlassian.net/browse/LTD-4894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -541,6 +541,11 @@ class Advice(TimestampableModel):
         except Advice.DoesNotExist:
             pass
 
+        has_no_team_set = self.team is None
+        has_user = self.user is not None
+        if has_no_team_set and has_user:
+            self.team = self.user.team
+
         super(Advice, self).save(*args, **kwargs)
 
         # We retain the denial reasons from the previous object because we're effectively really doing an edit here by
--- a/api/cases/tests/test_create_advice.py
+++ b/api/cases/tests/test_create_advice.py
@@ -1,6 +1,5 @@
 import pytest
 
-from django.test import override_settings
 from django.urls import reverse
 from parameterized import parameterized
 from rest_framework import status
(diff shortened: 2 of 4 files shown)


api/cases/models.py
```diff
@@ -541,6 +541,9 @@ def save(self, *args, **kwargs):
         except Advice.DoesNotExist:
             pass
 
+        if not self.team and self.user:
```
hnryjmes, line 544: you could consider putting parentheses around these to make it clear to others if you are writing
```
not (self.team and self.user)
```
or
```
(not self.team) and self.user
```

kevincarrogan: Changed, but in a different way.

hnryjmes (approved):

api/cases/tests/test_create_advice.py
```diff
@@ -698,6 +697,41 @@ def test_final_refusal_has_criteria_audit(self):
         assert criteria_audit_text == " added refusal criteria."
         assert criteria_additional_text == "WMD, 1."
 
+    @parameterized.expand(
+        [
+            AdviceType.APPROVE,
+            AdviceType.PROVISO,
+            AdviceType.REFUSE,
+            AdviceType.NO_LICENCE_REQUIRED,
+            AdviceType.NOT_APPLICABLE,
```
saruniitr, line 706: I think this is legacy Advice1.0 type

saruniitr (approved):

api/cases/tests/test_create_advice.py
kevincarrogan: I just copied this from another test in this file. Do you think we should remove it from both?

saruniitr: yeah it can be removed.

PR #1913, opened 2024-04-10
LTD-4887: Update page size to have precedents populated for all products

saruniitr:
## Change description

The endpoint that returns goods precedents using a default page_size of 25. In case of applications that has more goods than this size then only 25 of them are populated with previous assessments.

Increase this to 200 to cover more number of products.

[LTD-4887](https://uktrade.atlassian.net/browse/LTD-4887)

[LTD-4887]: https://uktrade.atlassian.net/browse/LTD-4887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/views/views.py
+++ b/api/cases/views/views.py
@@ -10,6 +10,7 @@ from rest_framework.exceptions import ParseError
 from rest_framework.generics import ListCreateAPIView, UpdateAPIView, ListAPIView, RetrieveAPIView
 from rest_framework.views import APIView
 
+
 from api.applications.models import GoodOnApplication
 from api.users.models import BaseNotification, ExporterUser
 from api.applications.serializers.advice import (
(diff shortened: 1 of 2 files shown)


api/cases/views/views.py
```diff
@@ -1264,3 +1265,16 @@ def get_queryset(self):
                 "control_list_entries",
             )
         )
+
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
```
currycoder, line 1270: Is there a test for this endpoint somewhere already?

saruniitr: yes there is,
https://github.com/uktrade/lite-api/blob/dev/api/cases/tests/test_good_precedents_view.py#L101

PR #1914, opened 2024-04-11
Readme amendments

markj0hnst0n:
Minor amendments to readme file mostly to include homebrew package manager for mac


hnryjmes (approved):
yeah this is fine. a while ago we would have suggested differently, as we all used to use different machines for dev work, then sometime in 2023 we all got DBT macs and so I think it's fine to have the `brew` commands in our README as they should apply to everyone

markj0hnst0n:
closed due to failing tests.  will retry.

PR #1918, opened 2024-04-12
LTD-4873-migration-denial-denailentity 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4873

This migration transforms and copies the data from DenialEntity into Denial 

This is to progress to an state model with DENIAL -- DENIALENTITY 

Eventually we will delete the redundant fields in DenialEntity when we have transitioned over search. 

This migration also attempts to delete any Bad data in PROD i.e a regime_ref which is black or null 
Also before we can merge and roll this out we need to fix the data in PROD i.e Records that are considered duplicate Denial Records. These are being fixed by the business and may require an additional load.


hnryjmes (approved):

api/external_data/migrations/0023_denials_data_migration.py
```diff
@@ -0,0 +1,64 @@
...
+
+    total_null_regime_reg_ref =  DenialEntity.objects.filter(Q(regime_reg_ref__isnull=True) | Q(regime_reg_ref=''))
+    log.info(
+            "Delete null regime_reg_ref total -> %s",
+            total_null_regime_reg_ref.count(),
+    )
+    total_null_regime_reg_ref.delete()
+    duplicate_denial_errors = []
+    with transaction.atomic():
+        sid = transaction.savepoint()
```
saruniitr, line 40: Could you clarify the purpose of this?
It is already under a transaction so if there are any errors then nothing is committed anyway.

api/external_data/migrations/tests/test_0023_denials_data_migration.py
```diff
@@ -0,0 +1,59 @@
+import pytest
+
+from django_test_migrations.contrib.unittest_case import MigratorTestCase
+
+
+test_data = [
+{"reference":"DN2010\/0057","regime_reg_ref":"reg.123.123","name":"name 1","address":"address 1","notifying_government":"UK","country":"UK","item_list_codes":"all","item_description":"desc a","end_use":"use 1","reason_for_refusal":"a"},
+{"reference":"DN2010\/0057","regime_reg_ref":"reg.123.1234","name":"name 2","address":"address 2","notifying_government":"UK","country":"UK","item_list_codes":"all","item_description":"desc a","end_use":"use 1","reason_for_refusal":"a"},
```
saruniitr, line 8: this is test data - what is the plan to verify this when deployed to an env that has actual denial data?

saruniitr (commented):
few comments for my understanding

api/external_data/migrations/tests/test_0023_denials_data_migration.py
depsiatwal: i've disabled the core transaction 
this is to prevent the migration from failing and hence no deployment if we see a duplicate  
rather then raise IntegrityError and everything fails we log it and no data will be migrated but we can check the logs to see why and what duplicates existed. 
if we don't see any duplicates and no errors we proceed with the complete migration.

saruniitr: Sorry I don't follow what you mean by disabling the core transaction? Is this the explanation for the comment above?

> this is to prevent the migration from failing and hence no deployment if we see a duplicate
we want the migration to fail if there is a duplicate right?

Could this exercise be done first on a dev environment with actual data and figure out how duplicates exist etc? then we can update migrate to handle all of those errors correctly.

This feels similar to back populating ARS data with prefix, subject ids etc - could we not test this all on an environment before adding these to `dev`?

depsiatwal: If the migration fails because of duplicate we don't want the exception to bubble as they would cause the whole deployment to fail. I've tested this with real data locally and all seems to work without duplicates and this will be tested on DEV/UAT as it makes it's way through the environments. 

Since we can never be 100% we could end-up with integrity error the fall back is to log all the denials that caused an issue and would would have to correct them before trying them again.  

Hence why this logging is there as a fallback if everything succeed then we update within the transaction. 

If we don't have the fall back it would be too dangerous to allow it to continue  as the whole deployment would become blocked.

depsiatwal: > this is test data - what is the plan to verify this when deployed to an env that has actual denial data?

Yeah this has been tested again UAT Copy and latest denials data, 
In this instance had to make an exception as local is only real safe way to test both as we load the full set of real denials.

api/external_data/migrations/tests/test_0024_denials_data_migration.py
```diff
@@ -0,0 +1,59 @@
...
+    
+
+
+    def test_0023_denials_data_migration(self):
+        DenialEntity = self.new_state.apps.get_model("external_data", "DenialEntity")
+        Denial = self.new_state.apps.get_model("external_data", "Denial")
+
+        assert DenialEntity.objects.all().count() == 4
+        assert Denial.objects.all().count() == 2
+        assert Denial.objects.get(regime_reg_ref='reg.123.1234').denial_entity.count() == 3
```
kevincarrogan, line 36: These should be `self.assert*` style assertions.

kevincarrogan (approved):

PR #1921, opened 2024-04-15
LTD-4927-spire-denials-import

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4927

New import that accommodates existing records. 

 We need to do a fresh import as the duplicates have been fixed in spire.  Need to make some changes so that the existing records that have a match aren't reimported or touched.


api/external_data/management/commands/ingest_denials.py
```diff
@@ -63,15 +63,22 @@ def handle(self, *args, **options):
...
             DenialEntity.objects.all().exclude(id__in=matched_denial_ids).delete()
 
-        errors = []
         for i, row in enumerate(data, start=1):
+            # This is required so we don't reload the same denial entity and load duplicates
+            has_fields = bool(row.get("regime_reg_ref") and row.get("name"))
+            if has_fields:
+                exists = DenialMatchOnApplication.objects.filter(
+                    denial__regime_reg_ref=row["regime_reg_ref"], denial__name=row["name"]
+                ).exists()
```
hnryjmes, line 79: I think this is fine, however - in my code for CSV upload I've been treating a DenialEntity record as a match if all of the `reference`, `regime_reg_ref`, `name`, `address` are the same, and if not, considering it a new DenialEntity

as I wasn't sure what to choose here if you're confident that we only need to check `regime_reg_ref` and `name` here, then happy to go with that instead.

depsiatwal: yeah I think in this case it's fine as we only checking a hand full.  Also this isn't about checking for duplicates but more about we preserve the denial record that has a match and not to reload as it would fail when we do the migration.  Using the current loader and migration I can confirm this works fine, given the current dataset.

api/external_data/management/commands/ingest_denials.py
```diff
@@ -63,15 +63,22 @@ def handle(self, *args, **options):
...
         if data:
             # Lets delete all denial records except ones that have been matched
             matched_denial_ids = DenialMatchOnApplication.objects.all().values_list("denial_id", flat=True).distinct()
             DenialEntity.objects.all().exclude(id__in=matched_denial_ids).delete()
 
-        errors = []
         for i, row in enumerate(data, start=1):
+            # This is required so we don't reload the same denial entity and load duplicates
+            has_fields = bool(row.get("regime_reg_ref") and row.get("name"))
+            if has_fields:
```
saruniitr, line 76: is `reference` not required in this check?
The result of `and` is already a `boolean` so further casting is not required.

depsiatwal: no need for reference as that's the spire reference , regime_reg_ref which is our unique denial record is sufficient

I did check but the and was returning the result of  row.get("name")

saruniitr (approved):

PR #1924, opened 2024-04-16
LTD-4896 Add basic CSV upload functionality

hnryjmes:
### Aim

These are the changes required for enabling basic CSV upload functionality for denial records. Mostly the existing serializer `DenialFromCSVFileSerializer` was updated to work with our new data model. The function `validate_csv_file` validates the data coming from the csv and uses `update_or_create` methods on the `Denial` and `DenialEntity` objects to update or create the objects. As a `DenialEntity` object has a foreign key reference to `Denial` it makes sense to do these operations at the same time so that the records can be linked.

I have tested this locally with example data and <100 rows of sample data derived from the data extract taken from Spire. This data was loaded into my local copy of LITE and then the objects were deleted afterwards after it was verified to be working and not creating duplicate records.

There will be further PRs to come later to do more work on error handling and improving the UI for CSV upload.

As there were some failing tests in lite-routing, I have also added some backwards compatibility which is why we need to keep the old serializer around. The tests in lite-routing create sample DNs data and they use the old style headers, so we have the classic lite-api / lite-routing circular dependency issue. There is a ticket to remove these backwards compatible changes and the old serializer once lite-routing has been updated (also a separate ticket).

[LTD-4896](https://uktrade.atlassian.net/browse/LTD-4896)

[LTD-4896]: https://uktrade.atlassian.net/browse/LTD-4896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/external_data/serializers.py
```diff
@@ -104,6 +106,95 @@ def add_bulk_errors(errors, row_number, line_errors):
...
+        "item_list_codes",
+        "item_description",
+        "consignee_name",
+        "end_use",
+        "reason_for_refusal",
+        "spire_entity_id",
+    ]
+
+    @transaction.atomic
+    def validate_csv_file(self, value):
```
depsiatwal, line 129: I think you will need to populate
data with two additional columns in the CSV 

END_USER_FLAG, CONSIGNEE_FLAG

Then once we have the new enity_type we can use these to populate that field.

api/external_data/tests/denial_valid.csv
```diff
@@ -1,5 +1,5 @@
-reference,name,address,notifying_government,country,item_list_codes,item_description,consignee_name,end_use,end_user_flag,consignee_flag,other_role
-FOO123,Jim Example,123 fake street,France,Germany,ABC123,Foo,Fred Food,used in car,true,true,false
-BAR123,Jak Example,123 fake street,France,Germany,ABC123,Foo,Fred Food,used in car,false,true,false
-BAG124,Bob Example,123 fake street,France,Germany,ABC123,Foo,Fred Food,used in car,false,false,true
-BAT123,James Jones,Bob Avenue,France,Germany,ABC123,Foo,Fred Food,used in car,false,false,true
+reference,regime_reg_ref,name,address,notifying_government,country,item_list_codes,item_description,consignee_name,end_use,reason_for_refusal,spire_entity_id
```
depsiatwal, line 1: we need END_USER_FLAG/CONSIGNEE_FLAG here OTHER_ROLE

check this https://github.com/uktrade/lite-api/blob/910d0f824ae491392f9e72c4c06b61a42a9e9f7b/api/external_data/helpers.py#L4

api/external_data/serializers.py
```diff
@@ -44,15 +44,24 @@ class Meta:
 
     def validate(self, data):
         validated_data = super().validate(data)
-        if validated_data.get("is_revoked") and not validated_data.get("is_revoked_comment"):
-            raise serializers.ValidationError({"is_revoked_comment": "This field is required"})
+        if validated_data.get("is_revoked"):
+            if not validated_data.get("is_revoked_comment"):
+                raise serializers.ValidationError({"is_revoked_comment": "This field is required"})
```
saruniitr, line 49: Is this not handled in the frontend form validation?

api/external_data/serializers.py
```diff
@@ -44,15 +44,24 @@ class Meta:
...
+        else:
+            # TODO: this uses DenialFromCSVFileOldSerializer instead of
+            # DenialFromCSVFileSerializer which is for backwards compatibility,
+            # this should be updated once lite-routing test data has been updated
+            if not (set(DenialFromCSVFileOldSerializer.required_headers)).issubset(set(validated_data.keys())):
+                raise serializers.ValidationError("Missing required fields in data")
```
saruniitr, line 55: This also feels like something that can be handled in the form validation.

saruniitr: We control the template that users download so from that we know the header fields, can this be used to figure out which fields are missing and raise errors in the frontend itself?

api/external_data/serializers.py
```diff
@@ -104,6 +113,95 @@ def add_bulk_errors(errors, row_number, line_errors):
...
+                    "reason_for_refusal": serializer.validated_data["reason_for_refusal"],
+                }
+                denial, is_denial_created = models.Denial.objects.update_or_create(
+                    regime_reg_ref=regime_reg_ref, defaults=denial_data
+                )
+
+                if is_denial_created:
+                    logging.info(f"Created new Denial record at row {i}")
+                else:
+                    logging.info(f"Updated existing Denial record at row {i} based on regime_reg_ref")
```
saruniitr, line 170: assuming this is going to be removed later because depending on the number of records uploaded this can make the process very slow.
Also logging should use old format string formatting rather than f-strings.

api/external_data/serializers.py
```diff
@@ -104,6 +113,95 @@ def add_bulk_errors(errors, row_number, line_errors):
...
+
+                # We assume that a DenialEntity object already exists if we can
+                # match on all of the following fields
+                denial_entity_lookup_fields = {
+                    "reference": serializer.validated_data["reference"],
+                    "regime_reg_ref": regime_reg_ref,
+                    "name": serializer.validated_data["name"],
+                    "address": serializer.validated_data["address"],
```
saruniitr, line 178: in some other place remember seeing `address` is not included in identifying if a denial entity already exists - Could this be made consistent across?

api/external_data/serializers.py
```diff
@@ -104,6 +113,95 @@ def add_bulk_errors(errors, row_number, line_errors):
...
+                denial_entity, is_denial_entity_created = models.DenialEntity.objects.update_or_create(
+                    defaults=serializer.validated_data, denial=denial, **denial_entity_lookup_fields
+                )
+
+                if is_denial_entity_created:
+                    logging.info(f"Created new DenialEntity record at row {i}")
+                else:
+                    logging.info(
+                        f"Updated existing DenialEntity record at row {i} based on reference, regime_reg_ref, name, address"
+                    )
```
saruniitr, line 190: same as above.
If logging is really necessary then perhaps batch and emit them at the end.

api/external_data/serializers.py
hnryjmes: it's an interesting point that I haven't really considered properly. we were talking recently about doing more validation in the frontend, so maybe this is a good opportunity to do that. so you would have the frontend form validation read the csv headers only but not load the whole file, before the API call is made? seems like it would work.

@depsiatwal what do you think?

api/external_data/serializers.py
hnryjmes: I chose this way because I think in the case that there are two similar records with `reference`, `regime_reg_ref`, `name` matching but `address` differing, I think we should keep two records to avoid data loss. then the TAU user can decide if the information is relevant or not. we already have the Denial object which is the way of grouping this information for a unique `regime_reg_ref`.

e.g.

(1)
```
DN2000/0000,AB-CD-EF-000,Organisation Name,"1000 Street Name, City Name"
```
(2)
```
DN2000/0000,AB-CD-EF-000,Organisation Name,"2000 Road Name, Other City Name"
```

would be two denial entities. so my preference would be to change other places this is done rather than change what I've done here, but happy to be persuaded otherwise.

api/external_data/serializers.py
depsiatwal: Yeah I don't think we need to record the denial number here but may be worth logging the regime_reg so we have a audit cause. I don't think we need to be too concerned about speed this CSV upload is designed for a handful delta uploads

api/external_data/serializers.py
depsiatwal: Yeah we could add to an array and log at the end  to speed up

api/external_data/serializers.py
hnryjmes: have added this info to https://uktrade.atlassian.net/browse/LTD-4936 as we will probably do this in that ticket

api/external_data/tests/denial_valid.csv
hnryjmes: have added this info to https://uktrade.atlassian.net/browse/LTD-4936 as we will probably do this in that ticket

api/external_data/serializers.py
hnryjmes: the logging has now been updated, both the counts and the `regime_reg_ref` values are collected and then logged at the end of the for-loop, hopefully this should not slow things down too much.

api/external_data/serializers.py
hnryjmes: updated

api/external_data/serializers.py
saruniitr: I think updating form validation is better which makes this serializer simpler.
If you encounter any issues you will have to update frontend as well as validation in API which increases the work, it will like lite-forms.

api/external_data/serializers.py
saruniitr: I think it was in one of your earlier PRs where address is not considered to determine if it is a duplicate or not.

api/external_data/serializers.py
saruniitr: There is a form in the frontend,
https://github.com/uktrade/lite-frontend/blob/dev/caseworker/external_data/forms.py#L6

I think by doing validation in the API it is going to increase work in future if you need to make any changes. Similar to lite-forms.
I don't know how frequently they upload or how bulky these uploads are but I think doing more of validation in the frontend makes it easier and API endpoint can just create records.

api/external_data/serializers.py
depsiatwal: ahh yeah that was very different to stop loading new entity for a handful of DenialEntity Matches that's because we need a key just to make sure we don't load that particular denial entity object again

depsiatwal (approved):
Obviously not ready to go live just yet but step towards right direction

api/external_data/serializers.py
hnryjmes: @saruniitr 

I'm happy to try adding some CSV validation in the frontend such as checking required headers, as we can do that based on the example csv which is already in the frontend.

There's some other tests in lite-routing (BDD tests) that already exist that directly make requests to the api, and I am also updating those soon. The validation I'm doing in the api in this PR ensures that those tests continue to be useful.

If the api trusted anything it got because the validation was all done in the frontend, then it seems to me those routing tests could do whatever they liked to create denial data, including badly formed data, and then those routing tests would appear to pass, but that just feels weird to me. Maybe this is a temporary problem we are experiencing because we are allowing almost all fields to be nullable and blankable (hence serializer validation is not that useful yet) because we're still working out the data model and doing data migration of existing records etc. 

My proposal is to try adding some CSV validation to the frontend at a later time, in the form you linked, and then, if it seems like we have redundant validation going on, there can be later work to remove the validation from api. Does that work?

depsiatwal: @saruniitr issue we have here is that we are sticky trying to go fast for the first parse. This has been sold as the happy path low validation flow. 

As already mentioned by @hnryjmes  this is the first change of many this is not seen ready to go to PROD yet. We have further fields to add also so my take is we add in the validation/fields.

saruniitr: ok, please include a follow-up in the epic to add validation in the frontend.

hnryjmes: done https://uktrade.atlassian.net/browse/LTD-4943

saruniitr (approved):

PR #1926, opened 2024-04-16
[LTD-4893] Fix bug relating to submission of re-opened applications

currycoder:
### Aim

This change fixes an issue where a re-opened application could not be submitted.

The issue manifests in the following situation;
- The case has two GoodOnApplication objects pointing to the same Good.
- The case has had final advice.
- The application is re-opened for editing.
- One of the GoodOnApplication records has been removed.

This would result in an IndexError at the point of generating the application document during submission.

This change fixes the issue by skipping any final advice where a matching GoodOnApplication cannot be found.

[LTD-4893](https://uktrade.atlassian.net/browse/LTD-4893)

[LTD-4893]: https://uktrade.atlassian.net/browse/LTD-4893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/letter_templates/context_generator.py
+++ b/api/letter_templates/context_generator.py
@@ -885,6 +885,8 @@ def _get_goods_context(application, final_advice, licence=None):
     # Ensure that for each proviso final advice record, we add a record to the goods
     #  context
     for advice in final_advice:
+        # Ignore final advice records where we have no associated good on application
+        # - either our mapping value is missing or is an empty list so skip it.
         if not good_ids_to_goods_on_application.get(advice.good_id):
             continue
         # Grab the next GoodOnApplication for this Good.id - this ensures that
--- a/api/letter_templates/context_generator.py
+++ b/api/letter_templates/context_generator.py
@@ -885,11 +885,12 @@ def _get_goods_context(application, final_advice, licence=None):
     # Ensure that for each proviso final advice record, we add a record to the goods
     #  context
     for advice in final_advice:
-        if advice.good_id in good_ids_to_goods_on_application:
-            # Grab the next GoodOnApplication for this Good.id - this ensures that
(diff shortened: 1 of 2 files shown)


saruniitr (approved):

api/letter_templates/context_generator.py
```diff
@@ -885,11 +885,12 @@ def _get_goods_context(application, final_advice, licence=None):
...
-            # Grab the next GoodOnApplication for this Good.id - this ensures that
-            #  each GoodOnApplication is present once on the end licence
-            good_on_application = good_ids_to_goods_on_application[advice.good_id].pop(0)
-            goods_context[advice.type].append(_get_good_on_application_context_with_advice(good_on_application, advice))
+        if not good_ids_to_goods_on_application.get(advice.good_id):
+            continue
+        # Grab the next GoodOnApplication for this Good.id - this ensures that
+        #  each GoodOnApplication is present once on the end licence
+        good_on_application = good_ids_to_goods_on_application[advice.good_id].pop(0)
+        goods_context[advice.type].append(_get_good_on_application_context_with_advice(good_on_application, advice))
```
kevincarrogan, line 893: This has taken me a while to get as I keep reading the two if statements as equivalent.

I'm presuming that actually what's happening here is that eventually the list in the dict is popping itself into an empty list at some point and the `.get` here is both checking whether there is an item in the dictionary _and_ if there is a list that it has items in it?

currycoder: Exactly that.  I will add a comment to explain as such

kevincarrogan (approved):

PR #1930, opened 2024-04-17
LTD-4933-es-search-model-changes

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4933

Changes denial search to use new fields from Denial Table 

Also the serialiser which is used to View on the UI and update on the UI has been updated to use the fields from Denial rather then Denial Entity. 

I've had to go alot further in this PR then I would expected.  The CSV loader and ingest_denials breaks as we change the default entity serializer as it become ready_only. This is deliberately done like this so we don't break the UI and require further changes. 
  

- We have already ingested a clean set of DenialEntity Objects that were release last week. Which fixed all the duplicates. 
- This PR deals with copying them across to Denial Model  through we are certain there's no duplicates (confirmed by testing locally)
- Wrapped in a transaction just incase so we can roll back and not copy any data into Denials.

### DIFF
--- a/api/external_data/documents.py
+++ b/api/external_data/documents.py
@@ -1,7 +1,7 @@
 from django.conf import settings
 from django_elasticsearch_dsl import Document, fields
 from django_elasticsearch_dsl.registries import registry
-from elasticsearch_dsl import analysis, InnerDoc
+from elasticsearch_dsl import analysis
 from api.external_data import models
 from api.search.application.documents import lowercase_normalizer
 
@@ -72,7 +72,8 @@ postcode_normalizer = analysis.normalizer(
     filter=["lowercase", "asciifolding"],
 )
 
-class DenialEnitytDocument(Document):
+
+class DenialEntityDocument(Document):
     id = fields.KeywordField()
     name = fields.TextField()
     address = fields.TextField(
@@ -84,7 +85,7 @@ class DenialEnitytDocument(Document):
         },
     )
     data = DataField()
-    is_revoked = fields.BooleanField()
+    is_revoked = fields.BooleanField(attr="denial.is_revoked")
 
     denial = fields.ObjectField(
         attr="denial",
@@ -125,7 +126,6 @@ class DenialEnitytDocument(Document):
                     "raw": fields.KeywordField(),
                 },
             ),
-            "is_revoked": fields.BooleanField(),
         },
     )
 
@@ -142,6 +142,11 @@ class DenialEnitytDocument(Document):
 
     class Django:
         model = models.DenialEntity
+        related_models = [models.Denial]
+
+    def get_instances_from_related(self, related_instance):
+        if isinstance(related_instance, models.Denial):
+            return related_instance.denial_entity.all()
 
 
 class SanctionDocumentType(Document):
@@ -176,4 +181,4 @@ class SanctionDocumentType(Document):
 
 
 if settings.LITE_API_ENABLE_ES:
-    registry.register_document(DenialEnitytDocument)
+    registry.register_document(DenialEntityDocument)
--- a/api/external_data/documents.py
+++ b/api/external_data/documents.py
@@ -72,11 +72,6 @@ postcode_normalizer = analysis.normalizer(
     filter=["lowercase", "asciifolding"],
 )
 
-
-class Denial(InnerDoc):
-    regime_reg_ref = fields.KeywordField()
-
-
 class DenialEnitytDocument(Document):
     id = fields.KeywordField()
     name = fields.TextField()
@@ -96,12 +91,6 @@ class DenialEnitytDocument(Document):
         properties={
             "regime_reg_ref": fields.TextField(
                 attr="regime_reg_ref",
-                fields={
-                    "raw": fields.KeywordField(),
-                },
-            ),
-            "notifying_government": fields.TextField(
-                attr="notifying_government",
                 fields={
                     "raw": fields.KeywordField(),
                 },
--- a/api/external_data/documents.py
+++ b/api/external_data/documents.py
@@ -1,8 +1,7 @@
 from django.conf import settings
 from django_elasticsearch_dsl import Document, fields
 from django_elasticsearch_dsl.registries import registry
-from elasticsearch_dsl import analysis
-
+from elasticsearch_dsl import analysis, InnerDoc
 from api.external_data import models
 from api.search.application.documents import lowercase_normalizer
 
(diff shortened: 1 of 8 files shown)


api/external_data/documents.py
```diff
@@ -74,27 +73,67 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
...
         },
     )
-    item_list_codes = fields.TextField()
-    item_description = fields.TextField()
-    consignee_name = fields.TextField()
-    end_use = fields.TextField()
     data = DataField()
     is_revoked = fields.BooleanField()
 
+    denial = fields.ObjectField(
```
saruniitr, line 94: Is `NestedField()` not going to work here? then you can directly reference the `Denial` document here.
There are examples in product search,
https://github.com/uktrade/lite-api/blob/dev/api/search/product/documents.py#L136

One thing to note about nested fields is that you won't be able to do full text searches. We had this issue in Product search so we flattened them and included in the main document itself to be able to do full text search on those fields.

api/external_data/documents.py
```diff
@@ -74,27 +73,67 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
 )
 
 
-class DenialDocumentType(Document):
+class Denial(InnerDoc):
```
saruniitr, line 76: this is not used anywhere yet

api/external_data/serializers.py
```diff
@@ -48,6 +57,15 @@ def validate(self, data):
             raise serializers.ValidationError({"is_revoked_comment": "This field is required"})
         return validated_data
 
+    def update(self, instance, validated_data):
+        # This is required because the flattened columns are required to support the older denial screen
+        # Serialisers don't support dotted fileds so we need to override. is_revoked and is_revoked_comments
+        # are the only updates we support so this is ok.
+        instance.denial.is_revoked = validated_data["denial"]["is_revoked"]
+        instance.denial.is_revoked_comment = validated_data["denial"]["is_revoked_comment"]
+        instance.denial.save()
```
saruniitr, line 66: is it not possible to add `denial` field with its own serializer for this?

depsiatwal: yeah I thought about that and tired it. 
Issue it changes the structure as everything will be in another dict {denial:[]}

hence did this to maintain backward compatibility it allows is to remove the redundant fields  
then depending on where we go with the search/display of denials  this will probably change again

api/external_data/documents.py
depsiatwal: removed

api/external_data/documents.py
depsiatwal: That's ok at present their used for display only the field we do filter on is_revoked is flattened   now

api/external_data/documents.py
```diff
@@ -74,26 +73,61 @@ def get_value_from_instance(self, instance, field_value_to_ignore=None):
 )
 
 
-class DenialDocumentType(Document):
+class DenialEnitytDocument(Document):
```
hnryjmes, line 76: I think this should be `DenialEntityDocumentType`

depsiatwal: Yeah I seen Type on the end on some but why ?

why type ?

hnryjmes: fair point. I looked at the examples in the django-elasticsearch-dsl and elasticsearch-dsl and they are just named things like `CarDocument` so there is no Type anywhere

`DenialEntityDocument` is fine then. I just didn't like the typo you have currently, `DenialEnitytDocument` - this should be changed

api/external_data/serializers.py
```diff
@@ -27,7 +59,6 @@ class Meta:
             "country",
             "item_list_codes",
             "item_description",
-            "consignee_name",
```
hnryjmes, line 30: will this be removed from the model also?

depsiatwal: yeah there's another Jira to do after all this is stable and remove all the old fields.

api/external_data/serializers.py
```diff
@@ -87,22 +138,26 @@ def validate_csv_file(self, value):
...
-            if serializer.is_valid() and isinstance(serializer.validated_data, dict):
+            denial_serializer = DenialSerializer(data=denial_data)
+            is_valid_denial_data = denial_serializer.is_valid()
+
+            if len(denial_serializer.errors) == 1 and denial_serializer.errors.get("regime_reg_ref"):
+                # This is a bit of a hack presently we care about uniqueness but do want to validate the fields
+                # Hence we override the check
+                if denial_serializer.errors.get("regime_reg_ref")[0].code == "unique":
+                    is_valid_denial_data = True
+
```
hnryjmes, line 157: I can see that this works having the code for this hack inside the csv serializer itself. Did you also consider having this code live in the `validate()` function that you would override inside the `DenialSerializer`? 

Related question is if you would expect the issue that this hack is intended to fix to also appear when using the `ingest_denials` command or if it is only something you've seen when working with the csv file uploader

This is just a thought, not requesting that you change anything right now

hnryjmes (commented):
currently there is a test `test_denial_entity_search` which asserts that the number of returned results is as expected (https://github.com/uktrade/lite-api/pull/1930/files#diff-dffbd914b75de92b95bd382233c502170cf9e73a87dd8d26abde0dc05be4bbc4R314)

(or this link https://github.com/uktrade/lite-api/blob/LTD-4933-es-search-model-changes/api/external_data/tests/test_views.py#L314)

given the changes we are making to the search functionality I would ideally want more testing of specific data being returned from the search query e.g. the fields listed in the denial required headers and denial entity required headers.

not sure if it's appropriate to add that extra test coverage now but wanted to make this comment so it's here for others to think about.

api/external_data/serializers.py
depsiatwal: I did consider putting this and overriding validate() but I think that's a bad pattern. I expect we would use this DenialSerializer to create data from the UI one day where would want to ensure regime_reg_ref is unique. 

I think a better solution may be to have an update serializer and update validate there. But I think for now we just want to do first iteration and we may need to revisit this one day. 


I haven't really tested the recent ingest_denails it's not a common usercase once the denials are loaded we probably want to think about how we maintain the data going forwards and most likely that will be CSV. As ingest denials was more geared towards a one off load.

depsiatwal:
> currently there is a test `test_denial_entity_search` which asserts that the number of returned results is as expected (https://github.com/uktrade/lite-api/pull/1930/files#diff-dffbd914b75de92b95bd382233c502170cf9e73a87dd8d26abde0dc05be4bbc4R314)
> 
> (or this link https://github.com/uktrade/lite-api/blob/LTD-4933-es-search-model-changes/api/external_data/tests/test_views.py#L314)
> 
> given the changes we are making to the search functionality I would ideally want more testing of specific data being returned from the search query e.g. the fields listed in the denial required headers and denial entity required headers.
> 
> not sure if it's appropriate to add that extra test coverage now but wanted to make this comment so it's here for others to think about.

Fair point but probably not one for this PR it's already bigger then we want it to be.

api/external_data/serializers.py
```diff
@@ -87,22 +138,26 @@ def validate_csv_file(self, value):
...
             # Create a serializer instance to validate data
             serializer = DenialEntitySerializer(data=denial_entity_data)
-            if serializer.is_valid() and isinstance(serializer.validated_data, dict):
+            denial_serializer = DenialSerializer(data=denial_data)
+            is_valid_denial_data = denial_serializer.is_valid()
+
+            if len(denial_serializer.errors) == 1 and denial_serializer.errors.get("regime_reg_ref"):
+                # This is a bit of a hack presently we care about uniqueness but do want to validate the fields
+                # Hence we override the check
+                if denial_serializer.errors.get("regime_reg_ref")[0].code == "unique":
```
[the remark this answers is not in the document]
depsiatwal: I check for denial_serializer.errors.get("regime_reg_ref"): earlier and that   returns a validation class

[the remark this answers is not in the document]
depsiatwal: added further check and also since denial_serializer.errors.get("regime_reg_ref") removed the 2nd check to avoid confusion

PR #1932, opened 2024-04-18
[LTD-4939] Optimise assessment endpoint

currycoder:
### Aim

This PR ensures that elasticsearch documents which are created by the django-elasticsearch-dsl library are created asynchronously with celery.  The current way that documents are created synchronously means that write-heavy operations incur a significant time cost in lite-api.

For example, the assessment endpoint takes >30seconds when ~100 products are assessed.  This leads to 504 timeout errors on production.  With the changes in this PR, the assessment call is further reduced to ~10seconds.

There are further speed gains to be had by reviewing our custom elasticsearch document indexing, however the changes present in this PR seem to be enough for now.

[LTD-4939](https://uktrade.atlassian.net/browse/LTD-4939)

[LTD-4939]: https://uktrade.atlassian.net/browse/LTD-4939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conftest.py
currycoder, line 1: Relocated so that lite_routing tests would also benefit from some of the global setup that this does; specifically setting celery to always eager (which ensures that denial documents are present in the search index).

api/conf/settings.py
```diff
@@ -340,6 +340,7 @@ def _build_redis_url(base_url, db_number, **query_args):
     ELASTICSEARCH_DSL = {
         "default": {"hosts": env.str("ELASTICSEARCH_HOST")},
     }
+    ELASTICSEARCH_DSL_SIGNAL_PROCESSOR = "api.search.signals.ElasticsearchDSLSignalProcessor"
```
depsiatwal, line 343: just trying to figure out how this works 
guess it's a common pattern 
so this registers the signal and then the handle save is an event that gets attached ?

depsiatwal (commented):
nicely done 

Will this be tested in some test ENV (you may have already) 

also a quick q for my understanding

api/conf/settings.py
currycoder: It's just an optional configuration setting that the django-elasticsearch-dsl library offers; https://django-elasticsearch-dsl.readthedocs.io/en/latest/settings.html#elasticsearch-dsl-signal-processor

Initially, I was using their "off the shelf" CelerySignalProcessor - but I found that processor would queue celery tasks for *all model saves* regardless of whether we were indexing the model in question in elasticsearch.  So I've created an overridden processor which checks whether we care about the model before queueing the task.

saruniitr (dismissed):

depsiatwal (dismissed):
lgtm

saruniitr (approved):

depsiatwal (approved):
lgtm

PR #1936, opened 2024-04-19
Ltd 4935 migration to populate entity type

markj0hnst0n:
### Aim

Populate new DenialEntity entity_type field as a choice of 3 strings End-user, Consignee or Third Party based on json data currently stored in the data field.


api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 4.2.11 on 2024-04-18 12:00
+
+from django.db import migrations
+
+# from django.db import transaction
+# import logging
```
hnryjmes, line 6: you can get rid of these if they're not being used anymore

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,48 @@
...
+        entity_type = "End-user"
+    elif (
+        normalised_entity_type_dict.get("end_user_flag", "false") == "false"
+        and normalised_entity_type_dict.get("consignee_flag", "false") == "true"
+    ):
+        entity_type = "Consignee"
+    elif (
+        normalised_entity_type_dict.get("end_user_flag", "false") == "false"
+        and normalised_entity_type_dict.get("consignee_flag", "false") == "false"
+        and isinstance(normalised_entity_type_dict["other_role"], str)
```
hnryjmes, line 27: (this isn't a suggestion to change anything, just a comment)

I'm not sure how this bit of the code should behave if the value for `other_role` is an empty string. it seems like what we have here in this block is going to run if there's any value in that column including it being empty.

maybe if that's what we want we can communicate that more simply by not making this bit an `elif` and just using `else`. what do you think?

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,48 @@
...
+
+    return entity_type
+
+
+def set_denial_entity_type(apps, schema_editor):
+
+    DenialEntity = apps.get_model("external_data", "DenialEntity")
+    for denial_entity in DenialEntity.objects.all():
+        denial_entity.entity_type = get_denial_entity_type(denial_entity.data)
+        denial_entity.save()
```
hnryjmes, line 39: once this is committed to dev branch it will run as a migration every time the full set of migrations are run, and it seems to be that it's only really the first time that we need to set `entity_type`.

do we definitely want to set it based on the `data` dict each time the migration is run? if it just needs to be run once, maybe this bit of code can be conditional on `entity_type` not already existing on the found `DenialEntity` object which might make the migration quicker to run as it will be able to skip objects that don't need updating in future.

hnryjmes (commented):
looks good to me, happy to approve after you've had a chance to respond to the other comments

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 4.2.11 on 2024-04-18 12:00
+
+from django.db import migrations
+
+
+def get_denial_entity_type(data):
```
saruniitr, line 6: I think retrieving the value and also checking at the same time in conditions is making it hard to follow, especially the same thing being repeated in all conditions.
Please consider something like below to simplify,
```
entity_type_dict = {key.lower(): value.lower() for key, value in data.items()}

is_end_user = entity_type_dict.get('end_user_flag', 'false') == 'true'
is_consignee = entity_type_dict.get('consignee_flag', 'false') == 'true'
is_third_party = entity_type_dict.get('other_role', 'false') == 'true'

if is_end_user:
    entity_type = DenialEntityType.END_USER
if is_consignee:
    entity_type = DenialEntityType.CONSIGNEE

# noticed below in the data so we need to handle this as well
if is_end_user and is_consignee:
    entity_type = DenialEntityType.END_USER.   # please verify what is required here.

if is_third_party:
    entity_type = DenialEntityType.THIRD_PARTY
```

Also please add if there are any other combinations

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,43 @@
...
+    ):
+        entity_type = "Third-party"
+
+    return entity_type
+
+
+def set_denial_entity_type(apps, schema_editor):
+
+    DenialEntity = apps.get_model("external_data", "DenialEntity")
+    for denial_entity in DenialEntity.objects.all():
```
saruniitr, line 31: is `DenialEntity.objects.filter(entity_type__isnull=True)` better?

api/external_data/tests/test_0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,89 @@
...
+        "address": "1000 Street Name, City Name",
+        "notifying_government": "Country Name",
+        "country": "Country Name",
+        "item_list_codes": "0A00100",
+        "item_description": "Medium Size Widget",
+        "consignee_name": "Example Name",
+        "end_use": "Used in industry",
+        "reason_for_refusal": "Risk of outcome",
+        "spire_entity_id": 123,
+        "data": {"END_USER_FLAG": "true", "CONSIGNEE_FLAG": "false", "OTHER_ROLE": ""},
```
saruniitr, line 19: data indicates both `end_user_flag` and `consignee_flag` can be true so please test for that as well.

saruniitr (changes requested):

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,42 @@
...
+
+    return entity_type
+
+
+def set_denial_entity_type(apps, schema_editor):
+
+    DenialEntity = apps.get_model("external_data", "DenialEntity")
+    for denial_entity in DenialEntity.objects.filter(entity_type__isnull=True):
+        denial_entity.entity_type = get_denial_entity_type(denial_entity.data)
+        denial_entity.save()
```
saruniitr, line 33: For the first time this may still take quite a lot of time as `entity_type` is not set.
We can actually query based on fields in json data as well, perhaps best to investigate a bit where by querying fields within json `data` if we can filter all end_users, consginee etc then we can just `update()` with one query rather than `save()` for each item.

api/external_data/tests/test_0023_set_denial_entity_type.py
markj0hnst0n: I've used:

`class DenialEntityType:
    CONSIGNEE = "consignee"
    END_USER = "end_user"
    THIRD_PARTY = "third_party"

    choices = ((CONSIGNEE, "Consignee"), (END_USER, "End-user"), (THIRD_PARTY, "Third party"))
`
as the 3 options for the DenialEntity.entity_type.  Would it be appropriate to use one of them or give another field?
@saruniitr @depsiatwal

api/external_data/migrations/0023_set_denial_entity_type.py
markj0hnst0n: The unit tests I wrote tested for this scenario and you're right it does produce a "Third-Party" result with an empty string.  Should the response be something else?

api/external_data/migrations/0023_set_denial_entity_type.py
markj0hnst0n: Not sure what the output should be for end_user_flag and consignee_flag both being true.  Have queried below

api/external_data/tests/test_0023_set_denial_entity_type.py
saruniitr: No, it is not about adding another option.
If both end_user_flag and consignee_flag is 'true' then I think that is case where consignee and end-user are the same or there is no consignee so the entity_type is most likely 'end-user' in this case.
I think this needs to be discussed with UCD/business on how SPIRE uses these fields.

markj0hnst0n: I see.  Thanks!

api/external_data/migrations/0023_set_denial_entity_type.py
depsiatwal: hmmm why would it run every time ?
you mean locally ? when we migrate

api/external_data/migrations/0023_set_denial_entity_type.py
markj0hnst0n: @saruniitr  I've made changes based on this comment.  I think I have implemented what you mean!  Although maybe it should have an else statement which populates with an empty string after to prevent the code below the for statement firing every time?

api/external_data/tests/test_0023_set_denial_entity_type.py
markj0hnst0n: made amendments based on this but will seek further guidance

depsiatwal: @markj0hnst0n  can you reach out to Collin Harrall to get some more assistance ?

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,47 @@
...
+    is_consignee_flag = normalised_entity_type_dict.get("consignee_flag", "false") == "true"
+    third_party_string_has_value = isinstance(normalised_entity_type_dict["other_role"], str)
+
+    if is_end_user_flag and is_consignee_flag:
+        entity_type = DenialEntityType.END_USER
+    elif not is_end_user_flag and is_consignee_flag:
+        entity_type = DenialEntityType.CONSIGNEE
+    elif is_end_user_flag and not is_consignee_flag:
+        entity_type = DenialEntityType.END_USER
+    elif is_end_user_flag == False and is_consignee_flag == False and third_party_string_has_value == True:
```
saruniitr, line 22: `third_party_string_has_value` is always going to be `True` here because all of the fields values are strings.
I guess you only want this to be True if that is a non-empty string correct?

Looking at the data for 'other_role' there are no consistent values for this field.
Eg there is 'Not specified' also, so in this case can this be considered as third party?
I think this also need to be discussed with UCD.

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def get_denial_entity_type(data):
+
+    entity_type = ""
+    normalised_entity_type_dict = {keys.lower(): values.lower() for keys, values in data.items()}
+
+    is_end_user_flag = normalised_entity_type_dict.get("end_user_flag", "false") == "true"
+    is_consignee_flag = normalised_entity_type_dict.get("consignee_flag", "false") == "true"
+    third_party_string_has_value = isinstance(normalised_entity_type_dict["other_role"], str)
```
saruniitr, line 14:
```
other_role = normalised_entity_type_dict.get("other_role", "")
```

then further down you can check whether it is empty/not-empty etc and add conditions accordingly.

api/external_data/tests/test_0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,89 @@
...
+        for row in test_data:
+            DenialEntity.objects.create(**row)
+
+    def test_0023_set_denial_entity_type(self):
+        DenialEntity = self.new_state.apps.get_model("external_data", "DenialEntity")
+
+        assert DenialEntity.objects.all().count() == 4
+        assert DenialEntity.objects.filter(entity_type="end_user").count() == 2
+        assert DenialEntity.objects.filter(entity_type="consignee").count() == 1
+        assert DenialEntity.objects.filter(entity_type="third_party").count() == 1
```
saruniitr, line 89: we could use the same enums `DenialEntityType` here as well.

saruniitr (commented):
Thanks for your patience.
This is looking good, just need clarifications on the other_role conditions.

api/external_data/tests/test_0023_set_denial_entity_type.py
markj0hnst0n: done!

api/external_data/migrations/0023_set_denial_entity_type.py
markj0hnst0n: I think I've implemented this in a slightly different way

api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 4.2.11 on 2024-04-18 12:00
+
+from django.db import migrations
+from api.external_data.enums import DenialEntityType
+
+
+def get_denial_entity_type(data):
+
+    entity_type = ""
+    normalised_entity_type_dict = {keys.lower(): values.lower() for keys, values in data.items()}
```
depsiatwal, line 10: What would happen here if data is none I think we need to check to make sure data is a dict else migration would fail.

saruniitr (approved):
looks good to me.

hnryjmes (approved):

PR #1941, opened 2024-04-24
LTD-4950 Update lite-api test csv files to match example csv

hnryjmes:
### Aim

The example csv used in lite-frontend was updated to have 3 extra columns and also to reorder the columns (https://github.com/uktrade/lite-frontend/pull/1907) so this just does the same for the test csv file used in lite-api.

EDIT: As `consignee_name` as a column is not needed it has been removed from the example csv, and so a lot of the changes in this PR are now to keep things consistent now that this field is not expected to be used.

[LTD-4950](https://uktrade.atlassian.net/browse/LTD-4950)

[LTD-4950]: https://uktrade.atlassian.net/browse/LTD-4950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/external_data/serializers.py
+++ b/api/external_data/serializers.py
@@ -65,7 +65,6 @@ class DenialFromCSVFileSerializer(serializers.Serializer):
         "country",
         "item_list_codes",
         "item_description",
-        "consignee_name",
         "end_use",
         "reason_for_refusal",
         "spire_entity_id",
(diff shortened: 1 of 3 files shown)


markj0hnst0n (approved):
looks grand

depsiatwal (commented):
possible to remove consignee_name

though it's still in current code base my next PR removes that field since it's not used or populated.

depsiatwal (approved):
lgtm

PR #1942, opened 2024-04-25
Ltd 4935 handle empty strings

markj0hnst0n:
### Aim

Similar to the previous pull request on this ref but now it filters the db objects it's looking for by an empty string.  This is because the entity_type field has been populated with an empty string in the db from the previous migration.  Issue was noticed when checking dev environment but this should fix it.

### DIFF
--- a/api/external_data/migrations/0023_set_denial_entity_type.py
+++ b/api/external_data/migrations/0023_set_denial_entity_type.py
@@ -30,10 +30,11 @@ def set_denial_entity_type(apps, schema_editor):
 
     DenialEntity = apps.get_model("external_data", "DenialEntity")
 
-    for denial_entity in DenialEntity.objects.filter(entity_type__isnull=True):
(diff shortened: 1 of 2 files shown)


api/external_data/migrations/0023_set_denial_entity_type.py
```diff
@@ -30,7 +30,7 @@ def set_denial_entity_type(apps, schema_editor):
 
     DenialEntity = apps.get_model("external_data", "DenialEntity")
 
-    for denial_entity in DenialEntity.objects.filter(entity_type__isnull=True):
+    for denial_entity in DenialEntity.objects.filter(entity_type=""):
```
depsiatwal, line 33: This won't have any effect now as this has already been merged up. 
You would need to create a new migration.  I think this didn't work at least locally. So you might have to create a new one may be check for both entity_type__isnull=True or entity_type__is=""

depsiatwal (commented):
please create a new migration

PR #1945, opened 2024-04-25
Ltd-4935 added new migration to backfill entity_type empty strings

markj0hnst0n:
### Aim

Similar to the previous pull request on this ref but now it filters the db objects it's looking for by an empty string. This is because the entity_type field has been populated with an empty string in the db from a previous migration. Issue was noticed when checking dev and uat environment but this new migration should fix it.


depsiatwal (commented):
LGTM however I will test locally before merging

depsiatwal (approved):
tested and all worked locally

hnryjmes (approved):

PR #1947, opened 2024-04-26
Ltd 4938  ingest denials.py to use new entity type field

markj0hnst0n:
### Aim

Modify the ingest_denials management command to populate the new entity_type field in the database when this pulls data from Spire.

[LTD-4938](https://uktrade.atlassian.net/browse/LTD-4938)


api/external_data/serializers.py
```diff
@@ -11,8 +11,17 @@
 from api.flags.enums import SystemFlags
 
 
+class EntityTypeSerializerField(serializers.Field):
```
depsiatwal, line 14: not sure why we have a separate field here 

will it not just come from DB now like the other fields ?

PR #1954, opened 2024-04-29
Ltd 4938 ingest denials.py to use new entity type field

markj0hnst0n:
### Aim

Modify the ingest_denials management command to populate the new entity_type field in the database when this pulls data from Spire.

[LTD-4938](https://uktrade.atlassian.net/browse/LTD-4938)

[LTD-4938]: https://uktrade.atlassian.net/browse/LTD-4938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/external_data/helpers.py
```diff
@@ -15,3 +18,24 @@ def get_denial_entity_type(data):
...
+
+
+def get_denial_entity_type_value(data):
+
+    if isinstance(data, dict):
+        entity_type = ""
+
+        is_end_user_flag = data.get("end_user_flag", "false") == "true"
+        is_consignee_flag = data.get("consignee_flag", "false") == "true"
+        is_other_role = len(data.get("other_role", "")) > 0
```
saruniitr, line 30: just like in migrations is it not required to convert to `lower()`?
In the csv file if the case is different then these checks give a different value.

PR #1955, opened 2024-04-30
LTD: Add additional fields for Crypto application

saruniitr:
### Aim


api/applications/views/applications.py
```diff
@@ -272,7 +272,7 @@ def put(self, request, pk):
             save_and_audit_have_you_been_informed_ref(request, application, serializer)
             serializer.save()
 
-        if application.case_type.sub_type == CaseTypeSubTypeEnum.F680:
+        if application.case_type.sub_type in [CaseTypeSubTypeEnum.F680, CaseTypeSubTypeEnum.OPEN]:
```
Tllew, line 275: I have no problem with this, I've just noticed in a few places we've treated OPEN as oiels and am aware there's no other open licences currently but that might not be the plan going forward, happy with this for now just thinking this might need to change if we add another open licence

Tllew (approved):
looks good, linting failed

api/applications/views/applications.py
saruniitr: Restricted to crypto only

PR #1961, opened 2024-05-01
hotfix-migration-disable-build

depsiatwal:
HOTFIX in Prod where Migration times out due to ES search index rebuilding.


hnryjmes (approved):
looks the same as the ones already merged to uat and master so happy to approve

PR #1963, opened 2024-05-01
Use bulk updates and creates for denials migrations

kevincarrogan:
### Aim


Tllew (approved):
Looks good to me

PR #1964, opened 2024-05-01
Ltd 4941 remove data field and use entity type

markj0hnst0n:
### Aim

Stop using the data field to populate entity_type and use the newly created field

[LTD-4941](https://uktrade.atlassian.net/browse/LTD-4941)

### DIFF
--- a/api/external_data/serializers.py
+++ b/api/external_data/serializers.py
@@ -7,7 +7,8 @@ from django_elasticsearch_dsl_drf.serializers import DocumentSerializer
 from rest_framework import serializers
 
 from api.external_data import documents, models
-from api.external_data.helpers import get_denial_entity_type_from_db
+
+# from api.external_data.helpers import get_denial_entity_type_from_db
 from api.flags.enums import SystemFlags
 
 
@@ -232,7 +233,7 @@ class DenialFromCSVFileSerializer(serializers.Serializer):
 
 
 class DenialSearchSerializer(DocumentSerializer):
-    entity_type = serializers.SerializerMethodField()
+    entity_type = serializers.ReadOnlyField(source="denial.entity_type")
     regime_reg_ref = serializers.ReadOnlyField(source="denial.regime_reg_ref")
     reference = serializers.ReadOnlyField(source="denial.reference")
     notifying_government = serializers.ReadOnlyField(source="denial.notifying_government")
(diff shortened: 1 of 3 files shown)


markj0hnst0n:
This is currently failing routing lite routing tests and showing entity_type from it's db field and not 'data' but not in human readable form

api/external_data/tests/test_views.py
```diff
@@ -335,7 +335,7 @@ def test_populate_denial_entity_objects(self, page_query):
             self.assertEqual(response_json["results"][0][key], value)
         self.assertEqual(len(response_json["results"]), 2)
         self.assertEqual(response_json["total_pages"], 1)
-        assert "entity_type" in response_json["results"][0]
+        # assert "entity_type" in response_json["results"][0]
```
depsiatwal, line 338: Why the comment do we still need entity_type

Also what is UI expected here does this return choices data ?

markj0hnst0n: I think I've fixed this with recent changes.  Just testing now

markj0hnst0n:
Corresponding front end pr https://github.com/uktrade/lite-frontend/pull/1931

api/external_data/serializers.py
```diff
@@ -232,7 +232,6 @@ def add_bulk_errors(errors, row_number, line_errors):
 
 
 class DenialSearchSerializer(DocumentSerializer):
-    entity_type = serializers.SerializerMethodField()
```
saruniitr, line 235: when there are choices we usually use `KeyValueChoiceField()` which will give `key` (saved in db) and `value` (that can be displayed in UI).
In FE you can just display the value, you don't need `if` conditions.
Eg https://github.com/uktrade/lite-api/blob/dev/api/cases/serializers.py#L67

saruniitr (commented):
comment about serializer field

api/external_data/serializers.py
markj0hnst0n: When I do this I'm getting a dict showing on the front end I'm a get_entity_type_value method as described here Extra instance methods[¶](https://docs.djangoproject.com/en/5.0/ref/models/instances/#extra-instance-methods) but I'm not sure I'm on the right track

depsiatwal: you need a frontend change to display the actual value from the dict

saruniitr: you get a dict on the FE something like `{'key': 'some key', 'value': 'some value'}` and you just display value from that.

saruniitr (approved):

kevincarrogan (approved):

depsiatwal (approved):
lgtm

PR #1968, opened 2024-05-02
Send Crypto licence details to lite-hmrc

saruniitr:
### Aim


api/licences/serializers/hmrc_integration.py
```diff
@@ -117,15 +114,12 @@ def get_old_id(self, instance):
         )
 
     def get_countries(self, instance):
+        countries = []
         if hasattr(instance.case, "baseapplication") and hasattr(instance.case.baseapplication, "openapplication"):
-            countries = get_approved_countries(instance.case.baseapplication)
-        else:
-            countries = instance.case.opengenerallicencecase.open_general_licence.countries.order_by("name")
+            excluded_crypto_destinations = [entry[1] for entry in CryptoOIEL.EXCLUDED_COUNTRIES]
+            countries = Country.objects.filter(id__in=excluded_crypto_destinations)
```
currycoder, line 120: How do these get expressed as "exclude" when this is sent through to lite-hmrc?

saruniitr: I see couple of options,
 - As all of these are non-permitted as this list is stable, we could just format the country line as excluded when preparing the message in lite-hmrc
 - Update the payload that is sent from lite-api to lite-hmrc to include additional field that determines whether it is allowed/excluded and that can be used to format the message in lite-hmrc

In this PR just wanted to get the payload ready to be sent, likely some changes required on lite-hmrc for the serializer etc for this call to succeed on the other end, which can be looked at if we are spending more time on this.

currycoder (approved):

PR #1972, opened 2024-05-03
LTD-4989: Update data model to capture multiple ARS values for each line item

saruniitr:
## Change description

Some of the products need assessments with multiple report summary values. Currently we only support one value so update data model to be able to capture multiple values for each line item.

For each report summary we need to preserve the combination of prefix and subject hence a separate model is created for this.

The changes that updates endpoints, back population etc will follow in subsequent PRs.

[LTD-4989](https://uktrade.atlassian.net/browse/LTD-4989)

[LTD-4989]: https://uktrade.atlassian.net/browse/LTD-4989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/staticdata/report_summaries/models.py
```diff
@@ -18,3 +18,20 @@ class ReportSummarySubject(models.Model):
...
+
+
+class ReportSummary(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    prefix = models.ForeignKey(
+        ReportSummaryPrefix, blank=True, null=True, related_name="prefix", on_delete=models.DO_NOTHING
+    )
+    subject = models.ForeignKey(
+        ReportSummarySubject, blank=True, null=True, related_name="subject", on_delete=models.DO_NOTHING
+    )
```
currycoder, line 30: Just thinking through how this might be used... Say that we have a `ReportSummary` record as follows;
```
ReportSummary.objects.create(
    prefix=ReportSummaryPrefix.objects.get(name="components for"), 
    subject=ReportSummarySubject.objects.get("sniper rifles")
)
```

Is that same `ReportSummary` record used for many different assessments across many different cases/organisations?

If so, maybe it would be good to enforce that have at most one record for each unique combination of prefix/subject e.g. https://docs.djangoproject.com/en/5.0/ref/models/options/#unique-together

saruniitr: yes that is the plan.
As there are set prefixes, subjects we can have a predefined set of `ReportSummary` instances which are linked to line items. There are 38 prefixes and 100+ subjects so approx 40K report summary instances.

kevincarrogan, line 30: I think one of these is always going to have a value so I'm not sure they both need to be allowed to be null.

kevincarrogan (approved):

depsiatwal (approved):
lgtm

PR #1977, opened 2024-05-07
LTD-4995-filter-uk-denials

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4995

It's been confirmed that the business do not wish to see denials entered by UK issued government.


api/external_data/views.py
```diff
@@ -73,7 +73,7 @@ class DenialSearchView(DocumentViewSet):
     }
 
     def filter_queryset(self, queryset):
-        queryset = queryset.filter("term", is_revoked=False)
+        queryset = queryset.filter("term", is_revoked=False).exclude("term", notifying_government="United Kingdom")
```
depsiatwal, line 76: This is a bit concerning that the notifying government is currently loose text. At the moment this is fine as the extract from spire has been done against look up tables.  But with CSV upload we need to be more careful with case and what test is entered here.

kevincarrogan (approved):

PR #1980, opened 2024-05-08
LTD-4990: Update assessments endpoint to prepare for supporting multiple ARS for each line item

saruniitr:
## Change description

We have a requirement to support multiple ARS values for each line item. The current product assessment endpoint only supports a single value (prefix and subject). In preparation this PR updates the assessment endpoint to populate `report_summaries` field using the `prefix` and `subject` values provided.

We look for the input data for each line item and based on whether it is a legacy item, controlled or not the report summaries are populated. If we receive `report_summaries` then the same are used.

As there are predefined set of prefixes and subjects we can create predefined set of `ReportSummary` instances and they can be used to link in product assessments. This is done using a new migration.

In this PR a new serializer field `ReportSummaryField` is added which returns `ReportSummary` instance for the given prefix and subject. At the moment there will only be one but when UI is updated then there can be multiple.

[LTD-4990]: https://uktrade.atlassian.net/browse/LTD-4990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/staticdata/report_summaries/migrations/0007_populate_report_summaries.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@transaction.atomic
+def populate_report_summaries(apps, schema_editor):
+    ReportSummaryPrefix = apps.get_model("report_summaries", "ReportSummaryPrefix")
+    ReportSummarySubject = apps.get_model("report_summaries", "ReportSummarySubject")
+    ReportSummary = apps.get_model("report_summaries", "ReportSummary")
+
+    prefix_ids = [""] + list(map(str, ReportSummaryPrefix.objects.values_list("id", flat=True)))
+    subject_ids = list(map(str, ReportSummarySubject.objects.values_list("id", flat=True)))
```
kevincarrogan, line 13: I think these should be list comprehensions instead.

saruniitr: performance reasons?
using `map` evaluates queryset first before mapping?

depsiatwal: why not spend log some times and check the bottle necks ? may not much performance to be improved

api/staticdata/report_summaries/migrations/0007_populate_report_summaries.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 4.2.11 on 2024-05-07 21:13
+
+from django.db import migrations, transaction
+
+
+@transaction.atomic
+def populate_report_summaries(apps, schema_editor):
```
depsiatwal, line 7: when adding migrations I normally like to add some tests

saruniitr: yes, I have created a test but there is one issue during teardown.
I would like to get feedback while I address that.

saruniitr: that is now resolved, test added for the migration

api/assessments/serializers.py
```diff
@@ -20,6 +21,14 @@
 logger = logging.getLogger(__name__)
 
 
+def is_legacy_line_item(data):
```
depsiatwal, line 24: wondering if we could just use conditional clause

saruniitr: You mean use the condition directly? It is required in couple of places and there is a quite a lot of repetition of fields `report_summary, report_summary_prefix, report_summary_subject` which makes it hard to follow.

saruniitr: besides it makes it clear what the condition is about.

depsiatwal (approved):
lgtm

api/assessments/views.py
```diff
@@ -46,15 +41,39 @@ def get_application_line_numbers(self, instances):
 
         return line_numbers
 
+    def populate_report_summaries(self, assessment_data):
+        payload = []
```
currycoder, line 45: Are we going to be able to come back and drop this once we have the frontend updated with the `report_summaries` contract?

api/assessments/serializers.py
```diff
@@ -43,6 +52,36 @@ def validate(self, data):
...
+        prefix = data.get("prefix", None)
+        subject = data.get("subject", None)
+
+        if not subject:
+            raise serializers.ValidationError("You must include a report summary if this item is controlled.")
+
+        try:
+            if prefix:
+                return ReportSummary.objects.get(prefix=prefix, subject=subject)
+            else:
```
currycoder, line 76: I don't think you need the if/else here as you already gave prefix a sensible default above

api/assessments/serializers.py
```diff
@@ -43,6 +52,36 @@ def validate(self, data):
...
+        super().__init__(
+            queryset=ReportSummary.objects.all(),
+            many=kwargs.get("many"),
+            serializer=ReportSummarySerializer,
+            error_messages={},
+            **kwargs,
+        )
+
+    def to_internal_value(self, data):
+        prefix = data.get("prefix", None)
```
currycoder, line 67: Given that the serializer in use is `ReportSummarySerializer`...  Wouldn't the format for data here be;
```
{
    "prefix": {"id": "", "name": ""},
    "subject": {"id": "", "name": ""},
}
```

I can't see that working with the below ReportSummary lookup..

api/assessments/views.py
saruniitr: yes, once FE is updated in future this will be removed as part of cleanup

api/assessments/serializers.py
saruniitr: this is deriving from `PrimaryKeyRelatedSerializerField` so just getting the ids which is enough here.

api/assessments/serializers.py
saruniitr: actually this is for the case where `prefix=""`. I added a test for this particular case.
In the case where prefix is optional and it always comes as `None` then this is not required.

api/assessments/views.py
currycoder: Could a `# TODO: ..` comment be put in to explain this?  We sometimes aren't great at cleanup and it would be good to have a comment that tells us to remove it in future if we haven't already

api/assessments/serializers.py
currycoder: I think the thing that's tripping me up is that a few lines up we have `serializer=ReportSummarySerializer` and this field just doesn't match the contract defined by that serializer..

saruniitr: looking at this again serializer was expected because I was using `PrimaryKeyRelatedSerializerField`.
https://github.com/uktrade/lite-api/blob/dev/api/core/serializers.py#L28

I have changed it to use `PrimaryKeyRelatedField` which doesn't need a serializer as we are only validating an update request.

api/assessments/views.py
saruniitr: added comment

currycoder (approved):

PR #1983, opened 2024-05-10
Retry creating an exporter user with a different fake email on integr…

kevincarrogan:
Attempt to handle the case where a duplicate random email is re-used and tests fail as a result.


depsiatwal (approved):
lgtm

markj0hnst0n (approved):

test_helpers/clients.py
```diff
@@ -207,10 +208,21 @@ def create_exporter_user(self, organisation=None, first_name=None, last_name=Non
...
-            first_name=first_name, last_name=last_name, email=self.faker.email(), pending=False, type=UserType.EXPORTER
-        )
-        base_user.save()
+        max_integrity_retry_attempts = 5
+        for _ in range(max_integrity_retry_attempts):
+            try:
+                base_user = BaseUser(
+                    first_name=first_name,
+                    last_name=last_name,
+                    email=self.faker.email(),
```
saruniitr, line 217: Does it work if we just append random digits to the email here in the factory?
https://github.com/uktrade/lite-api/blob/dev/api/users/tests/factories.py#L18

If it is quicker, can this be moved to the factory itself?

kevincarrogan: https://faker.readthedocs.io/en/master/#unique-values turns out this exists. I'm going to give that a try.

PR #1986, opened 2024-05-10
LTD-5003: Update part number search to look for exact matches

saruniitr:
## Change description

Currently part number search in products uses a delimiter '-' when analyzing the search string. Because of this if a part number contains hyphens then it splits into multiple tokens and matches on individual tokens as well. This can bring multiple results and not possible to get an exact match.

Users are expecting for an exact match so consider the whole search string without using any delimiters.

[LTD-5003](https://uktrade.atlassian.net/browse/LTD-5003)

[LTD-5003]: https://uktrade.atlassian.net/browse/LTD-5003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/search/product/documents.py
+++ b/api/search/product/documents.py
@@ -19,7 +19,7 @@ address_analyzer = analysis.analyzer(
 
 part_number_analyzer = analysis.analyzer(
     "part_number_analyzer",
-    tokenizer=analysis.tokenizer("part_number_path_hierarchy", "path_hierarchy", delimiter="-"),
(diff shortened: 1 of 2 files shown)


depsiatwal (approved):
lgtm

kevincarrogan (approved):

PR #1990, opened 2024-05-14
LTD-4991: Back populate multiple ARS data in GoodOnApplication instances

saruniitr:
## Change description

Using the populated `report_summary_prefix_id` and `report_summary_subject_id` we back populate report_summaries
field in all `GoodOnApplication` instances.

The migration looks at all the controlled Goods on application and populate their `report_summaries`. As there can be large number of entries, instead of setting on each good on application instance we create entries in bulk on the through table
for GoodOnApplication and ReportSummary

[LTD-4991](https://uktrade.atlassian.net/browse/LTD-4991)

[LTD-4991]: https://uktrade.atlassian.net/browse/LTD-4991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/staticdata/report_summaries/migrations/0008_back_populate_multiple_ars_data.py
```diff
@@ -0,0 +1,57 @@
...
+    errors = []
+    valid_report_summaries = []
+    for goodonapplication_id, prefix_id, subject_id in queryset_values:
+        try:
+            rs = ReportSummary.objects.get(prefix_id=prefix_id, subject_id=subject_id)
+            valid_report_summaries.append((goodonapplication_id, rs.id))
+        except ReportSummary.ObjectDoesNotExist:
+            # this is unlikely as prefix, subject on GoodOnApplication are linked from a predefined
+            # set of objects and the same are used to create ReportSummary objects.
+            errors.append({"prefix": prefix_id, "subject": subject_id})
```
currycoder, line 35: This is going to be a lot of single queries..  Could it be better to grab all of the ReportSummary objects and cache like so...?

```
all_report_summaries = ReportSummary.objects.all()
all_report_summary_cache = {f"{rs.prefix_id}-{rs.subject_id}": rs for rs in all_report_summaries}
```

This cache could then be interrogated in memory to build the mappings as you are doing here.

kevincarrogan (approved):

api/staticdata/report_summaries/migrations/0008_back_populate_multiple_ars_data.py
```diff
@@ -0,0 +1,57 @@
...
+    if errors:
+        logger.info("ReportSummary objects does not exist for,\n%s", errors)
+
+    # Create the entries in through table
+    objects_to_create = [
+        ReportSummaryThrough(goodonapplication_id=goodonapplication_id, reportsummary_id=report_summary_id)
+        for (goodonapplication_id, report_summary_id) in valid_report_summaries
+    ]
+
+    ReportSummaryThrough.objects.bulk_create(objects_to_create)
```
currycoder, line 46: It would be good to test this migration on an environment somewhere with a recent DB refresh.  We have observed recently that migrations taking longer than 1 minute need to be reworked to work with the deployment tooling.

saruniitr: yeah sure will try this on an env.
Locally I checked the timings and it is taking less than a sec.

api/staticdata/report_summaries/migrations/0008_back_populate_multiple_ars_data.py
currycoder: Or even better using `values_list()`...

currycoder (approved):

PR #1996, opened 2024-05-20
LTD-5007-allow-searching-of-item-list-code

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5007

This is the first iteration of this where we can see the behaviour in PROD the next iteration will correct the field name and eventually more powerful search


api/external_data/serializers.py
```diff
@@ -268,6 +269,11 @@ def get_address(self, obj):
             return obj.meta.highlight.to_dict().get("address")[0]
         return obj.address
 
+    def get_item_list_codes(self, obj):
+        if hasattr(obj.meta, "highlight") and obj.meta.highlight.to_dict().get("item_list_codes"):
+            return obj.meta.highlight.to_dict().get("item_list_codes")[0]
```
saruniitr, line 274: this looks repetitive now for each field. Is this because you also have highlighting for this field?

depsiatwal: yeah it's the field highlighting 
good point about the repetition  i've broken it up a little now and also added tests

depsiatwal (commented):
ignore

api/external_data/views.py
```diff
@@ -70,6 +70,13 @@ class DenialSearchView(DocumentViewSet):
                 "post_tags": ["</mark>"],
             },
         },
+        "item_list_codes": {
+            "enabled": True,
+            "options": {
+                "pre_tags": ["<mark>"],
+                "post_tags": ["</mark>"],
+            },
+        },
```
kevincarrogan, line 79: I think there's a shorter syntax for this looking at the docs:

```
"highlight": {
    "pre_tags": [
      "<mark>"
    ],
    "post_tags": [
      "</mark>"
    ],
    "fields": {
      "item_list_codes": {},
      ...
    }
  }
```

https://opensearch.org/docs/latest/search-plugins/searching-data/highlight/#changing-the-highlighting-tags

kevincarrogan: Although in saying that I'm not sure how that translates through the `DocumentViewSet`.

depsiatwal: yeah it doesn't we have to use highlight_fields and i'm not able to find many short cuts except 
highlight_fields = {"*": {"enabled": True, "options": {"pre_tags": ["<b>"], "post_tags": ["</b>"]}}}
Which would attempt to highlight every field which actually may not be a band thing as we only search on 3.

kevincarrogan (approved):

markj0hnst0n (approved):

PR #2014, opened 2024-05-30
LTD-4940-remove-Denial-Entity-Redundant-fields 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4876
https://uktrade.atlassian.net/browse/LTD-4940

This PR is attempting to 

- Remove all fields that were migrated from DenialEntity to Denial 
- Rename the factories to make it more obvious which entities are being created
- Rename DenialApplicationMatch to clearly show we mark a Denial record against a DenialEntity and not Denial 
- Add constraints so DenialEntity must be linked to a denial now that the data has been migrated 

These could have technically be done in smaller PRs but we would have still needed to update all the tests. 
Also The whole of Denials will require regression and if we do this in many stages then we would have duplicate effort around regression


api/external_data/management/commands/ingest_denials.py
```diff
@@ -88,25 +91,35 @@ def load_denials(self, filename):
             denial_entity_data["data"] = row
             denial_entity_data["entity_type"] = get_denial_entity_enum(denial_entity_data["data"])
             denial_serializer = DenialSerializer(data=denial_data)
-            denial_entity_serializer = DenialEntitySerializer(data=denial_entity_data)
 
-            is_valid_denial = denial_serializer.is_valid()
-            is_valid_entity_denial = denial_entity_serializer.is_valid()
-
-            if is_valid_denial and is_valid_entity_denial:
+            if denial_serializer.is_valid():
```
depsiatwal, line 95: since Denial is required for DenialEntity the logic has been changed to reflect this change

markj0hnst0n (approved):

kevincarrogan (approved):

PR #2016, opened 2024-05-31
Fix env variable usage in settings

hnryjmes:
### Aim

This is something small I noticed while doing the spike for LTD-4996. Currently the if-statement doesn't use the bool created in the line above, but it uses a string literal, which will always be true, so I guess this was just a typo.

[LTD-4996](https://uktrade.atlassian.net/browse/LTD-4996)

[LTD-4996]: https://uktrade.atlassian.net/browse/LTD-4996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/conf/settings.py
+++ b/api/conf/settings.py
@@ -126,7 +126,7 @@ INSTALLED_APPS = [
 
 MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS = env("MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS")
 
-if "MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS":
+if MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS:
     INSTALLED_APPS += [
         "mock_virus_scan",
     ]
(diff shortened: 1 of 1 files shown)


api/conf/settings.py
```diff
@@ -473,7 +473,7 @@ def _build_redis_url(base_url, db_number, **query_args):
     {
         "un_sanctions_file": "https://scsanctions.un.org/resources/xml/en/consolidated.xml",
         "office_financial_sanctions_file": "https://ofsistorage.blob.core.windows.net/publishlive/2022format/ConList.xml",
-        "uk_sanctions_file": "https://assets.publishing.service.gov.uk/media/65ca02639c5b7f0012951caf/UK_Sanctions_List.xml",
+        "uk_sanctions_file": "https://assets.publishing.service.gov.uk/media/65ca02639c5b7f0012951caf/UK_Sanctions_List.xml",  # /PS-IGNORE
```
hnryjmes, line 476: this was to make the pii checker error go away, not related to the other change

api/conf/settings.py
```diff
@@ -126,7 +126,7 @@
 
 MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS = env("MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS")
 
-if "MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS":
+if MOCK_VIRUS_SCAN_ACTIVATE_ENDPOINTS:
```
kevincarrogan, line 129: Ok. This means that this was always defaulting to `True`, it would be good to check that this environment variable is set how we expect it to be so that we're not inadvertently setting this to `False` somewhere we shouldn't be.

hnryjmes: I just double checked in Vault, this isn't currently set anywhere on any of the environments

markj0hnst0n (approved):

PR #2018, opened 2024-05-31
Improve file processing error messages

hnryjmes:
### Aim

Updates the error messages shown when a document fails to be processed.

[LTD-4996](https://uktrade.atlassian.net/browse/LTD-4996)

[LTD-4996]: https://uktrade.atlassian.net/browse/LTD-4996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/constants.py
+++ b/api/applications/constants.py
@@ -1,25 +1,6 @@
 TRANSHIPMENT_AND_TRADE_CONTROL_BANNED_COUNTRIES = ["GB"]
 
 
-class Standard:
-    CONSIGNEE_DOCUMENT_PROCESSING = (
-        "We are still processing a consignee document. Try submitting again in a few minutes."
-    )
-    END_USER_DOCUMENT_PROCESSING = (
-        "We are still processing an end-user document. Try submitting again in a few minutes."
-    )
-    THIRD_PARTY_DOCUMENT_PROCESSING = (
-        "We are still processing a third party document. Try submitting again in a few minutes."
-    )
-    ULTIMATE_END_USER_DOCUMENT_PROCESSING = (
-        "We are still processing an ultimate recipient document. Try submitting again in a few minutes."
-    )
-    ADDITIONAL_DOCUMENTS_PROCESSING = (
-        "We are still processing an additional document. Try submitting again in a few minutes."
-    )
-    GOODS_DOCUMENT_PROCESSING = "We are still processing a good document. Try submitting again in a few minutes."
-
-
 class F680:
     ADDITIONAL_INFORMATION_FIELDS = (
         "expedited",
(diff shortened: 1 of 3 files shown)


api/applications/creators.py
```diff
@@ -66,7 +67,7 @@ def check_party_document(party, is_mandatory):
             return None
 
     if None in documents_qs:
-        return getattr(strings.Applications.Standard, f"{party.type.upper()}_DOCUMENT_PROCESSING")
+        return getattr(constants.Standard, f"{party.type.upper()}_DOCUMENT_PROCESSING")
```
hnryjmes, line 70: this is how the constants `END_USER_DOCUMENT_PROCESSING` etc get accessed

api/applications/constants.py
```diff
@@ -1,6 +1,25 @@
 TRANSHIPMENT_AND_TRADE_CONTROL_BANNED_COUNTRIES = ["GB"]
 
 
+class Standard:
```
kevincarrogan, line 4: All this seems to be doing is replacing the type of document in the search string each time. Could we maybe think about doing that in a different way e.g. we get the type of document from the object itself and just replace the test `f"We are still processing a {document.get_document_type()}."`

hnryjmes: this has now been updated

currycoder (approved):

api/applications/creators.py
```diff
@@ -312,3 +314,17 @@ def validate_application_ready_for_submission(application):
...
+    return f"We are still processing {document_type_description} document. Try submitting again in a few minutes."
+
+
+def get_document_type_description_from_party_type(party_type):
+    document_type_description = {
+        PartyType.CONSIGNEE: "a consignee",
+        PartyType.END_USER: "an end-user",
+        PartyType.THIRD_PARTY: "a third party",
+        PartyType.ULTIMATE_END_USER: "an ultimate end-user",
+    }
```
saruniitr, line 329: I think having this defined with complete error message gives more control,
eg
```
    document_type_description = {
        PartyType.CONSIGNEE: We are still processing a consignee document. Try submitting again in a few minutes.,
        PartyType.END_USER: "an end-user",
        PartyType.THIRD_PARTY: "a third party",
        PartyType.ULTIMATE_END_USER: "an ultimate end-user",
    }
```

saruniitr (approved):
one comment to consider

api/applications/creators.py
hnryjmes: I originally had all the error messages written out in full, then changed to use interpolation based on Kevin's comment above. I think both approaches work, it's just a matter of picking one of them, otherwise I'll go around in circles making changes based on suggestions.

saruniitr: ok np which is why it is already approved.

PR #2020, opened 2024-05-31
[LTD-5055] Amend application by copy

currycoder:
### Context

This change adds a new endpoint for creating an application amendment.

An application amendment is a copy of an original application (minus all caseworker commentary/assessments/advice/licences etc).  This is a new approach for "major" amendments in lite as major amendments were previously achieved by allowing the exporter to make changes to the existing application and then ensuring that the application was dragged back to the start of the application flow.  The previous approach for dragging cases back in lite results in a number of repetitive support issues as detailed here; https://docs.google.com/document/d/1tbbjlAb20TakYjKm0NC2lWNQBsCP_6T_DDjV69TSUHI/edit#heading=h.9hfmj8gn0e14

A companion PR exists for the frontend changes; https://github.com/uktrade/lite-frontend/pull/2013

### Flow

A summary of how the new endpoint works;
- The endpoint creates a cloned application using `StandardApplication.clone()`.  This includes all of the exporter-input details, but none of the caseworker commentary/related objects.
- The new amendment application is in a draft state, enabling the exporter to edit any information that they want to on it.
- All queue assignments are removed from the original application and it is placed in to a new "Superseded by amendment" terminal status.

### Technical notes

Technical things to call out:
- Cloning is achieved in a relatively DRY way by introducing a new `Clonable` model mixin.  This allows cloning to be configured per-model by way of `clone_exclusions` (things to exclude on the copied application), `clone_mappings` (telling the copying code how to map field names for relations), `clone_overrides` (enabling setting particular static values on cloned objects) fields.
    - This allows for a mostly declarative interface for setting up cloning.
    - Cloning of related objects is a little more complex and requires overriding the `clone()` method; e.g. to ensure that `FirearmDetails` objects are copied when a `GoodOnApplication` is cloned.
    - The overall pattern of moving more logic like this in to the model layer is something that will hopefully help us towards a cleaner codebase/less messy views over time.
- When an amendment application is created, the original application refers to it with the new `amendment_of` ForeignKey field.  It was decided to use this as it is more specific than the existing `copy_of` field.

### Follow-up work

This is very much a first step for the "amendment by copy" flow.  There are a few edge cases/enhancements to follow:
- Deciding upon audit trail additions; it would probably be wise to create an audit entry when an amendment is created.
- Validating that the application in question can be amended as part of the amendment endpoint.
- Deciding how to handle amendments for applications which have had a licence issued already - the flow currently would set the case in to the "Superseded by amendment" status without cancelling the existing licence.
- Right now an amendment is created immediately when the user chooses the "major edit" choice on the frontend.  At the same time, their existing application is moved to "superseded" status.  It may be desirable to only supersede the existing application after the user has submitted the amendment.

[LTD-5055](https://uktrade.atlassian.net/browse/LTD-5055)

[LTD-5055]: https://uktrade.atlassian.net/browse/LTD-5055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/applications/models.py
```diff
@@ -302,8 +306,72 @@ class StandardApplication(BaseApplication):
...
+    clone_overrides = {
+        "status_id": CaseStatusIdEnum.DRAFT,
+    }
+
+    def clone(self, exclusions=None, **overrides):
+        cloned_application = super().clone(exclusions=exclusions, **overrides)
+
+        application_documents = ApplicationDocument.objects.filter(application=self)
+        for application_document in application_documents:
+            application_document.clone(application=cloned_application)
```
saruniitr, line 344: I think this needs to be checked with UCD/Business.
One of the document type is RFD document. Depending upon when a clone is created significant amount of time might have passed and the documents may no longer be valid.
So is it best to ask them re-upload latest documents? Previous ones can be referred anyway in the old application.

api/applications/models.py
```diff
@@ -302,8 +306,72 @@ class StandardApplication(BaseApplication):
...
+
+        site_on_applications = SiteOnApplication.objects.filter(application=self)
+        for site_on_application in site_on_applications:
+            site_on_application.clone(application=cloned_application)
+
+        external_location_on_applications = ExternalLocationOnApplication.objects.filter(application=self)
+        for external_location_on_application in external_location_on_applications:
+            external_location_on_application.clone(application=cloned_application)
```
saruniitr, line 352: This is not related to SIEL. There are `0` instances in UAT.

api/applications/views/tests/test_amendments.py
```diff
@@ -0,0 +1,42 @@
...
+from rest_framework import status
+
+from api.applications.models import StandardApplication
+from api.applications.tests.factories import StandardApplicationFactory
+from api.organisations.tests.factories import OrganisationFactory
+
+from test_helpers.clients import DataTestClient
+
+
+class TestCreateApplicationAmendment(DataTestClient):
```
saruniitr, line 11: Is it worth having another test where creating amendment fails if application is not in editable state?

api/cases/models.py
```diff
@@ -156,6 +159,12 @@ def _reset_sub_status_on_status_change(self):
                 payload={"sub_status": None, "status": CaseStatusEnum.get_text(self.status.status)},
             )
 
+    @property
+    def superseded_by(self):
+        if not self.amendment.exists():
```
saruniitr, line 164: I wish we have `.empty()` or similar methods for these type of checks to avoid `not`.

saruniitr: I guess `.first()` returns `None` if there are no objects?

api/core/model_mixins.py
```diff
@@ -0,0 +1,33 @@
+from django.forms import model_to_dict
+
+
+class Clonable:
```
saruniitr, line 4: this is a nice name

saruniitr (commented):
Few questions and comments.

Noticed the deletion of migration tests which saves some time .

api/applications/models.py
currycoder: That's a great question Arun.  OK for now I will drop this and add a follow-up ticket where we can try to figure this out..

api/applications/views/tests/test_amendments.py
currycoder: Right now, there is no validation for this.  It's a follow-up ticket here; https://uktrade.atlassian.net/browse/LTD-5078

api/applications/models.py
currycoder: https://uktrade.atlassian.net/browse/LTD-5105

api/cases/models.py
currycoder: I could change it to `return self.amendment.first()` but personally I kind of find this a bit more readable as it is.

api/applications/models.py
saruniitr: sounds good

api/cases/models.py
saruniitr: Agree it hurts readability if we just use `.first()`.

saruniitr (approved):

api/applications/models.py
```diff
@@ -302,8 +306,70 @@ class StandardApplication(BaseApplication):
...
+        party_on_applications = PartyOnApplication.objects.filter(application=self)
+        for party_on_application in party_on_applications:
+            party_on_application.clone(application=cloned_application)
+
+        return cloned_application
+
+    @transaction.atomic
+    def create_amendment(self):
+        amendment_application = self.clone(amendment_of=self)
+        # TODO: Do we need a log on the audit trail?
```
hnryjmes, line 364: I forgot to ask in the dev meeting, is there a plan for adding audit functionality to the amend application work later on? what are your current thoughts about this?

api/cases/models.py
```diff
@@ -113,6 +113,9 @@ class Case(TimestampableModel):
     )
     case_officer = models.ForeignKey(GovUser, null=True, on_delete=models.DO_NOTHING)
     copy_of = models.ForeignKey("self", default=None, null=True, on_delete=models.DO_NOTHING)
+    amendment_of = models.ForeignKey(
+        "self", default=None, null=True, on_delete=models.DO_NOTHING, related_name="amendment"
+    )
```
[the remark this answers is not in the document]
hnryjmes: I think this suggestion sounds reasonable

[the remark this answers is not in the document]
hnryjmes: same as above, I think this suggestion sounds reasonable

api/applications/models.py
currycoder: There's not really a plan as yet, other than "we need to figure out the audit trail" which should be this follow-up: https://uktrade.atlassian.net/browse/LTD-5077

PR #2026, opened 2024-06-05
Update website field validation

hnryjmes:
### Aim

In lite-frontend a recent change was made to allow exporters to use URLs for their organisation registration without a scheme prefix such as "www.example.com" or "example.com", where previously these would have been rejected for not being valid URLs, as the Django URL validator expects a scheme prefix such as "http://".

The API also does some validation using the same Django URL validator as the Django `URLField` is used. The simplest way to allow partial URLs like "example.com" to be stored in the db without causing problems is to change the field type to a `TextField` which is what has been done here.

(An alternative might be to add "https://" or "http://" to the user input at some point in either lite-frontend or lite-api, but that risks the wrong scheme being chosen)

[LTD-2285](https://uktrade.atlassian.net/browse/LTD-2285)

[LTD-2285]: https://uktrade.atlassian.net/browse/LTD-2285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hnryjmes:
I think it's better to make a change on the frontend instead to force add the `https://` prefix, so will close this

PR #2029, opened 2024-06-06
LTD-5067-increase-page-size-denial-search 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5067

### DIFF
--- a/api/conf/pagination.py
+++ b/api/conf/pagination.py
@@ -19,3 +19,9 @@ class MaxPageNumberPagination(pagination.PageNumberPagination):
                 "results": data,
             }
         )
+
+
+class MaxFiftyPageSizePaginator(MaxPageNumberPagination):
+    page_size = 50
+    page_size_query_param = "page_size"
+    max_page_size = 50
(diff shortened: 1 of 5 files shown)


api/external_data/views.py
```diff
@@ -42,6 +42,7 @@ class DenialSearchView(QueryStringValidationMixin, DocumentViewSet):
     serializer_class = serializers.DenialSearchSerializer
     authentication_classes = (GovAuthentication,)
     pagination_class = MaxPageNumberPagination
+    pagination_class.page_size = 50
```
depsiatwal, line 45: I did think of placing this in settings under a dict 
but not sure how often we would need to change.

api/conf/pagination.py
```diff
@@ -19,3 +19,9 @@ def get_paginated_response(self, data):
                 "results": data,
             }
         )
+
+
+class MaxFiftyPageSizePaginator(MaxPageNumberPagination):
```
depsiatwal, line 24: As I seem to be just cutting and pasting this pagination it make more sense to shift it and give it a generic name

saruniitr (approved):

markj0hnst0n (approved):

PR #2032, opened 2024-06-10
LTD-5066-5065-content

seijihg:
https://uktrade.atlassian.net/browse/LTD-5066
https://uktrade.atlassian.net/browse/LTD-5065

### DIFF
--- a/api/applications/managers.py
+++ b/api/applications/managers.py
@@ -5,14 +5,14 @@ from api.staticdata.statuses.libraries.get_case_status import get_case_status_by
 
 
 class BaseApplicationManager(InheritanceManager):
-    def drafts(self, organisation):
+    def drafts(self, organisation, sort_by):
         draft = get_case_status_by_status(CaseStatusEnum.DRAFT)
-        return self.get_queryset().filter(status=draft, organisation=organisation).order_by("-created_at")
+        return self.get_queryset().filter(status=draft, organisation=organisation).order_by(sort_by)
 
-    def submitted(self, organisation, sort):
+    def submitted(self, organisation, sort_by):
         draft = get_case_status_by_status(CaseStatusEnum.DRAFT)
-        return self.get_queryset().filter(organisation=organisation).exclude(status=draft).order_by(sort)
+        return self.get_queryset().filter(organisation=organisation).exclude(status=draft).order_by(sort_by)
 
-    def finalised(self, organisation, sort):
+    def finalised(self, organisation, sort_by):
         finalised = get_case_status_by_status(CaseStatusEnum.FINALISED)
-        return self.get_queryset().filter(status=finalised, organisation=organisation).order_by(sort)
+        return self.get_queryset().filter(status=finalised, organisation=organisation).order_by(sort_by)
--- a/api/applications/managers.py
+++ b/api/applications/managers.py
@@ -9,10 +9,10 @@ class BaseApplicationManager(InheritanceManager):
         draft = get_case_status_by_status(CaseStatusEnum.DRAFT)
         return self.get_queryset().filter(status=draft, organisation=organisation).order_by("-created_at")
 
-    def submitted(self, organisation):
+    def submitted(self, organisation, sort):
         draft = get_case_status_by_status(CaseStatusEnum.DRAFT)
-        return self.get_queryset().filter(organisation=organisation).exclude(status=draft).order_by("-submitted_at")
+        return self.get_queryset().filter(organisation=organisation).exclude(status=draft).order_by(sort)
 
-    def finalised(self, organisation):
+    def finalised(self, organisation, sort):
         finalised = get_case_status_by_status(CaseStatusEnum.FINALISED)
-        return self.get_queryset().filter(status=finalised, organisation=organisation).order_by("-updated_at")
+        return self.get_queryset().filter(status=finalised, organisation=organisation).order_by(sort)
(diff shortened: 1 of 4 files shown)


markj0hnst0n (approved):
The ticket mentions that this change was dependent on LTD-5065 but happy to approve given that 5065's completion is not a blocker for this one.

hnryjmes (approved):

seijihg:
> Details

I have merged them in into the same PR :D

api/applications/views/applications.py
```diff
@@ -119,10 +119,15 @@ def get_queryset(self):
...
+        finalised = optional_str_to_bool(self.request.GET.get("finalised"))
+        sort = self.request.GET.get("sort", "-updated_at")
+
         if submitted is None:
             applications = BaseApplication.objects.filter(organisation=organisation)
+        elif submitted and finalised:
+            applications = BaseApplication.objects.finalised(organisation, sort)
         elif submitted:
-            applications = BaseApplication.objects.submitted(organisation)
+            applications = BaseApplication.objects.submitted(organisation, sort)
```
saruniitr, line 130: I think this can follow the same approach that we do with Case search.
For cases also we have separate tab such 'All cases', 'My cases', 'Open queries' and based on the selected tab we update filters.
eg https://github.com/uktrade/lite-api/blob/dev/api/cases/views/search/views.py#L34

I think here `submitted is None` check is mainly to filter out draft applications.
`submitted and finalised`: If it is finalised then it means it is already submitted so we should not be checking for submitted again.

it can be like this (please change names),

```
if finalised_tab:
    # finalised applications
elif submitted_tab:
    # submitted applications
else:
    # drafts
```

api/applications/views/applications.py
```diff
@@ -113,16 +113,20 @@ def get_queryset(self):
         Filter applications on submitted
         """
         try:
-            submitted = optional_str_to_bool(self.request.GET.get("submitted"))
+            submitted_tab = optional_str_to_bool(self.request.GET.get("submitted"))
         except ValueError:
             return BaseApplication.objects.none()
```
saruniitr, line 118: As per designs there are three tabs, 'submitted, 'finalised', 'drafts'
so bool is not enough for this.
At any time only one tab is active so you just get the selected tab from the user and return those applications.
You can get rid of this try catch and retrieve the selected tab.

I have linked this earlier how this pattern is used to display Cases for caseworkers,
https://github.com/uktrade/lite-api/blob/dev/api/cases/views/search/views.py#L135

once you get the selected tab then you can filter and sort as per the selected option.

api/applications/views/applications.py
```diff
@@ -113,16 +113,20 @@ def get_queryset(self):
...
-        if submitted is None:
+        if submitted_tab is None:
             applications = BaseApplication.objects.filter(organisation=organisation)
-        elif submitted:
-            applications = BaseApplication.objects.submitted(organisation)
+        elif finalised_tab:
+            applications = BaseApplication.objects.finalised(organisation, sort)
+        elif submitted_tab:
+            applications = BaseApplication.objects.submitted(organisation, sort)
         else:
```
saruniitr, line 130: you will have three branches.
consider renaming `sort` to `sort_by` or `order_by`.

api/applications/views/applications.py
seijihg: this change is bigger than I would like to investigate tbh, a lot of tests will fail since it requires changes on the FE as well since we are removing` booleans submitted, finalised `and replacing it with `selected_tab`

api/applications/views/applications.py
```diff
@@ -112,17 +111,15 @@ def get_queryset(self):
...
-            applications = BaseApplication.objects.filter(organisation=organisation)
-        elif submitted:
-            applications = BaseApplication.objects.submitted(organisation)
+        selected_tab = self.request.GET.get("selected_tab", "submitted_tab")
+        sort_by = self.request.GET.get("sort", "-submitted_at")
+
+        if selected_tab == "submitted_tab":
+            applications = BaseApplication.objects.submitted(organisation, sort_by)
+        elif selected_tab == "finalised_tab":
+            applications = BaseApplication.objects.finalised(organisation, sort_by)
```
kevincarrogan, line 122: Could we come up with a better name as opposed to using the term "_tab" here? The API shouldn't really be aware of what a tab is really so if we could come up with a more generic naming for this I think it would be better.

api/applications/managers.py
```diff
@@ -9,6 +9,10 @@ def drafts(self, organisation):
...
 
-    def submitted(self, organisation):
+    def submitted(self, organisation, sort_by):
         draft = get_case_status_by_status(CaseStatusEnum.DRAFT)
-        return self.get_queryset().filter(organisation=organisation).exclude(status=draft).order_by("-submitted_at")
+        return self.get_queryset().filter(organisation=organisation).exclude(status=draft).order_by(sort_by)
+
+    def finalised(self, organisation, sort_by):
+        finalised = get_case_status_by_status(CaseStatusEnum.FINALISED)
+        return self.get_queryset().filter(status=finalised, organisation=organisation).order_by(sort_by)
```
kevincarrogan, line 18: Are we sure that these methods aren't being used in the code other than the one place you've updated it?

Also, it feels to me that we should make these parameters here consistent across all of these methods, so `drafts` should probably have a `sort_by` field too.

api/applications/views/applications.py
seijihg: updated to a better name

hnryjmes (approved):

PR #2035, opened 2024-06-11
LTD-5049: Good archive restore

saruniitr:
## Change description

To help exporters better manage their product list we are going to provide an option to archive a product that is no longer actively used. This hides it from their active products. Once archived they also have an option to restore it.
User can archive/restore multiple times and for each action we want to save the action, timestamp and user details.

To track these changes this PR introduces `django-reversion` which maintains version history and allows to track updates for a model.

The archive history is exported to user which will be displayed on the product list page.

[LTD-5049](https://uktrade.atlassian.net/browse/LTD-5049)

[LTD-5049]: https://uktrade.atlassian.net/browse/LTD-5049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/conf/settings.py
```diff
@@ -139,6 +140,7 @@
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
+    "reversion.middleware.RevisionMiddleware",
```
hnryjmes, line 143: did we resolve the conversation about adding `ATOMIC_REQUESTS` to settings? seeing this middleware reminded me of the transaction middleware which you said was replaced by atomic requests. is it needed for revision to work, I guess not?

api/goods/tests/test_views.py
```diff
@@ -119,3 +124,44 @@ def test_view_good__query_filter_by_control_list_entry(self, control_list_entry,
...
+
+        # Create sample version history
+        for is_archived in good_archive_status:
+            response = self.client.put(edit_url, {"is_archived": is_archived}, **self.exporter_headers)
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        # ensure correct number of revisions are created
+        self.assertEqual(Version.objects.get_for_object(good).count(), len(good_archive_status))
+
+        # determine expected statuses which should not include duplicates
```
hnryjmes, line 149: could you explain what you mean here by "which should not include duplicates"?

saruniitr: Yes, it is mainly to filter versions where the value of `is_archived` has not changed.
Lets say we archived a product which creates a version with `is_archived=True`.
Now say in the next 3 versions other fields might have changed but the archived field remain the same.
Finally user restores the product so we have one more revision with is_archived=False.

If we simply return all the version history we get version where the value remained the same but we only want to show where the value really changed (either it is archived or restored) hence this filters those versions where it is constant.

Reversion itself doesn't have any methods to filter these so handling this for this field in the serializer and verifying in the test.

hnryjmes: ok, makes sense to me

hnryjmes (approved):

api/goods/serializers.py
```diff
@@ -914,6 +936,24 @@ def get_case_officer(self, instance):
...
+            for v in Version.objects.get_for_object(instance).order_by("revision__date_created")
+            if v.field_dict["is_archived"] is not None
+        ]
+
+        version_history = []
+        for current, next in zip_longest(versions, versions[1:], fillvalue=None):
+            current_status = current.field_dict["is_archived"]
+            next_status = next.field_dict["is_archived"] if next else None
+            if current_status != next_status:
+                version_history.append(current)
```
currycoder, line 953: Could this be something to push up as a method on the `Good` model?

e.g. 
```
def get_history(self, field):
    ....

archive_history = good_object.get_history("is_archived")
```

currycoder (commented):
Looks good.  I've added one suggestion for an enhancement

api/goods/serializers.py
saruniitr: Yeah considered adding as a generic method that works for all fields. let me check again as this can be useful for other fields also.

saruniitr: @currycoder it is now moved into the model.
Similar to `Clonable`, I have added a `Trackable` mixin, idea is to add such helper methods in this mixin which can be useful for all models registered with reversion.

PR #2037, opened 2024-06-11
LTD 4956 add test jobs to run against postgres13

markj0hnst0n:
### Aim

Create jobs to run our api tests against a postgres db running version 13 on CircleCI

[LTD-4956](https://uktrade.atlassian.net/browse/LTD-4956)

[LTD-4956]: https://uktrade.atlassian.net/browse/LTD-4956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docker-compose.yml
```diff
@@ -13,6 +13,18 @@ services:
     networks:
       - lite
 
+  pg13_db:
```
depsiatwal, line 16: Don''t think we need this yet
isn't this for our local docker ?

markj0hnst0n: Yes good point,  I was doing some local testing and didn't remove this.  It's gone now

.circleci/config.yml
```diff
@@ -19,6 +19,12 @@ image_postgres: &image_postgres
     POSTGRES_USER: postgres
     POSTGRES_PASSWORD: password
 
+image_pg13: &image_pg13
```
saruniitr, line 22: I think it is better to expand the name rather than abbreviate, `image_postgres13`

markj0hnst0n: amended this @saruniitr

.circleci/config.yml
```diff
@@ -19,6 +19,12 @@ image_postgres: &image_postgres
     POSTGRES_USER: postgres
     POSTGRES_PASSWORD: password
 
+image_postgres13: &image_postgres13
```
currycoder, line 22: This all looks good to me, the only thing I would suggest is maybe linking up with Gurdeep's work over here a bit: https://github.com/uktrade/lite-api/pull/2040/files#r1639502428

Perhaps there is some way to make these new jobs basically `"tests_dbt_platform"` - i.e. we have tests with all our versions as they will continue to be in gov PaaS and tests as they will be in DBT platform

currycoder (approved):

hnryjmes:
I noticed that the DBT platform docs say the following:

> "The current release of PostgreSQL is version 16 and this will receive support until the year 2029. While upgrading to PostgreSQL 16 is preferred, we will accept an upgrade to version 14, which will be supported until the year 2027."

https://platform.readme.trade.gov.uk/migrations/conducting-a-migration/02b-codebase-preparation/#vulnerability-scanning-remediation-patching

Is the idea that we will move to using postgres 14 eventually? Or is 13 where we're stopping?

@markj0hnst0n

markj0hnst0n:
> I noticed that the DBT platform docs say the following:
> 
> > "The current release of PostgreSQL is version 16 and this will receive support until the year 2029. While upgrading to PostgreSQL 16 is preferred, we will accept an upgrade to version 14, which will be supported until the year 2027."
> 
> https://platform.readme.trade.gov.uk/migrations/conducting-a-migration/02b-codebase-preparation/#vulnerability-scanning-remediation-patching
> 
> Is the idea that we will move to using postgres 14 eventually? Or is 13 where we're stopping?
> 
> @markj0hnst0n

Yes we're aware of this @hnryjmes  but I think @kevincarrogan had a conversation and they will accept Postgres13 for the moment.

hnryjmes (approved):

PR #2040, opened 2024-06-12
LTD-4958-switch-to-opensearch 

depsiatwal:
This is a shift to DBT Platform and reduce potential errors. 

This is one of the steps to move our local/tests from elastic search to opensearch to minimise any migration risks. 

Please note the scope of this PR to only migrate the containers and circle-ci. Eventually we should sift the DSL library CONG and mark tests away from elasticsearch.


markj0hnst0n (approved):
Saw all tests have passed on CircleCi so happy to approve

docker-compose.yml
```diff
@@ -81,7 +81,7 @@ services:
     cap_add:
       - IPC_LOCK
     volumes:
-      - elasticsearch-data:/usr/share/elasticsearch/data
+      - opensearch-data:/usr/share/opensearch/data
```
currycoder, line 84: This will presumably drop any data that we have in indexes on local copies.  Are we ok with that?

.circleci/config.yml
```diff
@@ -19,15 +19,15 @@ image_postgres: &image_postgres
     POSTGRES_USER: postgres
     POSTGRES_PASSWORD: password
 
-image_elasticsearch: &image_elasticsearch
-  image: docker.elastic.co/elasticsearch/elasticsearch:7.10.1
+image_opensearch: &image_opensearch
+  image: opensearchproject/opensearch:2.14.0
```
currycoder, line 23: The opensearch version that we have up on UAT (and I presume production) is `1.3.13`.  A couple of issues that surface because of that...

- Our test suite is now not on the same version as prod running on gov PAAS (I believe elasticsearch 7.X is functionally the same as the opensearch version we had before)
- It seems like there are backward incompatible breaking changes; https://opensearch.org/docs/latest/breaking-changes/#200 The notable one for us might be the removal of the `type` parameter - are we sure we don't use it in any of our ES calls?

Also, what version of opensearch is used on DBT platform?

api/search/product/tests/test_views.py
```diff
@@ -200,6 +200,8 @@ def test_product_search_by_assessing_officer(self, query, expected_count):
 
     @pytest.mark.elasticsearch
     def test_product_search_by_applicant(self):
+        # This is required else we appear to have previous test which impact this one
+        call_command("search_index", models=["applications.GoodOnApplication"], action="rebuild", force=True)
```
currycoder, line 204: I'm surprised this came up for this change of search versions/libraries...  What were you seeing before this change?

.circleci/config.yml
currycoder: I just started looking at this PR from @markj0hnst0n https://github.com/uktrade/lite-api/pull/2037/files
That PR has a bunch of new circleci jobs which additionally test postgres13.

I would suggest following a similar pattern here.  e.g.
- Keep our existing jobs and make them point to opensearch 1.3.13 - these jobs are aligned with gov PaaS deployments.
- Add new jobs which point to whatever opensearch version we are targetting in DBT platform.

Maybe it would be handy to align this work with what Mark is working on in the other PR.

docker-compose.yml
depsiatwal: Yeah are required to do that anyway else opensearch trying to migrate and fails so best to start from scratch.

.circleci/config.yml
depsiatwal: Yeah I did think about this but came to the conclusion 
- DBT are pushing for the higher versions so this will give us confidence 
- We have shifted from Elastic search to Opensearch so IMHO much better then the current tests. 
- As we will be eventually moving anyway (couple of months to DBT) probably not much benefit having the complexity of running both. As we have quite a few tests/markers etc so not sure how straightforward it would be for ES/OS.

api/search/product/tests/test_views.py
depsiatwal: It gave some issue with instance during one of the get methods clearning the index fixed it 
wonder if it's worth investigating further ?

.circleci/config.yml
depsiatwal: I think a good balance here may be to leave all local containers on dev etc 
but run just the search tests under both versions.

api/search/product/tests/test_views.py
depsiatwal: actually ignore above 
i've added the newer version to all postgres_13 tests and leaving the others on V1 so when we do shift over I guess the postgres_13 will be the only tests required with same versions matching DBT platform.

hnryjmes (commented):
I've been able to get this branch running locally, after cleaning out my old docker-compose cached images and builds etc, so I'm content that this is working alright. 

If you wanted to completely remove references to elasticsearch in this PR that would make a nice addition. I was thinking some of these bits could be rewritten to reference opensearch instead:

-  `settings.py` https://github.com/uktrade/lite-api/blob/dev/api/conf/settings.py#L332-L342
- `local.env` https://github.com/uktrade/lite-api/blob/dev/local.env#L92
- README file - just remove any references to elasticsearch now as they would be out of date

currycoder (approved):

PR #2044, opened 2024-06-17
Ltd 5029 migrate additional tables to data workspace

dejandbt:
### Aim
Adding additional tables to data-workspace

[LTD-](https://uktrade.atlassian.net/browse/LTD-5029)


api/data_workspace/organisations_views.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import viewsets
+from rest_framework.pagination import LimitOffsetPagination
+
+from api.organisations import models, serializers
+from api.core.authentication import DataWorkspaceOnlyAuthentication
+
+
+class OrganisationsView(viewsets.ReadOnlyModelViewSet):
```
saruniitr, line 8: We already export Organisations to DW,
https://github.com/uktrade/lite-api/blob/dev/api/data_workspace/views.py#L17

api/data_workspace/organisations_views.py
```diff
@@ -0,0 +1,20 @@
...
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    serializer_class = serializers.OrganisationListSerializer
+    pagination_class = LimitOffsetPagination
+    queryset = models.Organisation.objects.all()
+
+class SiteView(viewsets.ReadOnlyModelViewSet):
```
saruniitr, line 14: Is this required to be included in any reports?

saruniitr: Organisation serializer has a primary site field which contains address as well so we are export primary site details already to DW.
There could be other sites which this view can export but just trying to understand if this is for any reporting purpose or to export all of available data?

hnryjmes (commented):
I noticed there was an error to do with black formatting, have you set up `pre-commit` for this repo locally? when you do that it should run black for you

api/data_workspace/organisations_views.py
dejandbt: I remove it

kevincarrogan:
I see that there is still a failing test on this that I don't think would be fixed by re-running the tests.

Do you know what's happening here? We won't be able to merge this in until all of those tests are passing.

api/data_workspace/tests/test_address_views.py
```diff
@@ -0,0 +1,15 @@
+from django.urls import reverse
+from rest_framework import status
+from test_helpers.clients import DataTestClient
+
+
+class AddressDataWorkspaceTests(DataTestClient):
+    def test_addresses(self):
+        url = reverse("data_workspace:dw-address-response")
```
saruniitr, line 8: This test is failing because the url is not the expected one.
I understand this is how you added in `urls.py` file but it is updating slightly after it is registered with router.
If you use `django_extensions` then you get a management command called `show_urls`, that should give you the name to use in this case.

seijihg:
In the screenshot from Brendan, since your serializers use `fields = "__all__"`, your tests are missing the fields that were automatically added via `TimestampableModel ("created_at", "updated_at")`. Additionally, in the Organisation model, you are missing the `registration_number` field, which exists in the `Organisation` class that inherits from `TimestampableModel`. If these fields are not intended to be exposed to DW, the serializer should have been explicitly specified instead of using `__all__`.

api/data_workspace/organisations_views.py
dejandbt: We need site in order to create new reports.

currycoder (dismissed):

api/data_workspace/tests/test_address_views.py
```diff
@@ -0,0 +1,15 @@
...
+class AddressDataWorkspaceTests(DataTestClient):
+    def test_addresses(self):
+        url = reverse("data_workspace:dw-address-response-list")
+        expected_fields = {"id", "address_line_1", "address_line_2", "city", "region", "postcode", "country"}
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        results = response.json()["results"]
+        self.assertGreater(len(results), 0)
+        self.assertEqual(set(results[0].keys()), expected_fields)
```
currycoder, line 15: Would suggest that this test is tightened up.  Right now this is depending on there being at least one address present in the DB and it's checking that part of the response (the keys on the first object) is as expected.

Instead, could it be possible to firstly delete all addresses (which are themselves created by `DataTestClient`, I assume)
```
from api.addresses.models import Address
...
def test_addresses(self):
    Address.objects.all().delete()
    ...
```

Then create a very explicit record just for this test
```
from api.addresses.tests.factories import AddressFactory
...
address = AddressFactory(address_line_1="some first line", ...)
```

And finally assert that the whole response is as you expect it to be;
```
expected_response = [{
    "address_line_1": "some first line",
    ...
}]
self.assertEqual(response.json(), expected_response)
```

api/data_workspace/tests/test_organisations_views.py
```diff
@@ -0,0 +1,24 @@
...
+            "users",
+            "admin_users",
+            "is_used_on_application",
+        }
+
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        results = response.json()["results"]
+        self.assertGreater(len(results), 0)
+        self.assertEqual(set(results[0].keys()), expected_fields)
```
currycoder, line 24: Would be good for this to also follow the tighter test pattern suggested above.

api/data_workspace/urls.py
```diff
@@ -89,4 +91,6 @@
     basename="dw-audit-licence-updated-status",
 )
 router_v1.register("survey-response", views.SurveyResponseListView, basename="dw-survey-reponse")
+router_v1.register("address-response", address_views.AddressView, basename="dw-address-response")
+router_v1.register("site-response", organisations_views.SiteView, basename="dw-site-response")
```
currycoder, line 95: I think you can drop "response" from all of this - assuming this is a copy/paste leftover from above where the model name is "SurveyResponse"

currycoder (commented):
This looks pretty close, but a few suggestions in line that hopefully won't take too long.

api/data_workspace/urls.py
dejandbt: I changed the url

seijihg (approved):

api/data_workspace/tests/test_address_views.py
dejandbt: I don't know is this the right way to do it because as I see we are pre populating the database with values, and if I delete all records from the table I might break other tests.

kevincarrogan (approved):

api/data_workspace/tests/test_address_views.py
currycoder: This should be ok since our test cases each run in their own transaction.  The transaction is rolled back after each test case completes - so although you may delete some records that other test cases depend upon, they are re-instated at the end of the testcase.

I've proven the approach which tightens this testcase up in an example PR here; https://github.com/uktrade/lite-api/pull/2101 (lite-routing failing due to a flaky testcase)

PR #2047, opened 2024-06-17
LTD-5116-create-lu-cic-flag 

depsiatwal:
Create LU CIC Flags.


api/flags/migrations/0007_populate_lu_cic_flag.py
```diff
@@ -0,0 +1,30 @@
...
+    Flag = apps.get_model("flags", "Flag")
+    
+    if not Flag.objects.filter(name=FLAG_NAME):
+        Flag.objects.create(
+            name = FLAG_NAME,
+            level = 'Case',
+            team_id = LICENSING_TEAM_UNIT_ID,
+            blocks_finalising=True,
+            alias = 'LU_CHANGE_IN_CIRCUMSTANCES',
+        )
```
saruniitr, line 18: Migrations to create flags, queues etc (anything that can be related to routing) are added in `lite-routing`

depsiatwal: interesting why ?
probably one worth discussing in dev standup at some point @kevincarrogan

markj0hnst0n (approved):

currycoder (approved):

api/flags/migrations/0007_populate_lu_cic_flag.py
kevincarrogan: It's like this to handle the similar problem that you were having in denials where you'd update the tests in `lite_routing` to require a certain flag but if you add the migration in `lite-api` you have this awkward dependency between the two.

Adding flags to `lite_routing` as a general rule of thumb tends to solve this issue as flags are often added as part of wider routing changes.

PR #2050, opened 2024-06-18
[LTD-5120] [LTD-5122] Amendment by copy audit trail entries

currycoder:
### Aim

Add audit trail entries to "amend by copy" flow as per Jira tickets.

[LTD-5120](https://uktrade.atlassian.net/browse/LTD-5120)
[LTD-5122](https://uktrade.atlassian.net/browse/LTD-5122)

[LTD-5120]: https://uktrade.atlassian.net/browse/LTD-5120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-5122]: https://uktrade.atlassian.net/browse/LTD-5122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/models.py
```diff
@@ -198,25 +194,13 @@ def change_status(self, user, status: CaseStatus, note: Optional[str] = ""):
...
-        # Only allow the final decision if the user has the MANAGE_FINAL_ADVICE permission
-        if status.status == CaseStatusEnum.FINALISED:
-            assert_user_has_permission(user.govuser, GovPermissions.MANAGE_LICENCE_FINAL_ADVICE)
-
-        if not can_set_status(self, status.status):
-            raise ValidationError({"status": [strings.Statuses.BAD_STATUS]})
-
-        if not can_status_be_set_by_gov_user(user.govuser, old_status, status.status, is_mod=False):
-            raise ValidationError({"status": ["Status cannot be set by user"]})
-
```
currycoder, line 219: Note that this block has been moved out to the view which (up until this PR, exclusively) uses this method.

In actual fact, the first two if statements could be dropped as part of this refactor.  The first if statement is covered within `can_status_be_set_by_gov_user()` and the second if statement uses a legacy function `can_set_status()` which only applies to dropped compliance visit logic.

hnryjmes (approved):
looks good. those massive migrations to alter the audit choices are difficult to read but not sure we can do anything about that right now.

currycoder:
@hnryjmes Yes I would have liked the migrations to be smaller, but I also wanted to do atomic commits.  It's fine - just a lot of auto-generated code.

depsiatwal (approved):
lgtm

api/applications/models.py
```diff
@@ -199,6 +201,20 @@ class BaseApplication(ApplicationPartyMixin, Case):
     class Meta:
         ordering = ["created_at"]
 
+    def on_submit(self, old_status):
+        additional_payload = {}
+        if self.amendment_of:
+            # Add an audit entry to the case that was superseded by this amendment
+            audit_trail_service.create_system_user_audit(
+                verb=AuditType.EXPORTER_SUBMITTED_AMENDMENT,
```
saruniitr, line 209: We have `reversion` now, is it time to track this model as well to have history?
When we get to improve auditing we will have data.

currycoder: I think that's a good idea. Would you be ok if we added this to a follow-up ticket?  I'm mindful that it will extend the scope of an already fairly large PR

saruniitr: yeah as a follow-up only, should've made it clear.

saruniitr (approved):

PR #2054, opened 2024-06-19
LTD-4965-dbt-platform-health-check 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4965

As per documentation adding in health check for celery worker. 

https://platform.readme.trade.gov.uk/


api/conf/settings.py
```diff
@@ -527,3 +530,8 @@ def _build_redis_url(base_url, db_number, **query_args):
 
 DB_ANONYMISER_CONFIG_LOCATION = Path(BASE_DIR) / "conf" / "anonymise_model_config.yaml"
 DB_ANONYMISER_DUMP_FILE_NAME = env.str("DB_ANONYMISER_DUMP_FILE_NAME", "anonymised.sql")
+
+
+# DBT Platform Spectic config
```
hnryjmes, line 535: was this meant to say `# DBT Platform specific config` ?

hnryjmes (approved):

api/conf/settings.py
depsiatwal: DBT Platform specific config is the new platform away from GOV Paas

PR #2055, opened 2024-06-19
LTD-4962 Add in ASIM format logging functionality and env variable to enable s…

markj0hnst0n:
### Aim

Add in ASIM format logging functionality.

Logs to show in simple format for local Env, ECS Format for Gov Paas and ASIM format for dbt platform.

Used `is_copilot()` variable to ensure AWS Copilot is not present for Gov Paas and DBT Platform uses this.

Docs [here](https://github.com/uktrade/copilot-python)

[LTD-4962](https://uktrade.atlassian.net/browse/LTD-4962)

[LTD-4962]: https://uktrade.atlassian.net/browse/LTD-4962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/conf/settings.py
```diff
@@ -369,18 +370,24 @@ def _build_redis_url(base_url, db_number, **query_args):
...
-    "root": {"handlers": ["stdout", "ecs"], "level": env("LOG_LEVEL").upper()},
+    "root": {"handlers": ["stdout", "ecs", "asim"], "level": env("LOG_LEVEL").upper()},
     "loggers": {
         DENIAL_REASONS_DELETION_LOGGER: {"handlers": ["sentry"], "level": logging.WARNING},
     },
 }
 
+additional_logger_config = env.json("ADDITIONAL_LOGGER_CONFIG", default=None)
+if additional_logger_config:
+    LOGGING["loggers"] = additional_logger_config
```
currycoder, line 389: How will this be used?

markj0hnst0n: I've actually removed this based on a suggestion form @kevincarrogan on a jira ticket for the other repo.  Pattern will be just to add this new asim logger once we need to switch it on for dbt platform

kevincarrogan: Although here's a question for you. Why can't we just switch it on now? I think part of the early code prep stuff is that it's things that are generally backwards compatible with gov paas anyway.

kevincarrogan: Or in cases that it's not backward compatible isn't this the whole point of the `if is_copilot` helper thingy where we can toggle settings based on the return of that.

api/conf/settings.py
```diff
@@ -362,25 +363,47 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+if ENVIRONMENT == "local":
+    LOGGING["formatters"] = {
+        "simple": {"format": "{asctime} {levelname} {message}", "style": "{"},
+    }
+    LOGGING["handlers"] = {
+        "stdout": {"class": "logging.StreamHandler", "formatter": "simple"},
+    }
+    LOGGING["root"] = {"handlers": ["stdout"], "level": env.str("LOG_LEVEL", "info").upper()}
+
+elif VCAP_SERVICES:
```
kevincarrogan, line 387: Is there a reason why we're so apprehensive to use the `is_copilot` helper for things like this and choosing to use `VCAP_SERVICES` instead?

api/conf/settings.py
```diff
@@ -362,25 +363,47 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+LOGGING = {"version": 1, "disable_existing_loggers": False}
+
+if ENVIRONMENT == "local":
+    LOGGING["formatters"] = {
+        "simple": {"format": "{asctime} {levelname} {message}", "style": "{"},
+    }
+    LOGGING["handlers"] = {
+        "stdout": {"class": "logging.StreamHandler", "formatter": "simple"},
+    }
+    LOGGING["root"] = {"handlers": ["stdout"], "level": env.str("LOG_LEVEL", "info").upper()}
```
kevincarrogan, line 385: I'm wondering if `LOGGING.UPDATE{...}` might be more readable here.

api/conf/settings.py
```diff
@@ -362,25 +363,47 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+else:
+    LOGGING["formatters"] = {
+        "asim_formatter": {
+            "()": ASIMFormatter,
+        },
+    }
+    LOGGING["handlers"] = {
+        "asim": {"class": "logging.StreamHandler", "formatter": "asim_formatter"},
+    }
+    LOGGING["root"] = {"handlers": ["asim"], "level": env.str("LOG_LEVEL", "info").upper()}
```
kevincarrogan, line 405: What's kind of interesting here is that the only thing of significance that's actually changing here is that formatter but we seem to make lots of additional changes just because the name is changing, I wonder if we could condense this is some way by keeping a uniform key throught.

api/conf/settings.py
markj0hnst0n: I was thinking ASIM would be default behaviour so I wanted Gov Paas to be the exception and then it can just be removed after.  Also it feels premature to leverage something which is coupled to the use of AWS ECS when we're not doing that at the minute

api/conf/settings.py
markj0hnst0n: The handlers, formatters and in turn the root handlers are all quite tightly coupled so it would just be trading readability for dict value swapping.  I'm sure you have something in mind but this is reduced to as simple as I'd like to take it.

api/conf/settings.py
markj0hnst0n: I've considered this butI don't think it would make a huge deal of difference to add on another layer of complexity

markj0hnst0n: I've added this in

api/conf/settings.py
markj0hnst0n: added the `is_copilot() ` var in there

api/conf/settings.py
```diff
@@ -365,25 +368,36 @@ def _build_redis_url(base_url, db_number, **query_args):
...
-        "ecs": {"class": "logging.StreamHandler", "formatter": "ecs_formatter"},
         "sentry": {"class": "sentry_sdk.integrations.logging.EventHandler"},
     },
-    "root": {"handlers": ["stdout", "ecs"], "level": env("LOG_LEVEL").upper()},
     "loggers": {
         DENIAL_REASONS_DELETION_LOGGER: {"handlers": ["sentry"], "level": logging.WARNING},
     },
 }
 
+ENVIRONMENT = env.str("ENV", "")
```
hnryjmes, line 382: we already have `ENV` defined on line 423 https://github.com/uktrade/lite-api/blob/dev/api/conf/settings.py#L423

given that your additional code is before that variable gets defined, it makes sense to keep your variable assignment, but if you rename it to `ENV` that enables you to remove the one later down so there's just a single variable that we reference

markj0hnst0n: changed

api/conf/settings.py
```diff
@@ -362,28 +365,38 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+
+elif not is_copilot():
+    LOGGING.update({"formatters": {"ecs_formatter": {"()": ECSFormatter}}})
+    LOGGING["handlers"].update({"ecs": {"class": "logging.StreamHandler", "formatter": "ecs_formatter"}})
+    LOGGING.update({"root": {"handlers": ["ecs"], "level": env("LOG_LEVEL").upper()}})
+
+else:
+    LOGGING.update({"formatters": {"asim_formatter": {"()": ASIMFormatter}}})
+    LOGGING["handlers"].update({"asim": {"class": "logging.StreamHandler", "formatter": "asim_formatter"}})
+    LOGGING.update({"root": {"handlers": ["asim"], "level": env("LOG_LEVEL").upper()}})
```
currycoder, line 397: I realise this is going around in circles a bit now, but do we really need this level of complexity?  What prevented us from just adding the asim handler in the above `LOGGING` variable like so;

```
LOGGING = {
    "version": 1,
    "disable_existing_loggers": False,
    "formatters": {
        "simple": {"format": "{asctime} {levelname} {message}", "style": "{"},
        "ecs_formatter": {"()": ECSFormatter},
        "asim_formatter": {"()": ASIMFormatter},
    },
    "handlers": {
        "stdout": {"class": "logging.StreamHandler", "formatter": "simple"},
        "ecs": {"class": "logging.StreamHandler", "formatter": "ecs_formatter"},
        "sentry": {"class": "sentry_sdk.integrations.logging.EventHandler"},
        "asim": {"class": "logging.StreamHandler", "formatter": "asim_formatter"},
    },
    "root": {"handlers": ["stdout", "ecs", "asim"], "level": env("LOG_LEVEL").upper()},
    "loggers": {
        DENIAL_REASONS_DELETION_LOGGER: {"handlers": ["sentry"], "level": logging.WARNING},
    },
}
```

Is it that ASIM and ECS can explicitly *not* be used together on a single environment?

markj0hnst0n: @currycoder They can be but it's not the outcome we want in this case.  This outputs 3 versions of logs in each environment.  The pattern that Kevin had outlined would be most in keeping with how we want to move forward is:

'Logs to show in simple format for local Env, ECS Format for Gov Paas and ASIM format for dbt platform.'

currycoder (approved):

depsiatwal (approved):
lgtm

PR #2060, opened 2024-06-24
 LTD-4960-dbt-platform-application-settings

depsiatwal:
Add application setting for DBT platform


api/conf/settings.py
```diff
@@ -531,7 +493,87 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+        ]
+
+else:
+    # Database
+    DATABASES = {"default": env.db()}  # https://docs.djangoproject.com/en/2.1/ref/settings/#databases
+
+    # redis 
+
+    if "redis" in VCAP_SERVICES:
+        REDIS_BASE_URL = VCAP_SERVICES["redis"][0]["credentials"]["uri"]
```
markj0hnst0n, line 538: Would be a better idea to separate out the VCAP services references into one path rather than nest within the `is_copilot` block for easy deletion earlier.  Kevin had mentioned his preferences on this pattern on my PR [here](https://github.com/uktrade/lite-api/pull/2055)

depsiatwal: good point redone so it can just be deleted.

markj0hnst0n (approved):

api/conf/settings.py
```diff
@@ -520,29 +462,163 @@ def _build_redis_url(base_url, db_number, **query_args):
...
+DB_ANONYMISER_CONFIG_LOCATION = Path(BASE_DIR) / "conf" / "anonymise_model_config.yaml"
+DB_ANONYMISER_DUMP_FILE_NAME = env.str("DB_ANONYMISER_DUMP_FILE_NAME", "anonymised.sql")
+
+# DBT Platform Spectic config
+
+VCAP_SERVICES = env.json("VCAP_SERVICES", {})
+
+if IS_DBT_PLATFORM:
+    ALLOWED_HOSTS = setup_allowed_hosts(ALLOWED_HOSTS)
+
```
saruniitr, line 479: I think some of the comments from Brendan in the FE PR apply here as well.

These `if-else` blocks are quite long and not easy to follow, is it better to create separate settings file for PaaS and DBT platform?

hnryjmes (approved):
I recognise that some of the if-else branching looks a bit hard to follow now, but we won't keep this around very long, so I don't think it's that much of a risk

PR #2062, opened 2024-06-27
Update ApplicationPartyView put to check case status

hnryjmes:


api/applications/views/parties.py
```diff
@@ -136,6 +137,14 @@ def get(self, request, **kwargs):
 
     @authorised_to_view_application(ExporterUser)
     def put(self, request, **kwargs):
+        # Check if the case is in draft or applicant_editing status
+        case_status = self.application.get_case().status.status
+        if case_status not in [CaseStatusEnum.APPLICANT_EDITING, CaseStatusEnum.DRAFT]:
+            return JsonResponse(
+                data={"errors": [f"The case cannot be edited in {case_status} status"]},
```
saruniitr, line 144: The party cannot be edited? as this is about editing party details.
Could you also include party type to capture more details in sentry.

api/applications/views/parties.py
```diff
@@ -136,6 +137,14 @@ def get(self, request, **kwargs):
 
     @authorised_to_view_application(ExporterUser)
     def put(self, request, **kwargs):
```
saruniitr, line 139: I think you need similar check in `post` also for consignee and other parties,
https://github.com/uktrade/lite-api/blob/dev/api/applications/views/parties.py#L40

I see there is an entry point to this for editing consignee in FE.
https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/views/parties/consignees.py#L94
Edit in those cases is basically creating a new instance and replacing existing one.

kevincarrogan (approved):

PR #2065, opened 2024-06-28
LTD-5071-forward-slash-breaking-search

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5071

This has been tested on the case highlighted in the Jira and can confirm we can now match cases when we have a forward slash in the search parameters.

Originally this was in a frontend PR to test it quickly but that felt a bit hackey hence why it's been implemented as a backend change.

### DIFF
--- a/api/core/search/filter_backends.py
+++ b/api/core/search/filter_backends.py
@@ -13,6 +13,20 @@ class QueryStringSearchFilterBackend(BaseSearchFilterBackend):
 
     search_param = "search"
 
+    def get_search_query_params(self, request):
+        """Get search query params.
+
+        :param request: Django REST framework request.
+        :type request: rest_framework.request.Request
+        :return: List of search query params.
+        :rtype: list
+        """
+        query_params = request.query_params.copy()
+        # This is required as query_string is unable to handle a single /
+        query_string = query_params.get(self.search_param, "").replace("/", "//")
+        query_params[self.search_param] = query_string
+        return query_params.getlist(self.search_param, [])
+
     query_backends = [
         QueryStringQueryBackend,
     ]
--- a/api/search/product/tests/test_views.py
+++ b/api/search/product/tests/test_views.py
@@ -110,7 +110,8 @@ class ProductSearchTests(BaseProductSearchTests):
         [
             ({"search": "ABC"}, 0),
             ({"search": "ABC-123"}, 1),
-            ({"search": "IMG-1300"}, 1),
+            ({"search": "IMG/1300"}, 1),
+            ({"search": "IMG-1300"}, 0),
             ({"search": "867-"}, 0),
             ({"search": "867-5309"}, 1),
             ({"search": "H2SO4"}, 1),
--- a/api/applications/tests/test_matching_denials.py
+++ b/api/applications/tests/test_matching_denials.py
@@ -17,9 +17,12 @@ class ApplicationDenialMatchesOnApplicationTests(DataTestClient):
         file_path = os.path.join(settings.BASE_DIR, "external_data/tests/denial_valid.csv")
         with open(file_path, "rb") as f:
             content = f.read()
+            f.seek(0)
+            self.total_denials = len(f.readlines()) - 1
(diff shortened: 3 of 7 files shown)


Tllew (approved):

api/core/search/filter_backends.py
```diff
@@ -13,6 +13,20 @@ class QueryStringSearchFilterBackend(BaseSearchFilterBackend):
 
     search_param = "search"
 
+    def get_search_query_params(self, request):
+        """Get search query params.
```
saruniitr, line 17: This change also affects product search.
Could you add a test for product search also to see that it doesn't affect?

depsiatwal: added couple of tests

api/search/product/tests/test_views.py
```diff
@@ -114,6 +115,7 @@ def test_product_search_by_name(self, query, expected_count, expected_name):
             ({"search": "867-"}, 0),
             ({"search": "867-5309"}, 1),
             ({"search": "H2SO4"}, 1),
+            ({"search": "H2SO4/"}, 1),
```
saruniitr, line 118: thanks.
Could you update as below (or I can push these changes to this branch) so that it is in between the search string?
In these examples `/` is not part of the test data.
```
diff --git a/api/search/product/tests/test_helpers.py b/api/search/product/tests/test_helpers.py
index dcb8af98..cc280c0f 100644
--- a/api/search/product/tests/test_helpers.py
+++ b/api/search/product/tests/test_helpers.py
@@ -49,7 +49,7 @@ test_applications_data = [
             {
                 "good": {
                     "name": "Thermal camera",
-                    "part_number": "IMG-1300",
+                    "part_number": "IMG/1300",
                     "is_good_controlled": True,
                     "control_list_entries": ["6A005"],
                 },
diff --git a/api/search/product/tests/test_views.py b/api/search/product/tests/test_views.py
index 58e31b24..97a08622 100644
--- a/api/search/product/tests/test_views.py
+++ b/api/search/product/tests/test_views.py
@@ -111,11 +111,11 @@ class ProductSearchTests(BaseProductSearchTests):
         [
             ({"search": "ABC"}, 0),
             ({"search": "ABC-123"}, 1),
-            ({"search": "IMG-1300"}, 1),
+            ({"search": "IMG1300"}, 0),
+            ({"search": "IMG/1300"}, 1),
             ({"search": "867-"}, 0),
             ({"search": "867-5309"}, 1),
             ({"search": "H2SO4"}, 1),
-            ({"search": "H2SO4/"}, 1),
         ]
     )
     def test_product_search_by_part_number(self, query, expected_count):
```

depsiatwal: done

depsiatwal: @saruniitr  is this what you hand in mind ?

api/search/product/tests/test_views.py
```diff
@@ -110,10 +110,12 @@ def test_product_search_by_name(self, query, expected_count, expected_name):
         [
             ({"search": "ABC"}, 0),
             ({"search": "ABC-123"}, 1),
-            ({"search": "IMG-1300"}, 1),
+            ({"search": "IMG/1300"}, 1),
+            ({"search": "IMG-1300"}, 0),
             ({"search": "867-"}, 0),
             ({"search": "867-5309"}, 1),
             ({"search": "H2SO4"}, 1),
+            ({"search": "H2SO4/"}, 1),
```
saruniitr, line 118: I think this case is a bit misleading as there is no test product with this part number hence I removed it in the diff that I shared with you.
If the `/` is at the end it is working like it is escaping and matching on the previous string but if it is in the middle then it is being considered. Because of the suggestions in product search it is not much of an issue but I would like to investigate this a bit later. At the same time do not want to hold this further.

I am happy for this to be merged with this case of `H2SO4/` removed.

saruniitr (approved):
with the specified test case removed it looks good to me.

PR #2066, opened 2024-06-28
LTD-3688 add/amend units of measurement.

markj0hnst0n:
## N.B. Corresponding LITE HMRC PR must be released in PROD before this PR

I've made notes of the prod data impact with comments on the tickets relating to these issues.

Make the following amendments:

- Tonnes / TON - Add and use code 25
[LTD-3690](https://uktrade.atlassian.net/browse/LTD-3690)

- Milligrams / MIM - use code 111 and change to “MGM”
- Micrograms / MCM - use code 116 and change to “MCG”
- Microlitres / MCR - use code 110 and change to “MCL”
[LTD-3688](https://uktrade.atlassian.net/browse/LTD-3688)

- Millilitres / MIR - change to “MLT”
[LTD-3691](https://uktrade.atlassian.net/browse/LTD-3691)

Corresponding LITE HMRC PR [here](https://github.com/uktrade/lite-hmrc/pull/265)

[LTD-3688]: https://uktrade.atlassian.net/browse/LTD-3688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-3690]: https://uktrade.atlassian.net/browse/LTD-3690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-3691]: https://uktrade.atlassian.net/browse/LTD-3691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hnryjmes (approved):

seijihg (approved):

hnryjmes (dismissed):
I am generally happy with the code changes but before we continue to merge this in we should have:
- a good understanding of which if any PROD cases would be affected by this change (i.e. which cases in prod have goods making use of the units of measurement that will be removed)
- a release plan (can be informal) taking into account backwards compatibility if we are not able to release lite-api and lite-hmrc simultaneously

Tllew:
Agreed with henry's comment, maybe just a blanket migration that converts any values that have been altered to the updated version? 

It doesn't look like the migration we have does this, I could be wrong though

api/applications/migrations/0083_amend_existing_goodonapplication_unit_legacy_codes.py
```diff
@@ -0,0 +1,30 @@
...
+
+    for good in GoodOnApplication.objects.filter(unit__in=legacy_unit_codes):
+        if good.unit == legacy_unit_codes[0]:
+            good.unit = Units.MGM
+        elif good.unit == legacy_unit_codes[1]:
+            good.unit = Units.MCG
+        elif good.unit == legacy_unit_codes[2]:
+            good.unit = Units.MLT
+        elif good.unit == legacy_unit_codes[3]:
+            good.unit = Units.MCL
```
currycoder, line 20: This sort of pattern with quite repetitive if statements is usually a sign that there may be a better data structure to use. There's an opportunity here to use a dict instead to make the code more readable.  e.g.

```
unit_mapping = {
    "MIM": "MGM",
    ...
}

for good in GoodOnApplication.objects.filter(unit__in=unit_mapping.keys()):
    legacy_unit = good.unit    
    good.unit = unit_mapping[legacy_unit]
    good.save()
```

api/applications/migrations/0083_amend_existing_goodonapplication_unit_legacy_codes.py
```diff
@@ -0,0 +1,30 @@
...
+        good.save()
```
currycoder, line 21: Would recommend calling this `good_on_application`.  We have Good and GoodOnApplication objects and this kind of explicit variable naming makes things a bit easier to grok.

api/applications/migrations/tests/test_0083_amend_existing_goodonapplication_unit_legacy_codes.py
```diff
@@ -0,0 +1,50 @@
...
+        GoodOnApplication = self.new_state.apps.get_model("applications", "GoodOnApplication")
+        new_good_on_applications = GoodOnApplication.objects.all()
+
+        assert all(
+            new_good_on_application.unit not in self.old_unit_codes
+            for new_good_on_application in new_good_on_applications
+        )
+        assert all(
+            new_good_on_application.unit in self.new_unit_codes for new_good_on_application in new_good_on_applications
+        )
```
currycoder, line 50: This does work, and I see where it's coming from - but it's a little bit loosey goosey.  Right now, we just assert that every created record has a unit in the new list of codes.

It would be really good to see that each record that we prepared has changed from one expected value to another.  And for those checks to be very explicit. e.g. something like...

```
def prepare(self):
    expected_mappings = {
        "MIM": "MGM",
        "<unchanged>": "<unchanged>",
        ...
    }
    self.test_records = [
        {
             "id": old_state_good_on_application_factory(unit=old_unit),
             "expected_unit": new_unit,
         } for unit, new_unit in expected_mappings
    ]

def test_change_legacy_unit_codes(self):
    GoodOnApplication = self.new_state.apps.get_model("applications", "GoodOnApplication")
    for test_record in self.test_records:
        good_on_application = GoodOnApplication.objects.get(id=test_record["id"])
        assert good_on_application.unit == test_record["expected_unit"]
```

api/applications/migrations/0083_amend_existing_goodonapplication_unit_legacy_codes.py
markj0hnst0n: Agreed.  This is a lot nicer.

api/applications/migrations/0083_amend_existing_goodonapplication_unit_legacy_codes.py
markj0hnst0n: Agreed.  Lovely use of the term 'grok' also.

hnryjmes: agree, `good_on_application` or `goa` if you need to abbreviate it

api/applications/migrations/tests/test_0083_amend_existing_goodonapplication_unit_legacy_codes.py
markj0hnst0n: Thanks for this.  I was thinking it could be tighter but I'd been round the houses a bit trying even to get the migration testing working at all so I think I just jumped on the solution I had.

api/applications/migrations/tests/test_0083_amend_existing_goodonapplication_unit_legacy_codes.py
```diff
@@ -0,0 +1,60 @@
...
+    CaseStatus = self.old_state.apps.get_model("statuses", "CaseStatus")
+    Good = self.old_state.apps.get_model("goods", "Good")
+    GoodOnApplication = self.old_state.apps.get_model("applications", "GoodOnApplication")
+    StandardApplication = self.old_state.apps.get_model("applications", "StandardApplication")
+    Organisation = self.old_state.apps.get_model("organisations", "Organisation")
+    Case = self.old_state.apps.get_model("cases", "Case")
+    CaseType = self.old_state.apps.get_model("cases", "CaseType")
+
+    case_status = CaseStatus.objects.get(status="submitted")
+    case_type = CaseType.objects.get(type="application", reference="oiel", sub_type="open")
```
hnryjmes, line 16: why is the case type here OIEL? as far as I know we don't support OIELs in LITE...

api/applications/migrations/tests/test_0083_amend_existing_goodonapplication_unit_legacy_codes.py
```diff
@@ -0,0 +1,60 @@
...
+    migrate_to = ("applications", "0083_amend_existing_goodonapplication_unit_legacy_codes")
+
+    def prepare(self):
+
+        self.expected_mappings = {
+            "MIM": Units.MGM,
+            "MCM": Units.MCG,
+            "MIR": Units.MLT,
+            "MCR": Units.MCL,
+            "NAR": Units.NAR,
```
hnryjmes, line 40: is this included for regression testing? why this unit specifically and not others, given that NAR is not being changed?

markj0hnst0n: So NAR is not changing so I thought it would be useful to include as it proves that a unit that is in the db is not impacted by this migration.  NAR was chosen as the majority of PROD cases use this unit.

hnryjmes (approved):
approving, just one minor comment re case type

api/applications/migrations/tests/test_0083_amend_existing_goodonapplication_unit_legacy_codes.py
markj0hnst0n: The short answer to that is that it works as an oiel and it doesn't as a siel.  I think it's to do with how the 'fake' lookup of the casestatus enum works.

hnryjmes: ok, that makes sense to me... I've already approved but if you would like to get it working with `siel` as a matter of style it would probably be nicer. but not too important as we often delete migration tests like this once we've verified the migration has run successfully, so it's not like it will stick around forever or anything

markj0hnst0n: Totally agree.  It's something I'd like to get working correctly so I will have another stab at it.  I'm reluctant to merge this in at all until I have a firm timeline for the HMRC release so I think I have a bit of time.

markj0hnst0n: I've amended this now @hnryjmes to use the Siel case type.  The issue was around the `case_subtype`.  The previous one didn't work with siel but it did with oiel.

PR #2067, opened 2024-07-01
Remove relation

kevincarrogan:
### Aim

Stops audit entries that are related to a `GoodOnApplication` from being removed when a `GoodOnApplication` object is deleted.

### DIFF
--- a/api/applications/models.py
+++ b/api/applications/models.py
@@ -1,6 +1,5 @@
 import uuid
 
-from django.contrib.contenttypes.fields import GenericRelation
 from django.contrib.postgres.fields import ArrayField
 from django.db import models, transaction
 from django.utils import timezone
@@ -16,7 +15,7 @@ from api.applications.enums import (
 )
 from api.appeals.models import Appeal
 from api.applications.managers import BaseApplicationManager
-from api.audit_trail.models import Audit, AuditType
+from api.audit_trail.models import AuditType
 from api.audit_trail import service as audit_trail_service
 from api.cases.enums import CaseTypeEnum
 from api.cases.models import Case, CaseQueue
@@ -512,12 +511,6 @@ class GoodOnApplication(AbstractGoodOnApplication, Clonable):
     # Exhibition applications are the only applications that contain the following as such may be null
     item_type = models.CharField(choices=ItemType.choices, max_length=10, null=True, blank=True, default=None)
     other_item_type = models.CharField(max_length=100, null=True, blank=True, default=None)
-    audit_trail = GenericRelation(
-        Audit,
-        related_query_name="good_on_application",
-        content_type_field="action_object_content_type",
-        object_id_field="action_object_object_id",
-    )
     control_list_entries = models.ManyToManyField(ControlListEntry, through=GoodOnApplicationControlListEntry)
     regime_entries = models.ManyToManyField(RegimeEntry, through=GoodOnApplicationRegimeEntry)
 
--- a/api/applications/tests/test_audit_trail.py
+++ b/api/applications/tests/test_audit_trail.py
@@ -1,15 +1,22 @@
+from api.audit_trail.models import Audit
 from api.audit_trail.tests.factories import AuditFactory
+
 from test_helpers.clients import DataTestClient
 
 
 class GoodOnApplicationAuditTrailTests(DataTestClient):
     def test_removing_object_keeps_audit_trail(self):
         application = self.create_draft_standard_application(self.organisation)
+        Audit.objects.all().delete()
+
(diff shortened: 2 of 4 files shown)


api/applications/tests/test_audit_trail.py
```diff
@@ -0,0 +1,15 @@
...
+    def test_removing_object_keeps_audit_trail(self):
+        application = self.create_draft_standard_application(self.organisation)
+        good_on_application = application.goods.first()
+        audit = AuditFactory(
+            action_object=good_on_application,
+        )
+
+        good_on_application.delete()
+        audit.refresh_from_db()
+        self.assertIsNone(audit.action_object)
```
saruniitr, line 15: so previously this was not none and cascading to delete audit entries also?

Could it be better if the test creates an audit entries related to GoodOnApplication and ensure they are still present after the delete? Personally this is not clear to me that audit entries are retained.

api/applications/tests/test_audit_trail.py
```diff
@@ -0,0 +1,15 @@
...
+class GoodOnApplicationAuditTrailTests(DataTestClient):
+    def test_removing_object_keeps_audit_trail(self):
+        application = self.create_draft_standard_application(self.organisation)
+        good_on_application = application.goods.first()
+        audit = AuditFactory(
+            action_object=good_on_application,
+        )
+
+        good_on_application.delete()
+        audit.refresh_from_db()
```
currycoder, line 14: What would happen if the audit object was deleted and you did this?  Could it be a little more readable to do a totally new `Audit.objects.get()` instead?  e.g. to prove it's still present

currycoder (commented):
Would like to suggest that we also record `good_id` in the audit trail payload for assessments.  I think right now we will retain the product reviewed audit entries, but we will make it hard to actually correlate which product they were for if the `action_object` is None

Would be happy for this bit to be a separate follow-up ticket though

api/applications/tests/test_audit_trail.py
kevincarrogan: Sure. When you do the `refresh_from_db` and the object has disappeared you get a big error.

api/applications/tests/test_audit_trail.py
kevincarrogan: That's correct. It would cascade delete.

kevincarrogan: When you say "create an audit entries related to GoodOnApplication" is there a reason why using the factory doesn't explain that for you?

kevincarrogan:
> Would like to suggest that we also record `good_id` in the audit trail payload for assessments. I think right now we will retain the product reviewed audit entries, but we will make it hard to actually correlate which product they were for if the `action_object` is None
> 
> Would be happy for this bit to be a separate follow-up ticket though

Ultimately I'm thinking of just attaching the good object instead of good on application anyway as an alternative to this. At this stage I just wanted to at least retain these audit entries for now as changing any current audit entries would require a data migration that I wasn't keen to do right now.

api/applications/tests/test_audit_trail.py
kevincarrogan: Update to use an explicit `get`.

currycoder (approved):

api/applications/tests/test_audit_trail.py
saruniitr: As this came up mainly with assessment entries, in my mind it was like making an assessment which creates entries for that GoodOnApplication and the test can check the same but this version also works.

api/applications/models.py
```diff
@@ -512,12 +511,6 @@ class GoodOnApplication(AbstractGoodOnApplication, Clonable):
     # Exhibition applications are the only applications that contain the following as such may be null
     item_type = models.CharField(choices=ItemType.choices, max_length=10, null=True, blank=True, default=None)
     other_item_type = models.CharField(max_length=100, null=True, blank=True, default=None)
-    audit_trail = GenericRelation(
```
hnryjmes, line 515: what was the purpose of having `GenericRelation` to begin with? to make querying easier?

saruniitr (approved):

api/applications/models.py
kevincarrogan: Yes. It appears that it would have been implemented so you could say `good_on_application.audit_trail.all()` for example.

However, this adds an implicit cascade delete that you can't seem to do anything about and we really don't want to remove audit trails just because a good_on_application has been deleted.

In this case I saw that this was only really used in one place and I've replaced it with an alternative that doesn't require us to use this attribute and force us to have an implicit cascade delete.

PR #2069, opened 2024-07-01
got a little stuck reindexing, figured it was worth adding to the readme/makefile

Tllew:
got a little stuck reindexing, figured it was worth adding to the readme/makefile


makefile
```diff
@@ -67,3 +67,6 @@ stop-e2e:
 
 test:
 	./manage.py test
+
+rebuild-elastic:
+	docker exec -it api pipenv run ./manage.py search_index --rebuild
```
hnryjmes, line 72: I usually use the variant with the `-f` flag, so that there are no prompts for user input

hnryjmes (approved):

saruniitr (approved):

PR #2070, opened 2024-07-01
Readonly statuses

kevincarrogan:
### Aim

This closes off minor amendments from the API, which has already been done on the frontend.

This also brings in some other changes related to statuses:

  - Switch around logic for readonly to be writeable by inclusion
  - Remove `is_terminal` and `is_read_only` from being database fields in favour of having a single source of truth for these values in code
  - Makes major edit statuses its own explicit list in the case status enum


api/staticdata/statuses/enums.py
```diff
@@ -166,27 +154,7 @@ class CaseStatusEnum:
...
-    def get_read_only_choices(cls):
-        lu_countersign_statuses = []
-        lu_countersign_statuses.extend(
-            [
-                cls.FINAL_REVIEW_COUNTERSIGN,
-                cls.FINAL_REVIEW_SECOND_COUNTERSIGN,
-            ]
-        )
-
-        return cls._read_only_statuses + lu_countersign_statuses
```
hnryjmes, line 189: I'm glad this is being removed, I think this was left over from removing a feature switch for LU countersigning work, which is why it looks so unusual on its own.

api/staticdata/statuses/models.py
```diff
@@ -13,12 +19,28 @@ class CaseStatus(models.Model):
...
+
+    @queryable_property
+    def is_read_only(self):
+        return CaseStatusEnum.is_read_only(self.status)
+
+    @is_read_only.filter(boolean=True)
+    @classmethod
+    def is_read_only(cls):
+        return Q(status__in=CaseStatusEnum.read_only_statuses())
+
```
hnryjmes, line 43: this is a nice change. better than having a db variable which might get out of sync with the case status.

api/staticdata/management/commands/seedcasestatuses.py
```diff
@@ -46,7 +46,7 @@ def operation(self, *args, **options):
             if row["workflow_sequence"] == "None":
                 row["workflow_sequence"] = None
 
-        self.update_or_create(CaseStatus, status_csv)
+        self.update_or_create(CaseStatus, status_csv, exclude=["is_terminal", "is_read_only"])
```
hnryjmes, line 49: silly question - what does the `exclude=["is_terminal", "is_read_only"]` do here? why is it needed?

api/gov_users/tests/test_gov_user_notifications.py
```diff
@@ -13,7 +13,8 @@
 class GovUserNotificationTests(DataTestClient):
     def setUp(self):
         super().setUp()
-        self.case = self.create_standard_application_case(self.organisation, "Case")
+        self.case = self.create_draft_standard_application(self.organisation, "Case")
+        self.set_application_status(self.case, CaseStatusEnum.APPLICANT_EDITING)
```
hnryjmes, line 17: it doesn't seem like a realistic scenario that a draft application would be put into applicant editing status - is this needed for the tests to work? if so, is there another way we could write this?

kevincarrogan: A draft application is just one that is in the draft status, this is me just changing the status to be applicant editing.

api/staticdata/management/commands/seedcasestatuses.py
kevincarrogan: There is a seed command that tries to write to `is_terminal` and `is_read_only` which I've now removed, this stops us from doing that without me having to update all of the csvs.

hnryjmes (approved):

saruniitr (approved):

PR #2072, opened 2024-07-03
LTD-5154: Add dedicated endpoint to archive/restore a good

saruniitr:
## Change description

Exporter has a choice to archive/restore a good and this is only allowed for goods that are submitted/verified on an application.

However we have a restriction that goods cannot be edited once they are submitted on an application. Removing this restriction can cause various issues so to avoid this add a dedicated endpoint which only allows archiving/restoring a particular good.

Add tests to ensure that only archive status can be updated using this endpoint.

### DIFF
--- a/api/goods/tests/test_edit.py
+++ b/api/goods/tests/test_edit.py
@@ -7,6 +7,7 @@ from rest_framework.reverse import reverse
 from reversion.models import Version
 
 from api.goods.enums import (
+    GoodStatus,
     GoodPvGraded,
     PvGrading,
     MilitaryUse,
@@ -571,7 +572,11 @@ class GoodsEditDraftGoodTests(DataTestClient):
         self.assertEqual(good.design_details, "design details")
 
     def test_edit_archive_status(self):
-        good = GoodFactory(organisation=self.organisation, item_category=ItemCategory.GROUP1_COMPONENTS)
+        good = GoodFactory(
+            organisation=self.organisation,
+            status=GoodStatus.SUBMITTED,
+            item_category=ItemCategory.GROUP1_COMPONENTS,
+        )
         url = reverse("goods:archive_restore", kwargs={"pk": str(good.id)})
 
         for version_count, is_archived in enumerate([True, False], start=1):
@@ -613,7 +618,7 @@ class GoodsEditDraftGoodTests(DataTestClient):
         ]
     )
     def test_only_archive_field_can_be_updated(self, initial, data, expected):
-        good = GoodFactory(organisation=self.organisation, **initial)
+        good = GoodFactory(organisation=self.organisation, status=GoodStatus.SUBMITTED, **initial)
         url = reverse("goods:archive_restore", kwargs={"pk": str(good.id)})
 
         response = self.client.put(url, data, **self.exporter_headers)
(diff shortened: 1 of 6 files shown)


seijihg (approved):

api/goods/tests/test_edit.py
```diff
@@ -593,6 +593,37 @@ def test_edit_archive_status(self):
...
+                {"name": "Rifle", "is_archived": False},
+            ],
+            [
+                {"name": "Rifle", "is_archived": None},
+                {"name": "Rifle updated", "is_archived": False},
+                {"name": "Rifle", "is_archived": False},
+            ],
+        ]
+    )
+    def test_only_archive_field_can_be_updated(self, initial, data, expected):
```
currycoder, line 615: Was about to comment recommending something like this, but here it is!

api/goods/views.py
```diff
@@ -194,6 +195,12 @@ def get_queryset(self):
         return queryset.order_by("-updated_at")
 
 
+class GoodArchiveRestore(UpdateAPIView):
+    authentication_classes = (ExporterAuthentication,)
+    queryset = Good.objects.all()
```
currycoder, line 200: Could I suggest a custom queryset method which filters the eligible Goods down to those that are associated with the user's organisation.  e.g. a double defence against somebody being able to archive another org's goods in some way

saruniitr: yeah good idea, will further limit to 'submitted, verified' goods

api/goods/views.py
```diff
@@ -194,6 +195,19 @@ def get_queryset(self):
...
 
 
+class GoodArchiveRestore(UpdateAPIView):
+    authentication_classes = (ExporterAuthentication,)
+    serializer_class = GoodArchiveRestoreSerializer
+
+    def get_queryset(self):
+        organisation = get_request_user_organisation_id(self.request)
+
+        return Good.objects.filter(
```
depsiatwal, line 205: nicely done

api/goods/tests/test_edit.py
```diff
@@ -593,6 +598,70 @@ def test_edit_archive_status(self):
...
+                {"name": "Rifle", "is_archived": True},
+                {"is_archived": False},
+                {"name": "Rifle", "is_archived": False},
+            ],
+            [
+                {"name": "Rifle", "is_archived": None},
+                {"name": "Rifle updated", "is_archived": False},
+                {"name": "Rifle", "is_archived": False},
+            ],
+        ]
```
depsiatwal, line 618: Are we missing {"name": "Rifle", "is_archived": False}, ?

api/goods/views.py
```diff
@@ -194,6 +195,19 @@ def get_queryset(self):
...
+            organisation=organisation,
```
kevincarrogan, line 206: This looks like a fairly common pattern to filter by organisation. I wonder if extracting this out as a DRF FilterBackend would be useful.

api/goods/tests/test_edit.py
saruniitr: Actually for this test it doesn't really matter.
`is_archived = False` is for restoring case.

The main intention of this test is to ensure only `is_archived` is updated to either `True/False` and other fields remain unmodified.

depsiatwal (approved):
lgtm

api/goods/views.py
saruniitr: organisation id is not part of the url so more changes required, ok to take this up as a follow up?

PR #2073, opened 2024-07-03
LTD-5002-denial-matches-not-showing 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5002

This fixes an issue relates to some users unable to see end_user (denial matches) in the make recommendation tab. 
It was originally an intermittent issue. 

Turns out this is now permanently broken as a result of the denial changes.  The Link on the make recommendation page uses reference and if this is no present or blank then the link is invisible to the user. 

The fix here is to include the new unique reference regime_reg_ref field and use this to display a hyperlink in the UI. 

I have added tests in the UI also to check this is displayed as expected. 

API change is here 

These can be merged independently with no impact but the change will only work when both PRs have been merged.


api/cases/tests/test_case_search.py
```diff
@@ -1093,6 +1093,7 @@ def test_api_success(self):
                     "reference": self.denial_entity.denial.reference,
                     "category": self.denial_on_application.category,
                     "address": self.denial_entity.address,
+                    "regime_reg_ref": self.denial_on_application.denial_entity.denial.regime_reg_ref,
```
saruniitr, line 1096: This works but I think the right place is the case details view because the 'recommendation and decisions' tab get the case details.

depsiatwal: This was added due to break tests as they've been added to case details.

depsiatwal: Also added a new test to check in case details.

saruniitr (approved):

api/cases/tests/test_case_search.py
currycoder: What actually is `regime_reg_ref`?  It's not immediately obvious to me from the name..

api/cases/tests/test_get_case.py
```diff
@@ -399,3 +400,29 @@ def test_cases_endpoint_with_sub_status(self):
...
+        self.assertEqual(len(response_data["case"]["data"]["denial_matches"]), 1)
+        actual_denial_match = response_data["case"]["data"]["denial_matches"][0]
+        self.assertEqual(str(denial_match.denial_entity.id), actual_denial_match["denial_entity"]["id"])
+        self.assertEqual(str(denial_match.denial_entity.name), actual_denial_match["denial_entity"]["name"])
+        self.assertEqual(str(denial_match.denial_entity.address), actual_denial_match["denial_entity"]["address"])
+        self.assertEqual(
+            str(denial_match.denial_entity.denial.regime_reg_ref),
+            actual_denial_match["denial_entity"]["regime_reg_ref"],
+        )
+        self.assertEqual(str(denial_match.category), actual_denial_match["category"])
```
currycoder, line 428: Rather than all the individual `self.assertEqual` calls for each field in the response, could it be possible to assert that the whole value is as expected?

e.g.
```
self.assertEqual(response_data["case"]["data"]["denial_matches"], expected_denial_matches)
```

This kind of thing would make the test better cover the response format and so break immediately if we have future additions that aren't covered here.

api/cases/tests/test_case_search.py
depsiatwal: regime_reg_ref 
is the unique reference number for a denial (regime_regulator_reference)

api/cases/tests/test_get_case.py
depsiatwal: Good idea Done

currycoder (approved):

PR #2074, opened 2024-07-03
LTD: Add factory to create a draft application

saruniitr:
## Change description

The idea is to replace create_draft_standard_application() with this factory as that is heavily overloaded with many things.


api/applications/tests/factories.py
```diff
@@ -210,3 +212,29 @@ class SanctionMatchFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = SanctionMatch
+
+
+class DraftStandardApplicationFactory(StandardApplicationFactory):
```
depsiatwal, line 217: good idea reduced the coupling between our code as well

api/applications/tests/factories.py
```diff
@@ -210,3 +212,29 @@ class SanctionMatchFactory(factory.django.DjangoModelFactory):
...
+    def _create(cls, model_class, *args, **kwargs):
+        obj = model_class(*args, **kwargs)
+        obj.status = get_case_status_by_status(CaseStatusEnum.DRAFT)
+        obj.save()
+
+        GoodOnApplicationFactory(application=obj, good=GoodFactory(organisation=obj.organisation))
+
+        PartyOnApplicationFactory(application=obj, party=EndUserFactory(organisation=obj.organisation))
+
+        if kwargs["goods_recipients"] in [
```
depsiatwal, line 231: just wondering if we need this level of logic here or may be have another variation of the factory 
DraftStandardApplicationFactory

saruniitr: I think this is required to have a valid draft application.
I actually created this factory when reworking the validation parts.
Open to any alternatives to make this better.

api/applications/tests/test_view_application.py
```diff
@@ -31,7 +32,7 @@ def test_view_draft_standard_application_list_as_exporter_success(self):
         Ensure we can get a list of drafts.
         """
         self.exporter_user.set_role(self.organisation, self.exporter_super_user_role)
-        standard_application = self.create_draft_standard_application(self.organisation)
```
hnryjmes, line 34: is the intention to replace `create_draft_standard_application()` wherever it is used? it seems like there are 100s of usages of this function throughout lite-api

currycoder (approved):

api/applications/tests/test_view_application.py
saruniitr: yes but not immediately.
Initial goal is to stop using that and slowly replace as we go.

saruniitr: @hnryjmes you have any other comments on this?

api/applications/tests/factories.py
```diff
@@ -210,3 +212,29 @@ class SanctionMatchFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = SanctionMatch
+
+
+class DraftStandardApplicationFactory(StandardApplicationFactory):
+    goods_starting_point = StandardApplication.GB
```
hnryjmes, line 218: this is the same value as in the parent class, is this still needed?

api/applications/tests/factories.py
```diff
@@ -210,3 +212,29 @@ class SanctionMatchFactory(factory.django.DjangoModelFactory):
...
+    goods_recipients = StandardApplication.VIA_CONSIGNEE
```
hnryjmes, line 219: I think I might be missing some context here... why does a draft application imply that the goods recipients should be via consignee? Is it just the default that you chose here?

hnryjmes (approved):
the factory looks good to me. the default choice of VIA_CONSIGNEE I don't quite follow (how is that related to it being draft?), but this isn't an issue. given that you have that, the implementation of that using the other factories to create parties etc looks fine.

api/applications/tests/factories.py
saruniitr: Actually this was not set in parent class so added it here.

api/applications/tests/factories.py
saruniitr: correct, its just the default.
I wanted to have atleast end_user, consignee as defaults as opposed to just end_user.

saruniitr: btw this can be overridden using that kwarg in the tests.

saruniitr: @hnryjmes I am happy to change this to `DIRECT_TO_END_USER` if that is preferred?

hnryjmes: no that's ok, if it's just your choice of default because you think it's more  representative, that's definitely fine

PR #2077, opened 2024-07-03
Initiating major edits

kevincarrogan:
### Aim


api/staticdata/statuses/enums.py
```diff
@@ -174,6 +174,10 @@ def get_value(cls, status):
     def is_read_only(cls, status):
         return status in cls.read_only_statuses()
 
+    @classmethod
+    def is_editable(cls, status):
+        return not cls.is_read_only(status)
+
```
hnryjmes, line 180: is it your intention that this should be defined by comparison to the read only statuses, which are in turn defined by comparison to the writeable statuses? is there a reason to not just have `is_editable` be defined by the writeable statuses?

api/staticdata/statuses/models.py
```diff
@@ -41,6 +41,14 @@ def is_read_only(self):
     def is_read_only(cls):
         return Q(status__in=CaseStatusEnum.read_only_statuses())
 
+    @property
+    def is_major_editable(self):
+        return CaseStatusEnum.is_major_editable_status(self.status)
+
+    @property
+    def can_invoke_major_editable(self):
```
hnryjmes, line 49: I think naming this as `can_invoke_major_edit` is better, seeing the word "editable" here has been confusing in other places

kevincarrogan: I've been going around in circles about the naming of this. The problem with `can_invoke_major_edit` is it sounds like it's about making a major edit, which this isn't, it's about being able to put it into a state to be able to make a major edit.

If there's better wording for it then I'm all ears, but I don't think `can_invoke_major_edit` is it.

api/staticdata/statuses/enums.py
kevincarrogan: No real reason. It all gets to the same state. Overall I didn't necessarily want to have this `is_editable` at all but I found some code that wanted this `is_editable` and this state felt like it's the reverse of `is_read_only` and this code clarifies that quite literally, there's no room for messing this up as it's always the opposite of `is_read_only`.

hnryjmes (approved):
approving based on the conversation we are having in slack, and that this has been regression tested with `lite-frontend` at `master`

PR #2078, opened 2024-07-04
Allow good document stream for non draft statuses

depsiatwal:
### Aim

(slight rewording by @hnryjmes)

Currently, when a document for a good has been uploaded and the application is in draft status, the exporter is able to download and view the document. However once the application has been submitted, the good is also marked as submitted. So when the exporter goes into the product list they're unable to download and view the document. 

This PR shows that the issue can be resolved using the current API calls by removing the check for the draft status. However this would leave just a single check on that view which is that the good organisation id is equal to the request user organisation id.

Given that this check was there before we should be confident that removing the draft status check is safe, or if not, consider an alternative approach here like writing a new endpoint.

[LTD-5148](https://uktrade.atlassian.net/browse/LTD-5148)

[LTD-5148]: https://uktrade.atlassian.net/browse/LTD-5148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/goods/views.py
+++ b/api/goods/views.py
@@ -32,7 +32,6 @@ from api.goods.libraries.save_good import create_or_update_good
 from api.goods.models import Good, GoodDocument
 from api.goods.permissions import (
     IsDocumentInOrganisation,
-    IsGoodDraft,
 )
 from api.goods.serializers import (
     GoodAttachingSerializer,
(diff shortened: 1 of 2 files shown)


api/goods/views.py
```diff
@@ -478,16 +477,12 @@ def get(self, request, pk, doc_pk):
         """
         Returns a list of documents on the specified good
         """
+
         good = get_good(pk)
 
         if good.organisation_id != get_request_user_organisation_id(request):
```
hnryjmes, line 483: I think this is the check that I would expect for this view - that the good organisation id is the same as the request user's organisation id.

It seems sensible to me that an exporter user should be able to view documents associated with their own organisation and that this should not be forbidden if the case is in a non-draft state.

However this check must have been added for a reason and it's not clear to me what that original reason might have been.

hnryjmes (approved):

currycoder:
We discussed in slack that it would be good to convince ourselves that *only* exporters can add these types of documents before we remove this restriction.  e.g. to avoid a situation where exporters are able to see a document that they should not be able to.

hnryjmes:
@currycoder thanks Brendan, I have replied in slack to my most recent PR thread with some more relevant info that should hopefully help answer this

currycoder (approved):

saruniitr (approved):

PR #2081, opened 2024-07-05
[LTD-5152] Ensure that PartyOnApplication.clone creates a copy of the related Party

currycoder:
### Aim

The various model clone methods are currently used in the "amend by copy" feature.  Amend by copy ensures that when a user makes a major amendment to an application a new copy of an application is created and the original application put in to a superseded state.  We need to ensure that the application itself is copied and all related model records are also copied such that no records are re-used between the amendment application and the superseded application (with the exception of `Good` - which is used across multiple different applications).

To keep in line with this thinking, we need to also ensure that the new application has a full copy of any related `Party` records.

This change ensures that the `PartyOnApplication.clone()` method will associate the cloned PartyOnApplication object with a copy of the associated Party object.  The reason for this is to maintain the restriction where a single `Party` record is used for at most one application; a given `Party` record should not be reused for multiple applications.

[LTD-5152](https://uktrade.atlassian.net/browse/LTD-5152)

[LTD-5152]: https://uktrade.atlassian.net/browse/LTD-5152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/parties/tests/test_models.py
+++ b/api/parties/tests/test_models.py
@@ -1,0 +1,71 @@
+from django.forms import model_to_dict
+
+from api.flags.models import Flag
+from api.parties.tests.factories import PartyFactory
+from api.staticdata.countries.models import Country
+
+from test_helpers.clients import DataTestClient
+
+
+class TestParty(DataTestClient):
+
+    def test_clone(self):
+        original_party = PartyFactory(
+            name="some party",
+            address="123 fake st",
+            country=Country.objects.get(id="FR"),
+            website="https://www.example.com/foo",
+            signatory_name_euu="some signatory name",
+            type="end_user",
+            organisation=self.organisation,
+            role="intermediate_consignee",
+            role_other="some other role",
+            sub_type="government",
+            sub_type_other="some other sub type",
+            end_user_document_available=False,
+            end_user_document_missing_reason="some reason",
+            product_differences_note="some differences",
+            document_in_english=True,
+            document_on_letterhead=True,
+            ec3_missing_reason="some reason",
+            clearance_level="uk_official",
+            descriptors="some descriptors",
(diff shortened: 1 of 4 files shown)


api/parties/tests/test_models.py
```diff
@@ -0,0 +1,71 @@
...
+            copy_of=PartyFactory(),
+            phone_number="12345",
+            email="some.email@example.net",  # /PS-IGNORE
+            details="some details",
+        )
+        original_party.flags.add(Flag.objects.first())
+
+        cloned_party = original_party.clone()
+        assert original_party.id != cloned_party.id
+        assert model_to_dict(cloned_party) == {
```
saruniitr, line 42: Something to discuss with UCD.
The cloned party is identical to original party except the `id` so on the exporter side UI they both look the same without any differentiator. May be we need an option to archive parties.

saruniitr (approved):

api/parties/tests/test_models.py
```diff
@@ -0,0 +1,71 @@
...
+            "document_on_letterhead": True,
+            "ec3_missing_reason": "some reason",
+            "clearance_level": "uk_official",
+            "descriptors": "some descriptors",
+            "copy_of": original_party.id,
+            "phone_number": "12345",
+            "email": "some.email@example.net",  # /PS-IGNORE
+            "details": "some details",
+            "flags": [],
+        }, """
```
markj0hnst0n, line 67: Is it an idea to use a dict for this just to reduce the repetition?

markj0hnst0n (approved):

PR #2082, opened 2024-07-05
LTD-5167-turn-off-HMRC-cancellation-message-when-suspending-Licence

markj0hnst0n:
### Aim

LITE API currently sends a cancel message to HMRC via edifact file when a licence is suspended.  The purpose of this ticket is to remove that functionality.

This is already tested in the `api/licences/tests/test_api_to_hmrc_integration.py` file and continues to work as normal after removal based on these tests.

This only removes functionality going forward.  There are currently no Cases or Applications in PROD with this status.

[LTD-5167](https://uktrade.atlassian.net/browse/LTD-5167)

[LTD-5167]: https://uktrade.atlassian.net/browse/LTD-5167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/licences/enums.py
```diff
@@ -51,6 +51,5 @@ def to_str(cls, status):
     LicenceStatus.REINSTATED: HMRCIntegrationActionEnum.UPDATE,
     LicenceStatus.REVOKED: HMRCIntegrationActionEnum.CANCEL,
     LicenceStatus.SURRENDERED: HMRCIntegrationActionEnum.CANCEL,
-    LicenceStatus.SUSPENDED: HMRCIntegrationActionEnum.CANCEL,
```
saruniitr, line 54: This only removes the action mapping, do we need to remove below and the usages of it?
https://github.com/uktrade/lite-api/blob/dev/api/licences/models.py#L62

markj0hnst0n: A good point.  I've updated the model method of this now to reflect that there should be no message sent to HMRC.  While the initial change made sure there was no action to send a cancel message to HMRC stopping the initial firing of this is good to do also.

saruniitr: looks good but we really need tests

api/licences/tests/test_models.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def test_suspend_licence_cancel_message_not_sent_to_hmrc(self):
+
+        licence = Licence.objects.create(
+            case=self.standard_application,
+            status=LicenceStatus.ISSUED,
+            reference_code="GBSIEL/2024/0000001/P",
+            start_date=timezone.now().date(),
+            duration=10,
+        )
```
saruniitr, line 20: We have `StandardLicenceFactory()`

api/licences/tests/test_models.py
```diff
@@ -0,0 +1,25 @@
+from django.utils import timezone
+from test_helpers.clients import DataTestClient
+from api.licences.models import Licence
+from api.licences.enums import LicenceStatus, HMRCIntegrationActionEnum, licence_status_to_hmrc_integration_action
+
+
+class LicenceStatusTests(DataTestClient):
```
saruniitr, line 7: I was expecting some of the tests to fail as they all use suspended status also,
https://github.com/uktrade/lite-api/blob/dev/api/licences/tests/test_api_to_hmrc_integration.py#L61

Please update above tests as they all still use suspended status.
I would like to see tests where even if LITE tries to send message to `lite-hmrc` with suspend action then we detect and fail.

api/licences/models.py
```diff
@@ -59,7 +59,7 @@ def surrender(self, send_status_change_to_hmrc=True):
         self.status = LicenceStatus.SURRENDERED
         self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
 
-    def suspend(self, send_status_change_to_hmrc=True):
+    def suspend(self, send_status_change_to_hmrc=False):
```
hnryjmes, line 62: on L89 of this file we have
```
send_status_change_to_hmrc = kwargs.pop("send_status_change_to_hmrc", False)
```

so to make this action more explicit, I think we can go further than just changing `True` to `False` in this line 62, which is basically just changing the default value. probably you can just write  
```
def suspend(self):
    self.status = LicenceStatus.SUSPENDED
    self.save()
```

which should make it very clear that we don't want this to be overridden. that's just my opinion though, others might have a different one

hnryjmes: actually yeah, Arun below is saying that we should have a test that ensures that, if a message is sent to lite-hmrc for the suspend action we detect and fail, so probably ok to make the change I was talking about above

api/licences/tests/test_models.py
markj0hnst0n: I've added another test now and changed the suspend method

api/licences/celery_tasks.py
```diff
@@ -45,6 +45,11 @@ def send_licence_details_to_lite_hmrc(licence_id, action):
 
 def schedule_licence_details_to_lite_hmrc(licence_id, action):
     licence = Licence.objects.get(id=licence_id)
+
+    if licence.status == LicenceStatus.SUSPENDED and action == HMRCIntegrationActionEnum.CANCEL:
```
saruniitr, line 49: I would just check for `licence.status == LicenceStatus.SUSPENDED` and remove the action check completely because we removed suspended from the action map and also no other action makes sense for suspended licences.
We can update if things change in future.

api/licences/tests/test_models.py
```diff
@@ -0,0 +1,16 @@
...
+    def setUp(self):
+        super().setUp()
+        self.standard_application = self.create_standard_application_case(self.organisation)
+
+    def test_suspend_licence_cancel_message_not_sent_to_hmrc(self):
+        licence = StandardLicenceFactory(case=self.standard_application, status=LicenceStatus.ISSUED)
+        licence.suspend()
+
+        assert licence.status == LicenceStatus.SUSPENDED
+        assert licence_status_to_hmrc_integration_action.get(licence.status) != HMRCIntegrationActionEnum.CANCEL
```
saruniitr, line 16: Could this be update to below? as that is what you get now correct?
`assert licence_status_to_hmrc_integration_action.get(licence.status) is None`

saruniitr (commented):
Thanks for your patience.
Few last comments and rest all looks good to me.

api/licences/tests/test_models.py
markj0hnst0n: Changes made.  No thanks for your patience and help with all of this @saruniitr !

saruniitr (approved):

api/licences/celery_tasks.py
```diff
@@ -45,6 +45,11 @@ def send_licence_details_to_lite_hmrc(licence_id, action):
 
 def schedule_licence_details_to_lite_hmrc(licence_id, action):
     licence = Licence.objects.get(id=licence_id)
+
+    if licence.status == LicenceStatus.SUSPENDED:
+        logger.info("Licence %s is suspended, so cancel message should not be sent to HMRC", licence.reference_code)
```
depsiatwal, line 50: As we now send no message to HMRC I think it might be confusing to mention "so cancel message should not be sent to HMRC" Just because that's what we use to do. 

Hence I would right something in the lines of. 

Licence %s is suspended, Status not supported by HMRC so no message will be sent. 
This will allow future devs to know why we not sending anything.

markj0hnst0n: change made

depsiatwal (approved):
lgtm

PR #2087, opened 2024-07-08
[LTD-5078] Amend by copy validation

currycoder:
### Aim

This change adds validation to the new "create amendment" endpoint to ensure the following;
- The application to be amended is in a status that allows major edits to be invoked.
- The application to be amended has no non-draft Licence records associated to it.  If there exists an issued/cancelled/revoked/... licence for this application, the API endpoint responds with a 400 and does not allow the user to proceed.  We are anticipating that this will be quite rare and for now we will handle amending of applications which have licences by some other means.

[LTD-5078](https://uktrade.atlassian.net/browse/LTD-5078)

[LTD-5078]: https://uktrade.atlassian.net/browse/LTD-5078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/views/amendments.py
+++ b/api/applications/views/amendments.py
@@ -36,6 +36,11 @@ class CreateApplicationAmendment(CreateAPIView):
         return self.application.organisation
 
     def perform_create(self, serializer):
+        # Create a clone of the application in question and set the original application
+        # to a superseded status.  Amendment applications are new copies of the original
+        # which can be edited by exporters as if they were a completely new draft.
+        # Caseworker commentary is not copied to the amendment application but persists
+        # on the old superseded application
         self.amendment_application = self.application.create_amendment(self.request.user)
 
     @application_can_invoke_major_edit
--- a/api/applications/views/amendments.py
+++ b/api/applications/views/amendments.py
@@ -4,10 +4,12 @@ from rest_framework.generics import CreateAPIView
 from rest_framework import serializers
 from rest_framework import status
 
-from api.core.authentication import ExporterAuthentication
-from api.core.exceptions import NotFoundError
-from api.core.permissions import IsExporterInOrganisation
 from api.applications.libraries.get_applications import get_application
+from api.core.authentication import ExporterAuthentication
+from api.core.decorators import application_can_invoke_major_edit
+from api.core.exceptions import NotFoundError, BadRequestError
+from api.core.permissions import IsExporterInOrganisation
+from api.licences.models import Licence
 
 
 class AmendmentSerializer(serializers.Serializer):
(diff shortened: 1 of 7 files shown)


api/licences/managers.py
```diff
@@ -22,3 +22,6 @@ def get_draft_or_active_licence(self, application):
                 return self.get_active_licence(application)
             except self.model.DoesNotExist:
                 return None
+
+    def filter_non_draft_licences(self, application):
+        return self.filter(case=application).exclude(status=LicenceStatus.DRAFT)
```
saruniitr, line 27: Are you being cautious not to exclude revoked, exhaused licences?

api/applications/views/tests/test_amendments.py
```diff
@@ -51,3 +55,17 @@ def test_create_amendment_application_does_not_exist(self):
         self.application.delete()
         response = self.client.post(self.url, **self.exporter_headers)
         self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+
+    def test_create_amendment_application_wrong_status(self):
+        self.application.status = CaseStatus.objects.get(status="ogd_advice")
```
saruniitr, line 60: Is it worth parameterizing to few other invalid statuses as well.
I think atleast adding `under_final_review` is better where it gets finalised.

saruniitr (commented):
looks good to me, added few questions.

api/licences/managers.py
currycoder: Overall in this PR, I wanted to get to a place where major amendments could only be initiated for cases which have *never* had a Licence issued.  This is probably being quite cautious, but I felt like the best way to do that was to consider Licence record which had a status beyond "draft".

What do you think?  Would be very happy to be convinced if you feel we could consider cases with Licences in other statuses

saruniitr: No I think only excluding draft is sufficient for now.
If Licence is already exhausted hopefully Case officers won't allow it to re-open.

saruniitr (approved):

api/applications/views/amendments.py
```diff
@@ -29,15 +31,21 @@ def setup(self, request, *args, **kwargs):
             self.application = get_application(pk=self.kwargs["pk"])
         except (ObjectDoesNotExist, NotFoundError):
             raise Http404()
-        # TODO: More validation to prevent creating amendment of something we should not
 
     def get_organisation(self):
         return self.application.organisation
 
     def perform_create(self, serializer):
         self.amendment_application = self.application.create_amendment(self.request.user)
```
depsiatwal, line 39: is it worth commenting here what create_amendment use is ?
i.e does it create a new application with a ew version ?

depsiatwal (approved):
Been though it with a magnifier and all LGTM. 
Love the decorators BTW. 

One small comment about adding a comment but no biggie.

PR #2089, opened 2024-07-11
Changes to allow deployment of the Dev branch to AWS using platform tools

garethpn:
Configure application to run on DBT Platform while being backwards compatible with GOVPaas.


api/documents/libraries/s3_operations.py
```diff
@@ -23,14 +24,23 @@ def init_s3_client():
...
-        aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
-        region_name=settings.AWS_REGION,
-        config=Config(connect_timeout=settings.S3_CONNECT_TIMEOUT, read_timeout=settings.S3_REQUEST_TIMEOUT),
-        **additional_s3_params,
-    )
+
+    if is_copilot():
+        _client = boto3.client(
+            "s3",
+            region_name="eu-west-2",
```
depsiatwal, line 31: why we removing injecting the Key /ID here when instantiating boto client ?
Also concerned that were hard coding the region here

james-francis-MT: In aws the boto3 client uses whatever role the thing that is instantiating it is using so need need for access key and id

We could change the region to be set somewhere though

.copilot/phases/pre_build.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+# Exit early if something goes wrong
+set -e
+
+update_git_submodules() {
+  git_clone_base_url="https://codestar-connections.eu-west-2.amazonaws.com/git-http/730335529260/eu-west-2/192881c6-e3f2-41a9-9dcb-fcc87d8b90be/uktrade"
```
hnryjmes, line 7: is it ok for this to be in a public git repo?

depsiatwal: good point we need to remove this

PR #2090, opened 2024-07-11
[LTD-5105] Amend by copy document copying

currycoder:
### Aim

This PR ensures that documents are cloned appropriately during the "amend by copy" flow.

In particular this;
- Ensures that `PartyDocument` records are copied instead of shared between cloned `Party` records.
- Ensures that `ApplicationDocument` records are also cloned when an application is cloned.  This will cover firearm dealer certificates, section 5 certificates, supporting documents etc.

[LTD-5105](https://uktrade.atlassian.net/browse/LTD-5105)

[LTD-5105]: https://uktrade.atlassian.net/browse/LTD-5105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/models.py
+++ b/api/applications/models.py
@@ -354,10 +354,9 @@ class StandardApplication(BaseApplication, Clonable):
     def clone(self, exclusions=None, **overrides):
         cloned_application = super().clone(exclusions=exclusions, **overrides)
 
-        # TODO: Figure out whether it is desirable to clone ApplicationDocument records
-        # application_documents = ApplicationDocument.objects.filter(application=self)
-        # for application_document in application_documents:
-        #    application_document.clone(application=cloned_application)
+        application_documents = ApplicationDocument.objects.filter(application=self)
+        for application_document in application_documents:
+            application_document.clone(application=cloned_application)
 
         site_on_applications = SiteOnApplication.objects.filter(application=self)
         for site_on_application in site_on_applications:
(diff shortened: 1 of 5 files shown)


api/applications/models.py
```diff
@@ -354,10 +354,9 @@ class StandardApplication(BaseApplication, Clonable):
     def clone(self, exclusions=None, **overrides):
         cloned_application = super().clone(exclusions=exclusions, **overrides)
 
-        # TODO: Figure out whether it is desirable to clone ApplicationDocument records
-        # application_documents = ApplicationDocument.objects.filter(application=self)
-        # for application_document in application_documents:
-        #    application_document.clone(application=cloned_application)
+        application_documents = ApplicationDocument.objects.filter(application=self)
```
saruniitr, line 357: What do you think about filtering by `safe=True` here and in Party documents below?
All of these very likely to be safe otherwise they would've failed to upload but making it explicit while copying also.

currycoder: That's an excellent idea Arun!  Will make these changes.

depsiatwal:
Does is this make another copy of the S3 document or just a model copy. 
Just not spotting S3 copy anywhere ?

currycoder:
> Does is this make another copy of the S3 document or just a model copy. Just not spotting S3 copy anywhere ?

This doesn't copy the S3 document at all, we *only* copy our model.  I've tried to make this explicit with a comment in the document model to express that one raw S3 document may have multiple model copies associated to it.

We had a slack thread discussing this also here; https://ditdigitalteam.slack.com/archives/C06U7LD9T61/p1720616575793469

depsiatwal (approved):
lgtm

PR #2096, opened 2024-07-17
LTD 5213 email notification when a license is suspended

markj0hnst0n:
### Aim

Send notify email to exporter via notify template when Licence is suspended via change in case status.

[LTD-5213](https://uktrade.atlassian.net/browse/LTD-5213)

[LTD-5213]: https://uktrade.atlassian.net/browse/LTD-5213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gov_notify/payloads.py
```diff
@@ -57,6 +57,12 @@ class ExporterLicenceRevoked(EmailData):
     exporter_frontend_url: str
 
 
+@dataclass(frozen=True)
+class ExporterLicenceSuspended(EmailData):
+    user_first_name: str
+    application_reference: str
```
saruniitr, line 63: Could you clarify what this represents?
Is this the case reference code (GBSIEL/...) or the reference name that exporters enter when they create an application (text reference)?
Looks like the template assumes that licence reference_code is same as case reference_code.

From the template subject it appears that it refers to Licence `reference_code`
Looking at the tests it is the case `reference_code` but note that the Licence `reference_code` can be different to case reference code in some cases.

Also a case can have multiple licences (although only one is active) so I think we need to be clear on which licence is being suspended.

gov_notify/payloads.py
```diff
@@ -57,6 +57,12 @@ class ExporterLicenceRevoked(EmailData):
     exporter_frontend_url: str
 
 
+@dataclass(frozen=True)
+class ExporterLicenceSuspended(EmailData):
```
saruniitr, line 61: The template says we inform HMRC but I think we made changes not to send that message correct?
So will that be done separately by other means?

depsiatwal (approved):
LGTM pretty much spot on

gov_notify/payloads.py
markj0hnst0n: That is correct those changes have been merged to dev.  I will go back to UCD and query this as the text has been signed off by the business.

gov_notify/payloads.py
markj0hnst0n: These are good points.  Just to update.  I've change the notify function to explicitly reference the `licence.reference_code` as opposed to the case/app reference to be more specific.  I've asked UCD for clarity on what reference to present to the exporter so there could be further changes

gov_notify/payloads.py
markj0hnst0n: It has been confirmed that LU send a manual email to HMRC who then run a manual script at their end to suspend the licences on HMRC so that portion of the template is valid as it stands.

gov_notify/payloads.py
markj0hnst0n: I've confirmed that the licence_reference is the correct information to present on the template so I've renamed the variable to something more appropriate in the application and on the template.

api/cases/notify.py
```diff
@@ -82,6 +83,26 @@ def notify_exporter_licence_revoked(licence):
...
+    )
+
+
+def notify_exporter_licence_suspended(licence):
+    exporter = licence.case.submitted_by
+    _notify_exporter_licence_suspended(
+        exporter.email,
+        {
+            "user_first_name": exporter.first_name,
+            "license_reference": licence.reference_code,
```
saruniitr, line 101: there is a typo here, `licence_reference`

api/cases/tests/test_grant_licence.py
```diff
@@ -133,3 +133,46 @@ def test_grant_standard_application_licence_and_revoke(
...
+        self.assertEqual(self.standard_case.status, CaseStatus.objects.get(status=CaseStatusEnum.FINALISED))
+        for document in GeneratedCaseDocument.objects.filter(advice_type__isnull=False):
+            self.assertTrue(document.visible_to_exporter)
+
+        send_exporter_notifications_func.assert_called()
+        mock_notify_licence_issue.assert_called_with(self.standard_case.get_case())
+
+        self.change_status_url = reverse("applications:manage_status", kwargs={"pk": self.standard_case.id})
+        data = {"status": CaseStatusEnum.SUSPENDED}
+        response = self.client.put(self.change_status_url, data=data, **self.gov_headers)
```
saruniitr, line 173: There is a thread about this in slack whether this is the way to suspend a licence.
My understanding based on that discussion is that this is going to be changed/revisited.
Changing the case status to change licence status doesn't look correct to me.

api/cases/notify.py
markj0hnst0n: corrected

api/licences/models.py
```diff
@@ -61,10 +62,12 @@ def surrender(self, send_status_change_to_hmrc=True):
 
     def suspend(self):
         self.status = LicenceStatus.SUSPENDED
+        notify_exporter_licence_suspended(self)
         self.save()
 
     def revoke(self, send_status_change_to_hmrc=True):
         self.status = LicenceStatus.REVOKED
+        notify_exporter_licence_revoked(self)
         self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
```
saruniitr, line 71: Please move sending notifications after save() method.
If we send notification and if there are any errors during save() then it is not consistent.

saruniitr (approved):
One last change and it looks good to me.

api/licences/models.py
markj0hnst0n: Thanks Arun!  This is now done.

PR #2099, opened 2024-07-19
LTD-5223-tab-finalised

seijihg:
https://uktrade.atlassian.net/browse/LTD-5223

Removes the `finalised` applications from the `submitted` tab.

### DIFF
--- a/api/applications/managers.py
+++ b/api/applications/managers.py
@@ -1,3 +1,5 @@
+from django.db.models import Q
+
 from model_utils.managers import InheritanceManager
 
 from api.staticdata.statuses.enums import CaseStatusEnum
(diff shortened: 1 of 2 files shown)


kevincarrogan (approved):

hnryjmes (commented):
I am happy with the new code changes and I have tested this locally and all seems to be working. However I would also like it if the corresponding test for viewing finalised applications was updated to be as specific, to be consistent with the test for viewing submitted applications. I am happy to make that code change myself

hnryjmes:
actually, please ignore the above comment, I misunderstood the setup of the test. I'm happy with the way it's currently written.

hnryjmes (approved):
I am happy with all code changes and have tested this locally which seems to be working fine.

PR #2100, opened 2024-07-22
LTD: Add management command to generate report of extant licences with destinations

saruniitr:
## Change description

Generate report of all extant licences along with the destinations on those licences.
The cases where all the products are NLR are filtered out.

### DIFF
--- a/api/support/management/commands/extant_licences.py
+++ b/api/support/management/commands/extant_licences.py
@@ -1,0 +1,93 @@
+import csv
+import logging
+import time
+
+from django.core.management.base import BaseCommand
+from django.db.models import Prefetch
+from rest_framework import serializers
+
+from api.applications.models import GoodOnApplication, PartyOnApplication
+from api.staticdata.statuses.enums import CaseStatusEnum
+from api.licences.enums import LicenceStatus
+from api.licences.models import Licence
+
+
+class PartyOnLicenceSerializer(serializers.ModelSerializer):
+    reference_code = serializers.CharField(source="application.reference_code")
+    licence_reference = serializers.SerializerMethodField()
+    party_type = serializers.CharField(source="party.type")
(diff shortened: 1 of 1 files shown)


api/support/management/commands/extant_licences.py
```diff
@@ -0,0 +1,93 @@
...
+                    to_attr="prefetched_licences",
+                    queryset=active_licences_qs,
+                )
+            )
+            .order_by("-application__reference_code")
+        )
+
+        return parties_qs
+
+    def write_to_csv(self, filename, data, headers):
```
depsiatwal, line 77: Just thinking this will end up on the server. 
How do we get the file to the user ?
Should it be going to S3 bucket ?

hnryjmes: I'm also curious about how the file would get to the user. Presumably it's going to be a small file size, and this command will only be used rarely?

saruniitr: this is clarified in a thread in slack.
This command is only meant for iterating the report and not the actual solution.

api/support/management/commands/extant_licences.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def get_parties_on_licences(self):
+        applications_with_controlled_goods = GoodOnApplication.objects.filter(
+            is_good_controlled=True,
+        ).values_list("application_id", flat=True)
+
+        # TODO: Ideally we need to filter by sub_status value of approved also
+        # but most of the old cases won't have this value.
+        active_licences_qs = Licence.objects.filter(
+            case_id__in=applications_with_controlled_goods,
```
hnryjmes, line 54: is the purpose of this to filter out NLR cases? I was wondering why just checking the licence status and case status would not be enough

PR #2105, opened 2024-07-25
LTD-5247-new-endpoint-view-license 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5247

This is to support the new feature of allowing License(s) to be updated directly from the UI. 

This is adding a new endpoint to allow govuser to retrieval the details  of a License based on id.  New Serialiser is used since it's very light weight. 

Also We will leverage this end point to allow updates to an individual License. 

Hence this will be followed by a series of PRs

### DIFF
--- a/api/licences/tests/test_get_licences.py
+++ b/api/licences/tests/test_get_licences.py
@@ -1,9 +1,10 @@
 from django.urls import reverse
 from rest_framework import status
 
+from api.applications.tests.factories import StandardApplicationFactory
 from api.cases.enums import CaseTypeEnum, AdviceType
 from api.cases.models import CaseType
-from api.cases.tests.factories import FinalAdviceFactory, GoodCountryDecisionFactory
+from api.cases.tests.factories import FinalAdviceFactory
 from api.licences.enums import LicenceStatus
 from api.licences.models import Licence
 from api.licences.tests.factories import GoodOnLicenceFactory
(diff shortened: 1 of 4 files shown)


api/licences/tests/test_get_licences.py
```diff
@@ -182,3 +183,32 @@ def test_filter_by_active_only(self):
         response_data = response.json()["results"]
 
         self.assertEqual(len(response_data), 0)
+
+
+class LicencesDetailsTests(DataTestClient):
+    def setUp(self):
+        super().setUp()
+        self.standard_application = StandardApplicationFactory()
```
depsiatwal, line 191: shifting to using new factories and asset methods.

api/licences/views/main.py
```diff
@@ -78,6 +83,15 @@ def get_queryset(self):
         return Licence.objects.filter(case__organisation_id=get_request_user_organisation_id(self.request))
 
 
+class LicenceDetails(RetrieveAPIView):
```
depsiatwal, line 86: This will be eventually be extended to RetrieveUpdate with appropriate permissions hence the new view.

api/licences/urls.py
```diff
@@ -7,6 +7,7 @@
 urlpatterns = [
     path("", main.Licences.as_view(), name="licences"),
     path("<uuid:pk>/", main.ViewLicence.as_view(), name="licence"),
+    path("license_details/<uuid:pk>/", main.LicenceDetails.as_view(), name="licence_details"),
```
hnryjmes, line 10: you've used the `license` spelling for the path here when you should use `licence` spelling instead I think

hnryjmes (approved):
one comment re spelling, but otherwise looks good to me. the explanation that this will eventually be extended was helpful for understanding the PR.

api/licences/urls.py
depsiatwal: ahh fixed

api/licences/tests/test_get_licences.py
```diff
@@ -182,3 +183,32 @@ def test_filter_by_active_only(self):
         response_data = response.json()["results"]
 
         self.assertEqual(len(response_data), 0)
+
+
+class LicencesDetailsTests(DataTestClient):
```
markj0hnst0n, line 188: Might be worth changing this to `LicenceDetailsTests` just to have the uniform naming convention but happy to approve anyway as this is only a semantic change

markj0hnst0n (approved):

api/licences/tests/test_get_licences.py
depsiatwal: yeah can change in the next PR which adds on to this one

PR #2107, opened 2024-07-25
Ltd-5135: Applications archived tab api changes

saruniitr:
## Change description

Update Applications list endpoint to filter archived applications

Refactored method to be able to filter queryset better for more options in future.
Instead of adding more business logic in the view add filter backends to handle the filtering.

Frontend changes are here https://github.com/uktrade/lite-frontend/pull/2081

[LTD-5135](https://uktrade.atlassian.net/browse/LTD-5135)

[LTD-5135]: https://uktrade.atlassian.net/browse/LTD-5135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/applications/views/filters.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+        queryset = queryset.exclude(id__in=disallowed_applications)
+
+        return queryset
+
+
+class ApplicationStateFilter(filters.BaseFilterBackend):
+    FILTER_STATUS_MAP = {
+        "draft_tab": CaseStatus.objects.filter(status=CaseStatusEnum.DRAFT),
+        "draft_applications": CaseStatus.objects.filter(status=CaseStatusEnum.DRAFT),
```
saruniitr, line 26: For backward compatibility.
`draft_tab` can be removed once FE PR is merged.

api/applications/views/filters.py
```diff
@@ -0,0 +1,47 @@
...
+        "submitted_applications": CaseStatus.objects.exclude(
+            status__in=[
+                CaseStatusEnum.DRAFT,
+                CaseStatusEnum.FINALISED,
+                CaseStatusEnum.SUPERSEDED_BY_EXPORTER_EDIT,
+            ]
+        ),
+        "finalised_applications": CaseStatus.objects.filter(status=CaseStatusEnum.FINALISED),
+        "archived_applications": CaseStatus.objects.filter(status=CaseStatusEnum.SUPERSEDED_BY_EXPORTER_EDIT),
```
saruniitr, line 35: This can include other statuses going forward (eg withdrawn)

kevincarrogan (commented):
Now that we've extracted out this logic as filters I think we should have tests that check these filters in isolation outside of just this view.

api/applications/views/filters.py
```diff
@@ -0,0 +1,47 @@
...
+
+class ApplicationSiteFilter(filters.BaseFilterBackend):
+
+    def filter_queryset(self, request, queryset, view):
+        organisation = get_request_user_organisation(request)
+        users_sites = Site.objects.get_by_user_and_organisation(request.user.exporteruser, organisation)
+        disallowed_applications = SiteOnApplication.objects.exclude(site__id__in=users_sites).values_list(
+            "application", flat=True
+        )
+        queryset = queryset.exclude(id__in=disallowed_applications)
```
currycoder, line 18: The logic here does feel a bit backwards to me...  Get all the applications which *do not* include the users' site; exclude all of those applications.

Can the same be achieved with `SiteOnApplication.objects.filter()` instead...?  I feel like it could even be in a single query too like

```
queryset = queryset.filter(application_sites__in=users_sites)
```

api/applications/views/filters.py
```diff
@@ -0,0 +1,47 @@
...
+
+        return queryset
+
+
+class ApplicationStateFilter(filters.BaseFilterBackend):
+    FILTER_STATUS_MAP = {
+        "draft_tab": CaseStatus.objects.filter(status=CaseStatusEnum.DRAFT),
```
currycoder, line 25: I don't *think* this would cause a problem, but it does strike me a little odd kicking off a queryset at the class level (i.e. when we import the python file), rather than at runtime...  Maybe it could be good to pop this in to a method?

api/applications/views/filters.py
```diff
@@ -0,0 +1,47 @@
...
+        "archived_applications": CaseStatus.objects.filter(status=CaseStatusEnum.SUPERSEDED_BY_EXPORTER_EDIT),
+    }
+
+    def filter_queryset(self, request, queryset, view):
+        selected_filter = request.GET.get("selected_filter", "submitted_applications")
+        sort_by = request.GET.get("sort_by", "-submitted_at")
+
+        organisation = get_request_user_organisation(request)
+
+        statuses = self.FILTER_STATUS_MAP[selected_filter]
```
currycoder, line 44: This will raise a KeyError - i.e. 500 if the selected filter isn't a key - do we want to respond 400 instead?  Or even select some default..?

api/applications/views/filters.py
saruniitr: Yeah objects will be available at import time so no issues.
I have also deployed this to a dev env to verify this and there were no issues.

I can move this to a method.

api/applications/views/filters.py
saruniitr: Actually default is already set based on query params above which is `submitted_applications`

api/applications/views/filters.py
saruniitr: I have modified this as below to remove double exclude. Because we need to check if user has admin access for all sites we need few more changes. However I think this exposed a bigger issue.
With below changes the `SiteOnApplication.objects.filter()` was `[]` and so returning empty list of applications. It looks like we are not creating `SiteOnApplication` instances at all.

I tried creating an application locally and submit and there was no instance of `SiteOnApplication` for that application.

This needs further investigation as a follow-up so keeping the changes as they are now.

@currycoder thoughts?

```
+        try:
+            user_org = UserOrganisationRelationship.objects.get(
+                user=request.user.exporteruser, organisation=organisation
+            )
+        except UserOrganisationRelationship.DoesNotExist:
+            raise Http404
+
+        # sites specific to this user
+        user_sites = user_org.sites.all()
+
+        # if user is admin then all sites of organisation
+        admin_sites = Permission.objects.filter(id__in=[ExporterPermissions.ADMINISTER_SITES.name])
+        if user_org.role.permissions.intersection(admin_sites):
+            user_sites = organisation.site.all()
+
+        application_ids = SiteOnApplication.objects.filter(site__in=user_sites).values_list("application", flat=True)
+        queryset = queryset.filter(id__in=application_ids)
```

currycoder: That snippet looks great to me.

Agreed that further investigation should be a follow-up though

currycoder: Actually one recommendation for the snippet; I think you don't need to get the `application_ids` first and it can be a single queryset filter like my original comment

currycoder (approved):

api/applications/views/filters.py
saruniitr: we discussed offline and agreed to keep the code as is and fix this properly in the `SiteOnApplication` follow-up

api/applications/views/filters.py
```diff
@@ -0,0 +1,55 @@
...
+            "draft_applications": CaseStatus.objects.filter(status=CaseStatusEnum.DRAFT),
+            "submitted_applications": CaseStatus.objects.exclude(
+                status__in=[
+                    CaseStatusEnum.DRAFT,
+                    CaseStatusEnum.FINALISED,
+                    CaseStatusEnum.SUPERSEDED_BY_EXPORTER_EDIT,
+                ]
+            ),
+            "finalised_applications": CaseStatus.objects.filter(status=CaseStatusEnum.FINALISED),
+            "archived_applications": CaseStatus.objects.filter(status=CaseStatusEnum.SUPERSEDED_BY_EXPORTER_EDIT),
```
hnryjmes, line 38: ah I see, this is where we could add in more statuses like "withdrawn"

not sure why your previous comment here was removed Arun but I've seen it now

api/applications/views/filters.py
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        if selected_filter not in FILTER_STATUS_MAP:
+            selected_filter = "submitted_applications"
```
hnryjmes, line 42: is there test coverage for this? assuming that this is the fall-through option if for some reason `selected_filter` is not something in the `FILTER_STATUS_MAP`?

if it is, I wonder if the default should just be to show all applications, or if submitted is better, will leave that up to you

hnryjmes (approved):
looks good, have made a minor comment also. I pulled this branch and the frontend branch locally and had a small test of this also, I enabled the new flow for my organisation and went through the process until I could see the superseded application showing in the "Archived" tab. everything was working as expected.

depsiatwal (approved):
all looks very sensible

api/applications/views/filters.py
saruniitr: Yes there is a test for this in parameterized version, coverage also complained.
Default tab is submitted so following the same behaviour here.

PR #2112, opened 2024-07-30
Add and populate new UUID field on DenialReason objects

hnryjmes:
### Aim

This is part of the work to update the refusal criteria in LITE. For this work we have confirmed that we need to be able to store different versions of refusal criteria depending on policy changes at a particular point in time. Currently the DenialReason model uses a natural key and this makes it more difficult to store different versions of these objects. 

I have explained in more detail about the need for this change here: https://uktrade.atlassian.net/browse/LTD-5264

There are also examples of what this data looks like in both local and prod-like environments in the parent ticket here: https://uktrade.atlassian.net/browse/LTD-5164

For now, this PR just creates the UUID field and populates it, there will be other PRs to follow to change existing references, migrate existing cases, etc.

The existing seed script for denial reasons has also been updated to maintain functionality.

[LTD-5265](https://uktrade.atlassian.net/browse/LTD-5265)

[LTD-5265]: https://uktrade.atlassian.net/browse/LTD-5265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/staticdata/denial_reasons/migrations/0006_populate_uuid_field.py
```diff
@@ -0,0 +1,22 @@
+import uuid
+
+from django.db import migrations
+
+from api.staticdata.denial_reasons.constants import DENIAL_REASON_ID_TO_UUID_MAP
+
+
+def populate_uuid_field(apps, schema_editor):
+    DenialReason = apps.get_model("denial_reasons", "DenialReason")
+    for denial_reason in DenialReason.objects.all():
```
hnryjmes, line 10: I've been quite trusting of the data being in a particular state when writing this, only because this is our own reference data and we don't update it often

api/staticdata/management/commands/seeddenialreasons.py
```diff
@@ -1,9 +1,12 @@
+import uuid
+
 from django.db import transaction
 
 from api.staticdata.denial_reasons.models import DenialReason
 from api.staticdata.management.SeedCommand import SeedCommand
+from api.staticdata.denial_reasons.constants import DENIAL_REASON_ID_TO_UUID_MAP
 
-DENIAL_REASONS_FILE = "lite_content/lite_api/denial_reasons.csv"
+DENIAL_REASONS_FILE = "lite_content/lite_api/denial_reasons_update.csv"
```
hnryjmes, line 9: this is the most recent version of the denial reasons data, and it matches what is currently on our deployed envs. for some reason the seed script wasn't updated before to use this updated data, so I'm doing it now. 

eventually we will get rid of the seed scripts entirely (I think) so this is just to keep things working until we get around to doing everything via migrations.

api/staticdata/denial_reasons/migrations/0005_denialreason_uuid.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("denial_reasons", "0004_denial_reasons_update"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="denialreason",
+            name="uuid",
```
hnryjmes, line 16: eventually this can just be called `id` but for now it needs to not clash with the existing text field `id` so calling it `uuid` for now.

api/staticdata/denial_reasons/models.py
```diff
@@ -1,8 +1,11 @@
+import uuid
+
 from django.db import models
 
 
 class DenialReason(models.Model):
     id = models.TextField(primary_key=True, editable=False)
+    uuid = models.UUIDField(primary_key=False, default=uuid.uuid4, editable=False)
```
saruniitr, line 8: Will there be more changes later to make this as primary key?

saruniitr: Are we not ready to make this as primary key now itself?

hnryjmes: yes, there will be more PRs to come after this to do that change, plus make sure all Advice objects are properly migrated (Advice has `denial_reasons` which is a ManyToManyField linking to DenialReason).

for now this just adds the UUID field but it does nothing.

saruniitr (approved):

api/staticdata/denial_reasons/models.py
hnryjmes: > Are we not ready to make this as primary key now itself?

Sort of but I think I prefer separate PRs - once the primary key changes, we'll want to add some more tests to check the Advice objects have been migrated, and there are also impacts to data we send to data workspace. So preferred to keep it separate for ease of reviewing.

api/staticdata/denial_reasons/constants.py
```diff
@@ -0,0 +1,28 @@
...
+    "5": "73e063e8-ede0-43aa-b8c5-a9b7e546d67c",  # /PS-IGNORE
+    "5a": "49203257-8fca-4f7f-9f86-b8a6fa1dfa58",  # /PS-IGNORE
+    "5b": "aecd9fd0-287e-4b39-9f6a-4531a2e5848e",  # /PS-IGNORE
+    "5c": "913d7cb9-f04d-488a-b68c-554c8bfe6711",  # /PS-IGNORE
+    "6": "b6b0ce7c-8494-4c16-9dec-f2aa770d77ad",  # /PS-IGNORE
+    "6a": "df28a06c-c2c6-4a07-bd7e-1959bee1e7a8",  # /PS-IGNORE
+    "6b": "f45133c5-3695-4f12-a6c3-ae9897ce48d8",  # /PS-IGNORE
+    "7": "ec3794dc-6387-4e8a-bec8-293f1598a710",  # /PS-IGNORE
+    "8": "2674ef18-ff2b-4739-bde9-7b4a53e7aa49",  # /PS-IGNORE
+    "IRAN": "56b8649f-59d7-4fdf-848a-488b0cb4d52e",  # /PS-IGNORE
```
Tllew, line 24: Do we care that this identified IRAN as a denial reason?

Tllew (approved):
Looks good, in my head there's a better way to do this than a list of constants with UUIDs hardcoded, but I know we do this elsewhere, might be worth confirming we don't mind IRAN being shown as a denial reason

api/staticdata/denial_reasons/constants.py
hnryjmes: I've previously been assured that all refusal criteria are in the public domain, as they relate to published "written ministerial statements"

PR #2113, opened 2024-07-31
LTD-5249-licence-audit-table 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5249

Since License status can be edited directly we need a way of tracking these changes. 
Hence as per the goods/application models we have added the same pattern to Licence.


api/licences/models.py
```diff
@@ -88,7 +93,10 @@ def issue(self, send_status_change_to_hmrc=True):
         self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
 
     def save(self, *args, **kwargs):
-        self.end_date = add_months(self.start_date, self.duration, "%Y-%m-%d")
+
+        self.end_date = datetime.strptime(
```
depsiatwal, line 97: Since we were returning a string for a datetime field this was causing serialisation errors when writing to Audit table. Hence this has been fixed by returning a date.

api/core/model_mixins.py
```diff
@@ -42,3 +44,27 @@ class Trackable:
 
     def get_history(self, field):
         raise NotImplementedError()
+
+
+class TrackableMixin:
+
+    def get_history(self, field):
```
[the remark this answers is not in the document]
depsiatwal: Ahh think I misunderstood what it was trying to do reverted back now

depsiatwal:
> You have not added any tests

Add tests now to check to confirm revision.

depsiatwal:
As agreed the template change are being handled in this jira so we don't have change links switched. 
so the plan is 
1) Permissions Changes first 
2) Create views that utilise these permission and allows status change 
3) When we ready to test with uses add the link 
https://uktrade.atlassian.net/browse/LTD-5244

api/goods/tests/test_edit.py
```diff
@@ -595,7 +595,7 @@ def test_edit_archive_status(self):
             self.assertEqual(versions.count(), version_count)
             self.assertEqual(versions.first().field_dict["is_archived"], is_archived)
             # Ensure user is recorded in each revision
-            for version in versions:
+            for count, version in enumerate(versions, start=1):
```
saruniitr, line 598: guess this was missed?

api/licences/tests/test_get_licences.py
```diff
@@ -321,3 +322,21 @@ def test_update_license_details_case_licence_editable_states(self, licence_statu
 
         response = self.client.patch(self.url, self.status_data, **self.gov_headers)
         assert response.status_code == expected_status
+
+    def test_update_license_details_check_version(self):
+        data_items = [{"status": "reinstated"}, {"status": "suspended"}, {"status": "revoked"}]
+
```
saruniitr, line 328: Have commented this in other PR - I am not clear on the reasons for manually assigning the reinstated status.

api/licences/tests/test_get_licences.py
```diff
@@ -321,3 +322,21 @@ def test_update_license_details_case_licence_editable_states(self, licence_statu
...
+        response = self.client.patch(self.url, data_items[0], **self.gov_headers)
+        assert response.status_code == status.HTTP_200_OK
+        response = self.client.patch(self.url, data_items[1], **self.gov_headers)
+        assert response.status_code == status.HTTP_200_OK
+        response = self.client.patch(self.url, data_items[2], **self.gov_headers)
+        assert response.status_code == status.HTTP_200_OK
```
saruniitr, line 334: I think this test can be parameterized to read better.

Parameters can be, `status, expected_version_count, expected_status in most recent version`
The expected status in the most recent version can be compared with the `status`

```
versions = Version.objects.get_for_object(self.standard_application_licence)
self.assertEqual(versions.count(), expected_version_count).  # this will be initially 1 but increments with each run

self.assertEqual(versions.last().revision.user, self.gov_user.baseuser_ptr)
self.assertEqual(versions.last().field_dict["status"], status)
```

api/licences/tests/test_get_licences.py
depsiatwal: As per comments on that PR I have added below hopefully clarifies.  

From Front end some (not all) License actions are being ported to the Licence screen.

If a Licence is in Issued State then the user can Revoke/Suspend a License
If a License is in Suspend State then the user can set the Licence state to reinstated (this is to differentiation against Issued). The reason why we have a new method for this is because the action is being down against the licence as opposed to the Case. When we find latest License and the issue method changed it to Reinstate where application. This is an implicit action against a particular Licence.

api/licences/tests/test_get_licences.py
depsiatwal: The issue with running as a parameterized test is that each run of the test will reset the data. 
So you only get 1 version and your only testing for the latest change. 

The test above will check the historical 3 versions.

api/goods/tests/test_edit.py
depsiatwal: ahh reverted back

api/licences/tests/test_get_licences.py
depsiatwal: In addition the Audit Table is kind of independent with the API changes here. 
This is more about adding an Audit/Version to License and i'm using status to test this works correctly.

hnryjmes (approved):
happy with code changes, explanations in the comments make sense to me

api/licences/tests/test_get_licences.py
```diff
@@ -321,3 +322,21 @@ def test_update_license_details_case_licence_editable_states(self, licence_statu
...
+
+        # returns most recent first
+        versions = Version.objects.get_for_object(self.standard_application_licence)
+        self.assertEqual(versions.count(), 3)
```
saruniitr, line 338: ok in that can this be changed to `len(data_items)` and below also replace `3` with the length?

saruniitr (approved):
one last minor comment which I will leave it with you.

PR #2114, opened 2024-07-31
present case status in licence details

markj0hnst0n:
### Aim

### DIFF
--- a/api/licences/service.py
+++ b/api/licences/service.py
@@ -1,6 +1,7 @@
 from api.licences.enums import LicenceStatus
 from api.licences.models import GoodOnLicence, Licence
 from api.staticdata.control_list_entries.serializers import ControlListEntrySerializer
+from api.staticdata.statuses.serializers import CaseStatusSerializer
 
 from rest_framework import serializers
 
@@ -38,6 +39,7 @@ class LicenceSerializer(serializers.ModelSerializer):
 
     goods = GoodOnLicenceSerializer(many=True)
     status = serializers.SerializerMethodField()
+    case_status = serializers.SerializerMethodField()
 
     class Meta:
         model = Licence
@@ -45,8 +47,12 @@ class LicenceSerializer(serializers.ModelSerializer):
             "id",
             "reference_code",
             "status",
+            "case_status",
             "goods",
         )
+
+    def get_case_status(self, instance):
+        return CaseStatusSerializer.get_value(instance, instance.case.status)
 
     def get_status(self, instance):
         return LicenceStatus.to_str(instance.status)
@@ -55,9 +61,15 @@ class LicenceSerializer(serializers.ModelSerializer):
 def get_case_licences(case):
     licences = (
         Licence.objects.prefetch_related(
-            "goods", "goods__good", "goods__good__good", "goods__good__good__control_list_entries"
+            "goods",
+            "goods__good",
+            "goods__good__good",
+            "goods__good__good__control_list_entries",
+            "case",
+            "case__status",
         )
         .filter(case=case)
         .order_by("created_at")
     )
+
     return LicenceSerializer(licences, many=True).data


api/licences/tests/test_service.py
```diff
@@ -68,3 +69,11 @@ def test_get_application_licences_application_level_control_list_entry(self):
         self.assertEqual(data["goods"][0]["is_good_controlled"], False)
         # ignore order of control list entries
         self.assertEqual(set([x["rating"] for x in data["goods"][0]["control_list_entries"]]), {"ML1a", "ML13d1"})
+
+    def test_get_application_licences_case_status(self):
```
depsiatwal, line 73: I think we should test the actually CaseDetails URL explicitly to ensure we get what's returned on the licence object (s)

using this https://github.com/uktrade/lite-api/blob/dev/api/cases/tests/test_get_case.py#L151

api/licences/service.py
```diff
@@ -38,26 +39,37 @@ class LicenceSerializer(serializers.ModelSerializer):
...
 
 def get_case_licences(case):
     licences = (
         Licence.objects.prefetch_related(
-            "goods", "goods__good", "goods__good__good", "goods__good__good__control_list_entries"
+            "goods",
+            "goods__good",
+            "goods__good__good",
+            "goods__good__good__control_list_entries",
+            "case",
```
depsiatwal, line 68: wondering if we could do with out case in prefetch_related does it break ?

api/licences/service.py
```diff
@@ -38,26 +39,37 @@ class LicenceSerializer(serializers.ModelSerializer):
...
         fields = (
             "id",
             "reference_code",
             "status",
+            "case_status",
             "goods",
         )
 
+    def get_case_status(self, instance):
+        return CaseStatusSerializer.get_value(instance, instance.case.status)
```
depsiatwal, line 55: I think this should just return instance.case.status.status

api/licences/service.py
```diff
@@ -38,26 +39,37 @@ class LicenceSerializer(serializers.ModelSerializer):
 
     goods = GoodOnLicenceSerializer(many=True)
     status = serializers.SerializerMethodField()
+    case_status = serializers.SerializerMethodField()
```
saruniitr, line 42: You also need to get Case sub_status to determine if it is approval or refusal because in both case the status will be finalised.

markj0hnst0n: if it is refusal is it likely to have a licence to change the status of?

api/licences/service.py
markj0hnst0n: amended

api/licences/service.py
markj0hnst0n: it does not so I've removed

api/licences/service.py
depsiatwal: I don't think we're doing that check at the moment @saruniitr  it was even questionable if we need to check the state of the Licence. I thought we should hence why case_status has been added.

depsiatwal: ahh ok I see why we would need to check this also. 
@markj0hnst0n  might want to add sub_status to update the checks

saruniitr: Refusal won't have a licence. In this case sub status is not required as we are retrieving licence details, please ignore this.

api/licences/tests/test_service.py
```diff
@@ -68,3 +69,11 @@ def test_get_application_licences_application_level_control_list_entry(self):
...
+        self.licence.case.status = get_case_status_by_status(CaseStatusEnum.FINALISED)
+        self.licence.case.save()
+
+        data = get_case_licences(self.application)[0]
```
saruniitr, line 77: Could you also assert that there is only one licence?

saruniitr (approved):

depsiatwal (approved):
lgtm

PR #2115, opened 2024-07-31
LTD-5190: Include assessments from superseded cases in previous assessments

saruniitr:
## Change description

Consider assessments from superseded cases also to include in previous assessments.
When exporter amends an application then we create a new application by copy and mark the previous one as superseded.

Existing tests already cover this as they are written to cover any additions as well.
https://github.com/uktrade/lite-api/blob/dev/api/cases/tests/test_good_precedents_view.py#L101

[LTD-5190](https://uktrade.atlassian.net/browse/LTD-5190)

[LTD-5190]: https://uktrade.atlassian.net/browse/LTD-5190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


currycoder (commented):
Just looked at the old tests that we have for this and I think they are close. I would feel a little better about this though if we had a parameterized test case which showed that this works for each eligible status. It seems like the existing tests just create a stack of precedent GoodOnApplications rather than proving that a precedent in each specific status will work

saruniitr:
> Just looked at the old tests that we have for this and I think they are close. I would feel a little better about this though if we had a parameterized test case which showed that this works for each eligible status. It seems like the existing tests just create a stack of precedent GoodOnApplications rather than proving that a precedent in each specific status will work

@currycoder specific test added for this.

currycoder (approved):
The test you added is so good that it has me wondering if we should fully close the loop and with a test for the other side; for all non-precedent case statuses, can we prove no precedents will be returned?

Leaving this with you though, as it's a coverage enhancement rather than essential to me.

api/staticdata/statuses/enums.py
```diff
@@ -68,6 +68,11 @@ class CaseStatusEnum:
 
     compliance_visit_statuses = [OPEN, UNDER_INTERNAL_REVIEW, RETURN_TO_INSPECTOR, AWAITING_EXPORTER_RESPONSE, CLOSED]
 
+    non_precedent_statuses = [
+        SUBMITTED,
+        INITIAL_CHECKS,
+    ]
+
```
hnryjmes, line 75: is it possible to make this based off of `precedent_statuses` through a class method? like we have for read only statuses, something like:
```
@classmethod
def non_precedent_statuses(cls):
    return list(set(cls.all()) - set(cls.precedent_statuses))
```

or is it not appropriate to do that (e.g. we want precise control over non_precedent_statuses)?

saruniitr: @hnryjmes  Unfortunately that is not possible because all statuses include many other values as well.
We need list of all status that a case goes through and then we can subtract `precedent_statuses` from that list but don't want to rush it as part of this hence specified a list in the enum.
Initially I was planning to parameterize within the test itself but this atleast keeping it together with `precendent_statuses`.

hnryjmes (approved):

PR #2116, opened 2024-07-31
LTD-5248-update-license-endpoint-changes 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5255

This builds on the previous Update Licence API Endpoint. 

I have now added the supported actions against a License. 

The model support 3 actions and these in turn are called so that we send the appropriate AUDIT, EMAIL or HMRC Message.


api/licences/serializers/view_licence.py
```diff
@@ -213,6 +213,11 @@ def get_document(self, instance):
 
 
 class LicenceDetailsSerializer(serializers.ModelSerializer):
+    action_dict = {
+        "reinstated": lambda instance: Licence.reinstate(instance, send_status_change_to_hmrc=True),
```
saruniitr, line 217: Is this also a status that is manually assigned?
Currently when we are re-issuing an active licence then the current one gets cancelled and the new one is issued with this status.
If the flow is different now, under what conditions this will be assigned?

api/licences/models.py
```diff
@@ -74,6 +74,12 @@ def cancel(self, send_status_change_to_hmrc=True):
         self.status = LicenceStatus.CANCELLED
         self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
 
+    def reinstate(self, send_status_change_to_hmrc=True):
+        # This supersedes the issue method as it's called explicity on the license
+        # Hence the user explicty knows which license is being reinstated
+        self.status = LicenceStatus.REINSTATED
```
saruniitr, line 80: One more place where this status is assigned to this value (below in `issue()` method), will that also remain?

api/licences/serializers/view_licence.py
depsiatwal: From Front end some (not all) License actions are being ported to the Licence screen. 

If a Licence is in Issued State then the user can Revoke/Suspend a License 
If a License is in Suspend State then the user can set the Licence state to reinstated (this is to differentiation against Issued). The reason why we have a new method for this is because the action is being down against the licence as opposed to the Case. When we find latest License and the issue method changed it to Reinstate where application. This is an implicit action against a particular Licence.

api/licences/models.py
depsiatwal: For now the Issue method will remain in place for newly issue license. 
We are however removing Suspend/Revoked from the case. Hence not all states are being ported over in this scope.

api/licences/models.py
```diff
@@ -74,6 +74,12 @@ def cancel(self, send_status_change_to_hmrc=True):
         self.status = LicenceStatus.CANCELLED
         self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
 
+    def reinstate(self, send_status_change_to_hmrc=True):
```
saruniitr, line 77: As discussed, please remove or set this to `False` as we don't want to send message in this case.
Here as well as further below in `action_dict`.

If we send then the current licence gets cancelled.

depsiatwal: done

saruniitr (approved):

hnryjmes (approved):
happy with code changes

PR #2119, opened 2024-08-05
LTD-5267-template

seijihg:
https://uktrade.atlassian.net/browse/LTD-5267


assets/css/siel.css
```diff
@@ -3,12 +3,13 @@
   src: url("../fonts/Helvetica.ttf") format("truetype");
 }
 .document * {
-  font-size: 12px;
+  font-size: 10px;
   font-family: Helvetica;
   word-wrap: break-word;
+  word-break: break-word;
```
hnryjmes, line 9: MDN docs say that `break-word` is deprecated, do you know if that matters to us? is there an alternative?

api/letter_templates/templates/letter_templates/siel.html
```diff
@@ -264,7 +264,7 @@ <h1>Standard Individual Export Licence</h1>
       <tr>
         <td class="border-black" colspan="6">
           <span class="cell__heading">2b. Applicant reference</span>
-          <span class="cell__uppercase">{{ details.user_reference }}</span>
+          <span class="cell__uppercase width-text">{{ details.user_reference }}</span>
```
hnryjmes, line 267: it's interesting to me that we need this addition when we show the user reference here, but not in the other places in the document where we use
`<span class="cell__uppercase">{{ details.user_reference }}</span>`
and this doesn't seem to cause a problem (at least, looking at the example pdf attached to the ticket)

should we make these usages consistent? do we need to?

seijihg: other places have a lot more spaces only here we have a small cell; makes the issue with long reference. this limit the characters and wrap it below.

assets/css/siel.css
```diff
@@ -3,12 +3,12 @@
...
 }
 .document * {
-  font-size: 12px;
+  font-size: 11px;
   font-family: Helvetica;
   word-wrap: break-word;
 }
 .document h1 {
-  font-size: 24px;
+  font-size: 20px;
```
hnryjmes, line 11: can you confirm with someone from UCD that we're ok to make these changes?

hnryjmes (commented):
have tested the PR on dev2 and this does seem to fix the issue. however, I have made a comment about font size changes that I would prefer answered before approving

hnryjmes (approved):
approved as Jonathan (UCD) confirmed the changes were acceptable

assets/css/siel.css
```diff
@@ -49,6 +49,9 @@
   font-style: italic;
   margin-bottom: 5px;
 }
+.document .max-width {
+  max-width: 140px;
```
saruniitr, line 53: if the reference is wider than this then it wraps?

saruniitr (approved):

PR #2120, opened 2024-08-06
LTD-5242-add-change-status-view-API-change

depsiatwal:
This goes together with the new view to update licence details 
https://github.com/uktrade/lite-frontend/pull/2089/files

case_status is required to ensure we can edit the licence as this is part of the permission check. 

https://uktrade.atlassian.net/browse/LTD-5242


markj0hnst0n (approved):
looks grand as it builds on previous pattern

saruniitr (approved):

PR #2121, opened 2024-08-06
LTD-5083: Add endpoint to allow update of quantity value fields only

saruniitr:
## Change description

Add exporter specific endpoint to allow editing of quantity/value of goods added to an application.
This endpoint is restrictive and only quantity, value can be edited.

As we are moving towards separating exporter, caseworker urls a new directory is created with the intention of grouping
relevant views under that directory.

Permission checks are moved to their own custom permission classes.

Note: This ticket is mainly about auditing quantity and value changes but this PR doesn't add audit events.
Editing of quantity value was introduced as a minor edit earlier but now we have removed minor edits completely. To perform major edits application need to be in `draft` state at which point it is not associated to any `Case`. This is a problem for audit events because they are always associated to a `Case`. So even if we add audit entries when in `draft` state they don't show up under Notes and timeline.
Hence we agreed to add this exporter specific endpoint which is an improvement and revisit auditing later after discussing with UCD.

Relevant FE PR: https://github.com/uktrade/lite-frontend/pull/2092

[LTD-5083](https://uktrade.atlassian.net/browse/LTD-5083)

[LTD-5083]: https://uktrade.atlassian.net/browse/LTD-5083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/tests/test_edit_good_on_applications.py
+++ b/api/applications/tests/test_edit_good_on_applications.py
@@ -6,7 +6,7 @@ from rest_framework import status
 
 from api.applications.libraries.case_status_helpers import get_case_statuses
 from api.applications.enums import NSGListType
-from api.applications.tests.factories import GoodOnApplicationFactory
+from api.applications.tests.factories import DraftStandardApplicationFactory, GoodOnApplicationFactory
 from api.goods.tests.factories import GoodFactory, FirearmFactory
 from api.staticdata.statuses.enums import CaseStatusEnum
 from api.staticdata.statuses.libraries.get_case_status import get_case_status_by_status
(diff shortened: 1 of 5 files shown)


api/exporter/applications/views.py
```diff
@@ -0,0 +1,49 @@
...
+        IsApplicationEditable,
+    )
+
+    def patch(self, request, **kwargs):
+        good_on_application_pk = kwargs["good_on_application_pk"]
+        good_on_application = GoodOnApplication.objects.get(pk=good_on_application_pk)
+
+        data = request.data.copy()
+        serializer = self.serializer_class(instance=good_on_application, data=data, partial=True)
+        if not serializer.is_valid():
```
depsiatwal, line 41: In the sprit of keeping views lighter do we need all this code ?
can the serialiser no just handle the rest as it's UpdateAPIView

api/applications/tests/test_edit_good_on_applications.py
```diff
@@ -117,6 +129,84 @@ def test_edit_a_good_on_applicaton_invalid_organisation(self):
             original_calibre,
         )
 
+    @parameterized.expand(
+        [
+            [
+                {"quantity": 5, "value": 10.00},
```
depsiatwal, line 135: may be add some tests to ensure we can't edit any other goods fields.

saruniitr: That is already added, the next one,
https://github.com/uktrade/lite-api/pull/2121/files#diff-07df80ba2eab84af35b538fe0c8bc98f0cb421e03902b9fe829521999374f50bR182

depsiatwal: ah ok makes sense now

depsiatwal (approved):
approved

Tllew (approved):
Looks good to me

PR #2123, opened 2024-08-07
LTD-5115-nt-casenote-change 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5115

As part of the CIC work stream where LU users can now change the state of the licence. There is a requirement to add the an Audit for N&T the user that changed the state. 

Currently all Audit changes are created by a signal for the Licence object. As this is bad practice this will be removed.

The option where Licence state is changed by user we have the method will be overloaded to allow the user to be passed so that user audit logs can be created.


api/licences/tests/test_audit_licence.py
```diff
@@ -39,12 +39,11 @@ def test_update_existing_licence(self):
         )
         licence.status = LicenceStatus.REVOKED
         licence.save()
-
+        # Just one audit created since revoke should be called directly.
```
hnryjmes, line 42: if I am following correctly, this test now doesn't check that the "revoked" status appears in the audit records, it checks that does NOT happen. if so could the test be renamed so the purpose is more clear? the name at the moment is "test update existing licence"... I want to say, test that it does what?

api/licences/serializers/view_licence.py
```diff
@@ -237,12 +239,25 @@ class Meta:
...
+        # Create a user audit for N&T
+        request = self.context.get("request")
+        audit_trail_service.create(
+            actor=request.user,
+            verb=AuditType.LICENCE_UPDATED_STATUS,
+            action_object=instance,
+            target=instance.case.get_case(),
+            payload={"licence": instance.reference_code, "status": update_action, "previous_status": previous_status},
+        )
+
```
hnryjmes, line 260: if the ValidationError above is raised, does that mean this code is unreachable?

wondering if to make it clear when this is expected to run, this code could be moved into the try block somehow.

api/licences/models.py
```diff
@@ -60,9 +62,26 @@ def hmrc_mail_status(self):
...
+        try:
+            Licence.objects.get(id=self.id, status=self.status)
+        except Licence.DoesNotExist:
+            # Since we only want to add entries where previous status has changed
+            audit_trail_service.create_system_user_audit(
+                verb=AuditType.LICENCE_UPDATED_STATUS,
+                action_object=self,
+                target=self.case.get_case(),
+                payload={"licence": self.reference_code, "status": self.status},
+            )
```
hnryjmes, line 78: I'm not sure I follow here - why would we need to create the audit only when the licence does not exist?

hnryjmes (commented):
generally happy with the code changes, just left a few comments for you to consider

api/licences/models.py
depsiatwal: Because the only way to this code is when we interact with the Licence object directly. 
i.e Only the system is allowed to create Licence objects hence why we create a System Audit. 

When a user changed state then we want a user audit which is covered in the serializer

api/licences/tests/test_audit_licence.py
depsiatwal: renamed test hopefully a little clear now

api/licences/serializers/view_licence.py
depsiatwal: that's a very good point I have shifted this into the Try

api/audit_trail/formatters.py
```diff
@@ -228,9 +228,11 @@ def product_reviewed(**payload):
 def licence_status_updated(**payload):
     status = payload["status"].lower()
     licence = payload["licence"]
+    previous_status = payload.get("previous_status")
     if status == LicenceStatus.EXHAUSTED:
         return f"The products for licence {licence} were exported and the status set to '{status}'."
-
+    elif previous_status:
```
currycoder, line 234: This could just be `if` instead of `elif`, right?

api/licences/serializers/view_licence.py
```diff
@@ -237,12 +239,29 @@ class Meta:
...
+                actor=request.user,
+                verb=AuditType.LICENCE_UPDATED_STATUS,
+                action_object=instance,
+                target=instance.case.get_case(),
+                payload={
+                    "licence": instance.reference_code,
+                    "status": update_action,
+                    "previous_status": previous_status,
+                },
+            )
```
currycoder, line 259: This looks like there's a lot of overlap with `Licence.create_licence_change_system_audit_log()` - and in fact I think there's some actions where we end up with multiple audit logs from one endpoint call.

api/licences/models.py
```diff
@@ -60,9 +62,26 @@ def hmrc_mail_status(self):
...
+
     def surrender(self, send_status_change_to_hmrc=True):
         self.status = LicenceStatus.SURRENDERED
         self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
+        # Surrender is called from case status change so we log this as a system event
+        self.create_licence_change_system_audit_log()
```
currycoder, line 84: Could all of this duplication between these action methods be simplified like....

```python
def create_licence_change_audit_log(self, original_status=None, user=None):
    """
    Generates a system audit log for licence changes
    """
    audit_kwargs = {
        "verb": AuditType.LICENCE_UPDATED_STATUS,
        "action_object": self,
        "target": self.case.get_case(),
        "payload": {"licence": self.reference_code, "status": self.status, "previous_status": original_status},
    }
    if user:
        audit_trail_service.create(
            **audit_kwargs,
            actor=user,
        )
    else:
        audit_trail_service.create_system_user_audit(**audit_kwargs)

def _set_status(self, status, user=None, send_status_change_to_hmrc=True):
    original_status = self.status
    self.status = status
    status_changed = original_status != self.status
    if status_changed:
        self.create_licence_change_system_audit_log(original_status=original_status)

def surrender(self, user, send_status_change_to_hmrc=True):
    self._set_status(LicenceStatus.SURRENDERED, user, send_status_change_to_hmrc=send_status_change_to_hmrc)

....
```

api/licences/models.py
```diff
@@ -60,9 +62,26 @@ def hmrc_mail_status(self):
...
+        except Licence.DoesNotExist:
+            # Since we only want to add entries where previous status has changed
+            audit_trail_service.create_system_user_audit(
+                verb=AuditType.LICENCE_UPDATED_STATUS,
+                action_object=self,
+                target=self.case.get_case(),
+                payload={"licence": self.reference_code, "status": self.status},
+            )
+
     def surrender(self, send_status_change_to_hmrc=True):
```
currycoder, line 80: Should `send_status_change_to_hmrc` be a kwarg for this?  It kind of demands the callers to know whether this is something hmrc is interested in or not...  Instead this method could be the thing that knows and sets this on the save call, right?

api/audit_trail/formatters.py
depsiatwal: yeah changed

api/licences/models.py
```diff
@@ -60,31 +70,46 @@ def hmrc_mail_status(self):
...
-        self.status = LicenceStatus.REVOKED
-        self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
+    def revoke(self, user=None):
+        self._set_status(LicenceStatus.REVOKED, user=user, send_status_change_to_hmrc=True)
         notify_exporter_licence_revoked(self)
 
-    def cancel(self, send_status_change_to_hmrc=True):
-        self.status = LicenceStatus.CANCELLED
-        self.save(send_status_change_to_hmrc=send_status_change_to_hmrc)
+    def cancel(self, user=None, send_status_change_to_hmrc=True):
```
currycoder, line 102: Does this method and `issue()` still need to take `send_status_change_to_hmrc` as a kwarg or can it set that itself?

currycoder (approved):
One last comment which I will leave with you

api/licences/tests/test_get_licences.py
```diff
@@ -371,3 +374,24 @@ def test_update_license_details_check_version(self):
...
+            [LicenceStatus.REVOKED],
+        ]
+    )
+    def test_update_licence_details_audit_trail(self, licence_status):
+        data = {"status": licence_status}
+
+        response = self.client.patch(self.url, data, **self.gov_headers)
+
+        assert response.status_code == status.HTTP_200_OK
+        audit = Audit.objects.filter(verb=AuditType.LICENCE_UPDATED_STATUS).first()
```
currycoder, line 391: Rather than `.first()` could this just be `Audit.objects.get(verb=..)` - I think that should work and will also reassure that there's only one audit entry

depsiatwal: actually there's 2 audits here one for create and one for the last action in this case revoke.

hnryjmes (approved):

PR #2133, opened 2024-08-12
[LTD-5299] Prevent caseworkers setting superseded status

currycoder:
### Aim

This change ensures that caseworkers are unable to set the case status to "Superseded by exporter edit" via the API.

[LTD-5299](https://uktrade.atlassian.net/browse/LTD-5299)

[LTD-5299]: https://uktrade.atlassian.net/browse/LTD-5299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/caseworker/permissions.py
+++ b/api/applications/caseworker/permissions.py
@@ -21,6 +21,9 @@ class CaseStatusCaseworkerChangeable(permissions.BasePermission):
         if new_status == CaseStatusEnum.APPLICANT_EDITING:
             return False
 
+        if new_status == CaseStatusEnum.SUPERSEDED_BY_EXPORTER_EDIT:
+            return False
+
         if CaseStatusEnum.is_terminal(original_status) and not user.has_permission(GovPermissions.REOPEN_CLOSED_CASES):
             return False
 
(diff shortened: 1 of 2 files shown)


api/applications/caseworker/permissions.py
```diff
@@ -21,6 +21,9 @@ def has_object_permission(self, request, view, application):
         if new_status == CaseStatusEnum.APPLICANT_EDITING:
```
depsiatwal, line 21: feel like the if statements are getting messy 
may be for some of then even some config dict mapping of some sort. 
I know there's a note at the top is the plan to tidy these at a later state ?

dejandbt (approved):
LGTM

api/applications/caseworker/permissions.py
currycoder: Yes I'd really like to get all of this in to django-rules instead.  We've talked a bit recently about giving the status changes stuff some love with e.g. finite state machines and folding "who can change what status to what" in to that, so I think there will be some overlap there.

If you are happy, I will defer that thinking to when we pick that up?

depsiatwal (approved):
approved

PR #2135, opened 2024-08-13
LTD-5306 added reversion to parties

Tllew:
[LTD-5306](https://uktrade.atlassian.net/browse/LTD-5306)

[LTD-5306]: https://uktrade.atlassian.net/browse/LTD-5306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/parties/models.py
+++ b/api/parties/models.py
@@ -10,6 +10,7 @@ from api.goods.enums import PvGrading
 from api.organisations.models import Organisation
 from api.parties.enums import PartyType, SubType, PartyRole, PartyDocumentType
 from api.staticdata.countries.models import Country
+import reversion
 
 
 class PartyManager(models.Manager):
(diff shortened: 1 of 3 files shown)


markj0hnst0n (approved):

hnryjmes (changes requested):
please can you add tests

api/parties/models.py
```diff
@@ -45,6 +46,7 @@ def copy_detail(self, pk):
         return values
 
 
+@reversion.register()
 class Party(TimestampableModel, Clonable):
```
saruniitr, line 50: it needs tests to check versions are being created when we change object

hnryjmes (approved):
this looks much better with tests. the test coverage checks both that it is possible to create versions and that versions are created automatically when the Party object is updated.

PR #2138, opened 2024-08-14
Move API endpoints into separate version directories

kevincarrogan:
### Aim

Separate out versioned API endpoints into their own directories in preparation for adding a v2 set of endpoints.

[LTD-5314](https://uktrade.atlassian.net/browse/LTD-5314)

[LTD-5314]: https://uktrade.atlassian.net/browse/LTD-5314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/data_workspace/v0/tests/test_licence_views.py
```diff
@@ -0,0 +1,46 @@
...
+    @override_settings(HAWK_AUTHENTICATION_ENABLED=True)
+    def test_dw_view_licences(self):
+        sender = get_hawk_sender("GET", self.licences, None, settings.HAWK_LITE_DATA_WORKSPACE_CREDENTIALS)
+        self.client.credentials(HTTP_HAWK_AUTHENTICATION=sender.request_header, CONTENT_TYPE="application/json")
+        response = self.client.get(self.licences)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    @override_settings(HAWK_AUTHENTICATION_ENABLED=True)
+    def test_dw_view_licences_fail_incorrect_hawk_key(self):
+        sender = get_hawk_sender("GET", self.licences, None, "internal-frontend")
```
currycoder, line 29: Just so this is a little easier to scan, could I suggest `"internal-frontend"` becomes `"invalid-hawk-key"`?  Or is there something special about the string that it is right now?

currycoder (approved):

api/data_workspace/v1/licence_views.py
```diff
@@ -0,0 +1,21 @@
...
+    serializer_class = serializers.GoodOnLicenceReportsViewSerializer
+    pagination_class = LimitOffsetPagination
+    queryset = models.GoodOnLicence.objects.all()
+
+
+class LicencesList(viewsets.ReadOnlyModelViewSet):
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    serializer_class = LicenceWithoutGoodsSerializer
+    pagination_class = LimitOffsetPagination
+    queryset = models.Licence.objects.all()
```
saruniitr, line 21: Could we exclude `draft` licences?

saruniitr (approved):
All looks good to me.
Is it worth adding a comment about what each version means?

api/data_workspace/v0/tests/test_licence_views.py
kevincarrogan: This is how the test was before. All I did was move it.

I could make that change and see if it still passes though.

kevincarrogan:
> All looks good to me. Is it worth adding a comment about what each version means?

I'd need someone to tell me as these versions were already there, I just made the folder structure map to the two versions we already had.

api/data_workspace/v1/licence_views.py
kevincarrogan: This is what we currently push into Data Workspace, I'm not making any changes here, and I'm not sure if the current expectation is that we do or do not expose draft licences but I wasn't considering changing the current output of any of these endpoints.

PR #2144, opened 2024-08-19
LTD-5356: Display CLEs assessed for this application on the licence document

saruniitr:
## Change description

Display only the CLEs assessed for this application on the Licence

Currently we are using CLEs from Good when displaying them on the licence document. Usually this will be the same as recorded in GoodOnApplication but if the underlying good is re-used in other applications where it is assessed differently then they both will differ. In such cases we will be displaying these values from both the assessments.

Fix this by using the entries from GoodOnApplication.

Add test to ensure correct values are displayed.

[LTD-5356](https://uktrade.atlassian.net/browse/LTD-5356)

[LTD-5356]: https://uktrade.atlassian.net/browse/LTD-5356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


seijihg (approved):

api/cases/generated_documents/tests/test_generate_document.py
```diff
@@ -14,21 +15,45 @@
...
 from api.staticdata.statuses.enums import CaseStatusEnum
 from api.staticdata.statuses.models import CaseStatus
+from api.staticdata.regimes.models import RegimeEntry
+from api.staticdata.report_summaries.models import ReportSummarySubject, ReportSummaryPrefix
 from lite_content.lite_api import strings
 from test_helpers.clients import DataTestClient
 from api.users.models import ExporterNotification
 
 
+def get_tag(html, tag, id):
```
saruniitr, line 38: We don't have BeautifulSoup in API so a simple method to extract the tags we want in this test.

PR #2151, opened 2024-08-21
LTD-5342 special character management command

Tllew:
### Aim


api/applications/management/commands/update_party_address.py
```diff
@@ -0,0 +1,42 @@
...
+        audit_trail_service.create(
+            actor=system_user,
+            verb=AuditType.DEVELOPER_INTERVENTION,
+            target=party,
+            payload={
+                "address": {"new": new_address, "old": address},
+            },
+        )
+
+        assert party.address == address
```
currycoder, line 39: Do we have a test proving this works?  Might even be good to just raise an explicit `CommandError` instead when there's a mismatch

api/applications/management/commands/update_party_address.py
```diff
@@ -0,0 +1,42 @@
...
+    help = "Update address for multiple parties from a CSV file"
+
+    def add_arguments(self, parser):
+        parser.add_argument("csv_file", type=open, help="The path to the CSV file containing updates")
+
+    def handle(self, *args, **kwargs):
+        csv_file = kwargs["csv_file"]
+
+        reader = csv.DictReader(csv_file)
+        for row in reader:
```
currycoder, line 19: It could be good to do this inside a transaction, so that we get rollbacks for free if anything unexpected happens

api/audit_trail/payload.py
```diff
@@ -162,4 +162,5 @@ def format_payload(audit_type, payload):
     AuditType.EXPORTER_CREATED_AMENDMENT: " opened the application for a 'major edit'",
     AuditType.EXPORTER_SUBMITTED_AMENDMENT: formatters.exporter_submitted_amendment,
     AuditType.AMENDMENT_CREATED: formatters.amendment_created,
+    AuditType.DEVELOPER_INTERVENTION: "updated application information",
```
currycoder, line 165: I assume on the FE this formats as "Tomos Williams updated application information" - could "Tomos Williams (lite dev team) updated application information" be better?

api/applications/management/commands/update_party_address.py
```diff
@@ -0,0 +1,42 @@
...
+        party = Party.objects.get(id=party_id)
+        system_user = BaseUser.objects.get(id="00000000-0000-0000-0000-000000000001")
+
+        audit_trail_service.create(
+            actor=system_user,
+            verb=AuditType.DEVELOPER_INTERVENTION,
+            target=party,
+            payload={
+                "address": {"new": new_address, "old": address},
+            },
```
currycoder, line 36: Is it worth adding `additional_text` to this audit entry to give a bit more context e.g. "Agreed to update with management command in LTD-XXX".  That could even be another CLI argument

api/audit_trail/payload.py
hnryjmes: I don't see where the developer's name would come from - if the actor is `system_user`, won't this be `LITE system updated application information` ?

I think this is maybe not clear enough to inform caseworkers that there has been a developer intervention

api/applications/management/commands/update_party_address.py
hnryjmes: yes, I think `additional_text` can be used as a sort of case note for us to communicate exactly what we've done on the case in a human readable way

api/audit_trail/payload.py
Tllew: was just checking into this, I don't exist on prod so I'm not sure how we would add me as a user for the developer intervention, I was thinking this could be handled in additional_text

api/applications/management/commands/update_party_address.py
Tllew: Yeah this is a good call, I was thinking we add the additional text as a csv column

api/applications/management/commands/update_party_address.py
Tllew: added

hnryjmes (approved):
happy with code changes

api/applications/management/commands/update_party_address.py
```diff
@@ -0,0 +1,45 @@
...
+        with transaction.atomic():
+            audit_trail_service.create(
+                actor=system_user,
+                verb=AuditType.DEVELOPER_INTERVENTION,
+                target=party,
+                payload={"address": {"new": new_address, "old": address}, "additional_text": additional_text},
+            )
+
+            if party.address == address:
+                party.address = new_address
```
saruniitr, line 41: Are we only looking for special characters in address? not checking in `name`?

Tllew: This management command is just for address, there will be another to follow that updates name, I was chatting with brendan and wanted something that felt very deliberate and didn't just hack into good name and party.address depending on a flag/field variable

saruniitr (approved):

PR #2152, opened 2024-08-21
Update CLE queryset to ignore deprecated CLEs

hnryjmes:
### Aim

This PR adds a new field `deprecated` with `default=False` to the `ControlListEntry` model, and then updates the queryset used by its corresponding list view so that CLEs with `deprecated=True` are not included by default. When the migration runs on production, all CLEs without that field will be given the default value which is False.

The detail view is not changed, so a deprecated CLE can still be looked up that way. By not having it in the list view by default, it means that deprecated CLEs cannot be used for new product assessments.

There is also the option `include_deprecated` which can be passed as a param to the API call. If we have `include_deprecated=True` in the params, the original queryset is used which does not filter out the deprecated CLEs. This means that for specific uses in the frontend we can decide to include deprecated CLEs in the response data.

[LTD-5349](https://uktrade.atlassian.net/browse/LTD-5349)


api/staticdata/control_list_entries/models.py
```diff
@@ -10,6 +10,7 @@ class ControlListEntry(models.Model):
     parent = models.ForeignKey("self", related_name="children", default=None, null=True, on_delete=models.CASCADE)
     category = models.CharField(max_length=100, default="")
     controlled = models.BooleanField(default=True)
+    deprecated = models.BooleanField(default=False)
```
saruniitr, line 13: Is this to create new CLEs where this can be `True`?
What is the need to create if it is already deprecated?

hnryjmes: The reason for this change as a whole is that we need to be able to keep CLE objects in our db while also not showing them as options for new assessments. This is so that existing cases which have goods using these CLEs are not affected in any way. I think the best way to do this is a model change to add this `deprecated` field. I imagine that we will update various parts of lite-frontend so that we either grab the full set of CLEs including deprecated or not, depending on need.

e.g. 
- TAU tab in frontend, or the list of CLEs presented to TAU assessors, would not include deprecated CLEs
- case filters in sidebar on queue view, maybe would include deprecated CLEs, we should discuss that

In this PR I am choosing to make the "default" choice to be not to return deprecated CLEs. I think this is right but would welcome other suggestions.

Product search I think uses different endpoints so won't be affected by this immediately, although there might be a change needed to have that field available to users of that search.

saruniitr: Ok so there is.a requirement to mark some of the existing ones as invalid which makes sense.
I think we may need some changes in `lite-routing` also where CLEs are queried to only consider the active ones.

api/staticdata/control_list_entries/tests/test_views.py
```diff
@@ -0,0 +1,121 @@
...
+        self.url = reverse("staticdata:control_list_entries:control_list_entries")
+
+    def test_gov_user_control_list_entries_list_ignores_deprecated_cles(self):
+        cles_count_model = ControlListEntry.objects.all().count()
+
+        # Assert that we have at least 1 CLE returned by the db manager
+        self.assertTrue(cles_count_model > 0)
+
+        response = self.client.get(self.url, **self.gov_headers)
+        cles_data = response.json().get("control_list_entries")
```
saruniitr, line 24: Is it also better to check if `deprecated=True` is not set for any of the entries in the response?

api/staticdata/control_list_entries/tests/test_views.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Assert that we have at least 1 CLE returned by the db manager
+        self.assertTrue(cles_count_model > 0)
+
+        response = self.client.get(url, **self.gov_headers)
+        cles_data = response.json().get("control_list_entries")
+        cles_count_data = len(cles_data)
+
+        # Assert that we have at least 1 CLE returned by the view
+        self.assertTrue(cles_count_data > 0)
```
saruniitr, line 64: Not sure if this is necessary in this test again, this is already covered in above test

api/staticdata/control_list_entries/tests/test_views.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Assert that we have at least 1 CLE returned by the db manager
+        self.assertTrue(cles_count_model > 0)
+
+        response = self.client.get(self.url, **self.exporter_headers)
+        cles_data = response.json().get("control_list_entries")
+        cles_count_data = len(cles_data)
+
+        # Assert that we have at least 1 CLE returned by the view
+        self.assertTrue(cles_count_data > 0)
```
saruniitr, line 99: same here

api/staticdata/control_list_entries/views.py
```diff
@@ -13,15 +13,20 @@
 class ControlListEntriesList(APIView):
     authentication_classes = (SharedAuthentication,)
```
saruniitr, line 15: Is this an opportunity to split this into caseworker and exporter side?
Looking at the tests exporter side should not see deprecated CLEs so how can we be sure of it if this is common to both?

hnryjmes: I think this is a good idea. I won't be able to do all of the changes needed to make separate GovAuthentication and ExporterAuthentication work in this PR (there are frontend changes needed, and we need to have both running for a while) but I can return to this PR once I've got those done.

hnryjmes (commented):
putting back to draft as there are a few more changes needed

api/staticdata/control_list_entries/views.py
hnryjmes: since I wrote this, the exporter app has been updated to use a separate endpoint, which is why this endpoint can be changed to use GovAuthentication in this PR

api/staticdata/control_list_entries/tests/test_views.py
```diff
@@ -0,0 +1,121 @@
...
+        updated_cles_data = response.json().get("control_list_entries")
+        updated_cles_count_data = len(updated_cles_data)
+
+        # Assert that the count returned by the view has increased by 1
+        self.assertTrue(updated_cles_count_data == cles_count_data + 1)
+
+        # Assert that the data returned by the view contains the deprecated CLE
+        self.assertIn("rating123", [cle["rating"] for cle in updated_cles_data])
+
+    def test_exporter_user_control_list_entries_list_ignores_deprecated_cles(self):
```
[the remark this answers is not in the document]
hnryjmes: actually it's now possible for the exporter app to request deprecated CLEs using the extra param `include_deprecated=True` that can be added to the get request. we will work on this a bit more in https://uktrade.atlassian.net/browse/LTD-5399

saruniitr (approved):

api/staticdata/control_list_entries/tests/test_views.py
hnryjmes: I realised I never responded to this. We can't have a test assertion like this because the values we get are just `rating` and `text`. Hence why we just have to assert for the presence or absence of a CLE with that rating `rating123`.

Tllew (approved):
Looks good to me

PR #2155, opened 2024-08-22
same as previous commit but good name

Tllew:
### Aim

### DIFF
--- a/api/applications/management/commands/update_good_name.py
+++ b/api/applications/management/commands/update_good_name.py
@@ -1,5 +1,5 @@
 import csv
-from django.core.management.base import BaseCommand, CommandError
+from django.core.management.base import BaseCommand
 from api.goods.models import Good
 from api.users.models import BaseUser
 from api.audit_trail import service as audit_trail_service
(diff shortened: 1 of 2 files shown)


api/applications/management/commands/update_good_name.py
```diff
@@ -0,0 +1,45 @@
...
+            new_name = row["new_name"]
+            additional_text = row["additional_text"]
+
+            self.update_good_name(good_id, name, new_name, additional_text)
+
+    def update_good_name(self, good_id, name, new_name, additional_text):
+        good = Good.objects.get(id=good_id)
+        system_user = BaseUser.objects.get(id="00000000-0000-0000-0000-000000000001")
+
+        with transaction.atomic():
```
depsiatwal, line 32: Are we sure we want the transaction on each good being updated. 
If one item succeeds and another fails then you would need to remove the relevant items from the CSV. Is it better to have all or nothing ?

api/applications/management/commands/update_good_name.py
```diff
@@ -0,0 +1,45 @@
...
+        for row in reader:
+            good_id = row["good_id"]
+            name = row["name"]
+            new_name = row["new_name"]
+            additional_text = row["additional_text"]
+
+            self.update_good_name(good_id, name, new_name, additional_text)
+
+    def update_good_name(self, good_id, name, new_name, additional_text):
+        good = Good.objects.get(id=good_id)
```
depsiatwal, line 29: rather then check good.name below why not just include in query ?

api/applications/management/commands/update_good_name.py
```diff
@@ -0,0 +1,45 @@
...
+from api.audit_trail import service as audit_trail_service
+from api.audit_trail.enums import AuditType
+from django.db import transaction
+
+
+class Command(BaseCommand):
+    help = "Update name for multiple goods from a CSV file"
+
+    def add_arguments(self, parser):
+        parser.add_argument("csv_file", type=open, help="The path to the CSV file containing updates")
```
depsiatwal, line 14: where is it proposed we upload the CSV ?

Tllew: you specify a path to a csv here, there is no proposed location it works from the current working path

api/applications/management/commands/update_good_name.py
Tllew: are you suggesting moving the atomic to the for loop above?

sounds like a good idea

api/applications/management/commands/update_good_name.py
Tllew: done, this will now throw a DoesNotExist when the name is incorrect and should trigger the transaction to be cancelled.

api/applications/management/commands/update_good_name.py
depsiatwal: how would you upload the file to CF server has this been done before ?
is it auto generated ?

Else I think the only option is AWS S3 Bucket but this won't be able to load from there

Tllew: VI is available so you can just:
```
touch address.csv
vi address.csv
i
ctrl + v
:wq
```

saruniitr (approved):

api/applications/management/commands/update_good_name.py
hnryjmes: @Tllew yeah that's how I got my csv files off of lite-api prod for my special character stuff

hnryjmes (approved):

depsiatwal (approved):
lgtm

PR #2160, opened 2024-08-29
Create separate CLEs list view with ExporterAuthentication only

hnryjmes:
### Aim

This is a subtask of [LTD-5349](https://uktrade.atlassian.net/browse/LTD-5349). We would like to move away from using SharedAuthentication where possible, so this change is a step towards being able to do that for the CLEs list endpoint.

Once this is merged in, we can update the frontend to have the exporter app use this new endpoint instead. Once that change is released, we can update the original CLEs endpoint to use GovAuthentication only. Then work can continue on LTD-5349 to set up the endpoints to work with deprecated CLEs.

The reason that this looks a lot simpler than the endpoint it is replacing is because the params `group` and `include_parent` are only used by the caseworker app and so are not needed for the exporter app specific endpoint.

Additionally this PR now follows the directory structure pattern established by Brendan in his earlier PR here: https://github.com/uktrade/lite-api/pull/2106 where there is hard separation between `/caseworker/` urls and `/exporter/` urls.

[LTD-5377](https://uktrade.atlassian.net/browse/LTD-5377)

[LTD-5349]: https://uktrade.atlassian.net/browse/LTD-5349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-5377]: https://uktrade.atlassian.net/browse/LTD-5377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/staticdata/control_list_entries/urls.py
+++ b/api/staticdata/control_list_entries/urls.py
@@ -6,6 +6,7 @@ app_name = "control_list_entries"
 
 urlpatterns = [
     path("", views.ControlListEntriesList.as_view(), name="control_list_entries"),
+    path("exporter-list/", views.ExporterControlListEntriesList.as_view(), name="exporter_list"),
     path(
         "<str:rating>/",
         views.ControlListEntryDetail.as_view(),
--- a/api/staticdata/control_list_entries/urls.py
+++ b/api/staticdata/control_list_entries/urls.py
@@ -6,7 +6,6 @@ app_name = "control_list_entries"
 
 urlpatterns = [
     path("", views.ControlListEntriesList.as_view(), name="control_list_entries"),
-    path("exporter-list/", views.ExporterControlListEntriesList.as_view(), name="exporter_list"),
     path(
         "<str:rating>/",
         views.ControlListEntryDetail.as_view(),
--- a/api/staticdata/control_list_entries/views.py
+++ b/api/staticdata/control_list_entries/views.py
@@ -1,12 +1,15 @@
 from django.http import JsonResponse
-from rest_framework import permissions
+from rest_framework import generics, permissions
 from rest_framework.decorators import permission_classes
 from rest_framework.views import APIView
 
 from api.core.authentication import SharedAuthentication, ExporterAuthentication
 from api.staticdata.control_list_entries.helpers import get_control_list_entry, convert_control_list_entries_to_tree
 from api.staticdata.control_list_entries.models import ControlListEntry
-from api.staticdata.control_list_entries.serializers import ControlListEntrySerializerWithLinks
+from api.staticdata.control_list_entries.serializers import (
+    ControlListEntrySerializerWithLinks,
+    ExporterControlListEntriesListSerializer,
+)
 
 
 @permission_classes((permissions.AllowAny,))
@@ -44,17 +47,9 @@ class ControlListEntryDetail(APIView):
         return JsonResponse(data={"control_list_entry": serializer.data})
 
 
-class ExporterControlListEntriesList(APIView):
+class ExporterControlListEntriesList(generics.ListAPIView):
     authentication_classes = (ExporterAuthentication,)
-
-    def get_queryset(self):
-        return ControlListEntry.objects.filter(controlled=True)
-
-    def get(self, request):
-        """
-        Returns list of all Control List Entries
-        """
-
-        queryset = self.get_queryset()
-
-        return JsonResponse(data={"control_list_entries": list(queryset.values("rating", "text"))})
+    model = ControlListEntry
+    pagination_class = None
+    serializer_class = ExporterControlListEntriesListSerializer
+    queryset = ControlListEntry.objects.all()
--- a/api/staticdata/control_list_entries/views.py
+++ b/api/staticdata/control_list_entries/views.py
@@ -3,7 +3,7 @@ from rest_framework import permissions
 from rest_framework.decorators import permission_classes
 from rest_framework.views import APIView
 
-from api.core.authentication import SharedAuthentication
+from api.core.authentication import SharedAuthentication, ExporterAuthentication
 from api.staticdata.control_list_entries.helpers import get_control_list_entry, convert_control_list_entries_to_tree
 from api.staticdata.control_list_entries.models import ControlListEntry
 from api.staticdata.control_list_entries.serializers import ControlListEntrySerializerWithLinks
@@ -42,3 +42,19 @@ class ControlListEntryDetail(APIView):
         control_list_entry = get_control_list_entry(rating)
         serializer = ControlListEntrySerializerWithLinks(control_list_entry)
         return JsonResponse(data={"control_list_entry": serializer.data})
+
+
+class ExporterControlListEntriesList(APIView):
+    authentication_classes = (ExporterAuthentication,)
+
+    def get_queryset(self):
+        return ControlListEntry.objects.filter(controlled=True)
+
+    def get(self, request):
+        """
+        Returns list of all Control List Entries
+        """
+
+        queryset = self.get_queryset()
+
+        return JsonResponse(data={"control_list_entries": list(queryset.values("rating", "text"))})
--- a/api/staticdata/control_list_entries/views.py
+++ b/api/staticdata/control_list_entries/views.py
@@ -52,4 +52,4 @@ class ExporterControlListEntriesList(generics.ListAPIView):
     model = ControlListEntry
     pagination_class = None
     serializer_class = ExporterControlListEntriesListSerializer
-    queryset = ControlListEntry.objects.all()
+    queryset = ControlListEntry.objects.filter(controlled=True)
--- a/api/staticdata/control_list_entries/views.py
+++ b/api/staticdata/control_list_entries/views.py
@@ -1,15 +1,12 @@
 from django.http import JsonResponse
-from rest_framework import generics, permissions
+from rest_framework import permissions
 from rest_framework.decorators import permission_classes
 from rest_framework.views import APIView
 
-from api.core.authentication import SharedAuthentication, ExporterAuthentication
+from api.core.authentication import SharedAuthentication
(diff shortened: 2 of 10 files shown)


api/staticdata/control_list_entries/views.py
```diff
@@ -42,3 +42,19 @@ def get(self, request, rating):
...
+        return ControlListEntry.objects.filter(controlled=True)
+
+    def get(self, request):
+        """
+        Returns list of all Control List Entries
+        """
+
+        queryset = self.get_queryset()
+
+        return JsonResponse(data={"control_list_entries": list(queryset.values("rating", "text"))})
```
kevincarrogan, line 60: I think there are a few things here that could just be covered using other generic DRF views.

For example I would suggest the following:

  - Probably use whatever ListView DRF gives you
  - Set queryset as a view level attribute
  - Use a serializer instead of `.values`

hnryjmes: these are all good suggestions, will see what I can do

kevincarrogan: https://github.com/uktrade/lite-api/blob/dev/api/staticdata/regimes/views.py#L14-L21

This is probably a good example of what I mean.

api/staticdata/control_list_entries/views.py
```diff
@@ -42,3 +42,19 @@ def get(self, request, rating):
         control_list_entry = get_control_list_entry(rating)
         serializer = ControlListEntrySerializerWithLinks(control_list_entry)
         return JsonResponse(data={"control_list_entry": serializer.data})
+
+
+class ExporterControlListEntriesList(APIView):
+    authentication_classes = (ExporterAuthentication,)
+
+    def get_queryset(self):
+        return ControlListEntry.objects.filter(controlled=True)
```
saruniitr, line 51: Suggest adding `deprecated=False` as well to the filter.

api/staticdata/control_list_entries/views.py
hnryjmes: updated

api/staticdata/control_list_entries/urls.py
```diff
@@ -6,6 +6,7 @@
 
 urlpatterns = [
     path("", views.ControlListEntriesList.as_view(), name="control_list_entries"),
+    path("exporter-list/", views.ExporterControlListEntriesList.as_view(), name="exporter_list"),
```
kevincarrogan, line 9: Is this following the conventions of the other changes we've made where we've split out exporter/caseworker endpoints?

hnryjmes: I think maybe not? I don't love `exporter-list` but couldn't remember where to look hence leaving it like this. But on having a look again, I think this is where the convention got started:  https://github.com/uktrade/lite-api/pull/2106

> This is also a first attempt at proving how we might split our API endpoints down more explicitly in to exporter/caseworker subdivisions. The endpoint URLs start with /caseworker/ and /exporter/ to provide a hard separation, and serializers/permission classes/views are completely separated between the two. They both call a common Case.change_status() method to share business logic.

So, following this, we could refactor this significantly and move things into separate directories for `exporter` and `caseworker`. Then you wouldn't have this weird `exporter_foo` namespacing, which would probably be a lot nicer.

I'll have a look at following this convention and moving things into separate directories

hnryjmes: this has been updated. as far as possible I have followed the directory structure established by Brendan in that previous PR. the URL ends up being `/exporter/static/control-list-entries/` or `reverse("exporter_staticdata:control_list_entries:control_list_entries")`

api/staticdata/exporter/control_list_entries/views.py
```diff
@@ -0,0 +1,13 @@
...
+from api.staticdata.control_list_entries.models import ControlListEntry
+from api.staticdata.exporter.control_list_entries.serializers import ControlListEntriesListSerializer
+
+
+class ControlListEntriesList(generics.ListAPIView):
+    authentication_classes = (ExporterAuthentication,)
+    model = ControlListEntry
+    pagination_class = None
+    serializer_class = ControlListEntriesListSerializer
+    queryset = ControlListEntry.objects.filter(controlled=True)
```
currycoder, line 13: I remember we were talking about there being a `deprecated` flag or something similar for ControlListEntry...  Is that what this is?

api/staticdata/exporter/control_list_entries/serializers.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from api.staticdata.control_list_entries.models import ControlListEntry
+
+
+class ControlListEntriesListSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ControlListEntry
+        fields = ("rating", "text")
```
currycoder, line 9: Just for my own understanding, could you point to where in the frontend we expect to swap out the use of the `SharedAuthentication` endpoint for this one?

api/staticdata/exporter/control_list_entries/views.py
hnryjmes: Not this PR. All the `deprecated` stuff is happening in this other PR: https://github.com/uktrade/lite-api/pull/2152

this PR is just to set up a separate exporter-only CLEs endpoint. once that is there and working, I can go back to the other PR, and make that caseworker-only i.e. use `GovAuthentication` there.

api/staticdata/exporter/control_list_entries/tests/test_views.py
```diff
@@ -0,0 +1,23 @@
...
+        self.url = reverse("exporter_staticdata:control_list_entries:control_list_entries")
+        super().setUp()
+
+    def test_list_view_success(self):
+        response = self.client.get(self.url, **self.exporter_headers)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        for cle in response.json():
+            self.assertEqual(list(cle.keys()), ["rating", "text"])
```
currycoder, line 18: Could I suggest trying to assert that the *whole response* is as we expect it to be?  e.g.
```
self.assertEqual(response.json(), [{"rating": "foo", "text": "bar"}, {...}]
```


To do that I would recommend deleting all CLE objects and adding a couple of fixed ones with use of factories.

I believe because each test case runs inside a transaction, deleting all CLEs will not cause problems for subsequent test cases as the transaction will be rolled back and the data put back in to the OG state.

currycoder (commented):
A few minor suggestions and questions

api/staticdata/exporter/control_list_entries/views.py
currycoder: Ok cool - I also assume that when that happens, this filter will be updated to exclude deprecated CLEs too?

api/staticdata/exporter/control_list_entries/serializers.py
hnryjmes: The corresponding draft PR for frontend is here: https://github.com/uktrade/lite-frontend/pull/2133/files

You can see exactly where it gets used if you search for `get_control_list_entries` in your IDE and filter the results by the `exporter` directory only. It's referenced in 5 different files from what I can see.

Although now I'll need to update it to use the new URL `/exporter/static/control-list-entries/` since the refactor

currycoder: Nice!

api/staticdata/exporter/control_list_entries/views.py
hnryjmes: yes, correct

api/staticdata/exporter/control_list_entries/serializers.py
saruniitr: you don't need `id` in here for anything in FE?

api/staticdata/exporter/control_list_entries/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from api.staticdata.exporter.control_list_entries import views
+
+app_name = "control_list_entries"
+
+urlpatterns = [path("", views.ControlListEntriesList.as_view(), name="control_list_entries")]
```
saruniitr, line 7: If you add `,` after `name=....` it will be formatted by black as below,

```

urlpatterns = [
    path("", views.ControlListEntriesList.as_view(), name="control_list_entries"),
]
```

This makes the diff easier when we add additional entries

api/staticdata/exporter/control_list_entries/views.py
```diff
@@ -0,0 +1,13 @@
+from rest_framework import generics
+
+from api.core.authentication import ExporterAuthentication
+from api.staticdata.control_list_entries.models import ControlListEntry
+from api.staticdata.exporter.control_list_entries.serializers import ControlListEntriesListSerializer
+
+
+class ControlListEntriesList(generics.ListAPIView):
+    authentication_classes = (ExporterAuthentication,)
+    model = ControlListEntry
```
saruniitr, line 10: don't think `model` is really required as it takes from the serializer

saruniitr (approved):
few minor things, with that fixed it looks good to me

api/staticdata/exporter/control_list_entries/serializers.py
hnryjmes: @saruniitr yes, correct. Compare to the endpoint it is replacing: https://github.com/uktrade/lite-api/blob/dev/api/staticdata/control_list_entries/views.py#L32

You can see that just `rating` and `text` are needed.

api/staticdata/exporter/control_list_entries/tests/test_views.py
hnryjmes: updated, the extra test is there now

PR #2162, opened 2024-09-03
Ltd 5383 Good name and Party address validator

Tllew:
### Aim

[LTD-5383](https://uktrade.atlassian.net/browse/LTD-5383)

[LTD-5383]: https://uktrade.atlassian.net/browse/LTD-5383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/core/validators.py
+++ b/api/core/validators.py
@@ -1,6 +1,6 @@
 from django.utils.deconstruct import deconstructible
 from rest_framework.exceptions import ValidationError
-
+import re
 from api.staticdata.control_list_entries.models import ControlListEntry
 
 
@@ -22,3 +22,23 @@ class ControlListEntryValidator:
             ControlListEntry.objects.get(rating=value)
         except ControlListEntry.DoesNotExist:
             raise ValidationError(self.message, code=self.code)
+
+
+class EdifactStringValidator:
+    message = "Undefined Error"
+    regex_string = r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>]+$"
+
+    def __call__(self, value):
+        match_regex = re.compile(self.regex_string)
+        is_value_valid = bool(match_regex.match(value))
+        if not is_value_valid:
+            raise ValidationError(self.message)
+
+
+class GoodNameValidator(EdifactStringValidator):
+    message = "Product name must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes"
+
+
+class PartyAddressValidator(EdifactStringValidator):
+    regex_string = re.compile(r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>\r\n]+$")
+    message = "Address must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes"
--- a/api/goods/serializers.py
+++ b/api/goods/serializers.py
@@ -1,8 +1,8 @@
 from rest_framework import serializers
 from rest_framework.relations import PrimaryKeyRelatedField
-
 from api.core.helpers import str_to_bool
 from api.core.serializers import KeyValueChoiceField, ControlListEntryField, GoodControlReviewSerializer
+from api.core.validators import GoodNameValidator
 from api.documents.libraries.process_document import process_document
 from api.goods.enums import (
     FirearmCategory,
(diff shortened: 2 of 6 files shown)


api/goods/serializers.py
```diff
@@ -819,6 +819,16 @@ class GoodSerializerInternal(serializers.Serializer):
     is_precedent = serializers.BooleanField(required=False, default=False)
     product_description = serializers.CharField()
 
+    def validate_name(self, value):
+        match_regex = re.compile(r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>]+$")
```
hnryjmes, line 823: I think it's likely that we'll want to reuse this regex pattern throughout lite-api, if so, maybe it's best to define this in a constants file somewhere and explain exactly what it is with an appropriate comment.

If someone in the future finds this and doesn't have the full context, I don't think it would be obvious to them that this is the HMRC implementation of EDIFACT Level Set B character set

api/goods/serializers.py
```diff
@@ -894,6 +904,16 @@ class GoodSerializerExporter(serializers.Serializer):
     has_declared_at_customs = serializers.BooleanField()
     design_details = serializers.CharField()
 
+    def validate_name(self, value):
+        match_regex = re.compile(r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>]+$")
```
hnryjmes, line 908: relating to the above comment, once the regex pattern is defined somewhere else, it can be referenced here, so no need for a raw string literal

api/goods/tests/test_serializers.py
```diff
@@ -150,6 +151,61 @@ def test_report_summary_present(self):
         self.assertEqual(actual_subject["id"], str(self.good.report_summary_subject.id))
         self.assertEqual(actual_subject["name"], self.good.report_summary_subject.name)
 
+    @parameterized.expand(
+        [
+            "random good",
```
currycoder, line 156: These parameters appear to be duplicated across 3 separate test cases.  If you do end up with a helper function, I would recommend that tests for that function are the exhaustive ones with all these different parameters.  And these tests just become single parameter sanity checks.

I think there's absolutely value in checks at both levels, but if we got to a place where a single helper was reused it would be more DRY/maintainable all round IMO

api/core/validators.py
```diff
@@ -22,3 +22,21 @@ def __call__(self, value):
...
         except ControlListEntry.DoesNotExist:
             raise ValidationError(self.message, code=self.code)
+
+
+class EdifactStringValidator:
+    message = "Undefined Error"
+
+    def __call__(self, value):
+        match_regex = re.compile(r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>]+$")
+        is_value_valid = bool(match_regex.match(value))
```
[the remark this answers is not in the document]
hnryjmes: although your tests would suggest it's fine

api/parties/tests/test_serializers.py
```diff
@@ -1,24 +1,97 @@
...
+                "Party address must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes",
+            ),
+            (
+                "party\naddress",
+                "Party address must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes",
+            ),
+            (
+                "party\r\naddress",
+                "Party address must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes",
+            ),
```
hnryjmes, line 71: I'm a bit concerned about what this means for address input in LITE. I can see that in the frontend PR you've also got this validator so the characters `\r\n` can be used: https://github.com/uktrade/lite-frontend/pull/2140/files

We have supported multi-line address input in LITE for party address for a long time. It's maybe not ideal but it's a natural way to enter addresses and it's suggested here for use in some cases: https://design-system.service.gov.uk/patterns/addresses/

If we don't allow these characters that means that users may not be able to enter addresses how they normally would, i.e. separating information on separate lines. I think if we're going to have a change to LITE that breaks this we need to have a chat with UCD about what component we're using and how we want to handle this in general.

My guess for how we would solve this was that we would allow newline characters as an exception to the EDIFACT string standard, and then process the data later before it goes to HMRC. I don't think either of these changes allow that to happen.

api/core/validators.py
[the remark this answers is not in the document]
Tllew: the +$ repeats it, this was to keep the regex as close to the original as possible

api/parties/tests/test_serializers.py
Tllew: I think you're probably right in that the field should be separated into multiple fields, validation of the existing input fields is an immediate fix, we would need to address the further solution with HMRC to ensure fields can align.

hnryjmes (commented):
any idea why tests are failing?

hnryjmes (approved):

markj0hnst0n (approved):

PR #2163, opened 2024-09-03
LTD-5304-managment-cmd-remove-advice 

depsiatwal:
A new management command which allows all finial advice to be deleted. 

This is required when TAU change their minds and have issue various final advice on a case. We need to invalidate this so the case can go back to post-circulation

### DIFF
--- a/api/applications/tests/factories.py
+++ b/api/applications/tests/factories.py
@@ -1,5 +1,5 @@
 import factory
-
+import factory.fuzzy
 from faker import Faker
 
 from api.applications.enums import ApplicationExportType, ApplicationExportLicenceOfficialType
@@ -14,7 +14,8 @@ from api.applications.models import (
     GoodOnApplicationInternalDocument,
     StandardApplication,
 )
-from api.cases.enums import CaseTypeEnum
+from api.cases.enums import AdviceLevel, AdviceType, CaseTypeEnum
+from api.cases.models import Advice
 from api.external_data.models import Denial, DenialEntity, SanctionMatch
 from api.documents.tests.factories import DocumentFactory
 from api.staticdata.statuses.models import CaseStatus
@@ -238,5 +239,28 @@ class DraftStandardApplicationFactory(StandardApplicationFactory):
 
         if kwargs["goods_recipients"] == StandardApplication.VIA_CONSIGNEE_AND_THIRD_PARTIES:
             PartyOnApplicationFactory(application=obj, party=ThirdPartyFactory(organisation=obj.organisation))
+
+        return obj
+
+
+class AdviceFactory(factory.django.DjangoModelFactory):
+    user = factory.SubFactory(GovUserFactory)
+    case = factory.SubFactory(StandardApplicationFactory)
+    type = factory.fuzzy.FuzzyChoice(AdviceType.choices, getter=lambda t: t[0])
+    level = factory.fuzzy.FuzzyChoice(AdviceLevel.choices, getter=lambda t: t[0])
+
+    class Meta:
+        model = Advice
+
+
+class FinalAdviceOnApplicationFactory(StandardApplicationFactory):
+
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        obj = model_class(*args, **kwargs)
+        obj.status = get_case_status_by_status(CaseStatusEnum.UNDER_FINAL_REVIEW)
+        obj.save()
+
+        AdviceFactory(case=obj, level=AdviceLevel.FINAL)
 
         return obj
--- a/api/support/management/commands/remove_case_final_advice.py
+++ b/api/support/management/commands/remove_case_final_advice.py
@@ -1,0 +1,85 @@
+import logging
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+
+from api.audit_trail.enums import AuditType
+from api.cases.enums import AdviceLevel
+from api.cases.models import Case
+from api.queues.models import Queue
+from api.audit_trail import service as audit_trail_service
+
+from api.staticdata.statuses.enums import CaseStatusEnum
+from api.staticdata.statuses.libraries.get_case_status import get_case_status_by_status
+from lite_routing.routing_rules_internal.enums import QueuesEnum
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = """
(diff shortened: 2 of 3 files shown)


api/support/management/commands/remove_all_final_advice.py
```diff
@@ -0,0 +1,76 @@
...
+        dry_run = options.pop("dry_run")
+
+        with transaction.atomic():
+            try:
+                case = Case.objects.get(reference_code=case_reference)
+            except Case.DoesNotExist:
+                logger.error("Invalid Case reference %s, does not exist", case_reference)
+                return
+
+            is_refusal = case.advice.filter(level=AdviceLevel.FINAL, type=AdviceType.REFUSE).exists()
```
saruniitr, line 50: We can remove this line and the code that removes countersigning flags below based on this.
Because if it is ready to finalise means the countersigning flags are already removed

api/support/management/commands/remove_all_final_advice.py
```diff
@@ -0,0 +1,76 @@
...
+                logger.error("Invalid Advice data, no final advice on this case")
+                return
+
+            if not dry_run:
+                # Move the Case to 'LU Post-Circulation Cases to Finalise' queue
+                # as it needs to be in this queue to finalise and issue
+                case.queues.set(Queue.objects.filter(id=QueuesEnum.LU_POST_CIRC))
+
+                for item in final_advice:
+                    item.delete()
```
saruniitr, line 64: Before deleting we also need to ensure case status is `under_final_review` which it should be when it is on the post-circulation queue.

saruniitr (commented):
I think we take this opportunity to fix this in code correctly instead of adding this command.
Once we add this then we end up relying on it (similar to `refresh_goods_on_licence` command).

Also for these sort of issues the same command may not be sufficient in all cases.
Eg for the current support issue the status is not matching with the queue so in the command we are changing status and queue.

api/support/management/commands/remove_all_final_advice.py
```diff
@@ -0,0 +1,80 @@
...
+            except Case.DoesNotExist:
+                logger.error("Invalid Case reference %s, does not exist", case_reference)
+                return
+
+            final_advice = case.advice.filter(level=AdviceLevel.FINAL)
+
+            # Ensure final advice given by LU
+            if not final_advice.exists():
+                logger.error("Invalid Advice data, no final advice on this case")
+                return
```
currycoder, line 54: Could this raise a `CommandError` instead? https://docs.djangoproject.com/en/5.1/howto/custom-management-commands/#command-exceptions

api/support/management/commands/remove_all_final_advice.py
```diff
@@ -0,0 +1,80 @@
...
+                # also need to ensure the status is under final review.
+                case.queues.set(Queue.objects.filter(id=QueuesEnum.LU_POST_CIRC))
+                old_case_status = case.status.status
+                case.status = get_case_status_by_status(CaseStatusEnum.UNDER_FINAL_REVIEW)
+
+                for item in final_advice:
+                    item.delete()
+                case.save()
+                additional_text = f"Deleted final advice and changed case status deleted advice ids:{final_advice_ids}"
+                audit_trail_service.create(
```
currycoder, line 70: Could we also have an extra audit trail entry specifically for changing the status? e.g. https://github.com/uktrade/lite-api/blob/32a438131674c57c203104fd32394bf30f79a011/api/cases/models.py#L211-L219

Sometimes we have needed to use these audit entries to work things out (and I believe we may even use them programmatically somewhere), so would be good to have consistency for these too.

api/support/management/commands/remove_all_final_advice.py
```diff
@@ -0,0 +1,80 @@
...
+                    actor=BaseUser.objects.get(id="00000000-0000-0000-0000-000000000001"),
```
currycoder, line 71: This got me thinking it could be good to use the helper we have for creating system user audit trail entries;
https://github.com/uktrade/lite-api/blob/32a438131674c57c203104fd32394bf30f79a011/api/audit_trail/service.py#L57-L70

api/support/management/commands/remove_all_final_advice.py
```diff
@@ -0,0 +1,80 @@
+import logging
```
currycoder, line 1: `python manage.py remove_case_final_advice` could make folks feel better than `python manage.py remove_all_final_advice` :D

currycoder (commented):
A few suggestions inline, but this is otherwise looking good.

api/support/management/commands/remove_all_final_advice.py
saruniitr: +1

api/support/management/commands/remove_all_final_advice.py
depsiatwal: removed

api/support/management/commands/remove_all_final_advice.py
depsiatwal: also added

api/support/management/commands/remove_all_final_advice.py
```diff
@@ -0,0 +1,80 @@
...
+            final_advice = case.advice.filter(level=AdviceLevel.FINAL)
+
+            # Ensure final advice given by LU
+            if not final_advice.exists():
+                logger.error("Invalid Advice data, no final advice on this case")
+                return
+
+            if not dry_run:
+                final_advice_ids = list(final_advice.values_list("id", flat=True))
+                final_advice_ids = [str(o) for o in final_advice_ids]
```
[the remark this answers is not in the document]
depsiatwal: Yeah I did think about that may be as part of this change we allow the Advice table to be auditable that way we could look up the ids if we ever needed. from this notes and timelines entry

depsiatwal:
> I think we take this opportunity to fix this in code correctly instead of adding this command. Once we add this then we end up relying on it (similar to `refresh_goods_on_licence` command).
> 
> Also for these sort of issues the same command may not be sufficient in all cases. Eg for the current support issue the status is not matching with the queue so in the command we are changing status and queue.

I agree but at present support just doesn't have capacity atm to deal with both as the latter comes under continuous improvement. But I will take some timeout to write up a continuous improvement ticket which should deal with underlying issues and consequently remove this command.

saruniitr (approved):

currycoder (approved):

PR #2164, opened 2024-09-03
LTD-5341 Update Firearms Cert on Org Details Page issue

markj0hnst0n:
### Aim

Exporter had issue whereby on Organisation Details page if they updated their 'Registered firearms dealer' certificate the details didn't update to the new certificate.

This issue only occurred if they updated their certificate to one that had an expiry date earlier than their current certificate.  If their new cert had an expiry date in the future the issue did not occur.

I found that even though the documents were being saved correctly the `OrganisationDetailSerializer` was presenting the org details view with the documents listed in order of `expiry_date` instead of `created_by`.  Changing to order by `created_by` repaired the issue on the frontend as the logic is there which presents the documents correctly.

[LTD-5341](https://uktrade.atlassian.net/browse/LTD-5341)

[LTD-5341]: https://uktrade.atlassian.net/browse/LTD-5341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/organisations/tests/test_documents.py
+++ b/api/organisations/tests/test_documents.py
@@ -1,4 +1,4 @@
-import datetime
+from datetime import datetime, timedelta
 
 from unittest import mock
 
@@ -29,9 +29,12 @@ class OrganisationDocumentViewTests(DataTestClient):
         mock_virus_scan.return_value = False
 
         url = reverse("organisations:documents", kwargs={"pk": self.organisation.pk})
+
+        future_date = datetime.now() + timedelta(days=365)
+        future_date_formatted = future_date.strftime("%Y-%m-%d")
         data = {
             "document": {"name": "some-document", "s3_key": "some-document", "size": 476},
-            "expiry_date": "2026-01-01",
+            "expiry_date": future_date_formatted,
             "reference_code": "123",
             "document_type": OrganisationDocumentType.FIREARM_SECTION_FIVE,
         }
@@ -46,7 +49,8 @@ class OrganisationDocumentViewTests(DataTestClient):
         self.assertEqual(instance.document.s3_key, "some-document")
         self.assertEqual(instance.reference_code, "123")
         self.assertEqual(instance.document.size, 476)
-        self.assertEqual(instance.expiry_date, datetime.date(2026, 1, 1))
+
+        self.assertEqual(instance.expiry_date.strftime("%Y-%m-%d"), future_date_formatted)
         self.assertEqual(instance.document_type, OrganisationDocumentType.FIREARM_SECTION_FIVE)
         self.assertEqual(instance.organisation, self.organisation)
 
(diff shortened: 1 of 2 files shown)


api/organisations/tests/test_documents.py
```diff
@@ -151,11 +161,46 @@ def test_retrieve_organisation_documents(self):
...
+        url = reverse("organisations:organisation", kwargs={"pk": self.organisation.pk})
+
+        response = self.client.get(url, **self.exporter_headers)
+
+        # We check that the first document delivered to organisation details is the most recently created
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json()["documents"][0]["document"]["name"], "some-document-two")
+
     def test_retrieve_organisation_documents_invalid_organisation(self):
```
markj0hnst0n, line 199: This is the new test which focuses on the changes

api/organisations/tests/test_documents.py
```diff
@@ -1,4 +1,4 @@
-import datetime
```
markj0hnst0n, line 1: Discovered that the unit tests for documents had hardcoded dates which would break in the future so repaired these.

api/organisations/tests/test_documents.py
```diff
@@ -29,9 +29,12 @@ def test_create_organisation_document(self, mock_virus_scan, mock_s3_operations_
         mock_virus_scan.return_value = False
 
         url = reverse("organisations:documents", kwargs={"pk": self.organisation.pk})
+
+        future_date = datetime.now() + timedelta(days=365)
+        future_date_formatted = future_date.strftime("%Y-%m-%d")
         data = {
             "document": {"name": "some-document", "s3_key": "some-document", "size": 476},
-            "expiry_date": "2026-01-01",
+            "expiry_date": future_date_formatted,
```
currycoder, line 37: I think this change is totally fine, but just to surface an alternative;
https://github.com/uktrade/lite-api/blob/32a438131674c57c203104fd32394bf30f79a011/api/cases/tests/test_case_ecju_queries.py#L278

`freezegun.freeze_time` essentially patches the current date/time to be frozen for the duration of a test case.  It can be particularly handy for things like this where setting future/past datetimes are under test.

currycoder (approved):

saruniitr (approved):

kevincarrogan:
> ### Aim
> Exporter had issue whereby on Organisation Details page if they updated their 'Registered firearms dealer' certificate the details didn't update to the new certificate.
> 
> This issue only occurred if they updated their certificate to one that had an expiry date earlier than their current certificate. If their new cert had an expiry date in the future the issue did not occur.
> 
> I found that even though the documents were being saved correctly the `OrganisationDetailSerializer` was presenting the org details view with the documents listed in order of `expiry_date` instead of `created_by`. Changing to order by `created_by` repaired the issue on the frontend as the logic is there which presents the documents correctly.
> 
> [LTD-5341](https://uktrade.atlassian.net/browse/LTD-5341)

I'm just wanting to think out loud about this fix.

So we could possibly assume that the previous logic made sense because there was an assumption that an exporter would always upload an expiration date in the future if they uploaded a new certificate?

In this case they're uploading a certificate that has an expiry date before their current one and this goes out of order and fails?

markj0hnst0n:
@kevincarrogan 

That's a fair summary.  I think they may have just assumed that the exporter would always upload a firearms cert in the future as this is the most normal thing to do (when old one expires).  I think that the fact that this isn't a common issue supports that.

My mistake in the description.  It's not frontend it's that `distinct()` which means it only provides the first document with the doctype provided.  I think I did work that out at one point but I forgot it and had to investigate again there.

api/organisations/tests/test_documents.py
depsiatwal: yeah also recommend shifting to freeze gun

PR #2167, opened 2024-09-05
LTD-5355 remove GovNotification

markj0hnst0n:
### Aim

There is a system which creates a notification sent to all caseworkers each time and audit.create is called.  This system assumes there will always be one notification on a case and updates it when it needs to create a new notification.  Somehow the system can create more than one object and so it will then break any time the system calls a 'get' on the notification object as it returns 2 and this can break various things for the user (for this ticket the user was unable to go in to edit their application after it was opened for edit).

On investigation I could not observe anywhere where these notifications are used on the front end.  They seem to be used for a queue with an id named `UPDATED_CASES_QUEUE_ID` but I can't see anywhere that this case can be assigned on the front end.

This GovNotification system is not in use really the decision was taken to remove it completely so the objective of this pull request is to achieve that.

There's further information on the ticket.

UCD Testing on dev2 has confirmed that the removal is a fix for existing issue and does not break existing functionality.

[LTD-5355](https://uktrade.atlassian.net/browse/LTD-5355)

[Frontend PR here](https://github.com/uktrade/lite-frontend/pull/2144)

[LTD-5355]: https://uktrade.atlassian.net/browse/LTD-5355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal (commented):
As an over all change if we're confident GovNotifications aren't used then this change looks good. 

I know you place this on a dev environment is it worth  letting UCD loose on it to make sure nothing unexpected is broken ?

saruniitr (commented):
Pretty much all changes are about removing code.
As mentioned by @depsiatwal if UCD confirms nothing is broken then this looks good to me.

depsiatwal (approved):
LGTM 
But please squash commits before merging

PR #2171, opened 2024-09-06
Fix a race condition when multiple amendments are made

kevincarrogan:
### Aim

This stops a race condition where an amendment request hits our API multiple times simultaneously.

It's possible for a user to have submitted the "confirm major edit" action multiple times and we want to ensure that the request is only processed once.

To handle this we ensure that a lock is put on the application object and then check the status of the application after any simultaneous edits are being made.

To properly test this I've added a separate CI job due to this test needing to use `TransactionTestCase` and that not playing nicely with other tests.

[LTD-5392](https://uktrade.atlassian.net/browse/LTD-5392)

[LTD-5392]: https://uktrade.atlassian.net/browse/LTD-5392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/tests/test_models.py
+++ b/api/applications/tests/test_models.py
@@ -1,9 +1,16 @@
+import concurrent.futures
+import pytest
+
 from django.forms import model_to_dict
+from django.test import TransactionTestCase
 from django.utils import timezone
+
+from parameterized import parameterized
 
 from test_helpers.clients import DataTestClient
 
 from api.appeals.tests.factories import AppealFactory
+from api.applications.exceptions import AmendmentError
 from api.audit_trail.models import Audit
 from api.cases.models import CaseType, Queue
 from api.flags.models import Flag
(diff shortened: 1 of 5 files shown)


pytest.ini
```diff
@@ -9,3 +9,4 @@ markers =
 	elasticsearch: Tests that use elasticsearch
 	seeding: tests that check seed commands
 	performance: tests that check performance
+	requires_transactions: tests that require fine grained controls of transactions
```
hnryjmes, line 12: not strictly part of this PR, but now that we have this `requires_transactions` mark, is the idea that we would use this whenever we write a test class inheriting from `TransactionTestCase`? e.g. what I think is the only other example of this: https://github.com/uktrade/lite-api/blob/dev/api/anonymised_db_dumps/tests/test_anonymised_dumps.py#L47

hnryjmes: or was it more the use of `ThreadPoolExecutor` that you found wasn't working well with other tests?

kevincarrogan: Yes. That is the idea behind it. But it was more that I was forced to do that because running this particular test with the other tests always broke in interesting ways.

hnryjmes (approved):
happy with code changes and test coverage

api/applications/tests/test_models.py
```diff
@@ -574,3 +598,32 @@ def test_clone_with_party_override(self):
...
+            future_2 = executor.submit(_create_amendment, same_application)
+
+            amendment_1 = future_1.result()
+            amendment_2 = future_2.result()
+
+        self.assertEqual(
+            StandardApplication.objects.count(),
+            2,
+        )
+        self.assertEqual(amendment_1, amendment_2)
```
currycoder, line 629: Just to really prove this out, could we assert that `amendment_1.amendment_of` points back to `original_application`

currycoder (approved):
Looks great, one minor comment inline

PR #2172, opened 2024-09-10
LTD-5361-caseworker-add-user

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5361

This is a new end point for a caseworker to add a new exporter to an organisation. 
A new frontend is bing developed to make use of this endpoint. 

I've attempted to build this from the ground up. However i've had to reuse the existing serialiser due to the logic required to create a user. Hence attempting to balance what needs to be rewritten vs new strategic views.

This is the frontend  for this endpoint 
https://github.com/uktrade/lite-frontend/pull/2150


saruniitr (commented):
few comments, overall looks good to me.

api/audit_trail/enums.py
```diff
@@ -145,6 +145,7 @@ class AuditType(LiteEnum):
     EXPORTER_SUBMITTED_AMENDMENT = autostr()
     AMENDMENT_CREATED = autostr()
     DEVELOPER_INTERVENTION = autostr()
+    ADD_EXPORT_USER_ORGANISATION = autostr()
```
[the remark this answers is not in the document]
depsiatwal: agree Done

api/organisations/caseworker/views/tests/test_users.py
```diff
@@ -0,0 +1,98 @@
...
+from api.applications.tests.factories import OrganisationFactory
+
+
+from test_helpers.clients import DataTestClient
+from api.core.constants import GovPermissions, Roles
+from uuid import uuid4
+from faker import Faker
+
+
+class TestCreateStatus(DataTestClient):
```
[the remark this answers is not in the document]
depsiatwal: Done

api/organisations/caseworker/views/users.py
```diff
@@ -0,0 +1,41 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            response_data = serializer.data
+            notify_exporter_user_added(org_pk, serializer.validated_data["email"])
+            add_case_note_add_export_user(request, data)
+
+        else:
+            return JsonResponse(data=serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+        return JsonResponse(data=response_data, status=status.HTTP_201_CREATED)
```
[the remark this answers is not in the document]
depsiatwal: Done

api/organisations/caseworker/views/helpers.py
```diff
@@ -0,0 +1,26 @@
...
+
+def notify_exporter_user_added(organisation_pk, email):
+    organisation = Organisation.objects.get(id=organisation_pk)
+    notify.notify_exporter_user_added(
+        email,
+        {"organisation_name": organisation.name, "exporter_frontend_url": get_exporter_frontend_url("/")},
+    )
+
+
+def add_case_note_add_export_user(request, data):
```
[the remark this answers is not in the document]
depsiatwal: yeah agree tidied this up now

api/organisations/caseworker/views/users.py
```diff
@@ -0,0 +1,37 @@
+from uuid import UUID
+from django.http import JsonResponse
+from api.core.authentication import GovAuthentication
+from api.core.permissions import CanCaseworkersManageOrgainsation
+from api.users.serializers import ExporterUserCreateUpdateSerializer
```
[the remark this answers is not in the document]
depsiatwal: So i've tried to balance this by reusing some of what we had 
at least now it has a separate serialiser lowing the risk you mentioned above. 

I don't think we require a lot of the validation methods some of these should ideally be handled by DB constraints and the rest are handled by the frontend. I think a lot of these methods were legacy where we had the whaky liteforms and used the serializer errors to present errors to the user FE.

api/core/celery_tasks.py
```diff
@@ -13,14 +11,6 @@ def debug_add(x, y):
     return x + y
 
 
-@shared_task
-def debug_count_cases():
```
depsiatwal, line 17: I had to remove this function as it's creating unnecessary coupling to Case.
Hence if I try to import this with models then we get circular reference.

api/organisations/caseworker/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+
+from api.organisations.caseworker.views import users
+
+app_name = "caseworker_orgainisation"
```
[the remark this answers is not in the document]
depsiatwal: fixed

api/organisations/caseworker/views/users.py
```diff
@@ -0,0 +1,39 @@
...
+from rest_framework.generics import CreateAPIView
+
+
+class CreateExporterUser(CreateAPIView):
+    authentication_classes = (GovAuthentication,)
+
+    serializer_class = ExporterUserCreateSerializer
+    permission_classes = [CanCaseworkersManageOrgainsation]
+
+    def post(self, request, org_pk):
```
currycoder, line 19: Now this is a `CreateAPIView` I think you can stop overriding the `post()` method here and go with something a lot simpler like; https://github.com/uktrade/lite-api/blob/873172c7369718c56c23518dbf51f47224b8cb17/api/applications/views/applications.py#L835

e.g. something like
```
def perform_create(self, serializer):
    organisation = Organisation.objects.get(id=self.kwargs["org_pk"])
    serializer.save(organisation=organisation)
    organisation.notify_exporter_user_added(serializer.validated_data["email"])
    organisation.add_case_note_add_export_user(request.user, data["sites"], serializer.validated_data["email"])
```

I think the response and things will be handled for you automatically by using the serializer_class.

currycoder (commented):
This is brilliant - I just have one more suggestion for simplifying things.

saruniitr (approved):
With last comment from Brendan addressed it looks good to me.

api/organisations/caseworker/views/users.py
depsiatwal: done had to do some modifications to get things working 
some of which are not ideal but hopefully tider

api/organisations/caseworker/views/users.py
```diff
@@ -0,0 +1,42 @@
...
+            raise Http404()
+
+    def perform_create(self, serializer):
+        super().perform_create(serializer)
+
+        sites = serializer.initial_data["sites"]
+        email = serializer.validated_data["email"]
+
+        self.notify_exporter_user_added(email)
+        self.add_case_note_add_export_user(self.request.user, sites, email)
```
currycoder, line 32: These could just be `self.organisation.notify_exporter_user_added(email)` etc rather than needing proxy methods on this class, right?

api/organisations/caseworker/views/users.py
```diff
@@ -0,0 +1,42 @@
...
+
+        try:
+            self.organisation = Organisation.objects.get(pk=self.kwargs["org_pk"])
+        except Organisation.DoesNotExist:
+            raise Http404()
+
+    def perform_create(self, serializer):
+        super().perform_create(serializer)
+
+        sites = serializer.initial_data["sites"]
```
currycoder, line 28: Is there a reason this can't be `validated_data`?

currycoder (approved):
A couple of questions inline, but this is looking good so they are fairly minor!

api/organisations/caseworker/views/users.py
depsiatwal: Yeah I wasn't too sure tbh was kind of trying to follow the same pattern we have for Application View

api/organisations/caseworker/views/users.py
depsiatwal: yeah I really didn't like this but validated_data converts this into objects 
I could iterate it and convert back into list of id's thoughts ?

currycoder: Well we already convert back to objects in the `Organisation.add_case_note_add_export_user()` method - so I wonder if you can just make that accept objects and simplify things all round.

PR #2177, opened 2024-09-11
Update seedinternalusers command to accept team and default queue

saruniitr:
## Change description

The users created using this command are by default belong to Admin team and queue is set to 'All cases'. This is a problem in E2E tests as when we sign in as different users their team and queue are not correct. Because of this we are having to switch team, queue as part of the test and as these persist between tests it can cause test failures.

Better approach is to seed certain users with expected roles, teams, queues etc.


depsiatwal (commented):
based on tests looks like this is backwards compatible 
just confirming ?

depsiatwal (approved):
lgtm

hnryjmes (approved):
this looks fine to me. 

I am noting that it's a bit inconsistent that we refer to roles by their readable name like "Senior Manager" but refer to teams and queues by their id, but this is more of a style thing, not a big problem. I've commented elsewhere that maybe this is what aliases should be for but I don't think we have role aliases so that wouldn't be consistent either without lots of other changes which is outside the scope here.

saruniitr:
> this looks fine to me.
> 
> I am noting that it's a bit inconsistent that we refer to roles by their readable name like "Senior Manager" but refer to teams and queues by their id, but this is more of a style thing, not a big problem. I've commented elsewhere that maybe this is what aliases should be for but I don't think we have role aliases so that wouldn't be consistent either without lots of other changes which is outside the scope here.

@hnryjmes 
Yes this is not consistent. When lite_routing was initially created we added all teams, queues etc by migrations so that we have same ids across all env and roles got left out. So ids are not consistent in all envs hence going by the names.

PR #2183, opened 2024-09-13
Add helper for creating developer intervention audit entries

kevincarrogan:
### Aim

Create a helper to more easily generate developer intervention audit logs.


api/support/helpers.py
```diff
@@ -0,0 +1,30 @@
+from contextlib import contextmanager
+
+from django.db import transaction
+
+from api.audit_trail import service as audit_trail_service
+from api.audit_trail.enums import AuditType
+
+
+@contextmanager
+def developer_intervention(*, dry_run=True):
```
markj0hnst0n, line 10: I don't know if this matters but the fact that the flags on these commands follow the python variable naming conventions doesn't feel right to me.  The normal cli convention of `dry-run` etc feels more intuitive when you're typing them.

kevincarrogan: You wouldn't be running this directly on the cli and you can't have a parameter with `-` in it in code.

But I agree that any management command that were to use this should use use `--dry-run`.

Tllew (approved):
I hate that we're making it easier to do, but it's worth doing

hopefully we can keep this maintained, might be a good ADR/readme update/ something to document so we remember it exists

markj0hnst0n (approved):

PR #2185, opened 2024-09-16
LTD-5407 create management command and added tests

Tllew:
### Aim

[LTD-](https://uktrade.atlassian.net/browse/LTD-5407)


depsiatwal (approved):
lgtm

api/applications/management/commands/update_party_name.py
```diff
@@ -0,0 +1,42 @@
...
+                self.update_field_on_party(party_id, name, new_name, additional_text)
+
+    def update_field_on_party(self, party_id, name, new_name, additional_text):
+        party = Party.objects.get(id=party_id, name=name)
+        system_user = BaseUser.objects.get(id="00000000-0000-0000-0000-000000000001")
+
+        audit_trail_service.create(
+            actor=system_user,
+            verb=AuditType.DEVELOPER_INTERVENTION,
+            target=party,
```
depsiatwal, line 36: I would change to the case rather then party so it's displayed in the case N&T

markj0hnst0n (approved):
Agree with Gurdeep's comment about adding to case but happy to approve based on that change

PR #2189, opened 2024-09-17
LTD-5432 new serializer and tests for party.name

Tllew:
### Aim

[LTD-5432](https://uktrade.atlassian.net/browse/LTD-5432)

[LTD-5432]: https://uktrade.atlassian.net/browse/LTD-5432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/parties/tests/test_serializers.py
```diff
@@ -92,3 +92,55 @@ def test_validate_party_address_invalid(self, address, error_message):
...
+                "party@name",
+                "Party name must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes",
+            ),
+        ]
+    )
+    def test_party_name_invalid(self, name, error_message):
+        serializer = PartySerializer(
+            data={"name": name},
+            partial=True,
+        )
```
saruniitr, line 139: This test is ok but do we have tests for views that uses this serializer and check for this error?

Tllew: I'll take a look, I think I changed the tests to work for this previously when party.name was added as a serializer in error and didn't change it back to test_data, but I'll get some caught errors on the go too

api/applications/views/tests/test_parties.py
```diff
@@ -72,3 +73,117 @@ def test_other_status_cannot_update_party_detail(self):
...
+class TestApplicationPartyViewValues(DataTestClient):
+
+    def setUp(self):
+        super().setUp()
+        self.application = StandardApplicationFactory(organisation=self.organisation)
+        self.case_statuses = [
+            case_status.status
+            for case_status in CaseStatus.objects.all()
+            if case_status.status not in ["draft", "applicant_editing"]
+        ]
```
saruniitr, line 87: Doesn't seem to be used.

api/applications/views/tests/test_parties.py
```diff
@@ -72,3 +73,117 @@ def test_other_status_cannot_update_party_detail(self):
...
+        self.url = reverse(
+            "applications:party",
+            kwargs={"pk": str(self.application.pk), "party_pk": str(self.party_on_application.party.pk)},
+        )
+        party = Party.objects.get(id=self.party_on_application.party.pk)
+        versions = Version.objects.get_for_object(party)
+        self.assertEqual(versions.count(), 0)
+
+        self.application.status = CaseStatus.objects.get(status="draft")
+        self.application.save()
```
saruniitr, line 182: There is https://github.com/uktrade/lite-api/blob/dev/api/applications/tests/factories.py#L220 if only draft is required

api/applications/views/tests/test_parties.py
```diff
@@ -72,3 +73,117 @@ def test_other_status_cannot_update_party_detail(self):
...
+
+        response = self.client.put(self.url, **self.exporter_headers, data=data)
+        if error:
+            self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+            self.assertEqual(response.json()["errors"], error_message)
+        else:
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
```
saruniitr, line 189: We usually avoid `if-else` as we are already parameterizing it, may I suggest separating success and failure as two separate tests? or we can probably remove success cases as other tests cover those.

saruniitr (approved):
Few final comments that I leave it with you.

api/applications/views/tests/test_parties.py
Tllew: I was considering this, playing around with merging the two as it's fairly simple but I do prefer the look of a success and failure test cases so will update

PR #2190, opened 2024-09-17
[LTD-5399] Add caseworker-specific CLEs endpoint

currycoder:
### Aim

TODO

[LTD-5399](https://uktrade.atlassian.net/browse/LTD-5399)


Tllew (approved):
Looks good to me

api/staticdata/control_list_entries/views.py
```diff
@@ -9,24 +9,19 @@
 from api.staticdata.control_list_entries.serializers import ControlListEntrySerializerWithLinks
 
 
+# TODO: Remove this endpoint when all callers are calling exporter/caseworker variants
```
hnryjmes, line 12: nice

hnryjmes (approved):
everything looks perfect, happy to approve

api/staticdata/caseworker/control_list_entries/views.py
```diff
@@ -0,0 +1,18 @@
...
+    authentication_classes = (GovAuthentication,)
+    pagination_class = None
+    serializer_class = ControlListEntriesListSerializer
+
+    def get_queryset(self):
+        include_unselectable = self.request.GET.get("include_non_selectable_for_assessment", False)
+        if include_unselectable:
+            return ControlListEntry.objects.filter(controlled=True)
+
+        return ControlListEntry.objects.filter(controlled=True, selectable_for_assessment=True)
```
saruniitr, line 18: Is it worth creating a filter backend for this?

saruniitr (commented):
one comment

api/staticdata/caseworker/control_list_entries/views.py
currycoder: I thought about creating a filter, but I think it's kind of tricky because of this approach where we are only showing `selectable_for_assessment=True` by default and including the unselectable entries due to a GET parameter.  If you can think of a way to add a filter that handles that then I'm happy to take a look but I couldn't figure out something that was clean in this case.

saruniitr: we discussed internally. Happy to merge this.

PR #2192, opened 2024-09-18
Ltd 5378 licence list bug

kevincarrogan:
### Aim

Fix licence list page when a good has been used multiple times on an application.

The issue was around advice objects being generated multiple times for the same good and one of the serializers falling over as a result.

The serializer for this page actually didn't need to load the advice at all so I've created a separate serializer that just sends what's required for the licence list page.

I've also separated out the Data Workspace serializer as well as this was also loading up information that it didn't need to.

[LTD-5378](https://uktrade.atlassian.net/browse/LTD-5378)

[LTD-5378]: https://uktrade.atlassian.net/browse/LTD-5378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/licences/serializers/view_licence.py
+++ b/api/licences/serializers/view_licence.py
@@ -15,7 +15,10 @@ from api.goods.enums import GoodControlled
 from api.goodstype.models import GoodsType
 from api.licences.enums import LicenceStatus
 from api.licences.helpers import serialize_goods_on_licence, get_approved_countries
-from api.licences.models import Licence
+from api.licences.models import (
+    GoodOnLicence,
+    Licence,
+)
 from api.parties.enums import PartyRole
 from api.parties.models import Party, PartyDocument
 from api.staticdata.control_list_entries.serializers import ControlListEntrySerializer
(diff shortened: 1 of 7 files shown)


api/licences/serializers/view_licence.py
```diff
@@ -392,22 +395,39 @@ def get_destinations(self, instance):
             ]
 
 
+class GoodOnLicenceLicenceListSerializer(serializers.ModelSerializer):
+    good_on_application_id = serializers.UUIDField(source="good.id")
+    control_list_entries = ControlListEntrySerializer(source="good.good.control_list_entries", many=True)
+    assessed_control_list_entries = ControlListEntrySerializer(source="good.control_list_entries", many=True)
```
saruniitr, line 401: Is this to future proof where these can be different in future?

kevincarrogan: I actually think that it's wrong that we're displaying `good.good.control_list_entries` on the licence page now so I want to change that but I wanted to do this in a backward compatible way by adding this new field first.

kevincarrogan: I'm also adding name to the serializer, which wasn't there before, as that field is now mostly blank so I also want to change the frontend to use name.

depsiatwal (commented):
makes sense to me 
is there any UI changes to accompany this ?

kevincarrogan:
> makes sense to me is there any UI changes to accompany this ?

Not yet. I want to deploy this first and then I can update the frontend.

depsiatwal:
> > makes sense to me is there any UI changes to accompany this ?
> 
> Not yet. I want to deploy this first and then I can update the frontend.

But we've checked it's backward compatible I mean ?

kevincarrogan:
> > > makes sense to me is there any UI changes to accompany this ?
> > 
> > 
> > Not yet. I want to deploy this first and then I can update the frontend.
> 
> But we've checked it's backward compatible I mean ?

Yes. It's backwards compatible.

depsiatwal (approved):
Just logged in as exporter and can confirm I can see Licences and Also caseworker side looks good also.

saruniitr (approved):

dejandbt (commented):
lgtm

PR #2207, opened 2024-09-25
Add separate serializer for exporter licence details

kevincarrogan:
### Aim

This creates a separate serializer for the exporter licence detail endpoint. This stops an issue with the original `get_advice` method that would fail given more than one advice object for a good on a licence.

Having an independent serializer here means that we are only sending the relevant and required information for the view on the frontend that doesn't require advice.

[LTD-5454](https://uktrade.atlassian.net/browse/LTD-5454)

[LTD-5454]: https://uktrade.atlassian.net/browse/LTD-5454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/licences/views/main.py
+++ b/api/licences/views/main.py
@@ -12,8 +12,8 @@ from api.core.decorators import authorised_govuser_roles, licence_is_editable
 from api.licences.enums import LicenceStatus
 from api.licences.models import Licence
 from api.licences.serializers.view_licence import (
+    ExporterLicenceViewSerializer,
     LicenceDetailsSerializer,
-    LicenceSerializer,
     NLRdocumentSerializer,
     LicenceListSerializer,
 )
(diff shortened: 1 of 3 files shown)


Tllew (approved):

api/licences/views/main.py
```diff
@@ -86,7 +86,7 @@ def get_queryset(self):
 
 class ViewLicence(RetrieveAPIView):
     authentication_classes = (ExporterAuthentication,)
-    serializer_class = LicenceSerializer
+    serializer_class = ExporterLicenceViewSerializer
 
     def get_queryset(self):
         return Licence.objects.filter(case__organisation_id=get_request_user_organisation_id(self.request))
```
currycoder, line 92: Did you consider breaking this out following the new exporter/caseworker API subdivision?
Asking the question, not saying we absolutely should.

kevincarrogan: I was trying to limit the amount of changes on this, especially considering this already was a separate view with its on `ExporterAuthentication`. Right now it's all nicely backwards compatible.

currycoder (approved):

PR #2210, opened 2024-09-26
working character checker for good name, party name and party address

Tllew:
### Aim

### DIFF
--- a/api/support/management/commands/get_lite_specialcharacters.py
+++ b/api/support/management/commands/get_lite_specialcharacters.py
@@ -1,0 +1,126 @@
+from api.applications.models import GoodOnApplication, PartyOnApplication, CaseStatusEnum
+from django.core.management.base import BaseCommand
+import csv
+import re
+from django.db.models import Q
+
+
+class SpecialCharacterFinder:
+    match_string = r"[^a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>]"
+    fieldnames = []
+
+    results = []
+
+    def __init__(self, filename, data):
+        self.filename = filename
+        self.results = self.check_data(data)
+        self.write_to_csv()
+
+    def check_regex(self, value):
+        match_regex = re.sub(self.match_string, "", value)
+        if len(match_regex) < len(value):
+            return set(value).difference(set(match_regex))
+
+    def get_value(self, entry):
+        return entry
+
+    def check_data(self, data):
+        results = []
+        for entry in data:
+            value = self.get_value(entry)
+            if match := self.check_regex(value):
+                results.append(self.format_results(entry, match))
+        return results
+
+    def format_results(self, data, match):
+        return {
+            "org_name": data.application.organisation.name,
+            "good_id": data.good.id,
+            "reference_code": data.application.reference_code,
+            "value": data.good.name,
+            "match": match,
+        }
+
+    def write_to_csv(self):
+        with open(f"{self.filename}.csv", "w", newline="") as csvfile:
+            writer = csv.DictWriter(csvfile, fieldnames=self.fieldnames)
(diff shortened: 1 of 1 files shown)


api/support/management/commands/get_lite_specialcharacters.py
```diff
@@ -0,0 +1,126 @@
...
+    def handle(self, *args, **options):
+
+        name_match_string = r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>]+$"
+        address_match_string = r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>\r\n]+$"
+
+        # get goods that don't match the string and are not finalised
+        goa = GoodOnApplication.objects.filter(
+            ~Q(good__name__iregex=name_match_string)
+            & ~Q(application__status__status__in=CaseStatusEnum._terminal_statuses)
+        )
```
saruniitr, line 110: Could this also be extended to `Good`? we can may be ask them to edit if possible or archive and create new ones before they are used on an application preventing this further?

api/support/management/commands/get_lite_specialcharacters.py
```diff
@@ -0,0 +1,142 @@
...
+    def check_regex(self, value):
+        match_regex = re.sub(self.match_string, "", value)
+        if len(match_regex) < len(value):
+            return set(value).difference(set(match_regex))
+
+    def get_value(self, entry):
+        return entry
+
+    def get_id(self, entry):
+        return entry
```
saruniitr, line 29: does this need to be `entry.id`?

api/support/management/commands/get_lite_specialcharacters.py
```diff
@@ -0,0 +1,142 @@
...
+        return {
+            "org_name": data.application.organisation.name,
+            "good_id": data.good.id,
+            "reference_code": data.application.reference_code,
+            "value": data.good.name,
+            "match": match,
+        }
+
+    def write_to_csv(self):
+        with open(f"{self.filename}.csv", "w", newline="") as csvfile:
```
saruniitr, line 52: I think it is better you take the complete filename `good_names.csv` so that it is clear that the output is csv

api/support/management/commands/get_lite_specialcharacters.py
```diff
@@ -0,0 +1,142 @@
...
+    def get_id(self, entry):
+        return entry
+
+    def check_data(self, data):
+        results = []
+        for entry in data:
+            id = self.get_id(entry)
+            if not self.unique_result.get(id):
+                value = self.get_value(entry)
+                if match := self.check_regex(value):
```
saruniitr, line 37: walrus operator, perhaps first usage in lite.

api/support/management/commands/get_lite_specialcharacters.py
```diff
@@ -0,0 +1,142 @@
...
+                    results.append(self.format_results(entry, match))
+                    self.unique_result[id] = True
```
currycoder, line 39: If this is just storing a flat single value, could it be a set instead e.g. 
```
self.unique_matches = set()
...
self.unique_matches.add(match)
```

api/support/management/commands/get_lite_specialcharacters.py
```diff
@@ -0,0 +1,142 @@
...
+        return results
+
+    def format_results(self, data, match):
+        return {
+            "org_name": data.application.organisation.name,
+            "good_id": data.good.id,
+            "reference_code": data.application.reference_code,
+            "value": data.good.name,
```
currycoder, line 47: is this column actually `good_name` instead of `value`?

currycoder (commented):
A couple of comments inline

PR #2216, opened 2024-09-30
Update the lock file

kevincarrogan:
This is to solve issues where the elasticsearch packages would cause issues when installing requirements.


depsiatwal (approved):
lgtm

PR #2217, opened 2024-10-01
Move flagging rule functions into lite_routing

kevincarrogan:
### Aim

This is the conclusion in moving the flagging rule functions into `lite_routing` that began with https://github.com/uktrade/lite-routing/pull/271.

The other PR duplicated the code that was originally in `lite-api` and this PR then switches all of the imports to point to that new code.

[LTD-5482](https://uktrade.atlassian.net/browse/LTD-5482)

[LTD-5482]: https://uktrade.atlassian.net/browse/LTD-5482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


markj0hnst0n:
The ReadMe still refers to the old place where the code was kept.  Not sure whether we want it pointing to the right place now or not.

markj0hnst0n:
Do you still need the `api_workflow_code_exists()` function in `test_flagging_engine.py` now that it's being deleted?

kevincarrogan:
> The ReadMe still refers to the old place where the code was kept. Not sure whether we want it pointing to the right place now or not.

I'm going to be doing a fair few incremental changes so I was planning on updating the documentation once that was done instead of me updating it for each change because some of it will become redundant in later changes anyway.

kevincarrogan:
> Do you still need the `api_workflow_code_exists()` function in `test_flagging_engine.py` now that it's being deleted?

I do whilst the `lite_routing` sha hasn't been updated in `lite_api`. It's just how our code checks out different branches that means I can't delete it yet.

markj0hnst0n (approved):

PR #2235, opened 2024-10-15
LTD: Add endpoint for licences list

saruniitr:
### Aim


api/data_workspace/v2/serializers.py
```diff
@@ -0,0 +1,18 @@
+from rest_framework import serializers
+
+from api.licences.models import Licence
+
+
+class LicenceStatusSerializer(serializers.Serializer):
+    name = serializers.CharField(source="*")
+
+
+class SIELLicenceSerializer(serializers.ModelSerializer):
```
kevincarrogan, line 10: Let’s just call this Licence for now.

api/data_workspace/v2/serializers.py
```diff
@@ -0,0 +1,18 @@
+from rest_framework import serializers
+
+from api.licences.models import Licence
+
+
+class LicenceStatusSerializer(serializers.Serializer):
```
kevincarrogan, line 6: Let’s not include status in this first prototype.

api/data_workspace/v2/tests/bdd/scenarios/licence_statuses.feature
```diff
@@ -0,0 +1,18 @@
...
+        | expired     |
+        | issued      |
+        | reinstated  |
+        | revoked     |
+        | surrendered |
+        | suspended   |
+    Then there are no other licence statuses
+
+
+Scenario: Issued licence is included in the extract
```
kevincarrogan, line 16: Can we have this in a separate BDD scenario file? I’d like it if each endpoint had its own scenario file.

api/data_workspace/v2/views.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class SIELLicencesListView(viewsets.ReadOnlyModelViewSet):
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    pagination_class = LimitOffsetPagination
+    renderer_classes = tuple(api_settings.DEFAULT_RENDERER_CLASSES) + (PaginatedCSVRenderer,)
+    serializer_class = SIELLicenceSerializer
+
+    def get_queryset(self):
+        return Licence.objects.exclude(status=LicenceStatus.DRAFT)
```
kevincarrogan, line 31: It would be good to see a test for this somewhere. Possibly as a distinct scenario in the BDD file for the endpoint.

api/data_workspace/v2/serializers.py
saruniitr: part of cherry-picked commit, will remove

api/data_workspace/v2/tests/bdd/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+@pytest.fixture(autouse=True)
+def system_user(db):
+    return BaseUserFactory(id=SystemUser.id)
+
+
+@pytest.fixture()
+def unpage_data(client):
+    def _unpage_data(url):
+        unpaged_results = []
+        while True:
```
hnryjmes, line 18: this works fine and looks great, although I am surprised there's not a way of doing this using the existing libraries we have

hnryjmes (approved):
everything looks good to me, happy to approve

api/data_workspace/v2/tests/bdd/scenarios/licences.feature
```diff
@@ -0,0 +1,10 @@
+@db
+Feature: Licences
+
+Scenario: Check that draft licences are not included in the extract
+    Given a standard draft licence is created
+    Then the draft licence is not included in the extract
+
+Scenario: Issued licence is included in the extract
+    Given a standard licence is issued
+    Then the issued licence is included in the extract
```
kevincarrogan, line 10: I feel that it might be nice to have at test that shows that a licence transitioning from draft to issued makes a difference.

api/data_workspace/v2/views.py
```diff
@@ -0,0 +1,42 @@
...
+        # When an application with all goods as NLR is finalised then the current code
+        # creates a licence however the goods on this licence will be empty. This
+        # will skew licence data hence exclude them
+        return (
+            Licence.objects.prefetch_related("goods")
+            .annotate(num_licensed_goods=Count("goods"))
+            .exclude(
+                status__in=[
+                    LicenceStatus.DRAFT,
+                    LicenceStatus.CANCELLED,
```
kevincarrogan, line 31: Could we extract these statuses out into a named thing? For CaseStatus we have a range of named lists so I think it would be good to do the same here.

api/data_workspace/v2/views.py
```diff
@@ -0,0 +1,42 @@
...
+                ]
+            )
+            .exclude(num_licensed_goods=0)
+            .filter(
+                case__status__status__in=[
+                    CaseStatusEnum.FINALISED,
+                    CaseStatusEnum.SUPERSEDED_BY_EXPORTER_EDIT,
+                ],
```
kevincarrogan, line 39: Could we come up with a named list for these statuses?

api/data_workspace/v2/views.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class LicenceDecisionViewSet(viewsets.ReadOnlyModelViewSet):
+    authentication_classes = (DataWorkspaceOnlyAuthentication,)
+    pagination_class = DisableableLimitOffsetPagination
+    queryset = (
+        Case.objects.filter(
+            licences__generatedcasedocument__template_id=SIEL_TEMPLATE_ID,
+            licences__generatedcasedocument__visible_to_exporter=True,
+            licences__generatedcasedocument__safe=True,
```
hnryjmes, line 31: is this just an extra precaution? the document is generated internally so I wouldn't think this would ever not be safe=True ?

saruniitr: `safe=True` indicates no virus found

api/data_workspace/v2/views.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    refused_qs = Case.objects.filter(
+        id__in=GeneratedCaseDocument.objects.filter(
+            template_id=SIEL_REFUSAL_TEMPLATE_ID,
+            visible_to_exporter=True,
+            safe=True,
+        ).values_list("case", flat=True)
+    )
+
+    queryset = (issued_qs | refused_qs).distinct().order_by("-reference_code")
```
kevincarrogan, line 39: Isn't this just a very long-winded way to do the same thing as 

```
Case.objects.filter(
        licences__generatedcasedocument__template_id__in=[SIEL_TEMPLATE_ID, SIEL_REFUSAL_TEMPLATE_ID],
        licences__generatedcasedocument__visible_to_exporter=True,
        licences__generatedcasedocument__safe=True,
    ).distinct()
```

api/data_workspace/v2/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+        model = Case
+        fields = (
+            "id",
+            "reference_code",
+            "decision",
+            "decision_made_at",
+        )
+
+    def get_decision(self, case):
+        if case.licences.count():
```
kevincarrogan, line 32: Can we name this in a way that clarifies what the `case.licences.count()` is doing and why it's important.

api/data_workspace/v2/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+            return LicenceDecisionType.ISSUED
+        else:
+            return LicenceDecisionType.REFUSED
```
kevincarrogan, line 35: This looks like it's mutually exclusive, but is it possible that something has been issued and then refused or vice versa?

saruniitr: I checked that, currently there are no such cases.
Because of amendments it is not possible going forward also but there is a possibility with Appeals.

api/data_workspace/v2/views.py
saruniitr: Actually no because there won't be a licence for refused cases.
That was the initial change and as expected refusal cases were not filtered.

api/data_workspace/v2/serializers.py
saruniitr: You mean just by adding a comment or annotate to a field and use that?

api/data_workspace/v2/views.py
kevincarrogan: Are both generated documents attached to the case and we could just skip the `licences__` bit entirely for both?

kevincarrogan: It would also be good to have some joined up thinking based on https://github.com/uktrade/lite-api/pull/2254/files#r1819089904 as it looks like we are diverging slightly with this particular heuristic and the future implementation.

PR #2243, opened 2024-10-22
LTD-5537-Licence-document-cutting-off-details

markj0hnst0n:
### Aim

Split good name onto different lines on Licence pdf by comma to ensure full good details appear on licence pdf.

Investigated and there are 2727 out of 14574 good records which contain commas so this a lighter touch on the existing data than adding splitting across any character types although this could be introduced.

[LTD-5537](https://uktrade.atlassian.net/browse/LTD-5537)

[LTD-5537]: https://uktrade.atlassian.net/browse/LTD-5537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/letter_templates/templates/letter_templates/refusal.html
+++ b/api/letter_templates/templates/letter_templates/refusal.html
@@ -13,7 +13,7 @@
   <ul class="govuk-list govuk-list--bullet">
     {% for good in goods.refuse %}
       <li>
-        {{good.good.name}}
+        {{good.good.name|split_good_name_across_lines}}
       </li>
     {% endfor %}
   </ul>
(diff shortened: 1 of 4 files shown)


Tllew (approved):
Is there a minimum length you'd want to implement before doing the replace?

Looks good to me, happy to approve

markj0hnst0n:
@Tllew I was thinking about this and I figured we have as much space below as we like (within reason) but not to the right margin so any new line is acceptable.

PR #2249, opened 2024-10-23
[Spike] Add LicenceDecision model

hnryjmes:
### Aim

[LTD-5577](https://uktrade.atlassian.net/browse/LTD-5577)

[LTD-5577]: https://uktrade.atlassian.net/browse/LTD-5577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/enums.py
```diff
@@ -408,3 +408,19 @@ class EnforcementXMLEntityTypes:
         (SITE, "site"),
         (ORGANISATION, "organisation"),
     ]
+
+
+class LicenceDecisionType:
+    ISSUED = "issued"
+    REFUSED = "refused"
+    NO_LICENCE_REQUIRED = "no_licence_required"
+    WITHDRAWN = "withdrawn"
```
saruniitr, line 417: this is exporter decision so not necessarily part of this.

api/cases/views/views.py
```diff
@@ -1006,6 +1005,17 @@ def put(self, request, pk):
             },
         )
 
+        for decision in required_decisions:
```
saruniitr, line 1008: this does not include 'revoked' which needs handling separately.

api/cases/enums.py
kevincarrogan: I would remove all of the decisions that we aren't sure of for now, which would mean all we are left with is "ISSUED".

api/cases/models.py
```diff
@@ -686,3 +687,20 @@ class EnforcementCheckID(models.Model):
     id = models.AutoField(primary_key=True)
     entity_id = models.UUIDField(unique=True)
     entity_type = models.CharField(choices=EnforcementXMLEntityTypes.choices, max_length=20)
+
+
+class LicenceDecision(models.Model):
+    case = models.ForeignKey(Case, on_delete=models.DO_NOTHING)
+    decision = models.CharField(choices=LicenceDecisionType.choices, max_length=50, null=False, blank=False)
+    decision_made_at = models.DateTimeField(null=False, blank=False)
```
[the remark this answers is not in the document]
kevincarrogan: Agreed.

api/cases/models.py
```diff
@@ -686,3 +687,20 @@ class EnforcementCheckID(models.Model):
...
+
+    # These are the AdviceType types that can be mapped to a licence decision.
+    # The remaining types 'not_applicable', 'conflicting', 'inform' cannot be
+    # mapped to a licence decision.
+    advice_to_licence_decision_mapping = {
+        AdviceType.APPROVE: LicenceDecisionType.ISSUED,
+        AdviceType.PROVISO: LicenceDecisionType.ISSUED,
+        AdviceType.REFUSE: LicenceDecisionType.REFUSED,
+        AdviceType.NO_LICENCE_REQUIRED: LicenceDecisionType.NO_LICENCE_REQUIRED,
+    }
```
kevincarrogan, line 706: Ultimately I don't think this mapping would work.

api/cases/views/views.py
kevincarrogan: I think we should strip this back just so it's dealing with issued instead of the other licence types for now.

hnryjmes:
work has moved to https://github.com/uktrade/lite-api/pull/2254

PR #2253, opened 2024-10-25
Temporarily skip rerun routing rules test

kevincarrogan:
### Aim

Temporarily skip rerun routing rules test.

I'm stuck in a bit of a chicken and egg situation where this rerun routing rule test is becoming very difficult to maintain in a backward compatible way.

However, this is only a problem for the tests and not a problem in reality due to the fact that I have a fix for this that will go with the sha update that matches against lite-routing.

Given that this is just a testing problem, it's temporary and it's also in an area of the code that we've "turned off" I've taken the decision that skipping this test is the most sensible option.

[LTD-5521](https://uktrade.atlassian.net/browse/LTD-5521)

[LTD-5521]: https://uktrade.atlassian.net/browse/LTD-5521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/tests/test_rerun_routing_rules.py
+++ b/api/cases/tests/test_rerun_routing_rules.py
@@ -1,14 +1,10 @@
-import pytest
+import unittest
 
 from django.urls import reverse
 from rest_framework import status
 
-from api.audit_trail.enums import AuditType
-from api.audit_trail.models import Audit
 from api.staticdata.statuses.enums import CaseStatusEnum
 from api.staticdata.statuses.libraries.get_case_status import get_case_status_by_status
-from api.staticdata.statuses.models import CaseStatus
-from api.workflow.routing_rules.models import RoutingRule
 from test_helpers.clients import DataTestClient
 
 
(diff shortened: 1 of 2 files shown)


api/cases/tests/test_rerun_routing_rules.py
```diff
@@ -27,6 +23,7 @@ def setUp(self):
             additional_rules=[],
         )
 
+    @unittest.skip("Skipping due to backwards compatability issues")
     def test_rules_rerun(self):
```
currycoder, line 27: Didn't we get rid of this button anyway?

kevincarrogan: We wrapped it in an always deny rule with the aim that we would remove it if no one complained about it, but we've not done that yet.

https://github.com/uktrade/lite-frontend/blob/bdf6a8aa4fcad691126d48d177bf680a1ea76b06/caseworker/core/rules.py#L203

That's one of the main reasons why I'm not too fussed about temporarily skipping this test and feeling that it's not worth putting in additional effort to work around this problem with the test.

saruniitr (approved):

PR #2254, opened 2024-10-25
Add LicenceDecision model an back populate

hnryjmes:
## Change description

Add LicenceDecision model to concretely define a case outcome.
A Case can have multiple outcomes so when the case is finalised we create instances of this model for each outcome.

This is mainly used for reporting purposes.
Add another migration to back populate these instances for all Cases that have reached terminal status.

View is refactored to make it simpler and group related code together in the model.

[LTD-5588]: https://uktrade.atlassian.net/browse/LTD-5588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/views/views.py
```diff
@@ -941,6 +938,8 @@ def put(self, request, pk):
                     },
                 )
 
+                LicenceDecision.objects.create(case=case, decision=LicenceDecisionType.ISSUED, licence=licence)
```
kevincarrogan, line 941: I wonder if it would be good to set an explicit audit entry as well to show that a licence decision was made.

kevincarrogan: In the prototype we are using the point in time when the `GeneratedCaseDocument` is generated and using that as a proxy.

If I look at line 924 I can see that it's not creating the document but finding it so this doesn't quite match our current heuristic that the `LicenceDecision` is being made at the point the document is generated.

Is there any reason why we are changing this logic?

hnryjmes: I'm proposing it as a way we could make our data more accurate, although it would introduce a small inconsistency between historical and future data (and if only the date makes it into the final report, possibly this wouldn't be noticed by anyone).

By having it in the finalise view, as close to the point in time as possible to when the document is made visible to the exporter, in my mind this is a more accurate representation of the date and time the licence was issued. In that sense it can be seen as improving the quality of the data.

However if we'd rather prioritise things being consistent over maximising accuracy, we could move this code to somewhere like: https://github.com/uktrade/lite-api/blob/dev/api/cases/generated_documents/views.py#L128 which is right when the GeneratedCaseDocument is created.

kevincarrogan: Do these two events actually happen at the same time?

What is the potential difference if we take one over the other?

Why are we not just using 

```
                audit_trail_service.create(
                    actor=request.user,
                    verb=AuditType.CREATED_FINAL_RECOMMENDATION,
                    target=case,
                    payload={
                        "case_reference": case.reference_code,
                        "decision": AdviceType.APPROVE,
                        "licence_reference": licence.reference_code,
                    },
                )
```

as the indicator as opposed to the GeneratedDocument to then make everything consistent?

kevincarrogan: Having these things inconsistent does lead to an issue if/when we have a data migration to back populate this as we would be generating models based on one heuristic when other rows have already been generated for a different one.

It would be good to see the data migration in this PR as well so that if we did put this live it would all happen at once.

hnryjmes: @saruniitr hi, I think you said before you'd thought a bit about the data migration we'd need to backpopulate historical records, before I start my own work on that, can you share where you got up to? then hopefully I can build on what you've done or take your advice on it

hnryjmes: also FYI, this PR has now been updated so that the object creation is in the `generated_documents` view and not the `finalise` view, as I want to try to keep things simple for us in the short term

api/cases/models.py
```diff
@@ -307,6 +309,123 @@ def set_sub_status(self, sub_status_id):
...
+        case = self.get_case()
+        required_decisions = get_required_decision_document_types(case)
+        breakpoint()
+
+        # Inform letter isn't required for finalisation
+        if AdviceType.INFORM in required_decisions:
+            required_decisions.remove(AdviceType.INFORM)
+
+        # Check that each decision has a document
+        # Excluding approve (done in the licence section below)
```
saruniitr, line 325: need to remove this comment

api/cases/migrations/0068_populate_licence_decisions.py
```diff
@@ -0,0 +1,58 @@
...
+SIEL_LICENCE_TEMPLATE_ID = "d159b195-9256-4a00-9bc8-1eb2cebfa1d2"
+SIEL_REFUSAL_TEMPLATE_ID = "074d8a54-ee10-4dca-82ba-650460650342"
+
+
+@transaction.atomic
+def populate_licence_decisions(apps, schema_editor):
+    GeneratedCaseDocument = apps.get_model("generated_documents", "GeneratedCaseDocument")
+    LicenceDecision = apps.get_model("cases", "LicenceDecision")
+
+    case_statuses = [*CaseStatusEnum.terminal_statuses(), CaseStatusEnum.REOPENED_DUE_TO_ORG_CHANGES]
```
saruniitr, line 19: The organisation name has been updated for some of the finalised cases. When that is done the current code updates the status to this value. This is extended to consider those cases also.

api/cases/models.py
```diff
@@ -307,6 +309,123 @@ def set_sub_status(self, sub_status_id):
...
+
+    @transaction.atomic
+    def finalise(self, request):
+        from api.licences.models import Licence
+        from api.cases.libraries.finalise import (
+            get_required_decision_document_types,
+            remove_flags_on_finalisation,
+            remove_flags_from_audit_trail,
+        )
+        from api.audit_trail import service as audit_trail_service
```
saruniitr, line 342: there are circular dependencies.
Also this is mainly lift and shift from the view, further improvements possible.

api/cases/models.py
```diff
@@ -307,6 +309,123 @@ def set_sub_status(self, sub_status_id):
             payload={"sub_status": self.sub_status.name, "status": CaseStatusEnum.get_text(self.status.status)},
         )
 
+    def validate_decision_documents(self):
+        from api.cases.libraries.finalise import get_required_decision_document_types
+        from api.cases.generated_documents.models import GeneratedCaseDocument
+
+        case = self.get_case()
+        required_decisions = get_required_decision_document_types(case)
+        breakpoint()
```
hnryjmes, line 318: breakpoint still here

PR #2262, opened 2024-10-31
Add DRF CSV package

kevincarrogan:
### Aim

This allows us to exporter CSV files from our endpoints.

I am adding this as I'm having to maintain a long lived branch that keeps conflicting because of the Pipfile.lock differences.

[LTD-5597](https://uktrade.atlassian.net/browse/LTD-5597)


Pipfile
```diff
@@ -79,6 +79,7 @@ dj-database-url = "~=2.2.0"
 certifi = "~=2024.7.4"
 pytz = "~=2024.1"
 drf-spectacular = "~=0.27.2"
+djangorestframework-csv = "*"
```
hnryjmes, line 82: would it be possible to change this to use `~=` ?

kevincarrogan: Yep. I always forget that.

kevincarrogan: Done.

hnryjmes (approved):

currycoder (approved):

PR #2271, opened 2024-11-07
LTD-5588: Introduce Licence decision model and back populate

saruniitr:
## Change description

Add `LicenceDecision` to record a case outcome and accurately report the data.
A case can have multiple outcomes and we create instances of this model for each outcome.

Back populate the data so the reports can be generated for historic cases.

This model records the decision and the time this decision is made. This is taken as the time when the case is finalised and decision documents are published. For some of the historic cases this time not available in audit logs so we take the document generation time instead.

[LTD-5588](https://uktrade.atlassian.net/browse/LTD-5588)

[LTD-5588]: https://uktrade.atlassian.net/browse/LTD-5588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/models.py
```diff
@@ -307,6 +318,109 @@ def set_sub_status(self, sub_status_id):
...
+        from api.licences.models import Licence
+
+        try:
+            licence = Licence.objects.get_draft_licence(self)
+        except Licence.DoesNotExist:
+            # This is not an error as there won't be a licence for refusal cases
+            licence = None
+
+        if AdviceType.APPROVE in decisions and licence:
+            licence.decisions.set([Decision.objects.get(name=decision) for decision in decisions])
```
hnryjmes, line 351: we talked about this before, so I know you're aware of this, but I still think the old model `Decision` could be renamed to make clearer what it is, or maybe we can do away with it entirely. it seems to relate to `AdviceType` only so `AdviceDecision` perhaps could be a better name if we need to keep it? 

I can see there are many types there not in `LicenceDecision` e.g. `proviso` so no possibility of just replacing this with `LicenceDecision` unfortunately.

saruniitr: Actually I want to remove this line completely but need to see if it breaks anything.
This is already a significant change so do not want to add more changes than necessary in this PR.
Renaming/removing that `Decision` model will be handled as a follow up as it is not critical for this work.

api/cases/enums.py
```diff
@@ -408,3 +411,36 @@ class EnforcementXMLEntityTypes:
...
+        AdviceType.APPROVE: ISSUED,
+        AdviceType.REFUSE: REFUSED,
+    }
+
+    @classmethod
+    def templates(cls):
+        return {
+            cls.ISSUED: SIEL_LICENCE_TEMPLATE_ID,
+            cls.REFUSED: SIEL_REFUSAL_TEMPLATE_ID,
+            cls.REVOKED: None,
```
hnryjmes, line 437: this is more of a business question I suppose - is there really no document when a licence is revoked? how does the exporter hear about the decision?

saruniitr: We send a notification to exporter. I believe they are also informed by other channels when a decision is made to revoke.

kevincarrogan (approved):

saruniitr:
@hnryjmes can you take another look? any other comments on this?

hnryjmes (approved):
nice work

PR #2279, opened 2024-11-14
LTD-5639: Attach licence to historical Licence decision instances

saruniitr:
## Change description

When Licence decisions are back populated the licence field is not set for the issued decisions. Add migration to populate this field.

As per requirement we take the first licence in case where there are multiple licences (eg re-issued case)

[LTD-5639](https://uktrade.atlassian.net/browse/LTD-5639)

[LTD-5639]: https://uktrade.atlassian.net/browse/LTD-5639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevincarrogan (approved):

api/cases/migrations/0069_attach_licence_to_licence_decision.py
```diff
@@ -0,0 +1,42 @@
...
+from api.licences.enums import LicenceStatus
+
+
+def get_licence_for_decision(licence_decision):
+    if licence_decision.decision != LicenceDecisionType.ISSUED:
+        return None
+
+    licences = licence_decision.case.licences.exclude(status=LicenceStatus.DRAFT)
+
+    return licences.earliest("created_at")
```
kevincarrogan, line 15: I’ve just realised that this is slightly incorrect because it can’t directly be ported from the serializer to the migration. 

The problem here is that a case with multiple issued decisions will only ever get the first licence attached when it may not be the right licence. 

In the context of the serializer where I mocked this in the first place that works because we do some filtering beforehand that ensures we are only dealing with the first issued decision which we know definitely relates to the first licence, however in this case we aren’t doing any filtering.

PR #2282, opened 2024-11-14
LTD-5634-edit-create-user-api-changes 

depsiatwal:
This is a API backend changes for larger part of a user story where privilege escalation means users can elevate their privileges. 

This endpoint is to be used with the new frontend endpoints 

A new endpoint was created to make this cleaner and more inline with our new /caseworkers/ pattern. 
I appreciate this is still using the GovUserCreateOrUpdateSerializer Serializer but it feels a bit too heavy weight as the moment to shift that at the same time. 

- List end point is used to make sure we don't have duplicate emails via form validation. The idea is we may use this endpoint further to filter users using the new endpoint.  
- Update is method is used by the new Caseworker Edit view

Frontend PR that's uses new endpoints 

https://github.com/uktrade/lite-frontend/pull/2237


api/gov_users/caseworker/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+    serializer_class = GovUserCreateOrUpdateSerializer
+    lookup_field = "pk"
+
+    def update(self, request, pk):
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data, partial=True, is_creating=False)
+        if serializer.is_valid():
+            # Remove user from assigned cases
+            if request.data.get("status") == GovUserStatuses.DEACTIVATED:
```
currycoder, line 41: Can you get this from `serializer.cleaned_data` instead of grabbing the unsanitized data from request?

api/gov_users/caseworker/views.py
```diff
@@ -0,0 +1,47 @@
...
+                instance.unassign_from_cases()
```
currycoder, line 42: This should probably be in the serializer.  I get the desire to not create a new serializer for this but it does mean that we are starting off with a bit of an untidy View.  If we had our own new serializer then we could probably drop this `update` method entirely

api/gov_users/caseworker/views.py
```diff
@@ -0,0 +1,47 @@
...
+from api.gov_users.serializers import GovUserCreateOrUpdateSerializer
+from api.users.models import GovUser
+from api.users.serializers import GovUserViewSerializer
+from rest_framework.generics import ListAPIView, UpdateAPIView
+from .permissions import CanCaseworkersManageUser
+
+
+class GovUsersList(ListAPIView):
+    authentication_classes = (GovAuthentication,)
+    serializer_class = GovUserViewSerializer
```
currycoder, line 15: I've looked at this serializer and it is **whacky** with the caching stuff it does.  IMO this is an easy one to replace and could even retain caching-like functionality with use of a `select_related()`

api/gov_users/caseworker/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+class GovUserUpdate(UpdateAPIView):
+    authentication_classes = (GovAuthentication,)
+    permission_classes = [
+        CanCaseworkersManageUser,
+    ]
+
+    queryset = GovUser.objects.all()
+
+    serializer_class = GovUserCreateOrUpdateSerializer
```
currycoder, line 33: Looked at this serializer also.  Agreed that it's more complex to replace however looking at it I think `create()` and `update()` would mostly stay the same and `validate()` could get split out in to field-level validation methods instead.  `self.is_creating` can be dropped and replaced by `is_creating = bool(self.instance)`

currycoder (commented):
Love to see the new API endpoint for this.  Would just **really** love to see serializers moved over too.  Realise that there is a time element to this work though so perhaps a follow-up?

depsiatwal:
> Love to see the new API endpoint for this. Would just **really** love to see serializers moved over too. Realise that there is a time element to this work though so perhaps a follow-up?

Yeah agree hence seializer in this case is doing too much validation when we need to make further changes I suggest we shift serializer when business need arises.

api/gov_users/caseworker/permissions.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import permissions
+from api.core.constants import Roles
+from api.core.permissions import assert_user_in_role
+
+
+class CanCaseworkersManageUser(permissions.BasePermission):
+    def has_permission(self, request, view):
+        if request.data.get("default_queue") and len(request.data) < 2:
+            return True
```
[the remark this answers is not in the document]
depsiatwal: Users can only update default_queue queue they're not allowed to change anything else.

api/gov_users/caseworker/serializer.py
```diff
@@ -0,0 +1,103 @@
...
+        model = GovUser
+        fields = (
+            "id",
+            "email",
+            "first_name",
+            "last_name",
+            "status",
+            "team",
+            "role",
+            "default_queue",
```
currycoder, line 101: The original GovUserViewSerializer had this as a `TinyQueueSerializer`, do we not need that anymore?  Just the ID is ok?

api/gov_users/caseworker/serializer.py
```diff
@@ -0,0 +1,103 @@
...
+        default_queue = value or self.instance.default_queue
+        team = self.initial_data.get("team") or self.instance.team_id
+        is_system_queue = str(default_queue) in SYSTEM_QUEUES.keys()
+        try:
+            if not is_system_queue:
+                queue = Queue.objects.get(id=default_queue)
+                if str(queue.team_id) != team:
+                    raise serializers.ValidationError("select a valid queue for team")
+        except Queue.DoesNotExist:
+            raise serializers.ValidationError("select a valid queue")
```
currycoder, line 59: Minor thing but would suggest the try/except goes only around the `Queue.objects.get()` line.  Saves a little bit of indentation.

api/gov_users/caseworker/views.py
```diff
@@ -0,0 +1,27 @@
...
+
+class GovUsersList(ListAPIView):
+    authentication_classes = (GovAuthentication,)
+    serializer_class = GovUserViewSerializer
+
+    def get_queryset(self):
+        queryset = GovUser.objects.all().select_related("role", "team", "baseuser_ptr")
+        email = self.request.GET.get("email")
+        if email is not None:
+            queryset = queryset.filter(baseuser_ptr__email=email)
```
currycoder, line 17: Worth making this `__icontains=` to avoid upper/lower case matching issues?

currycoder (approved):
Thanks for your patience - this is looking great 🙌

api/gov_users/caseworker/serializer.py
depsiatwal: for the purpose of the view where this is used what we have is ok. 
We can add extra (obviously avoid Monmouth serialisers as when we require the data. )

api/gov_users/caseworker/serializer.py
```diff
@@ -0,0 +1,103 @@
...
+from api.teams.serializers import TeamSerializer
+from api.users.models import GovUser, Role
+
+
+class GovUserUpdateSerializer(serializers.ModelSerializer):
+
+    # from baseuser
+    id = serializers.ReadOnlyField(source="baseuser_ptr_id")
+
+    # status = serializers.ChoiceField(choices=GovUserStatuses.choices, default=GovUserStatuses.ACTIVE)
```
markj0hnst0n, line 19: Is this still required?

api/users/tests/test_view.py
```diff
@@ -55,3 +62,145 @@ def test_user_select_organisations_exclude_rejected_orgs(self):
         active_organisations = response.json()["organisations"]
         self.assertEqual(len(active_organisations), 1)
         self.assertNotIn(organisation.name, [name for name in active_organisations])
+
+    def test_gov_user_list_filtered(self):
+        GovUserFactory()
```
markj0hnst0n, line 67: I think this can be removed?

markj0hnst0n (approved):
A few minor comments but happy to approve

api/gov_users/caseworker/serializer.py
depsiatwal: good point removed

api/users/tests/test_view.py
depsiatwal: nah the first one creates another entry for filtering

PR #2283, opened 2024-11-14
LTD-5642: Add unique constraint for licence and good in GoodOnLicence model

saruniitr:
## Change description

It should not be possible to have the same good_on_application twice on a given licence.

[LTD-5642](https://uktrade.atlassian.net/browse/LTD-5642)

[LTD-5642]: https://uktrade.atlassian.net/browse/LTD-5642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevincarrogan (approved):

depsiatwal (approved):
lgtm

api/licences/models.py
```diff
@@ -158,3 +158,6 @@ class GoodOnLicence(TimestampableModel):
     usage = models.FloatField(null=False, blank=False, default=0)
     quantity = models.FloatField(null=False, blank=False)
     value = models.DecimalField(max_digits=15, decimal_places=2, null=False, blank=False)
+
+    class Meta:
+        unique_together = ("licence", "good")
```
currycoder, line 163: Are there any records in the DB currently which break this constraint?

saruniitr: No, it is also no possible because of the way we pass data to create these in the API.
Please see my comment above.
Our code is already doing this implicitly but we also want to enforce this at db level.

kevincarrogan: The reason why I wanted this was because as part of creating the conceptual data model I couldn't easily find the answer to the question as to whether a GoodOnApplication could appear multiple times on a licence.

For me this was more about documentation than enforcing the constraint, although that enforcement is valuable to literally enforce that the documentation is accurate.

PR #2291, opened 2024-11-18
LTD-5654 Create simple endpoint for dbt platform healthcheck

markj0hnst0n:
### Aim

DBT platform team needs to hit an endpoint and get a 200 to make sure that the django app is up and running

[LTD-5654](https://uktrade.atlassian.net/browse/LTD-5654)

[LTD-5654]: https://uktrade.atlassian.net/browse/LTD-5654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/conf/settings.py
```diff
@@ -139,12 +140,17 @@
 
 if not IS_ENV_DBT_PLATFORM:
     INSTALLED_APPS += [
-        "health_check.db",
         "health_check.cache",
+        "health_check.db",
         "health_check.storage",
-        "health_check.contrib.migrations",
     ]
```
kevincarrogan, line 146: If we are splitting out the healthchecks into subsets then I think this whole chunk where we toggle healthchecks on and off for DBT Platform can go away.

api/conf/settings.py
```diff
@@ -139,12 +140,17 @@
...
 
+HEALTH_CHECK = {
+    "SUBSETS": {
+        "startup-liveness-probe": ["MigrationsHealthCheck"],
+    },
+}
+
```
kevincarrogan, line 153: Is the plan that DBT Platform will hit this endpoint?

Is there another endpoint that runs all of the checks that we could use for our own purposes or do we need to create another subset?

api/conf/settings.py
```diff
@@ -137,13 +141,12 @@
...
-if not IS_ENV_DBT_PLATFORM:
-    INSTALLED_APPS += [
-        "health_check.db",
-        "health_check.cache",
-        "health_check.storage",
-        "health_check.contrib.migrations",
-    ]
+
+HEALTH_CHECK = {
+    "SUBSETS": {
```
depsiatwal, line 146: for my understanding how is this used and called ?
Also worth adding comment so dev understand what this does in the future.

markj0hnst0n: Basically splits out the selected health checks in the list into a callable url. 

`BASEURL.com/healthcheck/startup-liveness-probe/`

kevincarrogan (approved):

depsiatwal (commented):
LGTM possible to add a test to ensure it returns as expected

depsiatwal (approved):
lgtm

PR #2294, opened 2024-11-20
Add destination endpoint

hnryjmes:
The basic test is for the response data from the endpoint being in the right format as expected. There is another test for deleting a party. When a case is in non-draft state, deleting a party uses a different `.delete()` method which actually does not delete the record, but instead sets a `deleted_at` field with the current datetime. It is important that these "deleted" parties do not appear as destinations in the published data so the test checks they are not included.

[LTD-5650](https://uktrade.atlassian.net/browse/LTD-5650)

[LTD-5650]: https://uktrade.atlassian.net/browse/LTD-5650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/data_workspace/v2/urls.py
+++ b/api/data_workspace/v2/urls.py
@@ -1,8 +1,8 @@
-from api.data_workspace.v2 import views
-
 from api.data_workspace.metadata.routers import TableMetadataRouter
+from api.data_workspace.v2 import views
 
 
 router_v2 = TableMetadataRouter()
 router_v2.register(views.LicenceDecisionViewSet)
 router_v2.register(views.CountryViewSet)
+router_v2.register(views.DestinationViewSet)
--- a/api/data_workspace/v2/views.py
+++ b/api/data_workspace/v2/views.py
@@ -6,15 +6,19 @@ from rest_framework_csv.renderers import PaginatedCSVRenderer
 
 from django.contrib.postgres.aggregates import ArrayAgg
 from django.db.models import F
+
+from api.applications.models import PartyOnApplication
 from api.cases.models import Case
 from api.core.authentication import DataWorkspaceOnlyAuthentication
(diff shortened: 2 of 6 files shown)


api/data_workspace/v2/views.py
```diff
@@ -65,3 +69,12 @@ def get_queryset(self):
 class CountryViewSet(BaseViewSet):
     serializer_class = CountrySerializer
     queryset = Country.objects.all().order_by("id", "name")
+
+
+class DestinationViewSet(BaseViewSet):
+    serializer_class = DestinationSerializer
+    queryset = (
+        PartyOnApplication.objects.filter(deleted_at__isnull=True)
+        .exclude(application__status__status=CaseStatusEnum.DRAFT)
```
kevincarrogan, line 78: I think we're missing a BDD test for draft applications.

hnryjmes: updated

kevincarrogan (approved):

api/data_workspace/v2/urls.py
```diff
@@ -1,8 +1,8 @@
-from api.data_workspace.v2 import views
-
 from api.data_workspace.metadata.routers import TableMetadataRouter
+from api.data_workspace.v2 import views
 
 
 router_v2 = TableMetadataRouter()
 router_v2.register(views.LicenceDecisionViewSet)
 router_v2.register(views.CountryViewSet)
+router_v2.register(views.DestinationViewSet)
```
markj0hnst0n, line 8: This is my first time noticing this approach.  Looks nice and clean.

markj0hnst0n (approved):

PR #2296, opened 2024-11-20
Automated metadata endpoint

kevincarrogan:
### Aim

This provides an automated metadata endpoint for registering Data Workspace specific endpoints.

The aim of this is to dynamically generate the tables on the Data Workspace side with all of the configuration coming from LITE instead of duplicating the configuration in multiple places.

[LTD-5664](https://uktrade.atlassian.net/browse/LTD-5664)

[LTD-5664]: https://uktrade.atlassian.net/browse/LTD-5664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/data_workspace/metadata/tests/serializers.py
+++ b/api/data_workspace/metadata/tests/serializers.py
@@ -1,0 +1,48 @@
+import datetime
+
+from typing import Optional
+
+from rest_framework import serializers
+
+
+class HiddenFieldSerializer(serializers.Serializer):
+    hidden_field = serializers.HiddenField(default="")
+
+
+class UUIDFieldSerializer(serializers.Serializer):
+    uuid_field = serializers.UUIDField()
+    nullable_uuid_field = serializers.UUIDField(allow_null=True)
+
+
+class CharFieldSerializer(serializers.Serializer):
+    char_field = serializers.CharField()
+    nullable_char_field = serializers.CharField(allow_null=True)
+
+
+class SerializerMethodFieldSerializer(serializers.Serializer):
+    returns_string = serializers.SerializerMethodField()
+    returns_optional_string = serializers.SerializerMethodField()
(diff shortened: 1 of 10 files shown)


api/data_workspace/v2/views.py
```diff
@@ -34,6 +34,28 @@ class BaseViewSet(viewsets.ReadOnlyModelViewSet):
 class LicenceDecisionViewSet(BaseViewSet):
     serializer_class = LicenceDecisionSerializer
 
+    class DataWorkspace:
```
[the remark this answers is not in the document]
kevincarrogan: Done.

api/data_workspace/metadata/tests/serializers.py
```diff
@@ -0,0 +1,48 @@
...
+
+class UUIDFieldSerializer(serializers.Serializer):
+    uuid_field = serializers.UUIDField()
+    nullable_uuid_field = serializers.UUIDField(allow_null=True)
+
+
+class CharFieldSerializer(serializers.Serializer):
+    char_field = serializers.CharField()
+    nullable_char_field = serializers.CharField(allow_null=True)
+
```
hnryjmes, line 20: will we ever need any numeric fields like integer, float? or do these all get converted to text?

kevincarrogan: Probably. But as they don't exist right now I haven't added them.

At the moment I'm not trying to add every single possible field but we can add ones as we go. At the moment I'm covering all of the fields we currently have in our v2 serializers.

saruniitr (approved):
I think it is beneficial to have tests that use real viewsets to ensure table definition we send to DW matches the fields in serializer but we can add as a follow up.

PR #2303, opened 2024-11-22
LTD-5670: Productionise latest licence decision changes

saruniitr:
## Change description

Update `LicenceDecision` to support all required licence decisions for statistics.
- Add migrations and back populate
- Add code changes to create new instances when decisions are made on the Case
- Add bdd tests for all decision types

[LTD-5670](https://uktrade.atlassian.net/browse/LTD-5670)

[LTD-5670]: https://uktrade.atlassian.net/browse/LTD-5670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevincarrogan (approved):

api/data_workspace/v2/serializers.py
```diff
@@ -1,40 +1,41 @@
...
 from api.cases.enums import LicenceDecisionType
-from api.cases.models import Case
+from api.cases.models import LicenceDecision
 from api.staticdata.countries.models import Country
 
 
 class LicenceDecisionSerializer(serializers.ModelSerializer):
-    decision = serializers.SerializerMethodField()
-    decision_made_at = serializers.SerializerMethodField()
+    application_id = serializers.CharField(source="case.id")
```
kevincarrogan, line 13: This should be `serlializers.UUIDField`

api/data_workspace/v2/serializers.py
```diff
@@ -1,40 +1,41 @@
...
+    decision_made_at = serializers.CharField(source="created_at")
```
kevincarrogan, line 14: This should be `serializers.DateTimeField`

api/data_workspace/v2/tests/bdd/licences/test_licence_decisions.py
```diff
@@ -55,39 +57,40 @@ def standard_licence_is_cancelled(standard_licence):
...
     licences = unpage_data(licence_decisions_list_url)
 
-    assert refused_case.reference_code in [item["reference_code"] for item in licences]
+    assert str(refused_case.id) in [item["application_id"] for item in licences]
 
 
 @given("a case is ready to be finalised", target_fixture="case_with_final_advice")
 def case_ready_to_be_finalised(standard_case_with_final_advice):
-    assert standard_case_with_final_advice.status.status == CaseStatusEnum.UNDER_FINAL_REVIEW
+    assert standard_case_with_final_advice.status == CaseStatus.objects.get(status=CaseStatusEnum.UNDER_FINAL_REVIEW)
```
hnryjmes, line 79: out of interest, why these changes? is it better to check equality of the CaseStatus objects rather than strings?

saruniitr: yes, because there is sub-status as well.
If we are only checking string it is possible that sub-status is different. By checking object we are indirectly ensuring the sub-status is correct as well.

api/cases/models.py
```diff
@@ -382,14 +382,33 @@ def finalise(self, request, decisions):
...
+                    # link up to previous instance if the decision remains same
+                    if previous_licence_decision.decision == current_decision:
+                        previous_decision = previous_licence_decision
+
                 LicenceDecision.objects.create(
                     case=self,
-                    decision=LicenceDecisionType.advice_type_to_decision(advice_type),
+                    decision=current_decision,
                     licence=licence,
+                    previous_decision=previous_decision,
```
hnryjmes, line 411: not something we need to fix now, but I don't like that these two kwargs `decision` and `previous_decision` sound similar but are actually different types, one is a string and the other is a `LicenceDecision`. reading the code, you have to keep this in mind in order to not get confused.

I don't have a great solution for this right now but maybe one we can pick up later.

hnryjmes (approved):

PR #2305, opened 2024-11-25
LTD-5522 fix unexpected search results

markj0hnst0n:
### Aim

Ensure that all hits for search results are presented for Search Products.

[LTD-5522](https://uktrade.atlassian.net/browse/LTD-5522)

[LTD-5522]: https://uktrade.atlassian.net/browse/LTD-5522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/search/product/views.py
```diff
@@ -102,9 +102,6 @@ def get_queryset(self):
                     "inner_hits": {
                         "size": 200,
                         "name": "related",
-                        "collapse": {
-                            "field": "context",
-                        },
                         "sort": [
```
markj0hnst0n, line 105: The search results were being 'collapsed' twice and so only the first 'collapse' on product name (first key in this dictionary) was being used.  The second collapse had the effect of hiding the search hits

depsiatwal: do you mind linking here the other collapse ?

markj0hnst0n: It was removed to enable the first collapse to work correctly.  Originally `get_queryset` looked like this:
image

markj0hnst0n: You can see collapse is the first key in the query dict

depsiatwal (approved):
lgtm

kevincarrogan (commented):
Could we get a test around this to make sure we don't accidentally break this again in future?

kevincarrogan (approved):

PR #2311, opened 2024-11-26
LTD-5673: Productionise goods ratings endpoint

saruniitr:
### Aim

[LTD-5673](https://uktrade.atlassian.net/browse/LTD-5673)

[LTD-5673]: https://uktrade.atlassian.net/browse/LTD-5673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/data_workspace/v2/views.py
```diff
@@ -141,3 +143,15 @@ class ApplicationViewSet(BaseViewSet):
...
         table_name = "applications"
+
+
+class AssessmentViewSet(BaseViewSet):
+    serializer_class = AssessmentSerializer
+
+    def get_queryset(self):
+        return ControlListEntry.objects.annotate(
+            good_id=F("goodonapplication__id"),
+        ).exclude(good_id__isnull=True)
```
kevincarrogan, line 154: I had to add an `order_by` here to make sure that the results were consistent for our tests to pass.

saruniitr: updated

kevincarrogan (approved):

api/data_workspace/v2/views.py
```diff
@@ -141,3 +143,19 @@ class ApplicationViewSet(BaseViewSet):
 
     class DataWorkspace:
         table_name = "applications"
+
+
+class AssessmentViewSet(BaseViewSet):
```
hnryjmes, line 148: can I ask why the name of the view / serializer / url doesn't match the name of the table it will be in Data Workspace? was it a name agreed to because there's something else called "rating"?

hnryjmes: I see that the endpoint actually used for TAU assessments is `assessments:make_assessments` so perhaps it's for that reason?

hnryjmes (approved):

api/data_workspace/v2/views.py
saruniitr: yes as these are product assessments

kevincarrogan: To be honest we should probably change this to `GoodsRatingSerializer` and `GoodsRatingViewSet` but I don't think this is a blocker for now and we can tidy that up after merging this in.

I think on the prototype branch I changed the name of this multiple times and just left it at one point.

PR #2313, opened 2024-11-26
LTD-5685-dbt-API-data-pull-config-changes

emileswarts:
### Aim

Add additional configurations to allow this codebase to run as a scheduled task on DBT Platform.


Procfile
```diff
@@ -1,5 +1,7 @@
 web: SWIG_LIB=/home/vcap/deps/0/apt/usr/share/swig4.0 CFLAGS=-I/home/vcap/deps/1/python/include/python3.9.18m pip3 install endesive==1.5.9 && python manage.py migrate && gunicorn --worker-class gevent -c api/conf/gconfig.py -b 0.0.0.0:$PORT api.conf.wsgi
 web-dbt-platform: python manage.py migrate && gunicorn -c api/conf/gconfig-dbt-platform.py -b 0.0.0.0:$PORT api.conf.wsgi
 dump-and-anonymise: python manage.py dump_and_anonymise
+rebuild-search-index: python manage.py search_index --rebuild -f
```
depsiatwal, line 4: we don't want to run these in PROD 
is this the commands only ?

emileswarts: Yeah this is a good point. It's currently not possible to run this in prod (job hasn't been deployed there) but we can restrict at this level as well. Will update

emileswarts: Fix has been added

depsiatwal (approved):
lgtm

Procfile
```diff
@@ -1,5 +1,7 @@
...
+seed-internal-users: ./bin/seed_internal_users.sh
```
saruniitr, line 5: do we want to run this on `prod`?

depsiatwal: it won't run on prod 

the command is there but if anyone attempts to run it exits 
https://github.com/uktrade/lite-api/blob/dev-migration-post-pause/bin/seed_internal_users.sh

saruniitr: ok, may be best to remove this altogether.
We can invite internal users from UI.

depsiatwal: St the moment it's part of the seed environment in non dev environments 
effectively the same  infrastructure we use currently

bin/seed_internal_users.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+if [[ "${ENV}" == "prod" ]]; then
+  echo "You can't re-index the prod OpenSearch instance"
```
saruniitr, line 6: this doesn't match with this command

saruniitr (approved):
error message need updating

bin/seed_internal_users.sh
depsiatwal: well spotted changed

PR #2319, opened 2024-11-27
LTD: Productionise units endpoint

saruniitr:
### Aim


api/data_workspace/v2/tests/bdd/scenarios/units.feature
```diff
@@ -0,0 +1,19 @@
+@db
+Feature: units Table
+
+Scenario: Units endpoint
```
saruniitr, line 4: Status endpoint can be similar to this so based on the feedback can be added in the same PR

api/data_workspace/v2/tests/bdd/scenarios/units.feature
```diff
@@ -0,0 +1,19 @@
...
+    Given endpoint exists for exporting `units`
```
kevincarrogan, line 5: I don't think this is something useful to say for the readers of this scenario. Is there a way of rewording it from the perspective that we're not discussing an endpoint (assume the readers of this scenario have no idea what an endpoint is).

api/data_workspace/v2/tests/bdd/conftest.py
```diff
@@ -364,3 +364,34 @@ def check_rows(client, parse_table, unpage_data, table_name, rows):
...
+
+
+@given(parsers.parse("endpoint exists for exporting `{table_name}`"))
+def given_endpoint_exists(client, table_name):
+    metadata_url = reverse("data_workspace:v2:table-metadata")
+    response = client.get(metadata_url)
+    assert table_name in [t["table_name"] for t in response.json()["tables"]]
+
+
+@then(parsers.parse("the `{table_name}` table should contain the following rows:{rows}"))
```
kevincarrogan, line 376: What's the difference here with the other step called `check_rows`?

api/data_workspace/v2/tests/bdd/scenarios/units.feature
saruniitr: yes technically just the `then` step is enough as it also checks for the endpoint first so something added to have a `given` step, may be,
`Give LITE exports `units` data to DW` ?

api/data_workspace/v2/tests/bdd/conftest.py
saruniitr: I had to separate mainly because of the difference in response.
For units we just get a list so cannot use `unpage_data()`

kevincarrogan: In that case I think we should actually get this to make the output the same as the other outputs otherwise this is possibly going to cause issues with our meta pipeline that assumes all of the outputs share the same output.

api/data_workspace/v2/tests/bdd/scenarios/units.feature
kevincarrogan: I think that works.

api/data_workspace/v2/tests/bdd/scenarios/units.feature
kevincarrogan: I think we don't need to add the status table now.

api/data_workspace/v2/views.py
```diff
@@ -147,22 +147,9 @@ class DataWorkspace:
...
-    def retrieve(self, request, pk):
-        units = dict(Units.choices)
-        try:
-            description = units[pk]
-        except KeyError:
-            raise Http404()
-        return Response(UnitSerializer({"code": pk, "description": description}).data)
+class UnitViewSet(BaseViewSet):
+    serializer_class = UnitSerializer
+    queryset = [{"code": code, "description": description} for code, description in Units.choices]
```
kevincarrogan, line 152: This is interesting. I wonder if there's going to be anywhere that really does expect this to be a queryset and it will end up breaking.

But if it works then this is a nice and quick resolution.

kevincarrogan (approved):

hnryjmes (approved):
happy to approve. saw the conversation about using a list vs a queryset but I think it's fine if we have the tests that we've got

PR #2333, opened 2024-12-06
Ltd 5681 update licence decision refusal criteria

kevincarrogan:
### Aim

This updates refusal licence decisions with the relevant refusal criteria.

This is so that the reason for refusal is directly attached to the licence decision model itself instead of being attached to the advice record.

The migration initially just uses the final advice from LU but in some cases older cases don't have this advice record so we fall back on using audit entries for this instead.

[LTD-5681](https://uktrade.atlassian.net/browse/LTD-5681)

[LTD-5681]: https://uktrade.atlassian.net/browse/LTD-5681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/migrations/0074_update_licence_decision_denial_reasons.py
```diff
@@ -0,0 +1,69 @@
...
+    Advice = apps.get_model("cases", "Advice")
+
+    denial_reasons = (
+        Advice.objects.filter(
+            case__licence_decisions__decision="refused",
+            level=AdviceLevel.FINAL,
+            team_id=LICENSING_UNIT_ID,  # Just care about LU advice
+        )
+        .only("denial_reasons__id", "case__licence_decisions__id")
+        .exclude(denial_reasons__id__isnull=True)  # This removes refusals without any criteria
```
saruniitr, line 23: refusal criteria is mandatory so ideally this should not be required for refuse decisions isn't it?

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
```diff
@@ -0,0 +1,69 @@
...
+        .values_list("denial_reasons__id", "case__licence_decisions__id")
+        .order_by()  # We need to remove the order_by to make sure the distinct works
```
saruniitr, line 25: the comment meant to say add `order_by`?

kevincarrogan: `order_by` without an argument actually removes the `ORDER BY` from the query.

In this case the model has a default `order_by` set in its meta and we want to actively remove it.

I'll make this clearer in the comment.

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
kevincarrogan: There is a difference in the counts with and without this. It's been a while since I first wrote this during the prototype/POC work.

Let me dig in a bit more as to why I originally added this in but it definitely makes a difference.

kevincarrogan: Excellent. I will filter by type as well as I feel that's a bit more self-documenting as to what's happening.

kevincarrogan: Updated.

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
kevincarrogan: Updated.

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
```diff
@@ -0,0 +1,68 @@
...
+
+    updated_cases = set()
+    for denial_reason, licence_decision_id in denial_reasons:
+        licence_decision = LicenceDecision.objects.get(pk=licence_decision_id)
+        licence_decision.denial_reasons.add(denial_reason)
+        updated_cases.add(licence_decision.case.pk)
+
+    Audit = apps.get_model("audit_trail", "Audit")
+    Case = apps.get_model("cases", "Case")
+    case_content_type = ContentType.objects.get_for_model(Case)
```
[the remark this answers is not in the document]
currycoder: Agreed with this

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
```diff
@@ -0,0 +1,75 @@
...
+
+    for denial_reason, licence_decision_id in denial_reasons:
+        licence_decision = LicenceDecision.objects.get(pk=licence_decision_id)
+        licence_decision.denial_reasons.add(denial_reason)
+
+    updated_cases = set()
+    for denial_reason, licence_decision_id in denial_reasons:
+        licence_decision = LicenceDecision.objects.get(pk=licence_decision_id)
+        licence_decision.denial_reasons.add(denial_reason)
+        updated_cases.add(licence_decision.case.pk)
```
currycoder, line 43: Looks like duplicate loops..?

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
```diff
@@ -0,0 +1,75 @@
...
+        target_content_type_id=case_content_type.pk,
+        target_object_id__in=updated_cases,
+    ).filter(verb=AuditType.CREATE_REFUSAL_CRITERIA)
+    for audit in refusal_criteria_audits:
+        case = Case.objects.get(pk=audit.target_object_id)
+        refusal_licence_decisions = case.licence_decisions.filter(decision="refused")
+        if not refusal_licence_decisions.exists():
+            continue
+        for licence_decision in refusal_licence_decisions:
+            denial_reasons = audit.payload["additional_text"].replace(".", "").replace(" ", "").split(",")
```
currycoder, line 60: These need to be valid DenialReason.id values, right?  Is it worth checking against a list of valid ids first?  Or is the thought that this will fail with an IntegrityError or similar on the `set()` below?

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
```diff
@@ -0,0 +1,75 @@
...
+from api.cases.enums import (
+    AdviceLevel,
+    AdviceType,
+)
+
+
+LICENSING_UNIT_ID = "58e77e47-42c8-499f-a58d-94f94541f8c6"
+
+
+def update_licencedecision_denial_reasons(apps, schema_editor):
```
currycoder, line 15: Has this been run against a recent anonymised copy?

test_helpers/factories.py
```diff
@@ -0,0 +1,32 @@
...
+            **field_value.defaults,
+        )
+        defaults[field_name] = factory.declarations.SubFactory(generated_factory)
+
+    updated_factory = factory.make_factory(
+        get_model_for_factory(factory_class, apps),
+        **defaults,
+        FACTORY_CLASS=factory_class,
+    )
+    return updated_factory
```
currycoder, line 32: How does this get used?

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
kevincarrogan: It has been.

test_helpers/factories.py
kevincarrogan: It doesn't currently. I can get rid of it.

saruniitr (approved):
with other comments addressed, it looks good to me.

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
kevincarrogan: It hasn't failed when running on the production-like data so far so I've not worried about this.

test_helpers/factories.py
currycoder: If you plan to use it somehow I'm ok with it staying, was just trying to work out if there was something else going on

kevincarrogan: It has now gone.

api/cases/migrations/0074_update_licence_decision_denial_reasons.py
kevincarrogan: Removed.

currycoder (approved):

PR #2335, opened 2024-12-06
hot-fix-dbt-endesive

depsiatwal:
HOT Fix for dbt-platform endesive not installed

this is required for UAT testing on PROD DBT Platform. 

This has been tested on DBT and will be tested by deployed to Support. 
No Gov-paas release is required as it has no impact..

### DIFF
--- a/.copilot/phases/pre_build.sh
+++ b/.copilot/phases/pre_build.sh
@@ -23,3 +23,6 @@ url = $git_clone_base_url/django-db-anonymiser.git
 EOF
 
 git submodule update --init --recursive
+
+sed -i 's/\[packages\]/[packages]\nendesive = "~=1.5.9"/' Pipfile
+pipenv lock


kevincarrogan (approved):

.copilot/phases/pre_build.sh
```diff
@@ -23,3 +23,6 @@ url = $git_clone_base_url/django-db-anonymiser.git
 EOF
 
 git submodule update --init --recursive
+
+sed -i 's/\[packages\]/[packages]\nendesive = "~=1.5.9"/' Pipfile
+pipenv lock
```
currycoder, line 28: Would suggest a big/obvious TODO comment here that we should push this in to the Pipfile itself once we have fully moved away from gov-paas.  It's one of those things that we could easily forget and assume has to be like this forever.

currycoder (approved):

PR #2337, opened 2024-12-09
LTD: Imrove validation during submission

saruniitr:
## Change description

We validate several things when a draft application is submitted. The current code assumes only one end-user which is not going to be the case for SIEL Temporary applications.

Before changing the code the idea is to simplify the current validation code. This is also not updated with the new fields that we added to the model. This PR doesn't add any new changes except that it simplifies validation.

Once this is merged then further changes will be done to allow for multiple end-users.


api/core/model_mixins.py
```diff
@@ -42,3 +42,20 @@ class Trackable:
 
     def get_history(self, field):
         raise NotImplementedError()
+
+
+class BaseApplicationValidator:
```
hnryjmes, line 47: I assume this is general enough that it could be used for other licence types than just SIELs (e.g. OIELs)?

saruniitr: yes

api/applications/validators.py
```diff
@@ -0,0 +1,177 @@
+from django.db.models import Q
+
+from api.applications.enums import ApplicationExportType
+from api.core.model_mixins import BaseApplicationValidator
+from api.goods.models import GoodDocument
+from api.parties.models import PartyDocument
+
+
+def siel_locations_validator(application):
```
hnryjmes, line 9: there are a couple of differences in the checks between this function and the original `_validate_siel_locations`, is that because they are not needed and so they can be removed?

saruniitr: yes, there was a check related to `GoodsTypeCategory.CRYPTOGRAPHIC` which is for OIELs and we don't support yet.

hnryjmes (commented):
I'm wondering if we can get this branch deployed to an env and have someone from UCD click around and do a few submissions as a regression test. Just because there's a lot changing here

saruniitr:
> I'm wondering if we can get this branch deployed to an env and have someone from UCD click around and do a few submissions as a regression test. Just because there's a lot changing here

@hnryjmes Verified these changes on `devdata` and compared with `uat` at the same time. In both cases the errors messages shown for the items in application task list is same. As the sections are filled then the errors are no longer shown.

api/applications/validators.py
```diff
@@ -0,0 +1,177 @@
...
+    return None
+
+
+def siel_end_user_validator(application):
+    error = None
+    if application.export_type == ApplicationExportType.TEMPORARY:
+        return None
+
+    if not application.end_user:
+        return "To submit the application, add an end user"
```
hnryjmes, line 34: it seems like this should be at the top, before the check for export type being temporary?

api/applications/validators.py
```diff
@@ -0,0 +1,177 @@
...
+
+    ultimate_end_user_required = application.goods.filter(
+        Q(is_good_incorporated=True) | Q(is_onward_incorporated=True)
+    ).exists()
+
+    if ultimate_end_user_required and application.ultimate_end_users.count() == 0:
+        error = "To submit the application, add an ultimate end-user"
+
+    return error
```
hnryjmes, line 104: should there be a check for ultimate end user documents here?

saruniitr: It is optional,
https://github.com/uktrade/lite-api/blob/dev/api/applications/creators.py#L154

hnryjmes (approved):
have checked all the validators and they look right to me by eye. if we can get product review from UCD on this later that would give us extra confidence but it looks to me like everything has been copied over correctly.

api/applications/models.py
```diff
@@ -268,12 +268,17 @@ def set_appealed(self, appeal, exporter_user):
...
+    def validate(self):
+        raise NotImplementedError("Validator to validate application attributes is not implemented")
+
     def create_amendment(self, user):
         raise NotImplementedError()
 
 
 # Licence     Applications
 class StandardApplication(BaseApplication, Clonable):
+    from api.applications.validators import StandardApplicationValidator
```
currycoder, line 280: Does this import need to be here?

api/applications/validators.py
```diff
@@ -0,0 +1,177 @@
...
+    recipient_choices = [item[0] for item in application.GOODS_RECIPIENTS_CHOICES]
+
+    if (
+        not SiteOnApplication.objects.filter(application=application).exists()
+        and application.export_type not in export_type_choices
+        and application.goods_starting_point not in starting_point_choices
+        and application.goods_recipients not in recipient_choices
+        and application.is_shipped_waybill_or_lading is None
+    ):
+        return "To submit the application, add a product location"
```
currycoder, line 23: It would be good to discuss; should the API validation errors be returned as end-user readable strings or as lite-frontend consumable error keys?
```
class StandardApplicationValidationErrors:
    MISSING_PRODUCT_LOCATION = "MISSING_PRODUCT_LOCATION"
...

return StandardApplicationValidationErrors.MISSING_PRODUCT_LOCATION
```

I feel a bit like this is a similar muddying of the API that we currently have for audit trail text formatting; the API does some text formatting rather than just responding with pure data.  That means when changing the format of audit trail messages, we always have to do work on the API rather than just adjusting the frontend.

Similarly for this, the API is getting unnecessarily tied to frontend structures/wording.  Instead, the frontend could translate the error key in to the user readable string.  Or have the freedom to do bullet points instead e.g.

```

To submit the application:
- Add a product location <a href="/place_to_fix_it">Here</a>
- Add an end user

```

saruniitr: Definitely worth considering, initially I only want to get parity with the current functionality. Once this is merged the add further improvements.

api/applications/models.py
saruniitr: actually added this to avoid circular dependencies but they are not there atm, I will move this outside of this.

PR #2338, opened 2024-12-10
LTD-5722 fix licence pdf css

markj0hnst0n:
### Aim

In licence document pdf, added fixed-layout for all table elements.  Removed unnecessary template tag.  Testing info in comments on ticket.

[LTD-5722](https://uktrade.atlassian.net/browse/LTD-5722)

[LTD-5722]: https://uktrade.atlassian.net/browse/LTD-5722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/letter_templates/templates/letter_templates/refusal.html
+++ b/api/letter_templates/templates/letter_templates/refusal.html
@@ -13,7 +13,7 @@
   <ul class="govuk-list govuk-list--bullet">
     {% for good in goods.refuse %}
       <li>
-        {{good.good.name|split_good_name_across_lines}}
+        {{good.good.name}}
       </li>
     {% endfor %}
   </ul>
--- a/assets/css/siel.css
+++ b/assets/css/siel.css
@@ -26,7 +26,7 @@
   margin: 0;
   padding: 5px;
 }
-.document table.fixed-layout {
+.document table {
   table-layout: fixed;
 }
 .document .border-black {
--- a/assets/css/siel.css
+++ b/assets/css/siel.css
@@ -20,14 +20,12 @@
 .document table {
   border-collapse: collapse;
   width: 100%;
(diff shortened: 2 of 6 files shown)


assets/css/siel.css
```diff
@@ -26,7 +26,7 @@
   margin: 0;
   padding: 5px;
 }
-.document table.fixed-layout {
+.document table {
   table-layout: fixed;
 }
```
kevincarrogan, line 31: Can this be put in with the other `.document table` declaration?

I was just reading about fixed layouts and that it needs a width and it took me a while to find it so I think it would be best if `table-layout` and `width` were together in the same declaration.

markj0hnst0n: Done!  Thanks for that.  I just didn't spot the other one in the sheet.

api/letter_templates/templates/letter_templates/refusal.html
```diff
@@ -13,7 +13,7 @@ <h2 class="govuk-heading-l">Products</h3>
   <ul class="govuk-list govuk-list--bullet">
     {% for good in goods.refuse %}
       <li>
-        {{good.good.name|split_good_name_across_lines}}
+        {{good.good.name}}
```
saruniitr, line 16: Is this to proactively fix on the refusal template?

markj0hnst0n: That was used on the refuse template to match the SIEL template's state previous to this PR.  I don't believe there were ever any issues reported with the refuse template so I don't think it was ever needed in the first place and such I removed it.

kevincarrogan (approved):

PR #2340, opened 2024-12-10
[LTD-5712] Add exclude_flags parameter to case search view

currycoder:
### Aim

Adds a filter to the case search view enabling callers to exclude cases which have specified flags.  Cases will be excluded if the case, any party or any good related to the case are associated with any flags in the `exclude_flags` value.

[LTD-5712](https://uktrade.atlassian.net/browse/LTD-5712)

### DIFF
--- a/api/cases/tests/test_case_search.py
+++ b/api/cases/tests/test_case_search.py
@@ -34,8 +34,6 @@ from test_helpers.clients import DataTestClient
 from api.users.enums import UserStatuses
 from api.users.models import GovUser
 from api.cases.tests import factories
-from api.cases.enums import AdviceType
-from api.staticdata.statuses.enums import CaseStatusEnum
 from api.teams.models import Team
 from api.cases.views.search.service import (
     get_case_status_list,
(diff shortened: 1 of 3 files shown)


api/cases/tests/test_case_search.py
```diff
@@ -55,6 +53,7 @@ def setUp(self):
         self.url = reverse("cases:search")
         statuses = [CaseStatusEnum.SUBMITTED, CaseStatusEnum.CLOSED, CaseStatusEnum.WITHDRAWN]
 
+        Case.objects.all().delete()
```
saruniitr, line 56: Pulled this branch locally and I see all tests in this file passing without this line, was this a debug leftover?

saruniitr (commented):
just a question, looks good to me otherwise.

saruniitr (approved):

api/cases/tests/test_case_search.py
currycoder: yep we can bin that I think too

hnryjmes (approved):

PR #2351, opened 2024-12-19
LTD: Create permission class to check if user can issue final advice

saruniitr:
## Change description

Create permission class instead of asserting permission at various places.

Add new tests in pytest style.
Delete few tests as they are related to legacy Advice1.0 code which is removed in this PR.


api/cases/views/views.py
```diff
@@ -446,52 +444,37 @@ def get_unique_documents(self, final_advice, pk):
...
 
     def get(self, request, pk):
-        """
-        Concatenates all advice for a case and returns it or just returns if final advice already exists
-        """
-        if len(self.final_advice) == 0:
-            assert_user_has_permission(request.user.govuser, constants.GovPermissions.MANAGE_LICENCE_FINAL_ADVICE)
-
-            group_advice(self.case, self.team_advice, request.user, AdviceLevel.FINAL)
-            final_advice = Advice.objects.filter(case=self.case).order_by("-created_at")
```
saruniitr, line 469: removing legacy Advice1.0 code.

api/cases/tests/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@pytest.fixture
+def submit_application(api_client, exporter_headers):
+    def _submit_application(draft_application):
+        response = api_client.put(
+            reverse(
+                "applications:application_submit",
+                kwargs={
+                    "pk": draft_application.pk,
```
saruniitr, line 118: Most of these are based on fixtures from official stats work, once we have more of these tests then we can collect them in a common place.

api/cases/tests/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+    gov_user.role = RoleFactory(name="Case officer", type=UserType.INTERNAL)
+    gov_user.role.permissions.set(
+        [
+            GovPermissions.MANAGE_LICENCE_FINAL_ADVICE.name,
+            GovPermissions.MANAGE_LICENCE_DURATION.name,
+            GovPermissions.REOPEN_CLOSED_CASES.name,
+        ]
+    )
+    gov_user.save()
+    return gov_user
```
kevincarrogan, line 81: I actually ran into some troubles with this setup on official stats but never got around to fixing them.

These shouldn't use the `gov_user` fixture directly as this actually will make changes on the `gov_user` itself so if you were to use both the `gov_user` fixture _and_ the `lu_user` or `lu_case_officer` fixture together in a test case you will end up with some confusing results.

Each one of these fixtures should create its own `GovUser` instead of relying on `gov_user`.

saruniitr: updated

api/core/permissions.py
```diff
@@ -47,3 +47,8 @@ def has_permission(self, request, view):
 class CanCaseworkersManageOrgainsation(permissions.BasePermission):
     def has_permission(self, request, view):
         return check_user_has_permission(request.user.govuser, GovPermissions.MANAGE_ORGANISATIONS)
+
+
+class CanCaseworkersIssueLicence(permissions.BasePermission):
```
markj0hnst0n, line 52: Is 'issue' the only action governed by this permission class?

saruniitr: Yes, this is only used in the view where we finalise and issue licence.
Suspend/Revoke handled by different views.

kevincarrogan (approved):

api/cases/tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@pytest.fixture()
+def exporter_user_permissions():
+    for permission in ExporterPermissions:
+        Permission.objects.get_or_create(id=permission.name, name=permission.value, type=UserType.EXPORTER.value)
+
+
+@pytest.fixture()
+def organisation(exporter_user_permissions, exporter_user):
```
depsiatwal, line 35: nice addition never actually thought about factories as fixtures.

depsiatwal (approved):
lgtm

PR #2355, opened 2024-12-23
LTD-5719-dbt-platform-post-migration-clean-up

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5719

Post DBT Platform Migration Clean-up.

This is broken down into 3 parts

Since additional tests were added for DBT Platform i.e different version of postgre and opensearch. We remove the Gov PAAS versions and rename the DBT Platform ones as no need to mention DBT platform.
Removing all is special conditions in code that require specific conditions for DBT/Gov Paas i.e S3 client config
The Final part is an installation of the Lib that was installed at runtime due to gov pass compatibility.

Have tested this on DBT Platform together with https://github.com/uktrade/lite-frontend/pull/2285
Document generation S3 uploads all work.


.circleci/config.yml
```diff
@@ -14,27 +14,11 @@ image_python_node: &image_python_node
   image: cimg/python:3.9.18-node
 
 image_postgres: &image_postgres
-  image: circleci/postgres:12
-  environment:
-    POSTGRES_USER: postgres
-    POSTGRES_PASSWORD: password
-
-image_postgres13: &image_postgres13
   image: circleci/postgres:13
```
currycoder, line 17: Would suggest changing the version of the `docker-compose.yml` to postgres 13 too.  Although worth letting everybody know when this happens as local DBs will be cleaned out without intervention.

Procfile
```diff
@@ -1,4 +1,4 @@
-web: SWIG_LIB=/home/vcap/deps/0/apt/usr/share/swig4.0 CFLAGS=-I/home/vcap/deps/1/python/include/python3.9.18m pip3 install endesive==1.5.9 && python manage.py migrate && gunicorn --worker-class gevent -c api/conf/gconfig.py -b 0.0.0.0:$PORT api.conf.wsgi
+web: SWIG_LIB=/home/vcap/deps/0/apt/usr/share/swig4.0 CFLAGS=-I/home/vcap/deps/1/python/include/python3.9.18m && python manage.py migrate && gunicorn --worker-class gevent -c api/conf/gconfig.py -b 0.0.0.0:$PORT api.conf.wsgi
 web-dbt-platform: python manage.py migrate && gunicorn -c api/conf/gconfig-dbt-platform.py -b 0.0.0.0:$PORT api.conf.wsgi
```
currycoder, line 2: Do we still need two web processes mentioned here?

api/documents/libraries/s3_operations.py
```diff
@@ -25,22 +24,12 @@ def init_s3_client():
...
-        )
-    else:
-        _client = boto3.client(
-            "s3",
-            aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
-            aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
-            region_name=settings.AWS_REGION,
-            config=Config(connect_timeout=settings.S3_CONNECT_TIMEOUT, read_timeout=settings.S3_REQUEST_TIMEOUT),
-            **additional_s3_params,
-        )
```
currycoder, line 43: I think this would break our ability to use minio locally as a dummy S3; https://github.com/uktrade/lite-api/blob/dev/docker-compose.yml#L107-L118
Would suggest that you perhaps make this in to something like...

```
aws_kwargs = {
    "region_name": settings.AWS_REGION,
    "config": Config(connect_timeout=settings.S3_CONNECT_TIMEOUT, read_timeout=settings.S3_REQUEST_TIMEOUT),
    **additional_s3_params,
}
if settings.AWS_ACCESS_KEY_ID and settings.AWS_SECRET_ACCESS_KEY:
    aws_kwargs["aws_access_key_id"] = settings.AWS_ACCESS_KEY_ID
    aws_kwargs["aws_secret_access_key"] = settings.AWS_SECRET_ACCESS_KEY

_client = boto3.client("s3", **aws_kwargs)
```

api/documents/libraries/tests/test_s3_operations.py
```diff
@@ -72,33 +68,6 @@ def test_get_client_with_aws_endpoint_url(self, mock_Config, mock_boto3):
...
-            read_timeout=44,
-        )
-        mock_boto3.client.assert_called_with(
-            "s3",
-            aws_access_key_id="AWS_ACCESS_KEY_ID",
-            aws_secret_access_key="AWS_SECRET_ACCESS_KEY",
-            region_name="AWS_REGION",
-            config=config,
-            endpoint_url="AWS_ENDPOINT_URL",
-        )
```
currycoder, line 90: I think with the above suggestion, a lot of the removals in this file could be re-instated

Procfile
depsiatwal: ahh can be removed as it's here now https://github.com/uktrade/lite-deploy/blob/main/copilot/lite-api/manifest.yml#L65

api/documents/libraries/s3_operations.py
depsiatwal: good point i've reinstated some of them as suggested.

api/documents/libraries/tests/test_s3_operations.py
```diff
@@ -89,10 +90,13 @@ def test_get_client_with_aws_endpoint_url(self, mock_Config, mock_boto3):
...
-    @patch("api.documents.libraries.s3_operations.is_copilot")
     @patch("api.documents.libraries.s3_operations._client")
-    def test_get_client_with_is_copilot(self, mock_client, mock_is_copilot, mock_Config, mock_boto3):
-        mock_is_copilot.return_value = True
+    def test_get_client_with_no_aws_secret_keys(self, mock_client, mock_Config, mock_boto3):
+
+        # These are only used locally
+        delattr(settings, "AWS_ACCESS_KEY_ID")
+        delattr(settings, "AWS_SECRET_ACCESS_KEY")
+
```
currycoder, line 99: Could these just be set with the `@override_settings` decorator as above?

depsiatwal: I don't think so basically DBT Platform doesn't use or care about these variables. Hence they're mot even set which is why in the client I check for the existence of the attribute.

markj0hnst0n (approved):
looks grand

PR #2358, opened 2025-01-02
LTD-5760 remove CRN check against rejected organisations

markj0hnst0n:
### Aim

Remove validation agains a rejected CRN when and exporter registers.

as per ticket: 'The change here is to only validate against CRNs where the registration that are either pending or approved.'

[LTD-5760](https://uktrade.atlassian.net/browse/LTD-5760)

[LTD-5760]: https://uktrade.atlassian.net/browse/LTD-5760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/organisations/serializers.py
+++ b/api/organisations/serializers.py
@@ -504,8 +504,7 @@ class OrganisationRegistrationNumberSerializer(serializers.Serializer):
         if not self.instance:
             if (
                 Organisation.objects.filter(registration_number=value)
-                .exclude(status=OrganisationStatus.REJECTED)
-                .exclude(status=OrganisationStatus.DRAFT)
+                .exclude(status__in=[OrganisationStatus.REJECTED, OrganisationStatus.DRAFT])
                 .exists()
             ):
                 raise serializers.ValidationError("This registration number is already in use.")
--- a/api/organisations/serializers.py
+++ b/api/organisations/serializers.py
@@ -505,6 +505,7 @@ class OrganisationRegistrationNumberSerializer(serializers.Serializer):
             if (
                 Organisation.objects.filter(registration_number=value)
                 .exclude(status=OrganisationStatus.REJECTED)
+                .exclude(status=OrganisationStatus.DRAFT)
                 .exists()
             ):
                 raise serializers.ValidationError("This registration number is already in use.")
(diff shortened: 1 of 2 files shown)


api/organisations/serializers.py
```diff
@@ -502,7 +502,12 @@ class OrganisationRegistrationNumberSerializer(serializers.Serializer):
     def validate_registration_number(self, value):
         # Check for uniqueness only when creating a new Organisation
         if not self.instance:
-            if Organisation.objects.filter(registration_number=value).exists():
+            if (
+                Organisation.objects.filter(registration_number=value)
+                .exclude(status=OrganisationStatus.REJECTED)
+                .exclude(status=OrganisationStatus.DRAFT)
```
kevincarrogan, line 508: `.exclude(status__in=[OrganisationStatus.REJECTED, OrganisationStatus.DRAFT])`

would be slightly simpler.

api/organisations/tests/test_organisations.py
```diff
@@ -974,3 +972,13 @@ def test_validate_registration_number_fail(self):
...
             response.json(), {"errors": {"registration_number": ["This registration number is already in use."]}}
         )
+
+    @parameterized.expand([OrganisationStatus.REJECTED, OrganisationStatus.DRAFT])
+    def test_validate_registration_number_success_for_rejected_or_draft_org(self, org_status):
+        self.organisation.refresh_from_db()
+        self.organisation.status = org_status
+        self.organisation.save()
+        data = {"registration_number": self.organisation.registration_number}
+        response = self.client.post(self.url, data, **self.exporter_headers)
```
kevincarrogan, line 982: I'm not certain we should update this now but the API for the endpoint is really strange.

Having it post and then return a 200 when it's not found and a bad request when it's found feels a bit odd.

One to change another day.

kevincarrogan (approved):

api/organisations/serializers.py
markj0hnst0n: changed

depsiatwal (approved):
lgtm

PR #2361, opened 2025-01-07
[LTD-5768] Ensure that case flags are fully present on all cases returned by case search endpoint

currycoder:
### Aim

This change fixes a bug where the queue view was not displaying case level flags on all the cases that a flag was present on.

The way that we injected case level flags in to the search results was incorrect; the set mechanism we were using would result in a case level flag only being included in a single search result instead of all of the cases that the flag was present on.

This set mechanism has been replaced with a nested serializer instead.  Queryset prefetches have been adjusted to ensure that the number of SQL queries has not increased.

[LTD-5768](https://uktrade.atlassian.net/browse/LTD-5768)

[LTD-5768]: https://uktrade.atlassian.net/browse/LTD-5768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/views/search/service.py
```diff
@@ -50,27 +51,7 @@ def get_advice_types_list():
...
 
-def populate_other_flags(cases: List[Dict]):
-    from api.flags.models import Flag
-
-    case_ids = [case["id"] for case in cases]
-
-    union_flags = set(
-        [
-            *Flag.objects.filter(cases__id__in=case_ids).annotate(case_id=F("cases__id")),
-            *Flag.objects.filter(organisations__cases__id__in=case_ids).annotate(case_id=F("organisations__cases__id")),
```
currycoder, line 61: This old approach combined flags from both `Case.flags` and `Organisation.flags`.  The new approach using the nested serializer handles case flags only.  However this is ok as Organisation flags are not a thing in practice.

api/cases/views/search/tests/test_views.py
```diff
@@ -1148,12 +1148,33 @@ def test_api_success(self):
...
+
+        # Perform the search
+        response = self.client.get(self.url, **self.gov_headers)
+        response_data = response.json()["results"]
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response_data["cases"]), 2)
+        for search_result in response_data["cases"]:
+            self.assertEqual(len(search_result["flags"]), 1)
+            self.assertEqual(search_result["flags"][0]["alias"], flag_alias)
```
hnryjmes, line 1176: looks good and this test is fine.

out of interest, would this test have passed using the old set logic? from your description it suggests not as only one search result would have come back ("the set mechanism we were using would result in a case level flag only being included in a single search result instead of all of the cases that the flag was present on") - is that correct?

hnryjmes (approved):

api/cases/views/search/tests/test_views.py
currycoder: Yep exactly;

```
> git revert b61e48ea7145d04742dd9839e99f465360b106d5
> docker exec -it api pipenv run pytest api/cases/views/search/tests/ --reuse-db -k test_api_multiple_cases
....
======================================================= FAILURES ========================================================
__________________________________ SearchAPITest.test_api_multiple_cases_flags_correct __________________________________

self = <api.cases.views.search.tests.test_views.SearchAPITest testMethod=test_api_multiple_cases_flags_correct>

    def test_api_multiple_cases_flags_correct(self):
        # Create two cases..
        self._create_data()
        self._create_data()
        # Add a case flag to each case..
        flag_alias = "REFER_TO_FCDO_MEUC_CONCERNS"
        flag = Flag.objects.get(alias=flag_alias)
        all_cases = Case.objects.all()
        for case in all_cases:
            case.flags.add(flag)

        # Perform the search
        response = self.client.get(self.url, **self.gov_headers)
        response_data = response.json()["results"]

        self.assertEqual(len(response_data["cases"]), 2)
        for search_result in response_data["cases"]:
>           self.assertEqual(len(search_result["flags"]), 1)
E           AssertionError: 0 != 1

api/cases/views/search/tests/test_views.py:1175: AssertionError
================================================ short test summary info ================================================
FAILED api/cases/views/search/tests/test_views.py::SearchAPITest::test_api_multiple_cases_flags_correct - AssertionError: 0 != 1
```

kevincarrogan:
When this bug was originally found I presume it was highlighting that flags that were expected to be shown on the queue view weren't appearing.

I'm guessing that was because things were being selected in the flag filters that weren't then matching up on the queue view.

However, the thing that this has got me thinking about is, does that filter just work on case level flags and doesn't look at party or good flags?

The reason why I say this is because I'm presuming there is still a discrepancy here between the flags that are shown on the details page and the queue view as this only picks up the ones from the case and I think the ones at the top of the detail page are an amalgamation of case, goods and party flags.

kevincarrogan:
> When this bug was originally found I presume it was highlighting that flags that were expected to be shown on the queue view weren't appearing.
> 
> I'm guessing that was because things were being selected in the flag filters that weren't then matching up on the queue view.
> 
> However, the thing that this has got me thinking about is, does that filter just work on case level flags and doesn't look at party or good flags?
> 
> The reason why I say this is because I'm presuming there is still a discrepancy here between the flags that are shown on the details page and the queue view as this only picks up the ones from the case and I think the ones at the top of the detail page are an amalgamation of case, goods and party flags.

Although the moment I hit enter on this I noticed:

```
        service.populate_goods_flags(cases)
        service.populate_destinations_flags(cases)
```

So are we doing something where the flags are combined with the flags on the cases and then they're all being presented?

Basically, I feel like I don't really know what flags are or aren't showing on the queue view now.

currycoder:
> > When this bug was originally found I presume it was highlighting that flags that were expected to be shown on the queue view weren't appearing.
> > I'm guessing that was because things were being selected in the flag filters that weren't then matching up on the queue view.
> > However, the thing that this has got me thinking about is, does that filter just work on case level flags and doesn't look at party or good flags?
> > The reason why I say this is because I'm presuming there is still a discrepancy here between the flags that are shown on the details page and the queue view as this only picks up the ones from the case and I think the ones at the top of the detail page are an amalgamation of case, goods and party flags.
> 
> Although the moment I hit enter on this I noticed:
> 
> ```
>         service.populate_goods_flags(cases)
>         service.populate_destinations_flags(cases)
> ```
> 
> So are we doing something where the flags are combined with the flags on the cases and then they're all being presented?
> 
> Basically, I feel like I don't really know what flags are or aren't showing on the queue view now.

Yep fair enough.  It's not exactly obvious and IMO it's still fairly patched together.

As you've highlighted there are three fields containing flags per case; good level, destination level and case level.  These are stitched together on the frontend here; https://github.com/uktrade/lite-frontend/blob/dev/caseworker/templates/includes/case-row-flags.html

All that this change does is ensure that the case level flags are correct and accounted for.  There's no attempt made to join all these flags up in to a single field.

Filtering works by accounting for the different levels also, here; https://github.com/uktrade/lite-api/blob/dev/api/cases/managers.py#L126-L142

kevincarrogan (approved):

PR #2362, opened 2025-01-07
LTD-5763: Add endpoint to submit approve recommendation for multiple cases

saruniitr:
## Change description

This endpoint allows to create approve recommendation for multiple cases.
We receive list of case id values and using that we create approve `Advice` records for all these cases.

Add new audit event to record this event along with the user and list of cases. This event is not specific to any particular case so it won't show up under a case notes and timeline so Case activity view is updated to extract these events and combine. The view is also converted to make use of filter backends to make this easier.

Once the recommendation is added the cases are moved further as per routing rules.

This is only to be exposed to certain teams so a permission class is added to enforce this.

**v2**:
Following review comments below updates are made,

- Re-home this endpoint as a caseworker specific url
- Use serializer to validate input data before transforming into the required format. This strongly enforces API contract with the FE
- Add bulk approval audit entry for each case (previously it was just one entry for all cases). This helps removing special code to inject these events into each case when they are displayed in notes and timeline.
- Removed Filter backends added earlier as a result of above change, I will add these back as a separate PR.
- Refactor moving the case forward as a `Case` model method

Relevant FE changes https://github.com/uktrade/lite-frontend/pull/2294

[LTD-5763]: https://uktrade.atlassian.net/browse/LTD-5763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/queues/views/bulk_approval.py
```diff
@@ -0,0 +1,113 @@
...
+        for application in applications:
+            advice_data = self.get_advice_data(request, application)
+            payload.extend(advice_data)
+
+        return payload
+
+    def create_audit_event(self, request, cases):
+        audit_trail_service.create(
+            actor=request.user,
+            verb=AuditType.CREATE_BULK_APPROVAL_RECOMMENDATION,
```
saruniitr, line 86: As this is not specific to any particular case, only one event is added for all. This is also reduce the amount of work we do in this call. We already create many `Advice` instances.
However this causes another issue when showing this in N&T because it is not tied to a particular case. For this reason this event is extracted and combined with the case events.

currycoder: > For this reason this event is extracted and combined with the case events.
I don't quite follow what this means. Do you mean that this non-case specific event ends up in case-specific notes and timelines?

saruniitr: Yes, we want to show this in notes and timeline but the `target` is not the case so be default it doesn't show up.
`payload` contains list of cases that are bulk approved so that is used to extract event and added to display in notes and timeline.

currycoder: So how would that work from the notes/timelines frontend view perspective?  Are multiple calls made to the API to get case-specific events and these bulk approval events or would something else happen?

saruniitr: No additional calls, it all happens in the same call.
https://github.com/uktrade/lite-api/pull/2362/files#diff-20cfbc9968fed9259b01dc2f01468e304d652f6d6a26f86f3d6c8b496c62b164R57

currycoder: I can see why you've chosen this path as the single audit event reflects a single user action.  I do kind of feel like it's unfortunate that we then have to include this specific type of audit entry in the notes/timelines by a special mechanism though.

An alternative I could see would be to create this overall audit entry as you have done but create case-specific entries too.  It seems a bit off that right now we don't have the standard "added a recommendation" audit entries for each of these cases - maybe `CREATED_USER_ADVICE`/`CREATED_TEAM_ADVICE` could be added as appropriate with a payload addition to mark those entries as bulk?

This way you could even drop the special case code to inject the bulk recommendation audit entry on to case-specific entries.

api/queues/views/bulk_approval.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Run routing rules and move the case forward
+        user_queue_assignment_workflow([self.queue], case)
+
+        audit_trail_service.create(
+            actor=request.user,
+            verb=AuditType.UNASSIGNED_QUEUES,
+            target=case,
+            payload={"queues": [self.queue.name], "additional_text": ""},
+        )
```
currycoder, line 45: I feel like this is repeating something that should be a common utility and only defined once.  I bet we've already repeated this logic in a few places so could be nice to factor out.  It could even be a method on `Case` e.g. `case.move_case_forward(user, current_queue)`

api/queues/views/bulk_approval.py
```diff
@@ -0,0 +1,113 @@
...
+        self.case_ids = input_data.get("case_ids", [])
+        self.advice = input_data.get("advice", {})
+        self.cases = Case.objects.filter(id__in=self.case_ids)
+        payload = []
+        applications = StandardApplication.objects.filter(id__in=self.case_ids)
+        for application in applications:
+            advice_data = self.get_advice_data(request, application)
+            payload.extend(advice_data)
+
+        return payload
```
currycoder, line 81: This seems a little bit backwards to me; we handle raw request data and format it in to something that a serializer can accept.  I think that the serializer should reflect the format of the data getting POST'ed to this endpoint and then do the necessary transformations as part of `.create()`.  That would probably mean departing from `ModelSerializer` though

api/queues/views/bulk_approval.py
```diff
@@ -0,0 +1,113 @@
+from django.db import transaction
```
currycoder, line 1: Could I suggest re-homing this as a caseworker-specific endpoint?

I think the pattern we have landed on is captured fairly well in this commit; https://github.com/uktrade/lite-api/commit/c16eab85ed0b69172d3e3242c67c38ddb8f8d213

api/queues/tests/test_bulk_approval.py
```diff
@@ -0,0 +1,127 @@
+import pytest
```
currycoder, line 1: Could the file name/path here reflect the path of the view under test..  e.g. `queues/views/tests/test_bulk_approval.py`

api/queues/views/bulk_approval.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_advice_data(self, request, application):
+        subjects = [("good", good_on_application.good.id) for good_on_application in application.goods.all()] + [
+            (poa.party.type, poa.party.id) for poa in application.parties.all()
+        ]
+        proviso = self.advice.get("proviso", "")
+        advice_type = AdviceType.PROVISO if proviso else AdviceType.APPROVE
+        return [
+            {
+                "level": AdviceLevel.USER,
```
currycoder, line 59: Should this always be user-level advice for all queues in question?  I'm a little unclear on when something should be user level or team level..

saruniitr: Yeah all OGD advice is user level.
'Review and combine' is the team level and LU advice in post-circ is the final level.

api/queues/views/bulk_approval.py
saruniitr: Actually serializer reflects the data but it is done like this to avoid sending a huge payload from the FE.
As we create multiple `Advice` instances (one for each product and party), FE has to send the same data multiple times for each case.
Eg if user selects 25 cases and each have 5 products, 2 parties then the payload sent will be huge with most of it duplicated.

The only thing we need from FE is the approval text, user, team details and case_ids and in the API here we are creating all the records required before giving it to the serializer.

currycoder: OK I think you are saying that the serializer reflects the ORM model.  I am saying that the serializer should reflect the API endpoint contract - e.g. the format of the POST data.

Totally agree that for a bulk endpoint we should try to avoid the massive payload you are talking about.  But I think that we can achieve that while still having our serializer reflect/validate the incoming data.  It should then be responsible for the transformation/creation of the bulk objects after that.

e.g. something like
```
class BulkApprovalSerializer(Serializer):
    case_ids = PrimaryKeyRelatedField(many=True, queryset=Case.objects.all())
    advice = SomeAdviceSerializer()

    def create(self):
        # Transform the input in to creating multiple advice records
        ....
```

saruniitr: yeah this is fine, can update to use this pattern.

api/queues/views/bulk_approval.py
saruniitr: I though of this but not for 1.0. I will check if this can be refactored easily.

currycoder: Happy for that to be a follow-up!

api/queues/views/bulk_approval.py
saruniitr: Discussed internally with Kevin, I am changing it to add this entry for each case (rather than one for all). Payload contains details about level, queue etc.
If it is really going to affect performance then we can refine later but atleast we will have data for all cases and no need of special code to combine these.
One good thing about current version is I had to refactor activity view which now uses standard filter backends.

hnryjmes: I also agree that the audit event should be one per case, but I see this discussion is already resolved

api/queues/tests/test_bulk_approval.py
```diff
@@ -0,0 +1,127 @@
...
+    return _multiple_cases_ogd_queue
+
+
+def case_subjects(case):
+    application = StandardApplication.objects.get(id=case.id)
+    return list(application.goods.all()) + list(application.parties.all())
+
+
+def test_user_bulk_approves_cases(api_client, mod_officer_headers, mod_bulk_approval_url, multiple_cases_ogd_queue):
+    cases = multiple_cases_ogd_queue(QueuesEnum.MOD_CAPPROT, count=25)
```
hnryjmes, line 61: I see that here `count` is 25 and the default in the fixture is 5. Is there any technical limit on how many cases could be bulk approved at once? Do we expect the typical numbers of cases that will get bulk approved at once to be in a certain range (e.g. 2-25)?

api/queues/tests/test_bulk_approval.py
```diff
@@ -0,0 +1,127 @@
...
+            "text": "No concerns",
+            "proviso": "",
+            "note": "",
+            "footnote_required": False,
+            "footnote": "",
+            "team": TeamIdEnum.FCDO,
+        },
+    }
+    response = api_client.post(fcdo_bulk_approval_url, data=data, **fcdo_officer_headers)
+    assert response.status_code == 403
```
hnryjmes, line 115: I think I am following that this test passes because the FCDO queue is not in `BULK_APPROVE_ALLOWED_QUEUES`. I was wondering why can't FCDO bulk approve? Is it something others have decided already for this work?

api/queues/tests/test_bulk_approval.py
saruniitr: Our page size is 25 hence I used that as a maximum here. It is really upto users how many they select as they are likely to filter and select so it may be less than this.
There is a potential performance issue with higher number but that is something we will learn once they start using it.

api/queues/tests/test_bulk_approval.py
saruniitr: Correct, they have a different set of requirements and required different approach hence it is decided to not make it available for them.

api/queues/views/bulk_approval.py
saruniitr: updating this as it is a simple change.

api/audit_trail/formatters.py
```diff
@@ -351,3 +351,7 @@ def exporter_submitted_amendment(**payload):
 
 def amendment_created(**payload):
     return f"created the case to supersede {payload['superseded_case']['reference_code']}."
+
+
+def create_bulk_approval_recommendation(**payload):
+    return f"added bulk approval recommendation on {payload['queue']} queue."
```
saruniitr, line 357: Content likely going to be changed to 'added a recommendation using the Approve button in the queue.'
Still being discussed.

api/cases/models.py
```diff
@@ -336,6 +336,28 @@ def no_licence_required(self):
 
         notify_exporter_no_licence_required(self)
 
+    def move_case_forward(self, queue, user):
```
currycoder, line 339: Could there be separate unit tests for this?

currycoder (approved):
One last comment which I'll leave with you.  Thanks for your patience!

hnryjmes (commented):
happy with code changes so far, I'm still waiting to see the test coverage check come through and then I can approve

api/cases/models.py
saruniitr: Added a test for this.

Also updated the other tests to cover all OGD queues for the cases where feature is supported and not supported.

saruniitr:
> happy with code changes so far, I'm still waiting to see the test coverage check come through and then I can approve

@hnryjmes added few more tests

hnryjmes (approved):

PR #2367, opened 2025-01-10
 LTD-5582-fix-case-search 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5582

There was an issue with Lite search not matching the EOD reports.  A mismatch was reported when ECJU attempt to find cases that were finalised between particular dates which was yielding inconsistent results. 

This is because Lite was dependent on Advice objects and looking at the created data, Advice can be given but a case may be finalised at a much later date i.e Counter signing etc etc. 

The Search has been shifted to use the new LicenceDecision model which is populated when a case is finalised.


markj0hnst0n (approved):
Looks good

api/cases/managers.py
```diff
@@ -169,9 +169,17 @@ def with_submitted_range(self, submitted_from, submitted_to):
     def with_finalised_range(self, finalised_from, finalised_to):
         qs = self.filter(status__status=CaseStatusEnum.FINALISED)
         if finalised_from:
-            qs = qs.filter(advice__level=AdviceLevel.FINAL, advice__created_at__date__gte=finalised_from)
+            qs = qs.filter(
+                Q(licence_decisions__decision=LicenceDecisionType.ISSUED)
+                | Q(licence_decisions__decision=LicenceDecisionType.ISSUED_ON_APPEAL),
```
saruniitr, line 174: this can be joined with the above `finalise` status filter

api/cases/managers.py
```diff
@@ -169,9 +169,17 @@ def with_submitted_range(self, submitted_from, submitted_to):
...
+                licence_decisions__created_at__date__gte=finalised_from,
+            )
         if finalised_to:
-            qs = qs.filter(advice__level=AdviceLevel.FINAL, advice__created_at__date__lte=finalised_to)
+            qs = qs.filter(
+                Q(licence_decisions__decision=LicenceDecisionType.ISSUED)
+                | Q(licence_decisions__decision=LicenceDecisionType.ISSUED_ON_APPEAL),
```
saruniitr, line 180: same as above so we don't need to repeat this.

api/cases/managers.py
```diff
@@ -169,9 +169,17 @@ def with_submitted_range(self, submitted_from, submitted_to):
...
+                licence_decisions__created_at__date__lte=finalised_to,
+            )
```
kevincarrogan, line 182: This only includes issued but I think that refusals are also finalised cases.

kevincarrogan: Do we actually fully know what the expectations are of this filter? Using the licence decisions is a good idea but it is possible that this would bring back unexpected results as well.

depsiatwal: I did do some initial testing and all seemed sensible.  We could de-risk and get UCD to do some testing on a test env thoughts ?

depsiatwal: Further discussion on this at the Jira ticket. So Finalised cases should be received regardless of decision. 
This is probably the best option on the basis that this is fundamentally broken atm.

kevincarrogan (approved):

PR #2368, opened 2025-01-10
Update codestar connection step

james-francis-MT:
Update the pre build step to match the standard way of pulling submodules

### DIFF
--- a/.copilot/phases/pre_build.sh
+++ b/.copilot/phases/pre_build.sh
@@ -3,26 +3,22 @@
 # Exit early if something goes wrong
 set -e
 
-git_clone_base_url="https://codestar-connections.eu-west-2.amazonaws.com/git-http/$AWS_ACCOUNT_ID/eu-west-2/$CODESTAR_CONNECTION_ID/uktrade"
+if [ -f "./.gitmodules" ]; then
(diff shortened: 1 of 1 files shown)


.copilot/phases/pre_build.sh
```diff
@@ -3,26 +3,22 @@
...
-url = $git_clone_base_url/lite-routing.git
-branch = main
-[submodule "django_db_anonymiser"]
-path = django_db_anonymiser
-url = $git_clone_base_url/django-db-anonymiser.git
-EOF
+    sed -i "s|url = git@github.com:uktrade/\(.*\).git|url = $git_clone_base_url/\1.git|g" ./.gitmodules
 
-git submodule update --init --recursive
+    git submodule update --init --remote --recursive
```
currycoder, line 17: `--remote` was dropped from here for a reason, please see my comment in jira from last time; https://uktrade.atlassian.net/browse/LTD-5627?focusedCommentId=162298

Would suggest that we have a comment in the code here to state that we explicitly want to exclude `--remote` as this will ignore any locally locked submodule sha - which is critical for incremental deployments of the lite-routing submodule.

depsiatwal:
This is not a prerequisite for go-live hence  we should be freezing such changes now hence this will be picked up after go-live

PR #2369, opened 2025-01-10
LTD-5779 remove rejected and draft orgs from those checked when exporter trys …

markj0hnst0n:
### Aim

Remove rejected and draft organisations from those checked when exporter tries to register new Organisation.

[Previous PR](https://github.com/uktrade/lite-api/pull/2358) related to this ticket only changed code which validated CRN for user input textbox but not code which actually checked CRN on creation of new org after data was posted to api.

Updated create and added new test for these statuses.

[LTD-5760](https://uktrade.atlassian.net/browse/LTD-5760)

[LTD-5760]: https://uktrade.atlassian.net/browse/LTD-5760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/organisations/serializers.py
```diff
@@ -287,7 +287,11 @@ def validate_registration_number(self, value):
 
         # Check for uniqueness only when creating a new Organisation
         if not self.instance:
-            if Organisation.objects.filter(registration_number=value).exists():
+            if (
+                Organisation.objects.filter(registration_number=value)
+                .exclude(status__in=[OrganisationStatus.REJECTED, OrganisationStatus.DRAFT])
```
currycoder, line 292: Do you happen to know how an organisation could be in draft status?  I had a quick look but could only really find that internal users created active organisations and exporter users created organisations in review.

currycoder (approved):

api/organisations/serializers.py
markj0hnst0n: I'm not 100% that it's possible but just wanted to make sure I fulfilled the request on the ticket.  Colin was the reporter on the original ticket so maybe he might be able to shed some light on it.

Tllew (approved):

PR #2374, opened 2025-01-15
LTD: Refactor Case notes and timeline view to use filter backends

saruniitr:
## Change description

Update notes and timeline filters to use django filter backends.
In bulk approval there is a requirement to show new audit event but the problem was this was not specific to a particular case. One of the solution was to add a filter backend for that and inject in notes and timeline but later it was changed to a different solution where that event is made case specific. However we can use these backends for other filters we have (team, exporter, mentions etc).

Add unit tests for each filtering option. These tests are pytests.
We have pytests in few places and they all need core fixtures to work so they are defined in a central place `api/tests/unit/fixtures/core.py` which can be included as pytest plugin whereever needed. So this also removes some of the redundant `conftest.py` files. If the core fixtures need to moved to a different location, name etc please comment.


api/core/tests/test_filters.py
```diff
@@ -18,6 +18,8 @@
     ROOT_URLCONF="api.core.tests.urls",
 )
 class TestMisconfiguredParentFilter(SimpleTestCase):
+    databases = {"default"}
+
```
saruniitr, line 22: Tests were failing as `SimpleTestCase` cannot have database access and recommended to add this as a solution. The other option is to change it to transaction test case.

PR #2377, opened 2025-01-16
Allow good creation to be skipped when creating a test application

kevincarrogan:
### Aim

When creating a standard application in tests, by default, a good will be generated for you.

In some cases this isn't the desired behaviour and this allows it to be toggled accordingly.


depsiatwal (approved):
lgtm

PR #2378, opened 2025-01-17
LTD-5775: Display all destinations flags on Case detail page

saruniitr:
## Change description

This ensures we see same set of flags on the queue view and case detail page.

[LTD-5775](https://uktrade.atlassian.net/browse/LTD-5775)

[LTD-5775]: https://uktrade.atlassian.net/browse/LTD-5775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/libraries/get_flags.py
+++ b/api/cases/libraries/get_flags.py
@@ -27,6 +27,8 @@ def get_destination_flags(case, case_type):
     if case_type == CaseTypeSubTypeEnum.EUA:
         return Flag.objects.filter(parties__parties_on_application__application_id=case.id)
     elif case_type == CaseTypeSubTypeEnum.STANDARD:
+        # Usually destination level flags are attached to `Party` but some of the flags
+        # are attached to `PartyOnApplication` so gather all of them
         flags = Flag.objects.filter(
             (
                 Q(parties__parties_on_application__application_id=case.id)
--- a/api/cases/libraries/get_flags.py
+++ b/api/cases/libraries/get_flags.py
@@ -1,4 +1,4 @@
-from django.db.models import QuerySet, When, Case as DB_Case, IntegerField, BinaryField
+from django.db.models import Q, QuerySet, When, Case as DB_Case, IntegerField, BinaryField
 
 from api.cases.enums import CaseTypeSubTypeEnum
 from api.cases.models import Case
@@ -27,10 +27,14 @@ def get_destination_flags(case, case_type):
     if case_type == CaseTypeSubTypeEnum.EUA:
         return Flag.objects.filter(parties__parties_on_application__application_id=case.id)
     elif case_type == CaseTypeSubTypeEnum.STANDARD:
-        return Flag.objects.filter(
-            parties__parties_on_application__application_id=case.id,
-            parties__parties_on_application__deleted_at__isnull=True,
+        flags = Flag.objects.filter(
+            (
+                Q(parties__parties_on_application__application_id=case.id)
+                & Q(parties__parties_on_application__deleted_at__isnull=True)
+            )
+            | (Q(parties_on_application__application__pk=case.id) & Q(parties_on_application__deleted_at__isnull=True))
         )
+        return flags
 
     return Flag.objects.none()
 
--- a/api/cases/tests/conftest.py
+++ b/api/cases/tests/conftest.py
@@ -15,6 +15,7 @@ from api.users.tests.factories import (
     ExporterUserFactory,
     GovUserFactory,
     RoleFactory,
+    SystemUserFactory,
     UserOrganisationRelationshipFactory,
 )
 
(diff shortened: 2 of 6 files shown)


api/cases/tests/conftest.py
```diff
@@ -52,14 +53,31 @@ def exporter_headers(exporter_user, organisation):
...
 
 
+@pytest.fixture(autouse=True)
+def system_user():
+    if BaseUser.objects.filter(id=SystemUser.id).exists():
+        return BaseUser.objects.get(id=SystemUser.id)
+    else:
+        return BaseUserFactory(id=SystemUser.id)
+
+
```
kevincarrogan, line 63: I've been doing this a lot but I think this is baked into `factory_boy` itself using https://factoryboy.readthedocs.io/en/stable/orms.html#factory.django.DjangoOptions.django_get_or_create i.e. set `django_get_or_create = ("id",)` and it will automatically do this for us.

api/cases/tests/test_case_flags.py
```diff
@@ -0,0 +1,107 @@
...
+        for good_on_application in application.goods.all():
+            good_on_application.good.flags.add(*good_flags)
+
+        destination_flags = flags_data.get(FlagLevels.DESTINATION, [])
+        party_on_application_flags = flags_data.get(FlagLevels.PARTY_ON_APPLICATION, [])
+        for party_on_application in application.parties.all():
+            party_on_application.party.flags.add(*destination_flags)
+            party_on_application.flags.add(*party_on_application_flags)
+
+        return standard_case
```
kevincarrogan, line 47: Could we maybe use some sort of `factory_boy` factory for this instead of a fixture?

api/cases/tests/test_case_flags.py
```diff
@@ -0,0 +1,107 @@
...
+    response = api_client.get(all_cases_queue_url, **gov_headers)
+    assert response.status_code == 200
+
+    for case in response.json()["results"]["cases"]:
+        all_flags = [
+            item["id"] for flag_level in ["flags", "goods_flags", "destinations_flags"] for item in case[flag_level]
+        ]
+
+        for flags in flags_data.values():
+            assert set(flags).issubset(set(all_flags))
```
kevincarrogan, line 79: Why are we only checking a subset and not checking all of the flags are exactly identical?

saruniitr: Yes because when flagging rules are executed they apply few more flags hence checking the test flags are present or not.

api/cases/tests/test_case_flags.py
saruniitr: may be but all of these flags may not be relevant in other places so went with a fixture specific for these tests.

saruniitr: actually I see what you mean now, we can still pass these as arguments. I will update.

api/cases/libraries/get_flags.py
```diff
@@ -27,10 +27,14 @@ def get_destination_flags(case, case_type):
     if case_type == CaseTypeSubTypeEnum.EUA:
         return Flag.objects.filter(parties__parties_on_application__application_id=case.id)
     elif case_type == CaseTypeSubTypeEnum.STANDARD:
-        return Flag.objects.filter(
-            parties__parties_on_application__application_id=case.id,
-            parties__parties_on_application__deleted_at__isnull=True,
+        flags = Flag.objects.filter(
```
currycoder, line 30: Could I recommend a comment just to make it extremely obvious for the next developer that flags can be attached to Party **and** PartyOnApplication objects?

currycoder (approved):

api/cases/tests/test_case_flags.py
saruniitr: updated

api/cases/tests/test_case_flags.py
saruniitr: moved to a factory

api/cases/tests/conftest.py
saruniitr: This one needs a bit more work as it is creating issues with Exporter and GovUser factories.
`ExporterUserFactory` has `BaseUserFactory` as a sub-factory. With the above change it creates system user by default. As per documentation even if we supply new values it won't take it.
I will try with a separate SystemUserFactory or other options.

api/cases/tests/test_case_flags.py
```diff
@@ -0,0 +1,80 @@
...
+    for case in response.json()["results"]["cases"]:
+        all_flags = [
+            item["id"] for flag_level in ["flags", "goods_flags", "destinations_flags"] for item in case[flag_level]
+        ]
+
+        expected_flags = []
+        for flags in flags_data.values():
+            expected_flags.extend(flags)
+
+        assert set(all_flags).difference(set(expected_flags)) == set()
```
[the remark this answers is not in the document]
saruniitr: in some cases the order is not maintained hence using the `difference`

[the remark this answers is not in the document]
kevincarrogan: `assert set(all_flags) == set(flags_data.values())` would probably be fine then instead.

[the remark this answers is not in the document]
saruniitr: this is updated

kevincarrogan (approved):

PR #2382, opened 2025-01-21
LTD-5781 updated denial_reason description

Tllew:
Removing "Military End Use Control." from the end of the text.

[LTD-5781](https://uktrade.atlassian.net/browse/LTD-5781)

### DIFF
--- a/api/staticdata/denial_reasons/migrations/0007_criterion_1_description_update.py
+++ b/api/staticdata/denial_reasons/migrations/0007_criterion_1_description_update.py
@@ -1,0 +1,19 @@
+from django.db import migrations
+
+
+def update_denial_reason(apps, schema_editor):
+    DenialReason = apps.get_model("denial_reasons", "DenialReason")
+    denial_reason = DenialReason.objects.get(id=1)
+    if denial_reason:
+        denial_reason.description = "Respect for the UK's international obligations and commitments, in particular sanctions adopted by the UN Security Council, agreements on non-proliferation and other subjects, as well as other international obligations."
+        denial_reason.save()
+
+
+class Migration(migrations.Migration):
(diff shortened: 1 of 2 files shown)


api/staticdata/denial_reasons/migrations/0007_criterion_1_description_update.py
```diff
@@ -0,0 +1,19 @@
+from django.db import migrations
+
+
+def update_denial_reason(apps, schema_editor):
+    DenialReason = apps.get_model("denial_reasons", "DenialReason")
+    denial_reason = DenialReason.objects.get(id=1)
+    if denial_reason:
```
depsiatwal, line 7: Is this required denial_reason

get will raise an DenialReason DoesNotExist also do we want to add an additional test to just id to ensure we have the correct object ?

Tllew: it's populated from the csv previously so must exist and I'd rather a .get than a .filter.first() which is what's done in previous tests

the test does assert we have the correct object.

api/staticdata/denial_reasons/migrations/0007_criterion_1_description_update.py
```diff
@@ -0,0 +1,19 @@
+from django.db import migrations
+
+
+def update_denial_reason(apps, schema_editor):
+    DenialReason = apps.get_model("denial_reasons", "DenialReason")
```
currycoder, line 5: Could I suggest a longer-tail follow up to absorb DenialReason in to migrations from the seed commands?  Similar to what we did for CLEs but hopefully a lot less onerous

currycoder (approved):

PR #2388, opened 2025-01-27
LTD-5833-application-history-endpoint 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5817

The exporter case view is being changed to include the application history. 

Rather then expand and over whelm existing 'LARGE' views and endpoints a new endpoint has been created just to extract application history. This uses the new exporter API pattern. 

The history for each application within an amendment is expected to show the full history showing the latest at the top. Hence the data is the same regardless of which case is queried. The UI will remove the hyperlink for the case that's currently being viewed. 

Frontend PR 
https://github.com/uktrade/lite-frontend/pull/2343

### DIFF
--- a/api/applications/exporter/serializers.py
+++ b/api/applications/exporter/serializers.py
@@ -1,9 +1,57 @@
+from api.cases.models import Case
+from api.staticdata.statuses.libraries.get_case_status import get_status_value_from_case_status_enum
+from api.staticdata.statuses.models import CaseStatus
 from rest_framework import serializers
 
 from api.staticdata.statuses.enums import CaseStatusEnum
 
 
 class ApplicationChangeStatusSerializer(serializers.Serializer):
-
     status = serializers.ChoiceField(choices=CaseStatusEnum.all())
     note = serializers.CharField(required=False, allow_blank=True, allow_null=True, max_length=2000)
+
+
+class ApplicationStatusSerializer(serializers.ModelSerializer):
+    status_display = serializers.SerializerMethodField()
+
+    class Meta:
+        model = CaseStatus
+        fields = ("status", "status_display")
+
+    def get_status_display(self, obj):
+        return get_status_value_from_case_status_enum(obj.status)
+
+
+class CaseAmendmentSerializer(serializers.Serializer):
+    status = serializers.SerializerMethodField()
+    ecju_query_count = serializers.SerializerMethodField()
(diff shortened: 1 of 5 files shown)


api/applications/exporter/serializers.py
```diff
@@ -1,9 +1,51 @@
...
+
+    def get_amendment_history(self, instance):
+        amendments = []
+        case_amended = instance
+
+        # Get to the latest case where we have no further amendments
+        while case_amended.superseded_by:
+            case_amended = case_amended.superseded_by
+
+        while case_amended:
```
depsiatwal, line 43: The history  should be the same order and same data regardless of which case we view.

api/applications/exporter/serializers.py
```diff
@@ -1,9 +1,49 @@
...
```
currycoder, line 41: I think this is pretty neat.  Could I suggest making the comments super explicit though e.g.
```
# Travel forwards in amendment chain to find the latest amended case
....
# Go backwards through amendment chain until we find the original case
```

currycoder (commented):
One comment about comments but this looks good

api/applications/exporter/serializers.py
```diff
@@ -1,9 +1,49 @@
...
 
 
 class ApplicationChangeStatusSerializer(serializers.Serializer):
-
     status = serializers.ChoiceField(choices=CaseStatusEnum.all())
     note = serializers.CharField(required=False, allow_blank=True, allow_null=True, max_length=2000)
+
+
+class CaseAmendmentSerializer(serializers.Serializer):
+    status = serializers.SerializerMethodField()
```
currycoder, line 14: This could perhaps be a little nested serializer.  Maybe we have one somewhere already for this

currycoder: In my head just `"status": {"submitted": "Submitted"}` is kind of enough.  The whole `{"key": .., "value": ..}` pattern seems a bit naff

depsiatwal: Nested status serializer added with dedicated field for display.

api/applications/exporter/serializers.py
depsiatwal: added comments hopefully clearer now

currycoder (approved):

Tllew (approved):
Looks good to me

PR #2390, opened 2025-01-28
LTD-5850: Sort the case queryset based on selected sort option

saruniitr:
## Change description

We are adding option to sort the case list in queue view. Initially users can sort based on submission date (newest first or oldest first).
Relevant FE PR is https://github.com/uktrade/lite-frontend/pull/2340

Additional changes include to create pytest fixtures in a central places so it is easier to add pytest based unit tests.

[LTD-5850](https://uktrade.atlassian.net/browse/LTD-5850)

[LTD-5850]: https://uktrade.atlassian.net/browse/LTD-5850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/tests/conftest.py
```diff
@@ -1,154 +0,0 @@
-import pytest
```
currycoder, line 1: Is this file not used by a bunch of tests?  It seems to have been deleted

currycoder: Ah it's been superseded by `tests/unit/fixtures/core.py`

api/cases/views/search/tests/test_case_sorting.py
```diff
@@ -0,0 +1,46 @@
...
+
+pytest_plugins = [
+    "api.tests.unit.fixtures.core",
+]
+
+
+@pytest.fixture
+def cases_list(organisation, submit_application):
+
+    applications = [DraftStandardApplicationFactory(organisation=organisation) for i in range(25)]
```
currycoder, line 17: Could there be a fixture to create these drafts?  A bit like `submit_application` here but
`applications = [get_draft_application() for i in range(25)]`

api/cases/views/search/views.py
```diff
@@ -32,15 +32,16 @@ def get(self, request, *args, **kwargs):
...
         filters = self.get_filters(request)
-        page = self.paginate_queryset(
-            self.get_case_queryset(user, queue_id, is_work_queue, include_hidden, filters).annotate(
-                has_open_queries=Exists(
-                    EcjuQuery.objects.filter(
-                        case=OuterRef("pk"), raised_by_user__team_id=user.team.id, is_query_closed=False
-                    )
-                ),
-            )
+        sort_by_field = filters.get("sort_by", "-submitted_at")
```
currycoder, line 35: I'm curious what the default sorting was before?

api/cases/views/search/tests/test_case_sorting.py
saruniitr: yeah can be rearranged like this.

api/cases/views/search/views.py
saruniitr: So, on 'All cases' queue it was newest first but the otherway round in all of the teams queues.
In the code we never specified the order anywhere so I think it was sorting based on created/updated but cannot be sure.

api/cases/tests/conftest.py
saruniitr: yes, all the common fixtures are included in that file.

currycoder (approved):
Will leave the comment about the fixture change with you

api/tests/unit/fixtures/core.py
kevincarrogan, line 1: Why is this in `core.py` instead of `conftest.py`?

api/tests/unit/fixtures/core.py
```diff
@@ -203,3 +206,8 @@ def _submit_application(draft_application):
 @pytest.fixture
 def standard_case(draft_standard_application, submit_application):
     return submit_application(draft_standard_application)
+
+
+@pytest.fixture
+def final_advice_url(standard_case):
+    return reverse("cases:case_final_advice", kwargs={"pk": standard_case.pk})
```
kevincarrogan, line 213: I think we shouldn't promote this up if we don't need to.

api/tests/unit/fixtures/core.py
```diff
@@ -24,6 +24,11 @@
...
 
 
+@pytest.fixture(autouse=True)
+def django_db(db):
+    return db
+
+
 @pytest.fixture(autouse=True)
 def setup(gov_user):
     pass
```
kevincarrogan, line 34: I know this is from a previous commit but I think we should be very cautious about adding `autouse=True` in these high level fixtures.

api/tests/unit/fixtures/core.py
saruniitr: core fixtures, just a way of organising fixtures instead of everything adding in `conftest.py`.

kevincarrogan: But doesn't everything just end up in this file and it is a bit less obvious where the fixtures are?

Where do we include this in the actual `conftest.py`?

saruniitr: not necessarily, we can break the fixtures down in multiple files as required.
Whereever fixtures are required (eg in api/cases/search/tests etc) we can include these as below,
```
pytest_plugins = [
    "api.tests.unit.fixtures.core",
]
```
No need for additional conftest.py locally.

The idea for this PR atleast is to have something so that it is easier to add pytests. May be there is a better way of organising this, if it is easier I can move them under a top level conftest.py

kevincarrogan (approved):

PR #2393, opened 2025-01-29
Optimise query for licence decision dw endpoint

kevincarrogan:
### Aim


api/applications/serializers/standard_application.py
```diff
@@ -155,18 +155,120 @@ def get_is_amended(self, instance):
...
+
+    def get_superseded_by(self, application):
+        if not application.superseded_by:
+            return None
+        return str(application.superseded_by.pk)
+
+    def get_amendment_of(self, application):
+        if not application.amendment_of:
+            return None
+        return str(application.amendment_of.pk)
```
currycoder, line 270: Could you use `application.superseded_by_id` to skip the lazy load?  Same for `amendment_of_id`

api/data_workspace/v1/tests/test_applications_views.py
```diff
@@ -55,133 +55,42 @@ def test_dw_standard_application_GET(self):
...
+                        "have_you_been_informed": "yes",
                         "is_shipped_waybill_or_lading": True,
-                        "is_suspected_wmd": False,
                         "is_temp_direct_control": None,
-                        "last_closed_at": None,
-                        "licence": None,
-                        "military_end_use_controls_ref": None,
-                        "name": "Test",
-                        "non_waybill_or_lading_route_details": None,
-                        "organisation": {
```
currycoder, line 108: Is this change of contract going to cause any issues downstream?

api/data_workspace/v1/tests/test_good_views.py
```diff
@@ -22,34 +22,12 @@ def test_goods(self):
...
                 "item_category",
-                "is_military_use",
-                "is_component",
-                "uses_information_security",
-                "modified_military_use_details",
-                "component_details",
-                "information_security_details",
-                "is_document_available",
-                "is_document_sensitive",
-                "no_document_comments",
```
currycoder, line 47: Same question as above.. I assume there is some confirmation that these fields don't get used?

api/applications/serializers/standard_application.py
kevincarrogan: Isn't one of these not a real foreign key though? I think I can only do this for one of them.

api/data_workspace/v1/tests/test_good_views.py
currycoder: Resolved on slack

api/data_workspace/v1/tests/test_applications_views.py
currycoder: Resolved on slack

currycoder (approved):

api/applications/serializers/standard_application.py
currycoder: Yep you're right, it will only work for `amendment_of`, not `superseded_by`

saruniitr (approved):
all changes looks good to me.

PR #2403, opened 2025-02-03
LTD: Add model to track case queue movements

saruniitr:
## Change description

This model is used to track the queues the case been through. We create instance of this model when a case is assigned to a particular queue.

Back populate data by going through MOVE_CASE events.
The back population of `exit_date` is not done for the reasons mentioned in the comments. I am looking for comments/suggestions to help with this.

Further changes required in `lite_routing` and API after these changes are merged (https://github.com/uktrade/lite-routing/pull/310). As lite_routing is a submodule, I need these changes merged first to get access to the model there.

lite_routing PR is merged so sha is updated in this PR.

[LTD-5866]: https://uktrade.atlassian.net/browse/LTD-5866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/migrations/0077_populate_case_queue_movements.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@transaction.atomic
+def populate_case_queue_movements(apps, schema_editor):
+    Audit = apps.get_model("audit_trail", "Audit")
+    CaseQueueMovement = apps.get_model("cases", "CaseQueueMovement")
+
+    case_queue_movements = []
+
+    # Some of the early events doesn't have queue_ids in the payload so they are ignored
```
kevincarrogan, line 16: They did have queue names so in theory we could use that data although do you think it's possibly not worth it?

api/cases/migrations/0077_populate_case_queue_movements.py
```diff
@@ -0,0 +1,43 @@
...
+    move_case_qs = Audit.objects.filter(verb=AuditType.MOVE_CASE, payload__queue_ids__isnull=False)
+    for event in move_case_qs:
+        case_id = event.action_object_object_id
+        queue_id = event.payload["queue_ids"][0]
+
+        case_queue_movements.append(
+            CaseQueueMovement(
+                case_id=case_id,
+                queue_id=queue_id,
+            )
```
kevincarrogan, line 26: Where are we setting the dates on these objects?

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: yes, as it is very old case I think its not worth otherwise more queries in the migration.

kevincarrogan: I think the problem might be that I'm going to use this model for Data Workspace as well so I think that we might need that historical data.

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: yes, my bad, fixed it.

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: ok, I will pick up those events as well

api/cases/models.py
```diff
@@ -871,3 +871,10 @@ class Meta:
 
     def __str__(self):
         return f"{self.case.reference_code} - {self.decision} ({self.created_at})"
+
+
+class CaseQueueMovement(TimestampableModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    case = models.ForeignKey(Case, related_name="casequeuemovements", on_delete=models.DO_NOTHING)
+    queue = models.ForeignKey(Queue, related_name="casequeuemovements", on_delete=models.DO_NOTHING)
+    exit_date = models.DateTimeField(blank=True, null=True)
```
saruniitr, line 880: There is no migration added to back populate this field for the following reasons.
1. In some cases the Case went back to the same queue again, it is not possible to correct identify the correct instances and update this field.
2. We need to use `AuditType.UNASSIGNED_QUEUES` and `AuditType.UNASSIGNED` to populate this field. For the second event the queue details are not included in the `payload` so we don't have this data.

I am open to any suggestions to help with (1) to get this field populated.
Instead of populating incorrect data I am thinking it is better to avoid back population.

The idea of this field is we record start and end dates so we can accurately check the time spent on each queue.

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: this is updated

api/cases/migrations/0077_populate_case_queue_movements.py
```diff
@@ -0,0 +1,70 @@
...
+
+    move_case_qs = Audit.objects.filter(
+        verb=AuditType.MOVE_CASE,
+        action_object_object_id__isnull=False,
+        payload__queue_ids__isnull=False,
+        payload__queues__isnull=False,
+    )
+    for event in move_case_qs:
+        case_id = event.action_object_object_id
+        queue_id = event.payload["queue_ids"][0]
```
depsiatwal, line 25: have we sanity checked prod to make sure this payload is always there ?

api/cases/migrations/0077_populate_case_queue_movements.py
```diff
@@ -0,0 +1,70 @@
...
+
+            case_queue_movements.append(
+                CaseQueueMovement(
+                    case_id=case_id,
+                    queue_id=queue.id,
+                    created_at=event.created_at,
+                )
+            )
+
+    CaseQueueMovement.objects.bulk_create(case_queue_movements)
```
depsiatwal, line 56: have we also checked timings or test on PROD like data ?

Just thinking of migration timeouts though it may on be an issue on DBT Platform i'll double check where this migration takes place.

api/cases/migrations/0077_populate_case_queue_movements.py
kevincarrogan, line 25: Why are we only populating the first one of these?

api/cases/migrations/0077_populate_case_queue_movements.py
```diff
@@ -0,0 +1,70 @@
...
+    # Some of the early events only have queue name in the payload so they are processed separately
+    move_case_qs = Audit.objects.filter(
+        verb=AuditType.MOVE_CASE,
+        action_object_object_id__isnull=False,
+        payload__queue_ids__isnull=True,
+        payload__queues__isnull=False,
+    )
+    for event in move_case_qs:
+        case_id = event.action_object_object_id
+        queue_name = event.payload["queues"]
```
kevincarrogan, line 44: I believe this might be an array as well as just a stirng.

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: I am going to deploy this to `demo` and check.

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: It is only one event for each queue.
https://github.com/uktrade/lite-routing/blob/main/routing_rules_internal/routing_engine.py#L49

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: queue_name for some reason is not a list, it is not consistent.
eg https://github.com/uktrade/lite-routing/blob/main/routing_rules_internal/routing_engine.py#L49
https://github.com/uktrade/lite-api/blob/dev/api/workflow/user_queue_assignment.py#L66

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: It should be there, will be deploying to `demo` and test

api/cases/migrations/0077_populate_case_queue_movements.py
```diff
@@ -0,0 +1,70 @@
+# Generated by Django 4.2.16 on 2025-02-03 11:33
+
+from django.db import migrations, transaction
+
+
+from api.audit_trail.enums import AuditType
+
+
+@transaction.atomic
+def populate_case_queue_movements(apps, schema_editor):
```
currycoder, line 10: Has this migration been run on an environment with representative data?  e.g. one of our demo envs which has anonymised data

api/cases/models.py
currycoder, line 880: Something that could be good to record here is `orchestrator`.  e.g. lite routing, a user assigning the queue manually, etc.
Maybe not for right now, though

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: I am going to deploy to `demo` and test

api/cases/migrations/0077_populate_case_queue_movements.py
kevincarrogan:
```
In [9]: set([len(a) for a in Audit.objects.filter(verb="move_case").values_list("payload__queue_ids", flat=True) if a])
Out[9]: {1, 2, 3, 5}
```

saruniitr: this is updated

api/cases/migrations/0077_populate_case_queue_movements.py
saruniitr: deployed to DBT `demo` env, build and deploy successful. model instances created as well.

saruniitr:
@kevincarrogan @depsiatwal @currycoder 
Addressed comments, deployed to DBT `demo`, build and deploy successful so migration applied correctly.
I can see instances of the model created as well.
```
In [1]: from api.cases.models import *

In [2]: CaseQueueMovement.objects.count()
Out[2]: 108216

In [3]: from django.conf import settings

In [4]: settings.ENV
Out[4]: 'demo'
```

marcuspp (approved):

currycoder (approved):

PR #2405, opened 2025-02-04
Update codestar connection env vars

james-francis-MT:
Update the pre build step to match the standard way of pulling submodules


depsiatwal (approved):
lgtm
tested also on dbt platform

PR #2407, opened 2025-02-05
Use created at pagination for party on application

kevincarrogan:
This is to solve an issue where paging without a sort sometimes means that data goes missing when you page through the resultset

### Aim


Tllew (approved):
LGTM

api/cases/migrations/0076_licencedecision_cases_licen_created_850df1_idx.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.17 on 2025-02-05 12:55
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
saruniitr, line 6: 0076 and 0077 are migrations that add `CaseQueueMovement` and back population which are already merged.
Does these need updating?

saruniitr (approved):
A query on migrations, otherwise looks good to me.

api/cases/migrations/0076_licencedecision_cases_licen_created_850df1_idx.py
kevincarrogan: I swapped out the dependencies so that my migrations were at the end.

PR #2410, opened 2025-02-06
LTD-5892-improve-sanction-matching-exact-match

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5892

It has been agreed with the business
that sanction matches will be exact phrase search to reduce the high number of false positives.

We did explore and present a solution that matches in any order but this is preferred for now.

We should get a couple of problem cases from prod with end part signature names that causing issues. 
Create a case with the same names in that environment and present the results to business on a real case.

### DIFF
--- a/api/conf/settings.py
+++ b/api/conf/settings.py
@@ -405,7 +405,7 @@ SANCTION_LIST_SOURCES = env.json(
     {
         "un_sanctions_file": "https://scsanctions.un.org/resources/xml/en/consolidated.xml",
         "office_financial_sanctions_file": "https://ofsistorage.blob.core.windows.net/publishlive/2022format/ConList.xml",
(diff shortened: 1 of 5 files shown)


api/conf/settings.py
```diff
@@ -405,7 +405,7 @@
     {
         "un_sanctions_file": "https://scsanctions.un.org/resources/xml/en/consolidated.xml",
         "office_financial_sanctions_file": "https://ofsistorage.blob.core.windows.net/publishlive/2022format/ConList.xml",
-        "uk_sanctions_file": "https://assets.publishing.service.gov.uk/media/6776a7d49d03f12136308d1c/UK_Sanctions_List.xml",  # /PS-IGNORE
+        "uk_sanctions_file": "https://assets.publishing.service.gov.uk/media/679b8f9eabe77b74cc146c71/UK_Sanctions_List.xml",  # /PS-IGNORE
```
depsiatwal, line 408: updating sanction file at the same time

marcuspp (approved):

PR #2412, opened 2025-02-07
LTD: Update API to be able to sort based on when a case is assigned to queue

saruniitr:
## Change description

Users can now sort list of cases on a given queue based on when it is assigned to that queue.
This is achieved using the new model `CaseQueueMovement` whose instances are created when a queue is assigned to a case (when routing rules are executed). We also record exit_date when the queue is unassigned.

Add tests to cover case progression with and without LU countersigning, when a case is sent back to TAU and also the sorting options.

Relevent FE PR where the options are enabled https://github.com/uktrade/lite-frontend/pull/2353

[LTD-5866]: https://uktrade.atlassian.net/browse/LTD-5866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/tests/test_case_queue_movements.py
```diff
@@ -0,0 +1,300 @@
...
+
+            case.refresh_from_db()
+            assert case.status.status == next_status
+
+        # check queue movement instances are created and recorded correctly
+        obj = CaseQueueMovement.objects.get(case=case, queue=current_queue)
+        assert obj.exit_date.isoformat() == action_time
+
+        for queue in expected_queues:
+            assert CaseQueueMovement.objects.get(case=case, queue=queue, exit_date=None)
```
saruniitr, line 174: this is to ensure we have instances created when routing rules are executed.
The exit_date is checked when these are unassigned in the next iteration.

api/cases/tests/test_case_queue_movements.py
```diff
@@ -0,0 +1,300 @@
...
+            (
+                "2025-01-10T12:00:00+00:00",
+                CaseStatusEnum.SUBMITTED,
+                TeamIdEnum.LICENSING_RECEPTION,
+                QueuesEnum.LICENSING_RECEPTION_SIEL_APPLICATIONS,
+                CaseStatusEnum.INITIAL_CHECKS,
+                [QueuesEnum.TECHNICAL_ASSESSMENT_UNIT_SIELS_TO_REVIEW, QueuesEnum.ENFORCEMENT_UNIT_CASES_TO_REVIEW],
+            ),
+            (
+                "2025-01-11T12:00:00+00:00",
```
saruniitr, line 94: Instead of freezing the time to a constant value it is updated in each iteration. This is because with constant time the fallback routing rule is incorrectly determining that it has not gone through OGDs and routing it back to fallback queues instead of post-circulation. To avoid that the time is made different for each iteration.

api/cases/models.py
```diff
@@ -351,6 +363,11 @@ def move_case_forward(self, queue, user):
         # Run routing rules and move the case forward
         user_queue_assignment_workflow([queue], self)
 
+        if CaseQueueMovement.objects.filter(case=self, queue=queue, exit_date=None).count() == 1:
```
saruniitr, line 366: Usually there is going to be only one event for a queue whose `exit_date` need recording when that queue is unassigned but this can get a bit fiddly when cases go back and users manually assign the cases.
They can change status which triggers routing rules and new queues are assigned or they can manually assign and then try to change status.

In those instances we may see more than 1. To avoid failures and support issues in those corner cases this check is added.

api/applications/views/applications.py
```diff
@@ -342,7 +343,9 @@ def put(self, request, pk):
                     AutoGeneratedDocuments.APPLICATION_FORM,
                     request.build_absolute_uri(),
                 )
-                run_routing_rules(application)
+                queues_assigned = run_routing_rules(application)
+                for queue in queues_assigned:
+                    CaseQueueMovement.objects.create(case=application.case_ptr, queue_id=queue)
```
kevincarrogan, line 348: As we create these records the `created_at` time is going to be ever so slightly different for each of these.

Is it worth maybe setting a single date object first so that they all have the same created at?

api/cases/managers.py
```diff
@@ -334,6 +336,12 @@ def search(  # noqa
                     ),
                 ),
             )
+        ).annotate(
+            time_on_queue=Subquery(
+                CaseQueueMovement.objects.filter(case_id=OuterRef("pk"), queue=queue_id)
+                .values("created_at")
+                .order_by("-created_at")[:1]
+            )
```
kevincarrogan, line 344: Are there any performance implications of running this on each request of the queue list?

api/applications/views/applications.py
saruniitr: Yes, possible that it can be different at seconds level.
I will update to use the same date object but wondering if it really matters as it will be at seconds level.

api/cases/managers.py
saruniitr: I have just deployed to `demo` and not noticing any issues. Access times are similar to submitted at sort option for an OGD queue.

api/applications/views/applications.py
saruniitr: updated

kevincarrogan (commented):
Nothing to be changed here. But I think this probably should have gone out first and then the migrations after, in that way there would be no gaps in data.

kevincarrogan (approved):

Tllew (approved):
LGTM

PR #2415, opened 2025-02-11
LTD-5917: Add user details to CaseQueueMovement model

saruniitr:
## Change description

This field records the user details who assigned a queue to the case.
This can be a System user when routing rules are executed hence `BaseUser` is used.

Add migration to back-populate this field.

[LTD-5917](https://uktrade.atlassian.net/browse/LTD-5917)

[LTD-5917]: https://uktrade.atlassian.net/browse/LTD-5917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/migrations/0081_populate_user_case_queue_movements.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@transaction.atomic
+def populate_user_case_queue_movements(apps, schema_editor):
+    CaseQueueMovement = apps.get_model("cases", "CaseQueueMovement")
+
+    # looking at all MOVE_CASE events in UAT, all of the users are SystemUser
+    # so we can just update user for all instances of CaseQueueMovement
+    # as SystemUser without iterating through MOVE_CASE events.
+    CaseQueueMovement.objects.all().update(user_id=SystemUser.id)
```
saruniitr, line 15:
```
In [8]: move_case_qs = Audit.objects.filter(
   ...:     verb=AuditType.MOVE_CASE,
   ...:     action_object_object_id__isnull=False,
   ...:     payload__queues__isnull=False,
   ...: )

In [9]: move_case_qs.count()
Out[9]: 113168

In [10]: from api.users.models import *

In [11]: actors = move_case_qs.values_list('actor_object_id', flat=True)

In [12]: users = BaseUser.objects.filter(id__in=actors)

In [13]: len(users)
Out[13]: 1

In [14]: users[0].type
Out[14]: 'system'
```

kevincarrogan:
```
In [4]: len(set(Audit.objects.filter(verb="move_case").values_list("actor_object_id", flat=True)))
Out[4]: 59
```

kevincarrogan: You seem to be over filtering here and removing a bunch of queue movement as a result.

kevincarrogan: Looking at the data in the DW table, this also seems to agree with the fact that there are plenty of queue movements that weren't the system user and ideally I want to get to a point where the data in the that table is exactly the same as it is with the audit model.

saruniitr:
I have added more details about this in the slack thread.

PR #2426, opened 2025-02-20
Make audit move pipeline more efficient

kevincarrogan:
### Aim

This makes the audit move case endpoint more efficient.

There's quite a lengthy comment in the code as to why I'm doing it like this but essentially this is the quickest route to getting this endpoint efficient and we can spend more time in the future making it work with the new CaseQueueMovement table.

[LTD-5887](https://uktrade.atlassian.net/browse/LTD-5887)

[LTD-5887]: https://uktrade.atlassian.net/browse/LTD-5887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/data_workspace/urls.py
+++ b/api/data_workspace/urls.py
@@ -3,15 +3,11 @@ from django.urls import (
     path,
 )
 
-from api.data_workspace.v0.urls import router_v0
-from api.data_workspace.v1.urls import router_v1
-from api.data_workspace.v2.urls import router_v2
-
 
 app_name = "data_workspace"
 
 urlpatterns = [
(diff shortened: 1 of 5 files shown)


Tllew (approved):
Looks good to me

depsiatwal (approved):
for data_workspace queries I feel it's fine if this works. 

Sure there many ways that's neater, don't see an issue with going down to SQL and not use the DRM for these cases .

PR #2429, opened 2025-02-24
Add cursor based pagination for goods ratings

kevincarrogan:
### Aim

Update OS Data Workspace endpoints to ensure that they have the relevant cursor pagination where appropriate to solve for missing and/or duplicated data when being ingested into Data Workspace.

[LTD-5935](https://uktrade.atlassian.net/browse/LTD-5935)

[LTD-5935]: https://uktrade.atlassian.net/browse/LTD-5935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/data_workspace/v2/tests/test_serializers.py
```diff
@@ -0,0 +1,41 @@
...
+        subject__name="Just subject",
+    )
+
+    class MockGoodOnApplication:
+        pass
+
+    obj = MockGoodOnApplication()
+    obj.report_summary_prefix_name = None
+    obj.report_summary_subject_name = "Just subject"
+    obj.id = 12345
```
saruniitr, line 22: Is this mainly to avoid creating GoodOnApplication using factory and setting up Good etc?

api/data_workspace/v2/views.py
```diff
@@ -199,23 +206,19 @@ class DataWorkspace:
...
-            )
-            .exclude(good_id__isnull=True)
-            .order_by("rating")
-        )
+class GoodRatingViewSet(BaseViewSet):
+    pagination_class = AssessmentDateCursorPagination
+    serializer_class = GoodRatingSerializer
+    queryset = GoodOnApplication.objects.exclude(control_list_entries__isnull=True).annotate(
+        rating=F("control_list_entries__rating")
+    )
```
saruniitr, line 214: I was thinking we could add `is_good_controlled=False` also in the exclusion but when the control list entries are empty then it will always be False.

saruniitr (approved):
looks good to me

api/data_workspace/v2/tests/test_serializers.py
kevincarrogan: Pretty much. Yes.

api/data_workspace/v2/views.py
```diff
@@ -28,43 +31,42 @@
...
+    @property
+    def pagination_class(self):
+        # Your pagination class should be a cursor pagination based class.
+        # This is to avoid the issue where DW ends up losing or duplicating data
+        # when a query doesn't return consistent results across pages.
+        #
+        # It is also highly recommended that you read the DRF documentation
+        # about cursor based pagination so that you correctly pick the correct
+        # type of field to order on.
+        raise NotImplementedError("You must provide a pagination class that is ideally a cursor paginator.")
```
hnryjmes, line 68: just checking I understand - we prefer CursorPagination because we usually have a unique ordering of items (e.g. GoodDescription is ordered by assessment_date), but for some things (e.g. Units) there is no such ordering and thus LimitOffsetPagination is used ?

kevincarrogan: Yes. In most cases we should find a sensible key to order on that is preferably a date field. In cases where that doesn't really exist (and the dataset is small) we can just use limit offset.

hnryjmes (approved):

PR #2432, opened 2025-02-26
 LTD-5944-f680-doc-generation-migrations 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5944

Migration for f680 letter templates


kevincarrogan (approved):

currycoder (approved):

api/letter_templates/migrations/0011_f680_letter_templates.py
```diff
@@ -0,0 +1,42 @@
...
+
+    f680_approval_layout, _ = LetterLayout.objects.get_or_create(name="F680 Approval", filename="f680_approval")
+    f680_refusal_layout, _ = LetterLayout.objects.get_or_create(name="F680 Refusal", filename="f680_refusal")
+
+    f680_approval_template, _ = LetterTemplates.objects.get_or_create(
+        name="F680 Approval", layout=f680_approval_layout, visible_to_exporter=True, include_digital_signature=True
+    )
+    f680_refusal_template, _ = LetterTemplates.objects.get_or_create(
+        name="F680 Refusal", layout=f680_refusal_layout, visible_to_exporter=True, include_digital_signature=True
+    )
```
currycoder, line 23: You were discussing hardcoding the ID for these in future work, would suggest enforcing that right in the migration to make sure we have consistent ids everywhere.

currycoder (changes requested):

api/letter_templates/migrations/0011_f680_letter_templates.py
depsiatwal: i've decided to make an API call to get the templates based on case type avoiding any hard coding.

currycoder (approved):

PR #2435, opened 2025-02-27
LTD-5947-doc-generation-view

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5947

This is still WIP most likely the filter ability will change and we will add/remove items that are returned. 

This iteration is just returning the template details to the frontend so we can obtain the template id for next step in letter generation. The current view that supports this depends to much on case/advice/goods filtering.


api/letter_templates/views.py
```diff
@@ -276,3 +280,13 @@ def get(self, request, **kwargs):
             return JsonResponse(data=preview, status=status.HTTP_400_BAD_REQUEST)
 
         return JsonResponse(data={"preview": preview}, status=status.HTTP_200_OK)
+
+
+class LetterTemplatesFilter(generics.ListAPIView):
```
currycoder, line 285: Since this is a new endpoint, would suggest that we think about locating it under `api/letter_templates/caseworker/` - ideally I think for most things going forward we will do a caseworker or an exporter endpoint

api/letter_templates/views.py
```diff
@@ -276,3 +280,13 @@ def get(self, request, **kwargs):
...
+    authentication_classes = (GovAuthentication,)
+    serializer_class = LetterTemplateFilterSerializer
+    pagination_class = None
+
+    def get_queryset(self):
+        case_type = self.kwargs["case_type"]
+        return LetterTemplate.objects.filter(case_types__sub_type=case_type)
```
currycoder, line 292: Take a look at django-filter based API filtering; https://github.com/uktrade/lite-api/blob/8d23e0107b462aa5084c611fd71c6f5f804aa500/api/gov_users/caseworker/views.py#L24-L25

It seems like we could do this declaratively with a filter class instead of having to override the queryset

api/letter_templates/serializers.py
```diff
@@ -91,3 +91,11 @@ class LetterTemplateListSerializer(serializers.Serializer):
     case_types = CaseTypeReferenceListSerializer(many=True)
     layout = LetterLayoutReadOnlySerializer()
     updated_at = serializers.DateTimeField()
+
+
+class LetterTemplateFilterSerializer(serializers.ModelSerializer):
```
currycoder, line 96: Putting "Filter" in the name here doesn't seem right, it's just a `LetterTemplateSerializer` - right?

api/letter_templates/views.py
depsiatwal: yeah agree was thinking if this should follow the new pattern

api/letter_templates/serializers.py
depsiatwal: Yeah there there was a clash with  previous one most of these are very heavy weight shifting to new caseworker namespace will I guess solve this

api/letter_templates/caseworker/filters.py
```diff
@@ -0,0 +1,15 @@
...
+class LetterTemplateFilter(filters.FilterSet):
+    case_type = filters.CharFilter(field_name="case_types__sub_type", lookup_expr="exact")
+    decision = filters.CharFilter(field_name="decisions__name", lookup_expr="exact")
+
+    class Meta:
+        model = LetterTemplate
+        fields = (
+            "case_types",
+            "decisions",
+        )
```
currycoder, line 15: Kind of confused how this is working with the attribute at the top called `case_type` and the one in `fields` called `case_types`

api/letter_templates/caseworker/tests/test_view.py
```diff
@@ -0,0 +1,56 @@
...
+            ],
+            [
+                {"decision": "approve", "case_type": "f680_clearance"},
+                ["F680 Approval"],
+                ["Approve"],
+                1,
+            ],
+        ]
+    )
+    def test_letter_templates_list_filter(self, filter, expected_names, expected_descisions, expect_count):
```
currycoder, line 47: Would suggest a few more test methods;
non-authorised user gets a 403
when no templates match the filter, an empty response is given

currycoder (commented):
A few comments inline

api/letter_templates/caseworker/tests/test_view.py
depsiatwal: done

api/letter_templates/caseworker/filters.py
depsiatwal: The way I understand it, The keyword we passing in to the params is case_type
the filter then knows how to translate this field as filters.CharFilter(field_name="case_types__sub_type", lookup_expr="exact")

Same for decision which is a look up for decisions__name

saruniitr: I see this is very generic in the sense that we can simply provide decision as a parameter but it can return multiple types (siel, f680 etc). But in practice we always provide the `case_type` so we only get `decisions`.

saruniitr (approved):

api/letter_templates/caseworker/filters.py
```diff
@@ -0,0 +1,15 @@
+from django_filters import rest_framework as filters
+
+from api.letter_templates.models import LetterTemplate
+
+
+class LetterTemplateFilter(filters.FilterSet):
+    case_type = filters.CharFilter(field_name="case_types__sub_type", lookup_expr="exact")
```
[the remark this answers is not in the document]
depsiatwal: that's how we originally was but decided to implement the filter pattern
that way caller can decided what they want to filter on.

currycoder (approved):

PR #2436, opened 2025-02-28
[LTD-5946] F680 generate approval document

currycoder:
### Aim
This change ensures that we are able to generate an approval document for F680s.

- The approval document generation endpoint assumed that approval documents would have a draft licence.  As far as we know, we will not have a licence object for F680s so this has been made conditional with a new `get_feature()` method on the application manifest.
- The template for the F680 approval document was also enhanced so that we show more application details.

[LTD-5946](https://uktrade.atlassian.net/browse/LTD-5946)

[LTD-5946]: https://uktrade.atlassian.net/browse/LTD-5946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/application_manifest.py
+++ b/api/cases/application_manifest.py
@@ -1,6 +1,6 @@
 from api.applications.models import StandardApplication
 from api.applications.serializers.standard_application import StandardApplicationViewSerializer
-from api.cases.enums import CaseTypeSubTypeEnum
+from api.cases.enums import CaseTypeSubTypeEnum, ApplicationFeatures
 from api.f680.models import F680Application
 from api.f680.caseworker.serializers import F680ApplicationSerializer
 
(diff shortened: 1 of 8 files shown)


api/cases/generated_documents/helpers.py
```diff
@@ -105,8 +105,12 @@ def get_decision_type(advice_type, template):
 
 
 def get_draft_licence(case, advice_type):
```
depsiatwal, line 107: PR looks good however I don't see this feature being used in this PR. is it ear marked for future use ?

api/cases/application_manifest.py
```diff
@@ -1,23 +1,29 @@
...
+    features = {}
+
+    def has_feature(self, feature):
+        return self.features.get(feature, False)
 
 
 class StandardApplicationManifest(BaseManifest):
     model_class = StandardApplication
     caseworker_serializers = {"view": StandardApplicationViewSerializer}
+    features = {ApplicationFeatures.LICENCE_ISSUE: True}
```
saruniitr, line 20: Instead of defining features, could it be better if we define the feature dict specific to case type (`SIELApplicationFeatures, F680ApplicationFeatures` or similar) and just assign it here?
When we add new features then we don't need to change this.

saruniitr (approved):
One comment to consider.

api/cases/application_manifest.py
depsiatwal: yeah also feel this needs to be see how we pan this out and decide to use it.

depsiatwal (approved):
lgtm

PR #2437, opened 2025-02-28
Bad request debugging

kevincarrogan:
### Aim

Some of our endpoints return a 400 with error details and it's often hard to debug these

This middleware simply adds a debug log that can then be added as a logger to see the error output of these bad request responses

```
ADDITIONAL_LOGGER_CONFIG='{"api.conf.middleware": {"handlers": ["stdout"], "level": "DEBUG", "propagate": false}}'
```


depsiatwal (approved):
Nice as this can be annoying.

PR #2443, opened 2025-03-05
LTD-5958 F680 should not route to countersign queue

markj0hnst0n:
### Aim

Leveraging the new application manifest pattern stopped the F680 approval application from routing to an unnecessary countersign queue.  It now goes straight to the under final review queue once OGDs are done which is what we require.

[LTD-5958](https://uktrade.atlassian.net/browse/LTD-5958)

[LTD-5958]: https://uktrade.atlassian.net/browse/LTD-5958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/application_manifest.py
```diff
@@ -17,13 +17,19 @@ def has_feature(self, feature):
 class StandardApplicationManifest(BaseManifest):
     model_class = StandardApplication
     caseworker_serializers = {"view": StandardApplicationViewSerializer}
-    features = {ApplicationFeatures.LICENCE_ISSUE: True}
+    features = {
+        ApplicationFeatures.LICENCE_ISSUE: True,
+        ApplicationFeatures.CAN_REQUIRE_COUNTERSIGN: True,
```
saruniitr, line 22: I think we need a better name for this, perhaps `ROUTE_TO_COUNTERSIGNING_QUEUES`? as that is what we are allowing/not-allowing based on this feature flag?

Countersigning means different things for different teams and this currently doesn't differentiate.
Eg LU countersigning depends on the destinations on the application and this feature flag is not related to that.

api/workflow/tests/test_user_queue_assignment.py
```diff
@@ -100,6 +102,22 @@ def test_countersigning_queue(self):
         self.assertNotIn(self.queue, self.case.queues.all())
         self.assertEqual(self.case.status, old_status)
 
+    def test_unsupported_application_type_skips_countersigning_queue_assignment(self):
+        """
+        Tests that countersigning queues are not assigned when work queue removed if the application type does not
+        support countersigning
+        """
```
saruniitr, line 109: This is fine as a unit test but I think we need a integration test where we trigger move case forward and the case doesn't get routed to countersigning queue for F680.
Eg https://github.com/uktrade/lite-api/blob/dev/api/queues/caseworker/tests/test_bulk_approval.py#L151
After bulk approval internally it moves the case forward and ensures the queues are as expected. Not necessarily the same pattern but the idea is to move case forward and ensure it is going to the expected queues.

api/cases/application_manifest.py
markj0hnst0n: Changed 🙂

api/workflow/tests/test_user_queue_assignment.py
markj0hnst0n: I've added a test in here so hopefully I've understood the requirements!

saruniitr (approved):
looks good to me now

api/workflow/user_queue_assignment.py
```diff
@@ -44,15 +45,9 @@ def get_next_compliance_visit_status(case):
...
 
-def user_queue_assignment_workflow(queues: [Queue], case: Case):
-
-    # Remove case from queues where all gov users are done with the case
-    queues_without_case_assignments = set(queues) - get_queues_with_case_assignments(case)
-    case.queues.remove(*queues_without_case_assignments)
-
-    system_user = BaseUser.objects.get(id=SystemUser.id)
-
+def look_for_countersign_queues(case, queues_without_case_assignments):
```
currycoder, line 48: Would suggest calling this `assign_to_countersign_queues`

currycoder, line 48: Could we perhaps have a docstring explaining what this function does?

currycoder (approved):
A couple of comments that I'll leave with you

PR #2446, opened 2025-03-07
Add default value for end to end tests for uploading documents

kevincarrogan:
### Aim

Update the default local env files for easier initial setup.


depsiatwal (approved):
a

PR #2447, opened 2025-03-07
Raise exception if in debug mode when uploading documents in tests

kevincarrogan:
### Aim

When in debug mode, if there is a file upload issue then raise the exception.

This should make it easier to debug exactly what failed when an upload error occurs, especially when running e2e tests in CI.


api/staticdata/upload_document_for_tests/views.py
```diff
@@ -32,6 +32,8 @@ def get(self, request):
         try:
             s3.upload_file(file_to_upload_abs_path, settings.AWS_STORAGE_BUCKET_NAME, s3_key)
         except Exception as e:  # noqa
+            if settings.DEBUG:
```
hnryjmes, line 35: now that we use minio in CI tests, this should make things easier to debug

hnryjmes (approved):

currycoder (approved):

PR #2448, opened 2025-03-07
LTD-5982-supporting-documents 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5967

This is a new endpoint to support F680 supporting-documents on a application. 

This is moving to the new exporter structure.  Technically these may not have been required but feel it's cleaner.


api/applications/exporter/serializers.py
```diff
@@ -55,3 +60,24 @@ def get_amendment_history(self, instance):
...
+        model = ApplicationDocument
+        fields = "__all__"
+
+    def create(self, validated_data):
+        # This rich is moved here to avoid heavy helpers in the view keeping view as simple
+        # TODO move these to models
+        document = super().create(validated_data)
+        document.save()
+        process_document(document)
+        audit_trail_service.create(
```
depsiatwal, line 77: I've put the audit log here so we have same as existing functionality. 
But I question do we actually need this ?

api/applications/exporter/views/tests/test_applications.py
```diff
@@ -173,3 +175,105 @@ def test_get_history_application_wrong_organisation(self):
...
+            "size": 476,
+            "description": "banana cake 1",
+            "application": self.application.id,
+        }
+
+    def test_get_application_documents(self):
+
+        my_doc_1 = ApplicationDocumentFactory(application=self.application)
+        my_doc_2 = ApplicationDocumentFactory(application=self.application)
+        ApplicationDocumentFactory()
```
markj0hnst0n, line 204: Don't think this is used or required

markj0hnst0n (commented):
Just one thing

api/applications/exporter/views/tests/test_applications.py
depsiatwal: it creates a third application just want to make sure it returns the 2 I expected and not the 3rd

markj0hnst0n (approved):

api/applications/exporter/views/tests/test_applications.py
markj0hnst0n: I get you.  That makes sense.  Probably would be clearer if you labeled the document and asserted it was not in the response json but happy to approve and leave that comment with you.

api/applications/exporter/serializers.py
saruniitr, line 77: Is it better to move these to the view then there is no need of this `create` method?
As it is an existing audit event I think it is useful to log this.

api/applications/exporter/views/tests/test_applications.py
```diff
@@ -173,3 +175,105 @@ def test_get_history_application_wrong_organisation(self):
         response = self.client.get(url, **self.exporter_headers)
 
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+
+class TestApplicationDocuments(DataTestClient):
+
+    def setUp(self):
+        super().setUp()
+        self.application = StandardApplicationFactory(organisation=self.exporter_user.organisation)
```
saruniitr, line 184: Could we create pytest based tests?
Also it is creating Standard application only, do we need to test for F680 as well?

api/applications/exporter/serializers.py
depsiatwal: The general conscious recently was to move stuff away from the view to avoid fat views, 
similar pattern to other new endpoints we created. 

The view as it stands now is very learn talking advantage of Django API.

api/applications/exporter/views/tests/test_applications.py
depsiatwal: done

saruniitr (approved):

PR #2451, opened 2025-03-10
[LTD-5969] F680 related modelling

currycoder:
### Aim

This change adds three new models for F680-related records;
- Recipient - the entity receiving the product components/information
- Product - the overarching product for the F680; a single overarching product
- SecurityReleaseRequest - a request for a single component/information to go to a Recipient and be released at a certain security grading

These records are created on application submission by extracting the required fields from the application JSON.  Firstly the JSON is validated with a serializer.  The serializer/JSON contract could do with some improvements however this will follow in subsequent PRs. 

[LTD-5969](https://uktrade.atlassian.net/browse/LTD-5969)

[LTD-5969]: https://uktrade.atlassian.net/browse/LTD-5969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/f680/exporter/serializers.py
```diff
@@ -29,3 +29,82 @@ def create(self, validated_data):
         validated_data["status"] = self.context["default_status"]
         validated_data["case_type_id"] = self.context["case_type_id"]
         return super().create(validated_data)
+
+
+def required_fields(required_keys, data):
```
currycoder, line 34: Note that this is a bit of a necessary evil at the moment since we have a format like so;
```
"fields": [{"key": "my-key", "answer": ...}]
```

I'd like to switch this format so that the fields value is a dictionary instead.  That will allow us to get a lot more bang for buck out of DRF's serializers and avoid these kinds of oddities.

api/f680/enums.py
```diff
@@ -0,0 +1,78 @@
...
+    AGENT_BROKER = "agent-broker"
+    INTERMEDIATE_CONSIGNEE = "intermediate-consignee"
+    AUTHORISED_SUBMITTER = "authorised-submitter"
+    CONSULTANT = "consultant"
+    CONTACT = "contact"
+    EXPORTER = "exporter"
+    OTHER = "other"
+
+    choices = [
+        (AGENT_BROKER, "End-user"),
```
saruniitr, line 23: Is it meant to be "Agent-broker" instead of "End-user" here?

currycoder: Great spot Arun, fixed it

api/f680/exporter/serializers.py
```diff
@@ -29,3 +29,82 @@ def create(self, validated_data):
...
+                "security_classification",
+                "end_user_intended_end_use",
+            ],
+            data,
+        )
+        return data
+
+
+class UserInformationSerializer(serializers.Serializer):
+    type = serializers.ChoiceField(choices=["multiple"])
```
depsiatwal, line 71: I know we only have two choices but any value in having this in as a constants choices enum ?

api/f680/models.py
```diff
@@ -1,25 +1,111 @@
...
+    type = models.CharField(choices=enums.RecipientType.choices, max_length=50)
+    role = models.CharField(choices=enums.RecipientRole.choices, max_length=50, default=None, null=True)
+    role_other = models.TextField(null=True, default=None)
+    organisation = models.ForeignKey(Organisation, related_name="organisation_recipient", on_delete=models.CASCADE)
+
+
+# TODO: Eventually we may want to use this model more widely.  We can do that
+#   but for now baking it in to the f680 application avoids us having to guess
+#   at unknown futures
+class Product(models.Model):
```
saruniitr, line 91: Could this one and `Recipient` and `SecurityReleaseRequest` be a `TimestampableModel`?

depsiatwal: I think eventually we may want to derive from the historical log tables ?

api/f680/models.py
```diff
@@ -1,25 +1,111 @@
...
+            ),
+        )
+
+        # Create a Recipient and SecurityRelease for each.  In F680s caseworkers
+        #   will advise against SecurityRelease records
+        for item in self.application["sections"]["user_information"]["items"]:
+            item_fields = item["fields"]
+
+            recipient = Recipient.objects.create(
+                name=self.get_field_value(item_fields, "end_user_name"),
```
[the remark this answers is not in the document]
currycoder: Yes it is.  Probably the name could be better on the frontend.  I might update it in my subsequent ticket for adjusting the JSON contract.  We just obviously need to be careful about this sort of thing as we progress to production.

api/f680/models.py
currycoder: I've gone with `TimestampableModel` for now - I think it's a decent pattern that we should try to reuse as a supplement to the history/versioning we have elsewhere.

api/f680/exporter/serializers.py
depsiatwal: I think even if we had a class to define the choices and just use a constant 
[type.MULTIPLE]

depsiatwal: but no biggie for now approving will leave with you

currycoder: Ohhh I see.  Yeah I can do that

PR #2456, opened 2025-03-14
Fix settings in tests

kevincarrogan:
Fix settings in tests so they aren't based on local .env file and therefore may fail if you don't have exactly the right values set.


depsiatwal (approved):
lgtm

api/users/tests/test_create_exporter.py
```diff
@@ -64,7 +68,7 @@ def test_create_exporter_user_when_user_already_exists_doesnt_create_new_user(se
             self.data["email"],
             {
                 "organisation_name": other_org.name,
-                "exporter_frontend_url": "https://exporter.lite.service.localhost.uktrade.digital/",
+                "exporter_frontend_url": "https://exporter.lite.example.com/",
```
hnryjmes, line 71: can `exporter_frontend_url` be any value as long as its consistent across these tests?

hnryjmes: I assume it just needs to match `EXPORTER_BASE_URL`

kevincarrogan: Yes. It picks up the value from `EXPORTER_BASE_URL` but this may have been overridden in your .env file so tests could suddenly break.

This just makes sure that the tests are isolated from your local .env and have a consistent value regardless.

hnryjmes (approved):

PR #2458, opened 2025-03-16
LTD-5992: Add Recommendation model for F680 and view

saruniitr:
## Change description

Add new model to capture recommendation given by users on an F680 application.
Existing `Advice` is not suitable for this as F680 is not a licence.

Add view to handle creating, retrieving and deleting recommendation and unit tests.

[LTD-5992](https://uktrade.atlassian.net/browse/LTD-5992)


api/f680/caseworker/views.py
```diff
@@ -0,0 +1,46 @@
...
+            }
+            for item in request_data
+        ]
+
+    def create(self, request, *args, **kwargs):
+        data = self.prepare_data(request.data.copy())
+        serializer = self.get_serializer(data=data, many=True)
+        serializer.is_valid(raise_exception=True)
+
+        self.delete_user_recommendation(self.request.user)
```
depsiatwal, line 39: why do we delete the recommendation here ?

api/f680/caseworker/views.py
```diff
@@ -0,0 +1,46 @@
...
+from api.f680.caseworker.serializers import F680RecommendationSerializer
+
+
+class F680RecommendationViewSet(viewsets.ModelViewSet):
+    authentication_classes = (GovAuthentication,)
+    serializer_class = F680RecommendationSerializer
+    pagination_class = None
+
+    def get_queryset(self):
+        return Recommendation.objects.filter(case_id=self.kwargs["pk"])
```
currycoder, line 16: Could this be a filter class?

api/f680/caseworker/views.py
currycoder: Yeah I'm not sure about whether it feels right to delete an existing recommendation as a side effect of creating a new one.

I think as this is, we wouldn't be able to build a frontend flow that depended upon several incremental API calls for recommendations - each API call to create a recommendation would clear all the previous ones for that user.

What about for now just 400'ing if there already exists a recommendation for the user+team+security release ID?

api/f680/caseworker/views.py
```diff
@@ -0,0 +1,46 @@
...
+        serializer = self.get_serializer(data=data, many=True)
+        serializer.is_valid(raise_exception=True)
+
+        self.delete_user_recommendation(self.request.user)
+        self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    def destroy(self, request, *args, **kwargs):
+        self.delete_user_recommendation(request.user)
```
currycoder, line 45: I'm not sure if bluntly deleting all recommendations for the user+team+case is what we want, but OK for this to go in for now.

api/f680/caseworker/tests/test_views.py
```diff
@@ -0,0 +1,135 @@
...
+    f680_application = get_f680_application()
+
+    data = [
+        {
+            "type": RecommendationType.APPROVE,
+            "security_grading": SecurityGrading.OFFICIAL_SENSITIVE,
+            "conditions": f"Conditions for {rr.recipient.country.name}",
+            "refusal_reasons": "",
+            "security_release_request": str(rr.id),
+        }
```
currycoder, line 77: Could we have more tests that prove that validation happens correctly and a 400 is returned when the data sent to the API is unexpected?

api/f680/caseworker/tests/test_views.py
```diff
@@ -0,0 +1,135 @@
...
+    response = api_client.get(url, **headers)
+    assert response.status_code == 200
+    assert len(response.json()) == f680_application.security_release_requests.count()
+
+    assert Recommendation.objects.count() == (
+        f680_application.security_release_requests.count() + another_f680_application.security_release_requests.count()
+    )
+
+
+def test_POST_recommendation_success(api_client, get_f680_application, team_case_advisor_headers):
```
currycoder, line 67: Could we have tests for when the F680 application cannot be found?

api/f680/caseworker/views.py
currycoder: Could we enforce that recommendations only happen when the case is in OGD advice..?

api/f680/caseworker/tests/test_views.py
```diff
@@ -0,0 +1,135 @@
...
+        ]
+        for recipient in recipients:
+            F680SecurityReleaseRequestFactory(application=application, recipient=recipient)
+
+        return application
+
+    return _get_f680_application
+
+
+def test_GET_recommendation_success(api_client, get_f680_application, team_case_advisor_headers):
```
currycoder, line 44: Similarly, could we have tests to prove that this raises a 404 as appropriate.

api/f680/models.py
```diff
@@ -110,3 +113,18 @@ class SecurityReleaseRequest(TimestampableModel):
     approval_types = ArrayField(models.CharField(choices=enums.ApprovalTypes.choices, max_length=50))
     # We need details of the release, this doesn't appear to be in the frontend flows yet..
     intended_use = models.TextField()
+
+
+class Recommendation(TimestampableModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    case = models.ForeignKey(Case, related_name="recommendations", on_delete=models.CASCADE)
+    type = models.CharField(choices=enums.RecommendationType.choices, max_length=30)
```
currycoder, line 121: "Not applicable" is exciting

api/f680/caseworker/views.py
saruniitr: this added some confusion - this can be removed but then we need a method to handle editing of user recommendation.
It is unlikely the same user gives recommendation twice as FE simply presents edit option when that user is already given recommendation so this is not an issue.
I will remove this to address this.

api/f680/caseworker/views.py
saruniitr: it is only deleting the recommendation given by current user. This is only required if we want to give option to user to clear their recommendation and give new one.

api/f680/caseworker/views.py
saruniitr: we also need for under final review or you want to limit to ogd advice for now?

api/f680/caseworker/tests/test_views.py
currycoder: Yeah so passing in an unknown application ID, do we get a 404?  Hopefully the test is fairly simple to prove that

api/f680/caseworker/views.py
currycoder: Yeah I suppose what I'm hinting at is that it feels like clearing *all* user recommendations for the given team/case seems pretty blunt.  e.g. I could see a situation where a user may want to tick a few recommendations and clear those only.  But I suppose we can add that easily enough later if the need comes up!

api/f680/caseworker/views.py
currycoder: I think we may end up using a whole different model for final review.  Not totally sure yet, but seems ok to make this OGD advice only for now

api/f680/caseworker/views.py
saruniitr: That will cause issues as currently we create recommendation for all release requests. If we are going to selectively delete then first we need an option to create them separately as well.
We can add these later, for now we create for all release requests and clear for all as well.

api/f680/caseworker/tests/test_views.py
saruniitr: Previously this is not possible as application is not retrieved but now a permission class requires application so this test is possible and added now.

api/f680/caseworker/tests/test_views.py
```diff
@@ -0,0 +1,235 @@
...
+    def _recommendation_url(f680_application):
+        return reverse("caseworker_f680:recommendation", kwargs={"pk": f680_application.id})
+
+    return _recommendation_url
+
+
+@pytest.fixture
+def get_f680_application(organisation):
+
+    def _get_f680_application():
```
depsiatwal, line 36: I also forget about using functions in fixture very nicely done and handy :-)

api/f680/caseworker/permissions.py
```diff
@@ -0,0 +1,28 @@
...
+    TeamIdEnum.MOD_DSR,
+)
+
+
+class CaseCanAcceptRecommendations(permissions.BasePermission):
+    def has_permission(self, request, view):
+        return view.get_case().status.status == CaseStatusEnum.OGD_ADVICE
+
+
+class CaseUserMakeRecommendations(permissions.BasePermission):
```
depsiatwal, line 18: wonder if we can rename this to be more permission like 
little confusing 

CaseCanUserMakeRecommondation ?

saruniitr: Actually it is a typo, I wanted to name it as `CanUserMakeRecommondation`, will update this.

depsiatwal (approved):
lgtm

currycoder (approved):

PR #2459, opened 2025-03-17
Get the relevant case type value for an enum type

kevincarrogan:
### Aim

Get the case type identifier from the specific enum class from the case type object.

This is in preparation for routing changes where routing rules can be defined at different levels of the case type hierarchy.

[LTD-5979](https://uktrade.atlassian.net/browse/LTD-5979)

[LTD-5979]: https://uktrade.atlassian.net/browse/LTD-5979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/cases/enums.py
+++ b/api/cases/enums.py
@@ -79,6 +79,10 @@ class CaseTypeTypeEnum:
     def as_list(cls):
         return [{"key": choice[0], "value": choice[1]} for choice in cls.choices]
 
+    @classmethod
+    def get_case_type(cls, case_type):
+        return case_type.type
+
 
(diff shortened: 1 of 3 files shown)


Tllew (approved):
Looks good to me, if you happen to go near where this would be intended to be used I'd be interested to see what causes either a CaseTypeTypeEnum or a CaseTypeSubTypeEnum to be passed to that function

marcuspp (approved):

PR #2462, opened 2025-03-19
[LTD-5993] Change F680 JSON contract for fields to be dict

currycoder:
### Aim

Change the F680 application JSON contract so that `fields` is a dictionary as opposed to an array.  This allows us to have a much easier time with DRF serializer validation and additionally makes lookups much easier.

Companion frontend PR: https://github.com/uktrade/lite-frontend/pull/2419

**NOTE** On merge we will need to delete existing F680 applications on dev2 as this contract change is backwards incompatible.  We could do a data migration, but since we only have test applications on a dev environment this would be overkill.

[LTD-5993](https://uktrade.atlassian.net/browse/LTD-5993)

[LTD-5993]: https://uktrade.atlassian.net/browse/LTD-5993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/f680/exporter/serializers.py
+++ b/api/f680/exporter/serializers.py
@@ -11,7 +11,7 @@ class SectionType:
     MULTIPLE = "multiple"
 
 
-class F680ApplicationSerializer(serializers.ModelSerializer):  # /PS-IGNORE
+class F680ApplicationSerializer(serializers.ModelSerializer):
     status = serializers.CharField(source="status__status", read_only=True)
     organisation = RelatedOrganisationSerializer(read_only=True)
     submitted_by = RelatedExporterUserSerializer(read_only=True)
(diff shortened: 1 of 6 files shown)


api/f680/exporter/serializers.py
```diff
@@ -53,55 +44,54 @@ class FieldSerializer(serializers.Serializer):
     datatype = serializers.ChoiceField(choices=["list", "string", "boolean", "date"])
 
 
+class UserItemFieldsSerializer(serializers.Serializer):
+    entity_type = FieldSerializer()
```
depsiatwal, line 48: I see why we doing this validation etc etc 
but isn't this morning further away from a lose JSON contract ?
will we need to define all UI fields here now ?

That may be ok but just wondering how much more complexity we may need to add

currycoder: There's another couple of steps that I have in mind here: 
1) We further tighten the validation for some of these fields (e.g. making it so that `approval_type` values must conform to the choices allowed on the API-side)
2) We make it so that any 400 errors on submission are correctly fed back to the user so that they are invited to go back to certain sections of the application and update their answers.  This second one will hopefully get the users to make the JSON data forward compatible for us if we happen to miss doing it ourselves.

depsiatwal: ok that has convinced me tested locally all working well :-)

PR #2466, opened 2025-03-20
LTD-6014: F680 make recommendation endpoint updates

saruniitr:
## Change description

These are the changes identified while update the FE to use `Recommendation` model.

Relevant FE PR is https://github.com/uktrade/lite-frontend/pull/2423

[LTD-6014](https://uktrade.atlassian.net/browse/LTD-6014)

[LTD-6014]: https://uktrade.atlassian.net/browse/LTD-6014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/f680/caseworker/read_only_serializers.py
```diff
@@ -0,0 +1,72 @@
...
+    intended_use = serializers.CharField(read_only=True)
+
+
+class F680RecommendationViewSerializer(serializers.ModelSerializer):
+    case = serializers.ReadOnlyField(source="case_id")
+    user = GovUserViewSerializer()
+    team = TeamViewSerializer()
+    type = KeyValueChoiceField(choices=RecommendationType.choices)
+    security_grading = KeyValueChoiceField(choices=SecurityGrading.security_release_choices)
+    security_release_request = F680SecurityReleaseRequestSerializer()
```
currycoder, line 55: Thinking this through..  Could we not keep this endpoint much more lightweight and avoid API data duplication by making this a UUID reference instead?  I believe everything you need for security release requests is returned in the application JSON response which should be available on every frontend view.  Although you would need a step somewhere in the frontend to stitch the two together.

In general, I feel like it's a good time to start thinking about lightweight API requests as the caseworker API endpoints up until now have been fairly heavyweight on SIELs - with lots of nested/duplicated data everywhere.

saruniitr: I added these to avoid depending on the json sections.
I can update to use json instead if that is the preferred option.

currycoder: I'm not saying that we use the JSON sections, I'm saying that we just have references by ID for `security_release_request` instead of a bunch of nested objects.  That's because we have everything we need in the `case["data"]` for product/security release (based on the ORM records, not the application JSON)

currycoder: > is returned in the application JSON response 
This was the confusing bit - sorry.  I meant in the `case["data"]` response

saruniitr: It should be possible, will keep these as ids but need other bits like `user`, `team` etc as we group, filter.

currycoder: Agreed about `user` and `team`

api/f680/caseworker/views.py
```diff
@@ -52,6 +53,11 @@ def create(self, request, *args, **kwargs):
         headers = self.get_success_headers(serializer.data)
         return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
 
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+        serializer = F680RecommendationViewSerializer(queryset, many=True)
```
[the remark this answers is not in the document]
saruniitr: updated

currycoder (approved):
This looks good.  Thanks for your patience

depsiatwal (approved):
lgtm

PR #2468, opened 2025-03-21
Export licence case type

kevincarrogan:
### Aim

Allows export licence case type application to be submitted.

The main part of this is to take the current application endpoints and make the specific to SIELs (and now export licences). A lot of the changes are where I'm removing various `if` statements that are checking for the specific type of application and then doing something different.

[LTD-6040](https://uktrade.atlassian.net/browse/LTD-6040)

[LTD-6040]: https://uktrade.atlassian.net/browse/LTD-6040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/parties/serializers.py
```diff
@@ -94,9 +92,6 @@ def __init__(self, *args, **kwargs):
             else:
                 self.fields.pop("sub_type_other")
 
-            if application_type == CaseTypeSubTypeEnum.F680:
-                self.fields["clearance_level"].required = True
-
```
hnryjmes, line 99: do F680s have their own party serializer now

kevincarrogan: F680s does its own thing now so doesn't need this at all.

api/core/context_processors.py
```diff
@@ -3,13 +3,23 @@
...
-class ApplicationSerializerContextProcessor:
+def organisation_serializer_context_processor(request):
+    return {
+        "organisation": get_request_user_organisation(request),
+    }
+
+
+def draft_status_serializer_context_processor(request):
+    return {
+        "default_status": get_case_status_by_status(CaseStatusEnum.DRAFT),
```
hnryjmes, line 14: if this is static data which will be the same every time it's run does it need to be set in a context processor? is there a simpler way to do it?

kevincarrogan: This is the simplification as it is a reusable way to set the default status for serializers that desire to have that.

A serializer needs to set this at some point so this does that for you.

PR #2471, opened 2025-03-24
Updating organisation and party DW endpoints to use cursor pagination

kevincarrogan:
### Aim

Use cursor pagination for organisation and party endpoints to decrease the number of pipeline failures due to duplicate entries that are caused by non-deterministic results across pages.

[LTD-6042](https://uktrade.atlassian.net/browse/LTD-6042)

[LTD-6042]: https://uktrade.atlassian.net/browse/LTD-6042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/data_workspace/v1/views.py
+++ b/api/data_workspace/v1/views.py
@@ -1,6 +1,7 @@
 from rest_framework import viewsets
 from rest_framework.pagination import LimitOffsetPagination
 
+from api.conf.pagination import CreatedAtCursorPagination
 from api.core.authentication import DataWorkspaceOnlyAuthentication
 from api.organisations.models import Organisation
 from api.organisations.serializers import OrganisationDetailSerializer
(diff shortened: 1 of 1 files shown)


api/data_workspace/v1/views.py
```diff
@@ -17,27 +18,15 @@
...
-    pagination_class = LimitOffsetPagination
-    # We ensure that organisations with the same name always come out
-    # in the same order by additionally ordering by created_at
-    # and by id. If just name is used, the order is non-deterministic
-    # when multiple organisations have the same name. This can mean the
-    # same organisation being emitted on different pages, causing
-    # primary key violations in Data Workspace.
-    queryset = Organisation.objects.order_by("name", "created_at", "id")
+    pagination_class = CreatedAtCursorPagination
+    queryset = Organisation.objects.all()
```
hnryjmes, line 22: have you tested that this solves the problem, if so how?

hnryjmes (approved):
if it's broken currently, we should probably try and fix it

api/data_workspace/v1/views.py
depsiatwal: I know this has been discussed before 
but no unit tests either to test these scenarios ?

kevincarrogan: I've tried to write tests to test this scenario but I've not been able to do something that reproduces them locally.

depsiatwal (approved):
approved

PR #2473, opened 2025-03-24
Remove unnecessary lookup of goods flags

kevincarrogan:
### Aim

Remove unnecessary look for goods flags on search view.


Tllew (approved):
Does this need tests?

kevincarrogan:
> Does this need tests?

I'm really just removing something, that actually doesn't do anything at all, so I'm not certain that tests would be particularly useful here.

From the perspective of the output of this function nothing has changed.

currycoder (approved):

PR #2474, opened 2025-03-25
[LTD-6026] Add SecurityReleaseOutcome model, migration and factory

currycoder:
### Aim

Adds a new SecurityReleaseOutcome model as per the agreed F680 modelling plan.  Also adds a factory for the model which we may be able to use to decouple creating outcome letters from creating the outcome flows.

[LTD-6026](https://uktrade.atlassian.net/browse/LTD-6026)

[LTD-6026]: https://uktrade.atlassian.net/browse/LTD-6026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/f680/enums.py
```diff
@@ -88,3 +95,13 @@ class RecommendationType:
         (REFUSE, "Refuse"),
         (NOT_APPLICABLE, "Not applicable"),
     ]
+
+
+class SecurityReleaseOutcomes:
+    APPROVE = "approve"
+    REFUSE = "refuse"
```
saruniitr, line 102: Is it worth adding `NOT_APPLICABLE` as we have in recommendation type?

currycoder: I think at the point where MOD-ECJU advise it's either "yes" or "no"

saruniitr (approved):

PR #2475, opened 2025-03-25
LTD: Filter F680 recommendation queryset for the current case

saruniitr:
## Change description

Currently it is looking into all recommendation objects, because of this if user made recommendations in another case we fail in permission class and not allow to add recommendations on the current case.


api/f680/caseworker/views.py
```diff
@@ -30,6 +29,9 @@ def dispatch(self, request, *args, **kwargs):
 
         return super().dispatch(request, *args, **kwargs)
 
+    def get_queryset(self):
+        return Recommendation.objects.filter(case_id=self.kwargs["pk"])
```
currycoder, line 33: I thought `CurrentCaseRecommendationFilter` did this?

saruniitr: yes but in permission class it was using `view.queryset()` which goes through all recommendations.
Updating to use filter class.

markj0hnst0n (approved):

PR #2477, opened 2025-03-25
LTD-6054-allow-flags-muliple-teams 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6054

This allows multiple teams to be able to add and remove flags. 
We need the ability to be able to add flags which are owned by other teams for example on F680 (manpads) 
rather then create multiple versions for a flag that is essentially the same this allows 
1 Owner but allows multiple teams to apply the flag.  We could make this applicable by Case type but feel we want to get case type agnostic at the moment.


api/flags/views.py
```diff
@@ -68,7 +68,7 @@ def get_queryset(self):
             flags = flags.filter(priority=priority)
 
         if team and team != "None":
-            flags = flags.filter(team=team)
+            flags = Flag.objects.filter_by_applicable_team(team) & flags
```
currycoder, line 71: I'm not sure I follow this line, why do we need `& flags`?

Could `flags = flags.filter_by_applicable_team(team)` not work instead?

currycoder: Ah I think to do that we would need this to be a method on a custom queryset instead, that might be cleaner TBH.

api/flags/models.py
```diff
@@ -12,6 +13,10 @@ class FlagManager(models.Manager):
     def get_by_natural_key(self, name):
         return self.get(name=name)
 
+    def filter_by_applicable_team(self, team, **kwargs):
+        flags = self.get_queryset().filter(**kwargs)
+        return flags.filter(Q(team=team) | Q(applicable_by_team=team))
```
currycoder, line 18: As per the other comment, I think this would be better off on a custom queryset instead.  That way you could do...

```
queryset = Flag.objects.filter(...)
queryset = queryset.filter_by_applicable_team(team)
```
Instead of having to pass kwargs in to here at all and get an initial queryset.  Here's the normal pattern for that; https://docs.djangoproject.com/en/5.1/topics/db/managers/#calling-custom-queryset-methods-from-the-manager

currycoder (approved):

api/flags/tests/test_list_flags.py
```diff
@@ -41,6 +41,33 @@ def test_gov_user_can_see_only_filtered_case_level_and_team_flags(self):
...
+        owner_team = self.create_team("Team")
+        applicable_team = self.create_team("Team2")
+
+        flag_1 = self.create_flag(
+            "Flag1",
+            "Organisation",
+            owner_team,
+        )
+        self.create_flag("Flag2", "Organisation", self.team)
+        self.create_flag("Flag3", "Organisation", self.team)
```
saruniitr, line 59: Could these be replaced by factories?

api/cases/tests/test_managing_flags_on_case.py
```diff
@@ -77,6 +77,29 @@ def test_user_can_add_case_level_flags_from_their_own_team(self):
...
+        """
+        Given a Case with no Flags assigned
+        When a user attempts to add a case-level Flag which are applicable by their Team to the Case
+        Then the Flag is successfully added
+        """
+        owner_team = self.create_team("OwnerTeam")
+
+        applicable_flag_flag = self.create_flag("Flag1", "Case", owner_team)
+        applicable_flag_flag.applicable_by_team.set([self.team])
+        applicable_flag_flag.save()
```
saruniitr, line 90: Same here, factories can be used

saruniitr (approved):

api/cases/tests/test_managing_flags_on_case.py
depsiatwal: I did think of that however these are very old tests and style we probably need to shift the whole lot to pytest style

PR #2479, opened 2025-03-28
LTD-6067: Fix KeyError when clearing recommendations

saruniitr:
## Change description

The `delete()` in the endpoint doesn't return any content in the `Response`.
With Hawk authentication enabled it expects `Content-Type` header which is not included in this case and fails with `KeyError`. Switch this to use `HttpResponse` which includes the necessary headers.

Update all tests to run with and without Hawk authentication enabled to match the environment settings.

Recommend reviewing commit by commit as I have grouped the tests under classes which is showing large number of changes in the diff.

[LTD-6067]: https://uktrade.atlassian.net/browse/LTD-6067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/f680/caseworker/tests/test_views.py
```diff
@@ -23,6 +29,40 @@
...
+def get_api_client():
+    def _get_api_client(method, url, data=None):
+        client = Client()
+        if settings.HAWK_AUTHENTICATION_ENABLED:
+            url = parse.urljoin("http://testserver", url)
+            sender = get_hawk_sender(method, url, data, settings.HAWK_LITE_API_CREDENTIALS)
+            client.credentials(HTTP_HAWK_AUTHENTICATION=sender.request_header, CONTENT_TYPE="application/json")
+            return client, url
+        else:
+            return client, url
```
saruniitr, line 61: Instead of just running the delete test (the test that failed when hawk is enabled), I looked for a way to run all the tests with hawk enabled and disabled. It is possible with parameterized fixtures but `client` need to be updated with credentials when hawk is enabled so this is unavoidable.
The advantage is that we don't have to use `if-else` in each test and they are restricted to this function only.

api/f680/caseworker/tests/test_views.py
```diff
@@ -53,136 +93,100 @@ def _get_f680_application():
     return _get_f680_application
 
 
-@freeze_time("2025-01-01 12:00:01")
-def test_GET_recommendation_success(
-    api_client, get_f680_application, url, team_case_advisor, team_case_advisor_headers
-):
```
saruniitr, line 59: Test functionality is not changed, they are only grouped under classes hence the diff showing large number of changes.

api/f680/caseworker/tests/test_views.py
```diff
@@ -53,136 +93,100 @@ def _get_f680_application():
...
-            case=f680_application,
-            security_release_request=release_request,
-            type=RecommendationType.APPROVE,
-            security_grading="official",
-            conditions="No concerns",
-        )
-    for release_request in another_f680_application.security_release_requests.all():
-        F680RecommendationFactory(
-            case=another_f680_application, security_release_request=release_request, conditions="No concerns"
+@pytest.mark.usefixtures("hawk_authentication")
```
saruniitr, line 96: this is the main change

api/f680/caseworker/tests/test_views.py
```diff
@@ -23,6 +29,40 @@
...
+def hawk_authentication(request):
+    with override_settings(HAWK_AUTHENTICATION_ENABLED=request.param):
+        yield request.param
+
+
+@pytest.fixture
+def get_api_client():
+    def _get_api_client(method, url, data=None):
+        client = Client()
+        if settings.HAWK_AUTHENTICATION_ENABLED:
```
currycoder, line 55: Do we ever want to test these endpoints *without* hawk authentication?  Just thinking that there's a fair bit of conditionality here to allow these tests to work with or without hawk - but in production we *always* have hawk.  Could it be best to just force hawk to being on for endpoint tests all the time, given that we've discovered there are incompatibilities that can manifest in our application code?

api/f680/caseworker/tests/test_views.py
```diff
@@ -23,6 +29,40 @@
...
+
+        # Without this hawk sender gets data as string whereas
+        # receiver gets it as bytes resulting in failure because
+        # of hash mismatch. Our version on mohawk sender supports
+        # receiving in bytes so convert before sending it.
+        data = json.dumps(data).encode("utf-8")
+        return super().post(url, data=data, content_type="application/json", **kwargs)
+
+
+@pytest.fixture(params=[False, True])
```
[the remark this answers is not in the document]
saruniitr: This is just like parameterizing a test and we have examples in `advice` in FE unit tests.
Basically whereever this fixture is used, that test runs twice (based on the values specified as params which are two in this case). I hope that answers your question?

api/f680/caseworker/tests/test_views.py
saruniitr: The conditionality is only in one place.
If we always want to run with hawk enabled then this can easily be updated.

api/f680/caseworker/tests/test_views.py
```diff
@@ -23,6 +29,40 @@
 ]
 
 
+class Client(APIClient):
```
depsiatwal, line 32: is the intention to use this every where ?
how come it only lives in test_views.py ?

saruniitr: All other tests need updating to work with hawk enabled, I can probably move this to a central place.

api/f680/caseworker/tests/test_views.py
saruniitr: updated to run with Hawk enabled always.

api/f680/caseworker/tests/test_views.py
saruniitr: moved to common fixtures file

depsiatwal: why not conf.py ?

saruniitr: You mean `conftest.py`? It is in similar place. Instead of putting all fixtures in that file we can create separate files and refer that which gives better control.

depsiatwal: ok fair enough

depsiatwal (approved):
lgtm

currycoder (approved):

PR #2484, opened 2025-03-31
LTD:6017: Include F680 case data to display in queue view

saruniitr:
## Change description

Update serializer to include F680 details to be displayed in the queue view.

Relevant FE PR https://github.com/uktrade/lite-frontend/pull/2438

[LTD-6017]: https://uktrade.atlassian.net/browse/LTD-6017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/cases/serializers.py
```diff
@@ -121,6 +122,7 @@ class CaseListSerializer(serializers.Serializer):
     end_users = serializers.SerializerMethodField()
     sub_status = CaseSubStatusSerializer()
     flags = PrimaryKeyRelatedSerializerField(many=True, queryset=Flag.objects.all(), serializer=CaseListFlagSerializer)
+    f680_data = serializers.SerializerMethodField()
```
saruniitr, line 125: This serializer is common for all case type and the fields are very specific to SIEL hence all of the required F680 details are included in this field.
Many of the fields in the queue view itself are not applicable for F680 but since we are using the same queue for all types this is unavoidable.

api/f680/caseworker/tests/test_views.py
```diff
@@ -53,6 +53,56 @@ def _get_f680_application():
...
+    response = api_client.get(url, **headers)
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+    actual = response.json()["results"]["cases"][0]["f680_data"]
+
+    product = f680_application.get_product()
+    assert product.name == actual["product"]["name"]
+    assert product.description == actual["product"]["description"]
+
+    for index, release_request in enumerate(f680_application.security_release_requests.all()):
```
saruniitr, line 81: Would've been ideal to check all of the output but `id` and faker fields makes it difficult. We will have to create each of those instances with defined values.
To keep it simple all of the fields used in the queue view are verified here.

marcuspp (approved):
This could be handy when working out if a case is f680 or not.

api/cases/serializers.py
currycoder: I feel like ideally we might be better off trying to put case-type specific things under a new nested serializer, so that we can keep the variable fields contained to a single location - a bit like `case["data"]`.  What do you think?

Not saying that we should shift the existing SIEL-specific fields, but this might start us off on the right foot for new case types.

saruniitr: nested serializer instead of method field sounds good, I will update.

api/cases/serializers.py
```diff
@@ -182,6 +183,16 @@ def get_end_users(self, instance):
     def has_end_users(self, instance):
         return hasattr(instance, "baseapplication") and hasattr(instance.baseapplication, "end_user_parties")
 
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        representation["f680_data"] = {}
+        if instance.case_type.sub_type != CaseTypeSubTypeEnum.F680:
```
depsiatwal, line 189: should this line be below 
representation["f680_data"] = {}

I would say return earlier

api/f680/caseworker/tests/test_views.py
```diff
@@ -155,6 +155,51 @@ def test_GET_recommendation_raises_notfound_error(
         response = api_client.get(target_url, **headers)
         assert response.status_code == 404
 
+    def test_f680_cases_list(self, get_hawk_client, get_f680_application, team_case_advisor):
+        gov_user = team_case_advisor(TeamIdEnum.MOD_CAPPROT)
+        f680_application = get_f680_application()
+        for release_request in f680_application.security_release_requests.all():
```
depsiatwal, line 161: This appears in a few places I noticed and thoughts on 
placing this in get_f680_application()

saruniitr: this is done wherever we want to create recommendations with factory, we cannot move it to `get_f680_application()` because we don't want an application with recommendations prepopulated.

depsiatwal (approved):
lgtm

api/cases/serializers.py
```diff
@@ -182,6 +183,16 @@ def get_end_users(self, instance):
     def has_end_users(self, instance):
         return hasattr(instance, "baseapplication") and hasattr(instance.baseapplication, "end_user_parties")
 
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        representation["f680_data"] = {}
```
currycoder, line 188: Would it trip things up if we omitted `f680_data` completely if the case is not an F680?

Just thinking that otherwise we may end up with empty `siel_data`, `oiel_data`, ... in this JSON contract if we keep going down this path.

saruniitr: I think we can skip completely as we have separate case row templates for each type.
I will remove if there are no issues otherwise will keep it.

I am also going to merge once I verify this.

PR #2497, opened 2025-04-04
[LTD-6080] F680 ECJU Query emails

marcuspp:
### Aim

Send alternative emails for F680s ECJU Queries

[LTD-6080](https://uktrade.atlassian.net/browse/LTD-6080)

[LTD-6080]: https://uktrade.atlassian.net/browse/LTD-6080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/application_manifest.py
+++ b/api/applications/application_manifest.py
@@ -3,6 +3,7 @@ from api.applications.serializers.standard_application import StandardApplicatio
 from api.application_manifests.base import BaseManifest
 from api.application_manifests.registry import application_manifest_registry
 from api.cases.enums import ApplicationFeatures, CaseTypeSubTypeEnum
+from gov_notify.enums import TemplateType
 
 
 @application_manifest_registry.register(CaseTypeSubTypeEnum.STANDARD)
@@ -12,4 +13,12 @@ class StandardApplicationManifest(BaseManifest):
     features = {
         ApplicationFeatures.LICENCE_ISSUE: True,
         ApplicationFeatures.ROUTE_TO_COUNTERSIGNING_QUEUES: True,
+    }
+    ecju_max_days = 20
+    notification_config = {
+        "ecju_query": {"template": TemplateType.EXPORTER_ECJU_QUERY, "frontend_url": "/"},
+        "ecju_query_chaser": {
+            "template": TemplateType.EXPORTER_ECJU_QUERY_CHASER,
+            "frontend_url": "/applications/{case_id}/ecju-queries/",
+        },
     }
--- a/api/cases/celery_tasks.py
+++ b/api/cases/celery_tasks.py
@@ -160,7 +160,8 @@ def update_cases_sla():
 @shared_task
 def schedule_all_ecju_query_chaser_emails():
     """
-    Sends an ECJU 15 working days reminder
+    Sends an ECJU  reminder 5 days before Max days allowed
+    Max days for each case type set in the application manifest
     Runs as a background task daily at a given time.
     Can accommodate reruns and can send reminders if any have been missed upto 20 days after ECJU Query created
     """
@@ -174,7 +175,7 @@ def schedule_all_ecju_query_chaser_emails():
         )
 
         for ecju_query in ecju_queries:
-            if ecju_query.open_working_days >= 15 and ecju_query.open_working_days <= 20:
+            if ecju_query.should_send_chaser_email():
                 ecju_query_reminders.append(ecju_query.id)
 
         for ecju_query_id in ecju_query_reminders:
--- a/api/cases/models.py
+++ b/api/cases/models.py
@@ -849,6 +849,11 @@ class EcjuQuery(TimestampableModel):
         end_date = self.responded_at if self.responded_at else timezone.now()
         return working_days_in_range(start_date=start_date, end_date=end_date)
 
+    def should_send_chaser_email(self):
+        application_manifest = self.case.get_application_manifest()
+        max_days = application_manifest.ecju_max_days
+        return max_days - 5 <= self.open_working_days <= max_days
+
     notifications = GenericRelation(ExporterNotification, related_query_name="ecju_query")
 
     def save(self, *args, **kwargs):
--- a/api/cases/notify.py
+++ b/api/cases/notify.py
@@ -1,5 +1,3 @@
-from django.db.models import F
-
 from api.core.helpers import get_exporter_frontend_url
 from api.cases.models import Case, EcjuQuery
 from gov_notify.enums import TemplateType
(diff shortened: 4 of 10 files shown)


api/cases/notify.py
```diff
@@ -15,6 +15,19 @@
...
+F680_ECJU_QUERY_MAX_DAYS = 30
+STANDARD_APPLICATION_ECJU_QUERY_MAX_DAYS = 20
+
+
+def should_send_ecju_chaser_email(ecju_query):
+    if ecju_query.is_f680_query:
+        max_days = F680_ECJU_QUERY_MAX_DAYS
+    else:
+        max_days = STANDARD_APPLICATION_ECJU_QUERY_MAX_DAYS
+    return max_days - 5 <= ecju_query.open_working_days <= max_days
```
currycoder, line 30: This feels like a job for the application manifest; 
https://github.com/uktrade/lite-api/blob/10b4f4d0a6189b6f0359b9261beae19d7ab13424/api/applications/application_manifest.py#L8-L15

https://github.com/uktrade/lite-api/blob/10b4f4d0a6189b6f0359b9261beae19d7ab13424/api/f680/application_manifest.py#L8-L15

https://github.com/uktrade/lite-api/blob/10b4f4d0a6189b6f0359b9261beae19d7ab13424/api/applications/helpers.py#L22-L24

It's a little pattern we've started building on to try to avoid `if application_is_f680` type switches.  Instead, in places like this we get the correct manifest for the case type and refer to a common variable that the manifest is guaranteed to have for every case type.

I could see it being a good idea to perhaps add a new attribute to the manifests like...

`ECJU_QUERY_MAX_DAYS`

What do you think?

api/cases/notify.py
```diff
@@ -122,34 +135,53 @@ def notify_exporter_ecju_query(case_pk):
...
         },
+        case_info["case_type"],
     )
 
 
 def notify_exporter_ecju_query_chaser(ecju_query_id, callback):
     ecju_query = EcjuQuery.objects.get(id=ecju_query_id)
-
-    exporter_frontend_ecju_queries_url = get_exporter_frontend_url(f"/applications/{ecju_query.case_id}/ecju-queries/")
+    if ecju_query.is_f680_query:
```
currycoder, line 144: Same comment as above

api/cases/notify.py
```diff
@@ -122,34 +135,53 @@ def notify_exporter_ecju_query(case_pk):
...
 
 
-def _notify_exporter_ecju_query(email, data):
+def _notify_exporter_ecju_query(email, data, case_type_id):
     payload = ExporterECJUQuery(**data)
-    send_email(email, TemplateType.EXPORTER_ECJU_QUERY, payload)
+    if str(case_type_id) == str(CaseTypeEnum.F680.id):
+        email_template = TemplateType.EXPORTER_F680_ECJU_QUERY
+    else:
+        email_template = TemplateType.EXPORTER_ECJU_QUERY
```
currycoder, line 174: Probably another job for the manifest pattern

api/cases/notify.py
```diff
@@ -122,34 +135,53 @@ def notify_exporter_ecju_query(case_pk):
...
 
 
-def _notify_exporter_ecju_query_chaser(email, data, callback):
+def _notify_exporter_ecju_query_chaser(email, data, case_type_id, callback):
     payload = ExporterECJUQueryChaser(**data)
-    send_email(email, TemplateType.EXPORTER_ECJU_QUERY_CHASER, payload, callback)
+    if str(case_type_id) == str(CaseTypeEnum.F680.id):
+        email_template = TemplateType.EXPORTER_F680_ECJU_QUERY_CHASER
+    else:
+        email_template = TemplateType.EXPORTER_ECJU_QUERY_CHASER
```
currycoder, line 183: Probably another job for the manifest pattern

currycoder (commented):
Nice work!  I have a suggestion for getting rid of case type switches inline, hopefully that will enhance this a little bit

api/cases/models.py
```diff
@@ -849,6 +850,10 @@ def open_working_days(self):
         end_date = self.responded_at if self.responded_at else timezone.now()
         return working_days_in_range(start_date=start_date, end_date=end_date)
 
+    @property
+    def is_f680_query(self):
+        return self.case.case_type.id == CaseTypeEnum.F680.id
```
marcuspp, line 855: This is no longer required, i'll remove

api/cases/tests/test_case_ecju_queries.py
```diff
@@ -679,50 +684,94 @@ def setUp(self):
...
         )
+        self.assertEqual(ecju_f680_query_send_1.open_working_days, 15)
 
-        ecju_quey_send_1 = EcjuQueryFactory(
-            question="ECJU Query 15 days",
+        ecju_f680_query_send_2 = EcjuQueryFactory(
+            question="F680 ECJU Query reminder after 25 days (found)",
+            case=self.f680_case,
+            raised_by_user=self.gov_user,
+            created_at=datetime.datetime(2024, 1, 1, 12, 00, tzinfo=timezone.utc),
```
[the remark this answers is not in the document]
marcuspp: This is to avoid the warnings when the test is run if a naive date time is used. The other tests in this file all use UTC as their tz? So i followed suit. Should they be BST?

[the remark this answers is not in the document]
saruniitr: I think it should match the timezone that notifications are triggered with. `localtime()` is matching the local time zone.

api/applications/application_manifest.py
```diff
@@ -13,3 +14,9 @@ class StandardApplicationManifest(BaseManifest):
         ApplicationFeatures.LICENCE_ISSUE: True,
         ApplicationFeatures.ROUTE_TO_COUNTERSIGNING_QUEUES: True,
     }
+    ecju_max_days = 20
+    frontend_urls = {"ecju_queries": "/applications/{case_id}/ecju-queries/"}
+    email_templates = {
+        "ecju_query": TemplateType.EXPORTER_ECJU_QUERY,
+        "ecju_query_chaser": TemplateType.EXPORTER_ECJU_QUERY_CHASER,
+    }
```
currycoder, line 22: I wonder if these should ideally be collected under a single datastructure - e.g. 
```
notification_overrides = {
    "ecju_query": {
        "template": TemplateType.EXPORTER_ECJU_QUERY, 
        "frontend_url": "/applications/{case_id}/ecju-queries/",
     }
}
```

It kind of felt like `frontend_urls` was fairly specific to notifications (with the string formatting) instead of being something general purpose.

What do you think?

api/cases/notify.py
```diff
@@ -17,6 +15,12 @@
 from gov_notify.service import send_email
 
 
+def should_send_ecju_chaser_email(ecju_query):
+    application_manifest = ecju_query.case.get_application_manifest()
+    max_days = application_manifest.ecju_max_days
+    return max_days - 5 <= ecju_query.open_working_days <= max_days
```
currycoder, line 21: Could this be a method on the `ECJUQuery` ORM class?  It feels like over time we should be trying to put what logic we can on ORM classes

api/cases/celery_tasks.py
```diff
@@ -160,7 +160,7 @@ def update_cases_sla():
 @shared_task
 def schedule_all_ecju_query_chaser_emails():
     """
-    Sends an ECJU 15 working days reminder
+    Sends an ECJU  reminder 5 days before Max days allowed
```
currycoder, line 163: Could we specify that "max days" can be found on the manifest?

currycoder (commented):
This feels very close - just a few more suggestions for structural enhancements.  Happy to be persuaded that these are minor though

saruniitr (approved):
looks good to me

currycoder (approved):

PR #2501, opened 2025-04-10
Added support for cloning f680 application

marcuspp:
### Aim


api/f680/models.py
```diff
@@ -13,13 +14,63 @@
...
+    }
+
+    def clone(self, exclusions=None, **overrides):
+        cloned_application = super().clone(exclusions=exclusions, **overrides)
+        application_data = deepcopy(cloned_application.application)
+        try:
+            items = application_data["sections"]["user_information"]["items"]
+        except KeyError:
+            return cloned_application
+        else:
```
currycoder, line 65: Do we need this extra indentation here?  My understanding is, if `items` was not found due to `KeyError` we will return early.

api/f680/tests/test_models.py
```diff
@@ -71,3 +75,68 @@ def test_on_submit_name_missing_in_application_json(self):
         assert f680_application.name is None
         with pytest.raises(KeyError):
             f680_application.on_submit({})
+
+    def test_clone_application(self, data_application_json):
```
currycoder, line 79: The pattern that we use for other `test_clone` methods uses `model_to_dict` to assert the whole record's data structure is as expected; https://github.com/uktrade/lite-api/blob/d5ed63d25c85ababa0e18ffebbd47f325780595f/api/goods/tests/test_models.py#L103

The reason we did this was to make sure that any schema migrations are fully considered for the clone methods; e.g. the developer will get a failing test and have to decide whether the new field should be cloned or should be excluded.

Could I suggest the same pattern for this test?

api/cases/models.py
```diff
@@ -514,6 +519,52 @@ def get_application(self):
 
         return model_class.objects.get_prepared_object(pk=self.pk)
 
+    @transaction.atomic
+    def create_amendment(self, user):
```
currycoder, line 523: Very nice!

currycoder (approved):

markj0hnst0n (approved):

PR #2506, opened 2025-04-14
LTD-6126 added postcode serialisation to upper and strip

Tllew:
may need to move it to frontend if we don't want to automate

[LTD-6126](https://uktrade.atlassian.net/browse/LTD-6126)

[LTD-6126]: https://uktrade.atlassian.net/browse/LTD-6126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/addresses/tests/test_postcode_validation.py
+++ b/api/addresses/tests/test_postcode_validation.py
@@ -8,6 +8,7 @@ from test_helpers.clients import DataTestClient
 from api.organisations.enums import OrganisationType
 from lite_content.lite_api.strings import Addresses
 from api.addresses.models import Address
+from parameterized import parameterized
 
 
 class AddressSerializerPostcodeValidationTest(DataTestClient):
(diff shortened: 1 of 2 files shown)


api/addresses/tests/test_postcode_validation.py
```diff
@@ -76,14 +77,28 @@ def test_valid_postcodes(self, mock_gov_notification):
...
             data["site"]["address"]["postcode"] = postcode
             data["registration_number"] = "".join([str(randint(0, 9)) for _ in range(8)])
             response = self.client.post(
                 self.url, data, **{EXPORTER_USER_TOKEN_HEADER: user_to_token(self.exporter_user.baseuser_ptr)}
             )
 
             self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+            addr = Address.objects.get(postcode=expected_results[i])
+
+            self.assertEqual(addr.postcode, expected_results[i])
```
kevincarrogan, line 104: Could we start using parametrized for this instead now?

api/addresses/tests/test_postcode_validation.py
```diff
@@ -45,8 +47,20 @@ def test_invalid_postcodes(self, mock_gov_notification):
             self.assertEqual(response.json()["errors"]["site"]["address"]["postcode"][0], Addresses.POSTCODE)
             self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
+    @parameterized.expand(
+        [
+            ("BT32 4PX", "BT32 4PX"),
+            ("GIR 0AA", "GIR 0AA"),
+            ("BT324PX", "BT324PX"),
+            (" so11aa ", "SO11AA"),
+            (" so1  1aa ", "SO1  1AA"),
```
hnryjmes, line 56: is there a reason why we want to preserve the middle whitespace instead of just normalizing to be 0 or 1 character width?

hnryjmes: I get that it's not what you're testing but just something looked odd about this as a test case

Tllew: I was thinking this but I think this is an initial improvement and we can address it if we need to in the future, I think this example in particular only works because the postcode is technically shorter than normal

hnryjmes (approved):

kevincarrogan (approved):

PR #2508, opened 2025-04-14
LTD-6140: Add fields to capture Security release outcome validity dates

saruniitr:
## Change description

For each Security release outcome user need to be able to determine the expiry date so add fields to capture validity start and end dates.

Instead of a single expiry date it is captured as start and end dates because the letter includes the start date and duration in months. It is also possible to start date may not always be the date the outcome is issued. To allow for all these they are saved as two different fields.

FE changes are here https://github.com/uktrade/lite-frontend/pull/2465.

[LTD-6140](https://uktrade.atlassian.net/browse/LTD-6140)

[LTD-6140]: https://uktrade.atlassian.net/browse/LTD-6140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/f680/models.py
```diff
@@ -147,3 +147,5 @@ class SecurityReleaseOutcome(TimestampableModel):
         choices=enums.SecurityGrading.security_release_outcome_choices, max_length=50, blank=True, null=True
     )
     approval_types = ArrayField(models.CharField(choices=enums.ApprovalTypes.choices, max_length=50), default=list)
+    validity_start_date = models.DateField(blank=True, null=True, help_text="Date the outcome validity starts")
+    validity_end_date = models.DateField(blank=True, null=True, help_text="Date the outcome is no longer valid")
```
currycoder, line 151: I think we should force these fields to have values by making them not nullable.  Would suggest a data migration after this to set expiry start_date/end_date based on `created_at` and then a further schema migration to make this not-nullable.  That sequence should ensure that our existing test data works on local/dev2.

What do you think?

api/f680/enums.py
```diff
@@ -105,3 +105,7 @@ class SecurityReleaseOutcomes:
         (APPROVE, "Approve"),
         (REFUSE, "Refuse"),
     ]
+
+
+class SecurityReleaseOutcomeDuration:
+    DEFAULT_DURATION_MONTHS = 24
```
currycoder, line 111: If this is only used for tests, could it be possible to move it directly in to factories or similar?  Having it in the application code made me think we would use it in the api app somewhere.

api/f680/models.py
currycoder: I think it's just a data integrity thing.  We have gotten in to the habit on SIELs of making things nullable when they shouldn't be so it's something I'd like for us to start tightening up.

Just suggested the data migration to avoid any headaches when we come to deploy this to dev2 - but I'm happy if you would prefer to resolve that manually.

saruniitr: agree, I was only thinking about time.
I will create tickets for those additional migrations and pick up later.

currycoder: Sounds good Arun

currycoder (approved):

api/f680/models.py
currycoder: I think we will have exactly the same issue for adding `Recommendation.security_grading` back in - except that one is perhaps a little more troublesome because we can't easily guess at what the value should be from other data.

api/f680/enums.py
```diff
@@ -105,3 +105,7 @@ class SecurityReleaseOutcomes:
         (APPROVE, "Approve"),
         (REFUSE, "Refuse"),
     ]
+
+
+class SecurityReleaseOutcomeDuration:
```
marcuspp, line 110: Just an observation, this file is called `enums.py` yet none of these are python enums

marcuspp (approved):

PR #2510, opened 2025-04-15
[LTD-6144] Add case type to f680 Serializers

marcuspp:
### Aim

[LTD-6144](https://uktrade.atlassian.net/browse/LTD-6144)

### DIFF
--- a/api/f680/caseworker/serializers.py
+++ b/api/f680/caseworker/serializers.py
@@ -74,6 +74,12 @@ class F680ApplicationSerializer(serializers.ModelSerializer):
     submitted_by = RelatedExporterUserSerializer(read_only=True)
     security_release_requests = SecurityReleaseRequestSerializer(many=True)
     product = ProductSerializer(source="get_product")
+    case_type = serializers.SerializerMethodField()
+
+    def get_case_type(self, instance):
+        from api.cases.serializers import CaseTypeSerializer
+
+        return CaseTypeSerializer(instance.case_type).data
 
     class Meta:
         model = F680Application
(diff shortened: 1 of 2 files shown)


api/f680/caseworker/serializers.py
```diff
@@ -74,6 +74,12 @@ class F680ApplicationSerializer(serializers.ModelSerializer):
     submitted_by = RelatedExporterUserSerializer(read_only=True)
     security_release_requests = SecurityReleaseRequestSerializer(many=True)
     product = ProductSerializer(source="get_product")
+    case_type = serializers.SerializerMethodField()
+
+    def get_case_type(self, instance):
+        from api.cases.serializers import CaseTypeSerializer
```
depsiatwal, line 80: any reason for the import in mid method ?

api/f680/exporter/serializers.py
```diff
@@ -35,6 +37,11 @@ def create(self, validated_data):
         validated_data["case_type_id"] = self.context["case_type_id"]
         return super().create(validated_data)
 
+    def get_case_type(self, instance):
+        from api.cases.serializers import CaseTypeSerializer
```
depsiatwal, line 41: same comment as above ?

api/f680/caseworker/serializers.py
marcuspp: This one has a circular dependancy if moved to the top of the file.

depsiatwal: is it worth having a new serializer which is cleaner ?
generally what we attempt to do when shifting to new endpoints

marcuspp: I think maybe a future refactor as its important it returns the data the same way for all case types. It seems moving the `CaseTypeSerializer` out of `cases/serializers.py` would be the best solution as this pattern of importing this serializer mid method is common across the repo. ie. 
https://github.com/uktrade/lite-api/blob/f654cd9a63da4da13dbb8f137857f6dd6f52f701/api/compliance/serializers/ComplianceSiteCaseSerializers.py#L115 
I'd prefer to leave this for now if thats okaym and put a ticket on the backlog to move the serializer at a later date?

markj0hnst0n (approved):

api/f680/exporter/serializers.py
marcuspp: This one can be moved so have done so.

api/f680/caseworker/serializers.py
depsiatwal: Ok fair enough it's these kind of bad practices were trying to move away from in the new endpoints.

depsiatwal (approved):
Commented added happy to approve on the basis a follow up ticket created.

PR #2512, opened 2025-04-17
LTD-6115: Fix formatting for F680 outcome letters to match templates

saruniitr:
### Aim


api/f680/caseworker/serializers.py
```diff
@@ -178,6 +178,8 @@ class Meta:
 
 class SecurityReleaseOutcomeLetterSerializer(serializers.ModelSerializer):
     security_release_requests = SecurityReleaseRequestSerializer(many=True)
+    approval_types = serializers.MultipleChoiceField(choices=enums.ApprovalTypes.choices)
```
depsiatwal, line 181: lgtm I think these changes are breaking tests though

saruniitr: fixed failing tests

depsiatwal (approved):
LGTM

api/letter_templates/context_generator.py
```diff
@@ -612,6 +616,14 @@ def get_security_release_outcomes(self, obj):
             data[outcome] = SecurityReleaseOutcomeLetterSerializer(sros, many=True).data
         return data
 
+    def get_case_officer(self, obj):
+        case_officer = None
+        qs = obj.case_assignments.filter(queue_id=QueuesEnum.MOD_ECJU_F680_CASES_UNDER_FINAL_REVIEW)
```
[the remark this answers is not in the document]
currycoder: This feels a little bit brittle - e.g. if we choose to change the queue that cases route to in future then this will stop working.  If we just want to put in the details for the user generating the letter, can't we pass that through from the request to the template context?

api/letter_templates/templates/letter_templates/f680_approval.html
```diff
@@ -20,21 +20,22 @@
...
 
+  <h3 class="govuk-heading-l">Commercial exploitation agreements</h3>
   <p>If the item is designed by or for the MOD, you must have a commercial exploitation agreement before it can be manufactured or sold commercially. This is the case regardless of whether MOD owns the commercial design right.</p>
   <p>Get help with your commercial exploitation agreement by contacting:  </p>
   <ul>
     <li>IPRG, Poplar 2, MOD, Abbey Wood, Bristol BS34 8JH if the MOD owns intellectual property rights relating to the equipment</li>
     <li>the MOD Contracts Branch who issued the original contract if the company owns the design rights (but MOD can claim a levy) </li>
   </ul>
-  <h2>Export licences </h2>
+  <h3>Export licences </h3>
```
currycoder, line 38: This letter still ends with "signature here" - do we want that?

currycoder: As does the end of the refusal letter.  The lead in to the refusal criteria also feels a bit unfinished
```
Reasons for refusal
This application is refused under

{{refusal_reasons}}
```

api/letter_templates/context_generator.py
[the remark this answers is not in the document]
saruniitr: We could use `request.user` but this keep things within the serializer.
There is also a test which checks for this, if the queue changes then it fails.

api/letter_templates/templates/letter_templates/f680_approval.html
saruniitr: "signature here" need to be removed but left it as placeholder for now as we need to change the digital signature, I can remove in this PR.

saruniitr: `refusal_reasons` contains the combined text of all refusal criteria.
The template shows the criteria codes but we don't actually capture that separately which is why it is like that.
However we are going to include that as part of the text. eg it will look like this,

```
## Reasons for refusal

This application is refused under,

Criteria 1: text of this criteria

Criteria 5: text of this criteria
```

currycoder: OK so the refusal letter will get some attention in a different ticket?

saruniitr: We can probably extract criteria codes from this and include in that line but that will have to be a follow-up when the linked PR is merged.

saruniitr: correct, once the linked PR is merged then we can update this.

currycoder (approved):

PR #2519, opened 2025-04-24
LTD-6163 notes/timelines entry for f680 recommendation

markj0hnst0n:
### Aim

Create a new audit table log entry with information for debugging any future issues when adding and removing (clearing) a recommendation for an F680 application and ensure that it creates a message for caseworkers in Notes and Timelines.

[LTD-6165](https://uktrade.atlassian.net/browse/LTD-6165)

[LTD-6165]: https://uktrade.atlassian.net/browse/LTD-6165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/f680/caseworker/views.py
```diff
@@ -64,8 +89,17 @@ def list(self, request, *args, **kwargs):
     def destroy(self, request, *args, **kwargs):
         user = request.user
         qs = self.get_queryset().filter(user_id=user.id, team=user.govuser.team)
+
+        # A record for the audit table of the recommendation_object_ids which have been removed
+        recommendation_ids = [str(recommendation.id) for recommendation in qs]
```
currycoder, line 94: Could this just be for the security release request ids associated with these recommendation objects (same end payload as above)?  Thinking about it - the recommendation ID isn't useful if we've deleted the recommendation object!

api/f680/caseworker/views.py
```diff
@@ -47,12 +48,36 @@ def prepare_data(self, request_data):
...
+
+        if advice_type:
+            payload.update({"additional_text": f"With advice type of {advice_type}"})
+
+        audit_trail_service.create(
+            actor=self.request.user,
+            verb=audit_type,
+            target=self.application.get_case(),
+            payload=payload,
+        )
```
currycoder, line 62: I think it would also be good to prove that these audit entries exist as expected in some of the existing test cases that we have for these endpoints

api/f680/caseworker/views.py
```diff
@@ -47,12 +48,36 @@ def prepare_data(self, request_data):
...
         data = self.prepare_data(request.data.copy())
         serializer = self.get_serializer(data=data, many=True)
         serializer.is_valid(raise_exception=True)
 
         self.perform_create(serializer)
+
+        # A record of the security release objects impacted by the recommendation for the audit table
+        security_release_request_object_ids = [str(item["security_release_request"]) for item in serializer.data]
+        # Advice type records will all be the same type as serializer does not allowed mixed advice
+        advice_type = serializer.data[0]["type"]["value"]
```
saruniitr, line 74: We are using recommendation instead of advice in F680.

api/f680/caseworker/views.py
```diff
@@ -47,12 +48,36 @@ def prepare_data(self, request_data):
             for item in request_data
         ]
 
+    def create_audit(self, audit_type, advice_type=None, payload=None):
```
saruniitr, line 51: this is probably ok in this case as we only use it in create/destroy.

api/f680/caseworker/views.py
saruniitr: yeah, tests need updating to ensure relevant audit events are created with expected data.

api/f680/caseworker/views.py
markj0hnst0n: I've renamed this

api/f680/caseworker/views.py
markj0hnst0n: Good point!  I've made this change now.

currycoder (approved):

PR #2523, opened 2025-04-25
Optimise fetching most recent two activities on queue view

kevincarrogan:
The main issue here was that the original query was getting all of the activity updates, serializing them all and then chucking nearly all of the results away.

In testing there was a page that was serializing ~2000 activity updates and then pruning this down to ~50.

This does the pruning much earlier and thus only needs to serializer activities that it definitely needs.

[LTD-6176](https://uktrade.atlassian.net/browse/LTD-6176)

[LTD-6176]: https://uktrade.atlassian.net/browse/LTD-6176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/audit_trail/managers.py
+++ b/api/audit_trail/managers.py
@@ -48,19 +48,26 @@ class AuditManager(GFKManager):
     def get_latest_activities(self, case_ids: List, number_of_results):
         obj_type = ContentType.objects.get_for_model(Case)
 
-        latest_activities = (
-            self.get_queryset()
-            .filter(Q(action_object_content_type=obj_type) | Q(target_content_type=obj_type))
-            .annotate(
-                activity_case_id=DBCase(
-                    When(target_content_type=obj_type, then=F("target_object_id")),
-                    When(action_object_content_type=obj_type, then=F("action_object_object_id")),
-                )
-            )
-            .annotate(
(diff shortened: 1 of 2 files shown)


Tllew (approved):
Change looks good, there's some functionality changing here though with it going from "most recent 2 records only" to any activity_obj, are we able to/do we want to test this?

api/audit_trail/managers.py
```diff
@@ -40,20 +47,27 @@ def create(self, *args, **kwargs):
...
+                When(target_content_type=obj_type, then=F("target_object_id")),
+                When(action_object_content_type=obj_type, then=F("action_object_object_id")),
+            )
         )
+        case_related_activities = case_related_activities.filter(activity_case_id__in=case_ids)
+
+        # We can now find the latest `n` activities using a window function.
+        top_n_activities = case_related_activities.annotate(
+            case_row=Window(expression=RowNumber(), partition_by=["activity_case_id"], order_by="-created_at")
+        ).filter(case_row__lte=number_of_results)
```
currycoder, line 71: So much better

PR #2527, opened 2025-04-28
LTD-6116-add-mod-signature

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6116

Shifting the ENV Variable for digital signature to application manifest so we can support different application types. 
Will also create a tidy up Jira once we confirm this is the approach to remove the ENV Variables from all the environments.


api/cases/generated_documents/signing.py
```diff
@@ -38,12 +37,19 @@ def _load_certificate_and_key():
     )
 
 
-def _get_signature_text(date):
+def _get_back_ground_image(image_name):
```
currycoder, line 40: Is this really a background image?

api/cases/generated_documents/signing.py
```diff
@@ -86,25 +92,27 @@ def _get_signature_image(text):
     return image
 
 
-def sign_pdf(original_pdf: bytes):
+def sign_pdf(original_pdf: bytes, document_signing: dict):
```
currycoder, line 95: Can't we just make these proper arguments instead of a dictionary?  Would be a bit more explicit/readable

currycoder (commented):
Just a few suggestions, I think this looks fine though.

api/cases/generated_documents/signing.py
depsiatwal: updated

saruniitr (approved):

currycoder (approved):
Just tried this locally on both SIELs and F680s - hits the nail on the head!

PR #2529, opened 2025-04-29
Handle Export Licence reference codes

kevincarrogan:
### Aim

Allow reference codes to be generated for Export Licence applications.

This removes some of the old logic for case types that never went live but also does a bit of refactoring to allow us to more easily bring back new case types.

[LTD-6180](https://uktrade.atlassian.net/browse/LTD-6180)

[LTD-6180]: https://uktrade.atlassian.net/browse/LTD-6180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/applications/models.py
```diff
@@ -403,6 +403,17 @@ def get_required_decision_document_types(self):
 
         return required_decisions
 
+    def get_reference_code_prefix(self):
+        reference_code_prefix = super().get_reference_code_prefix()
+
+        return f"GB{reference_code_prefix}"
```
Tllew, line 409: I wonder if the GB should be a variable?

then again maybe it'll never change since it's a GB licence..

api/applications/models.py
```diff
@@ -403,6 +403,17 @@ def get_required_decision_document_types(self):
...
+
+    def get_reference_code_suffix(self):
+        return {
+            ApplicationExportType.PERMANENT: "P",
```
Tllew, line 413: should this be added to a constant file?

Tllew (approved):
Looks good, added a few comments but nothing that warranty holding the ticket up

api/applications/models.py
kevincarrogan: This is the only place this is used at the moment so I've just left it here.

If/when we use it in other places then I think we could move it to a constants file.

api/cases/models.py
```diff
@@ -150,12 +150,37 @@ def save(self, *args, **kwargs):
...
 
         super(Case, self).save(*args, **kwargs)
 
+    def get_reference_code_prefix(self):
+        case_type = CaseTypeEnum.reference_to_class(self.case_type.reference)
+        reference_code_prefix = getattr(case_type, "reference_code_identifier", case_type.reference)
+        return reference_code_prefix
+
+    def get_reference_code_suffix(self):
+        return ""
```
hnryjmes, line 165: what is the purpose of this

kevincarrogan: Have a look at https://github.com/uktrade/lite-api/pull/2529/files#diff-16db8c7666ec8d18224261f0f2fcd942c4fcdcacf9dfc8a1f4f48641afdde0ebR411-R415

hnryjmes: ah yeah I get it now

api/cases/tests/test_reference_code.py
```diff
@@ -1,67 +1,50 @@
...
+@pytest.mark.parametrize(
+    "application_factory, factory_kwargs, expected_reference_code",
+    (
+        (DraftStandardApplicationFactory, {}, None),
+        (StandardApplicationFactory, {"export_type": ApplicationExportType.PERMANENT}, "GBSIEL/2023/0000001/P"),
+        (StandardApplicationFactory, {"export_type": ApplicationExportType.TEMPORARY}, "GBSIEL/2023/0000001/T"),
+        (F680ApplicationFactory, {}, None),
+        (SubmittedF680ApplicationFactory, {}, "F680/2023/0000001"),
+        (DraftExportLicenceApplicationFactory, {}, None),
+        (ExportLicenceApplicationFactory, {"export_type": ApplicationExportType.PERMANENT}, "GBEL/2023/0000001/P"),
```
hnryjmes, line 36: I wonder if users will expect (incorrectly) that a case they have referenced as `GBEL/2026/0000042/P` will convert into a `GBSIEL/2026/0000042/P` when in fact the numberings are unrelated and the new reference will be whatever is the next number in the sequence

hnryjmes (approved):

saruniitr:
These Exporter licences, are these different application types that we are going to support besides OIELs?

kevincarrogan:
> These Exporter licences, are these different application types that we are going to support besides OIELs?

Not quite.

An Export Licence encompasses both OIELs and SIELs so it's a hierarchy.

What we're trying to do here though is not decide yet whether it's an OIEL or a SIEL but leave that later in the process so I'm using this more generic model up until the point it switches to an actual SIEL or OIEL.

api/cases/tests/test_reference_code.py
kevincarrogan: That's quite a conclusion you've jumped to.

It's likely that it will in fact just convert as you've suggested it will. Or at the very least the licence number will reflect the actual licence type.

There's no intention though to use the next number in the sequence for any of these changes so the reference number will largely be consistent.

PR #2530, opened 2025-04-29
LTD-6165-note-time-line-f680-outcome 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6165

Adds a NT Audit for adding/clearing outcomes


marcuspp (approved):

api/f680/caseworker/tests/test_views.py
```diff
@@ -618,6 +618,13 @@ def test_POST_create_single_item_group_success(
             "validity_period": 24,
         }
 
+        audit_record = Audit.objects.get(target_object_id=str(f680_application.id), verb=AuditType.CREATE_F680_OUTCOME)
```
saruniitr, line 621: do you think it will be useful to record `security_grading`?

depsiatwal: I thought about this but then thought where do we stop ?
I think this should really be user as action audit log really. 
Anything else we should we should use the Table history audit

saruniitr: Actually the validity period depends on the security grading which is why I think it helps to record this.

saruniitr (approved):

api/f680/caseworker/tests/test_views.py
depsiatwal: fair enough added

currycoder (approved):

PR #2534, opened 2025-04-30
LTD-6145 - F680 finalise (refactor)

marcuspp:
### Aim


api/cases/models.py
```diff
@@ -444,59 +434,50 @@ def finalise(self, user, decisions, note):
...
                 target=self,
                 payload={
-                    "case_reference": self.reference_code,
-                    "decision": advice_type,
-                    "licence_reference": licence_reference,
+                    "status": {"new": self.status.status, "old": old_status},
+                    "additional_text": note,
                 },
             )
+            logging.info("Case is now finalised")
```
kevincarrogan, line 467: Can we make this a name logger i.e. `logger = logging.getLogger(__name__)`?

Could we also add some information here as to what case was finalised, maybe the case id and/or reference?

api/cases/models.py
```diff
@@ -444,59 +434,50 @@ def finalise(self, user, decisions, note):
...
+        if application_manifest.has_feature(ApplicationFeatures.LICENCE_ISSUE):
+            licence = self.issue_licence(user, decisions)
 
-        # Finalise Case
         old_status = self.status.status
-        self.status = get_case_status_by_status(CaseStatusEnum.FINALISED)
-        self.save()
+        finalised_status = get_case_status_by_status(CaseStatusEnum.FINALISED)
+
+        if old_status != finalised_status:
```
kevincarrogan, line 454: Could we flip this to `==` and bale out early instead of having all of this logic underneath this if statement?

api/f680/models.py
```diff
@@ -149,6 +152,40 @@ def get_required_decision_document_types(self):
...
+        notify_exporter_f680_outcome_issued(self)
+
+    def manage_decisions(self, user, decisions, **kwargs):
+        decision_actions = self.get_decision_actions()
+        ordered_decisions = sorted(decisions, reverse=True)
+        for advice_type in ordered_decisions:
+            decision_actions[advice_type](self)
+            audit_trail_service.create(
+                actor=user,
+                verb=AuditType.CREATED_FINAL_RECOMMENDATION,
```
saruniitr, line 180: I think this is not a valid event for F680 because if the decision is approve then it adds "added a decision of licence approved" to the timeline which is not correct.

We have F680 specific ones here https://github.com/uktrade/lite-api/blob/dev/api/audit_trail/payload.py#L170 but as this is for finalisation may be we need a separate event for this specific to F680?

saruniitr (commented):
all looks good except for the audit event in finalise method for F680.

saruniitr (approved):

PR #2537, opened 2025-05-01
Ltd 6081 show f680 application name in draft list

markj0hnst0n:
### Aim

Any draft application on the exporter application list page currently shows as `None` even after the general application details with the name have been filled in.  To fix this this PR promotes the application name to the db field for name which the list uses.  It takes this information from the json blob sent on `partial_update`

[LTD-6081](https://uktrade.atlassian.net/browse/LTD-6081)


api/f680/exporter/views.py
```diff
@@ -41,6 +41,27 @@ class F680ApplicationViewSet(viewsets.ModelViewSet):
...
+            .get("fields", {})
+            .get("name", {})
+            .get("answer", "")
+        )
+
+        if not application.name and incoming_application_name:
+            application.name = incoming_application_name
+            application.save()
+
+        return self.update(request, *args, **kwargs)
```
currycoder, line 63: It would be good to keep our API views quite lean where we can - historically overriding methods like this has been somewhere that debt has accumulated.

I believe you could achieve the same outcome by overriding the serializer's save method instead.  You could extract the name, set it on the serializer's instance and then call through to the super's save and it should work in the same way as this

markj0hnst0n: Lifted and shifted this across to serializer

api/f680/exporter/serializers.py
```diff
@@ -40,6 +40,22 @@ def create(self, validated_data):
...
+            .get("sections", {})
+            .get("general_application_details", {})
+            .get("fields", {})
+            .get("name", {})
+            .get("answer", "")
+        )
+
+        if self.instance and not self.instance.name and incoming_application_name:
+            self.instance.name = incoming_application_name
+            self.instance.save()
```
currycoder, line 55: I believe you can drop this particular save line and defer to the call to super doing it.  Does that work?

markj0hnst0n: Changed this and it still works grand 👍

currycoder (approved):

saruniitr (approved):

PR #2542, opened 2025-05-02
LTD-5953-bdd-f680-routing-tests 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5953

helper method which is used to create a baseline F680 Application.

### DIFF
--- a/test_helpers/clients.py
+++ b/test_helpers/clients.py
@@ -8,6 +8,7 @@ from typing import Tuple
 import django.utils.timezone
 from django.db import connection
 from django.test import override_settings
+from api.f680.tests.factories import F680ApplicationFactory
 from rest_framework.test import APITestCase, URLPatternsTestCase, APIClient
 import pytest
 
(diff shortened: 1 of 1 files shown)


markj0hnst0n (approved):

api/f680/conftest.py
```diff
@@ -17,624 +20,7 @@ def data_uae_release_id():
 
 
 @pytest.fixture
-def data_application_json(data_australia_release_id, data_france_release_id, data_uae_release_id):
```
depsiatwal, line 20: This has been shifted to the file. 
I was originally using pytest_plugins to reference this in BDD Routing so we only have a single point of reference for all this JSON. However appears this is no longer allowed in lower level conf files. Hence i've shifted the data to a common file that can be referenced and loaded where required. 

[ns.html#pytest-plugins-in-non-top-level-conftest-files](https://docs.pytest.org/en/stable/deprecations.html#pytest-plugins-in-non-top-level-conftest-files)

test_helpers/clients.py
```diff
@@ -670,6 +671,20 @@ def create_draft_standard_application(
...
+        organisation: Organisation,
+        data_application_json,
+        reference_code=None,
+    ):
+
+        f680_application = F680ApplicationFactory(
+            organisation=organisation,
+            application=data_application_json,
+            reference_code=reference_code,
+        )
```
saruniitr, line 685: We can simply use this instead of helper?
I actually tried to get rid of standard application helper and replaced with factories directly where possible. The problem is it gets overloaded with many other things. In this case all we need is a json so we can directly use the factory.

PR #2543, opened 2025-05-02
LTD-6199 - F680 manually finalised

marcuspp:
### Aim

To allow an F680 caseworker to manually mark an application as finalised

[LTD-6199](https://uktrade.atlassian.net/browse/LTD-6199)

[LTD-6199]: https://uktrade.atlassian.net/browse/LTD-6199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/applications/caseworker/permissions.py
+++ b/api/applications/caseworker/permissions.py
@@ -13,10 +13,17 @@ class CaseStatusCaseworkerChangeable(permissions.BasePermission):
         user = request.user.govuser
 
         if new_status == CaseStatusEnum.FINALISED:
-            lu_user = str(user.team.id) == TeamIdEnum.LICENSING_UNIT
-            if lu_user and user.has_permission(GovPermissions.MANAGE_LICENCE_FINAL_ADVICE):
-                return True
-            return False
+            application_manifest = application.get_application_manifest()
+
+            is_managing_user = str(user.team.id) == application_manifest.managing_team_id
+            if not is_managing_user:
+                return False
+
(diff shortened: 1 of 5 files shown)


api/applications/caseworker/permissions.py
```diff
@@ -13,10 +13,17 @@ def has_object_permission(self, request, view, application):
...
         if new_status == CaseStatusEnum.FINALISED:
-            lu_user = str(user.team.id) == TeamIdEnum.LICENSING_UNIT
-            if lu_user and user.has_permission(GovPermissions.MANAGE_LICENCE_FINAL_ADVICE):
-                return True
-            return False
+            application_manifest = application.get_application_manifest()
+
+            is_managing_user = str(user.team.id) == application_manifest.managing_team_id
+            if not is_managing_user:
+                return False
```
currycoder, line 20: I like this block but wondering if it may be better situated on `Case`/`BaseApplication`?  e.g.
```
def is_managing_caseworker(self, gov_user):
    ....
```

We then have a reusable method for wherever we might need to check this in future.

currycoder (approved):
One comment inline

saruniitr:
For my understanding, what is the use for this?

marcuspp:
After further discussion it seems this is not required for F680s at this time as there is a specific use case it used for SIELs.

PR #2551, opened 2025-05-09
Use a single manifest for licence application types

kevincarrogan:
### Aim

Allow the Export Licence type to be pushed through the caseworker side by using a single manifest file for both Export Licence and Standard Licence.

If they diverge in the future I'll split them out again, however this is to make sure that these stay inline until that point as updates were made to the StandardApplication at one point that broke the Export Licence journey.

[LTD-6212](https://uktrade.atlassian.net/browse/LTD-6212)


depsiatwal (approved):
nice to have this on FE also

PR #2553, opened 2025-05-09
LTD-6208-f680-documents-tab-incorrect-user 

depsiatwal:
A tidy up of the serializer UI changes here fir impacted change 
https://github.com/uktrade/lite-frontend/pull/2500


markj0hnst0n (approved):
Looks grand

api/applications/caseworker/views/applications.py
```diff
@@ -36,6 +36,7 @@ def post(self, request, pk):
 
 class ApplicationDocumentView(CaseworkerApplicationMixin, ListAPIView):
     serializer_class = ApplicationDocumentSerializer
+    pagination_class = None
```
currycoder, line 39: This is a compatibility breaking change, so just triple checking - it's only the F680 frontend views which use this endpoint?

depsiatwal: correct newly created for F680s I wanted to tidy this up before we lose the opportunity

currycoder (approved):

PR #2555, opened 2025-05-09
LTD-6085 Review lite-api caseworker view dispatch methods

markj0hnst0n:
### Aim

There is some reusable functionality in the F680s recommendation/outcome views so this has been pulled out into a mixin and I've also used a backend filter so that only F680 cases are filtered and `get_queryset` to generate 404 when no case is found.

[LTD-6085](https://uktrade.atlassian.net/browse/LTD-6085)

[LTD-6085]: https://uktrade.atlassian.net/browse/LTD-6085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/f680/caseworker/filters.py
+++ b/api/f680/caseworker/filters.py
@@ -2,11 +2,6 @@ from rest_framework import filters
 from api.cases.enums import CaseTypeSubTypeEnum
 
 
-class CurrentCaseFilter(filters.BaseFilterBackend):
-    def filter_queryset(self, request, queryset, view):
-        return queryset.filter(case_id=view.kwargs["pk"])
-
-
 class F680CaseFilter(filters.BaseFilterBackend):
     def filter_queryset(self, request, queryset, view):
         return queryset.filter(case__case_type__sub_type=CaseTypeSubTypeEnum.F680)
--- a/api/f680/caseworker/filters.py
+++ b/api/f680/caseworker/filters.py
@@ -1,6 +1,12 @@
 from rest_framework import filters
+from api.cases.enums import CaseTypeSubTypeEnum
 
 
 class CurrentCaseFilter(filters.BaseFilterBackend):
     def filter_queryset(self, request, queryset, view):
         return queryset.filter(case_id=view.kwargs["pk"])
+
+
+class F680CaseFilter(filters.BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        return queryset.filter(case__case_type__sub_type=CaseTypeSubTypeEnum.F680)
--- a/api/f680/caseworker/filters.py
+++ b/api/f680/caseworker/filters.py
@@ -1,10 +1,4 @@
 from rest_framework import filters
-from api.cases.enums import CaseTypeSubTypeEnum
-
-
-class F680CaseFilter(filters.BaseFilterBackend):
-    def filter_queryset(self, request, queryset, view):
-        return queryset.filter(case__case_type__sub_type=CaseTypeSubTypeEnum.F680)
 
 
 class CurrentCaseFilter(filters.BaseFilterBackend):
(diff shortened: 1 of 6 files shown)


api/f680/caseworker/views.py
```diff
@@ -89,33 +76,23 @@ def destroy(self, request, *args, **kwargs):
...
-            raise Http404()
-
-        return super().dispatch(request, *args, **kwargs)
-
-    def get_case(self):
-        return self.application
+    def get_queryset(self):
+        return SecurityReleaseOutcome.objects.filter(
+            case=self.case,
+        )
```
currycoder, line 95: TBH I'm a much bigger fan of this than `CurrentCaseFilter`

currycoder (approved):
Just one comment inline which I'll leave with you

api/f680/caseworker/filters.py
```diff
@@ -1,6 +1,7 @@
 from rest_framework import filters
+from api.cases.enums import CaseTypeSubTypeEnum
 
 
-class CurrentCaseFilter(filters.BaseFilterBackend):
+class F680CaseFilter(filters.BaseFilterBackend):
     def filter_queryset(self, request, queryset, view):
-        return queryset.filter(case_id=view.kwargs["pk"])
+        return queryset.filter(case__case_type__sub_type=CaseTypeSubTypeEnum.F680)
```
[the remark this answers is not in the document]
markj0hnst0n: Added `CurrentCaseFilter` back in and removed `get_queryset`

api/f680/caseworker/views.py
```diff
@@ -1,43 +1,31 @@
...
-            raise Http404()
-
-        return super().dispatch(request, *args, **kwargs)
-
-    def get_case(self):
-        return self.application
+    def get_queryset(self):
+        return Recommendation.objects.filter(
+            case=self.case,
+            case__status__status=CaseStatusEnum.OGD_ADVICE,
```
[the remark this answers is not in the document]
markj0hnst0n: Added `CurrentCaseFilter` back in and used permission which was the previous mechanism for checking if Case is ready for recommendation by looking at this status.

api/f680/caseworker/filters.py
```diff
@@ -1,4 +1,10 @@
 from rest_framework import filters
+from api.cases.enums import CaseTypeSubTypeEnum
+
+
+class F680CaseFilter(filters.BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        return queryset.filter(case__case_type__sub_type=CaseTypeSubTypeEnum.F680)
```
saruniitr, line 7: I still think this is redundant as it is covered by `CurrentCaseFilter()` where we go by pk so we are targeting the correct case.

api/f680/caseworker/views.py
```diff
@@ -54,8 +38,7 @@ def create(self, request, *args, **kwargs):
         data = self.prepare_data(request.data.copy())
         serializer = self.get_serializer(data=data, many=True)
         serializer.is_valid(raise_exception=True)
-
-        self.perform_create(serializer)
+        serializer.save()
```
saruniitr, line 41: `perform_create()` does the same correct?

saruniitr (approved):
one last comment that I leave it with you.

api/f680/caseworker/filters.py
markj0hnst0n: Agreed.  Removed and it's cleaner now

api/f680/caseworker/views.py
markj0hnst0n: Yeah they both do the exact same thing.  I had changed this to get the output data from the `save()` for something else I was doing.

PR #2560, opened 2025-05-13
LTD-6181: Add picklist type specific for F680 provisos

saruniitr:
## Change description

We currently have a common bucket for provisos for all case types. Create a separate type specific for F680 which helps users when adding their recommendation.

Each team need to create their own picklist entries once this is merged.
FE needs updating to filter F680 entries which will be in a separate PR.

Add new pytest based tests to cover create and list picklists.

Relevant FE PR is https://github.com/uktrade/lite-frontend/pull/2507

[LTD-6181]: https://uktrade.atlassian.net/browse/LTD-6181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/picklists/tests/test_views.py
```diff
@@ -1,10 +1,149 @@
...
+                TeamIdEnum.MOD_DSTL,
+                [PicklistType.LETTER_PARAGRAPH],
+            ),
+            (
+                TeamIdEnum.NCSC,
+                [PicklistType.PROVISO, PicklistType.F680_PROVISO, PicklistType.REPORT_SUMMARY],
+            ),
+        ),
+    )
+    def test_ogds_retrieve_picklist_items(self, get_hawk_client, team_case_advisor_headers, team_id, item_types):
```
saruniitr, line 76: The idea is to create all types for all teams so they are spread across.

api/picklists/tests/test_views.py
```diff
@@ -74,25 +245,3 @@ def test_gov_user_can_see_items_by_ids_filter(self):
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(len(response_data["results"]), 2)
-
-
-class PicklistView(DataTestClient):
-    def setUp(self):
```
saruniitr, line 80: replaced by new test which is pytest based.

saruniitr:
> any frontend changes how does this manifest in the FE ?

@depsiatwal yes a simple change required in FE.
The place where we retrieve picklists, currently we use `proviso` which needs changing to new type `f680_proviso`.

api/picklists/tests/test_views.py
```diff
@@ -1,98 +1,277 @@
...
+            team_id=TeamIdEnum.MOD_CAPPROT, type=PicklistType.FOOTNOTES, status=PickListStatus.DEACTIVATED
         )
-        response_data = response.json()
-
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertEqual(len(response_data["results"]), 1)
-
-    def test_gov_user_can_see_filtered_picklist_items_excluding_deactivated(self):
-        response = self.client.get(self.url + "?type=" + PicklistType.REPORT_SUMMARY, **self.gov_headers)
-        response_data = response.json()
```
saruniitr, line 57: These are replaced with updated tests.
The new tests added uses hawk authentication by default and as these are in the same file these started failing because they don't use hawk.
These could've been moved to another file but used this opportunity to improve them.

depsiatwal (approved):
all looks good along with the FE changes.

api/picklists/tests/test_views.py
currycoder: Nice one paying off this debt

PR #2565, opened 2025-05-15
LTD-6209 and LTD-6225 Security classification prefix logic

markj0hnst0n:
### Aim

In an F680 application we have `security_classification_prefix` question field for product and security release request journeys.  This is now required and must be presented to caseworkers as part of the `security_classification` variable.

This PR takes that answer from the json blob and on submission promotes it to a field in the database as either the choice from the drop down the user has chosen other the text they have added in the 'other' textbox.

It then combines the `security_classification_prefix` and `security_classification` as a model property called `security_classification_final` for the serializers to use to present.

[LTD-6209](https://uktrade.atlassian.net/browse/LTD-6209)
[LTD-6225](https://uktrade.atlassian.net/browse/LTD-6225)

[LTD-6209]: https://uktrade.atlassian.net/browse/LTD-6209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LTD-6225]: https://uktrade.atlassian.net/browse/LTD-6225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/f680/exporter/serializers.py
```diff
@@ -73,6 +73,8 @@ class UserItemFieldsSerializer(serializers.Serializer):
     end_user_name = FieldSerializer()
     address = FieldSerializer()
     country = FieldSerializer()
+    prefix = FieldSerializer()
```
currycoder, line 76: Normally I think we'd need to be quite careful about adding a new required field to the submit promotion; old draft applications may not have this field set and may run in to issues on submit.  However since we have very few people making applications in production, we can probably get away with it without doing anything special.

Out of interest, what happens when you have an application created pre-branch and submitted post-branch?

api/f680/models.py
```diff
@@ -125,11 +138,17 @@ def on_submit(self, application_data):
                 ),
             )
 
+            security_grading_prefix = item_fields["prefix"]["raw_answer"]
+            if security_grading_prefix == enums.SecurityGradingPrefix.OTHER:
+                security_grading_prefix = (
+                    item_fields["other_prefix"]["raw_answer"] if "other_prefix" in item_fields else None
+                )
```
currycoder, line 145: I believe this will save free text in to the `security_grading_prefix` field which is created as having set choices according to the model.  Django might not explicitly complain, but we should ensure that a choice field only contains values from the possible choices.  Probably you will need to follow the pattern for security grading e.g. an additional field `security_grading_prefix_other`

api/f680/models.py
```diff
@@ -207,29 +226,57 @@ class Product(TimestampableModel):
...
     security_grading_other = models.TextField(null=True, default=None)
     organisation = models.ForeignKey(Organisation, related_name="organisation_product", on_delete=models.CASCADE)
 
+    @property
+    def security_grading_final(self):
+        if self.security_grading == enums.SecurityGrading.OTHER:
+            security_grading = self.security_grading_other.upper() if self.security_grading_other else ""
+            security_grading_final = f"{self.security_grading_prefix.upper()} {security_grading}"
+        else:
+            security_grading_final = f"{self.security_grading_prefix.upper()} {self.security_grading.upper()}"
```
currycoder, line 244: Maybe you could avoid ifs here with...

```
@property
def security_grading_display(self):
    return f"{self.composed_security_grading_prefix} {self.composed_security_grading}"

@property
def composed_security_grading_prefix(self):
    return self.security_grading_prefix_other if self.security_grading_prefix == "other" else self.get_security_grading_prefix_display()

@property
def composed_security_grading(self):
    return self.security_grading_other if self.security_grading== "other" else self.get_security_grading_display()
```

api/f680/models.py
```diff
@@ -207,29 +226,57 @@ class Product(TimestampableModel):
...
+
+        return security_grading_final
+
 
 class Recommendation(TimestampableModel):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     case = models.ForeignKey(Case, related_name="recommendations", on_delete=models.CASCADE)
     type = models.CharField(choices=enums.RecommendationType.choices, max_length=30)
+    # prefix = models.CharField(choices=enums.SecurityGradingPrefix.prefix_choices, blank=True, null=True, max_length=50)
+    # prefix_other = models.TextField(null=True, default="", blank=True)
```
currycoder, line 279: Suggest dropping these comments until we get to it in another ticket.

api/f680/models.py
currycoder: `get_<foo>_display()` is a little convenience method on django ORM records for getting the display value for a given choicefield btw

api/f680/exporter/serializers.py
markj0hnst0n: Submitting an application without a prefix doesn't break anything.  It just has nothing to present to the front end when creating the `security_classifcation_final`.

The default is set to `None` in the model so I think that's what happens on create when it doesn't have that input.

api/f680/models.py
markj0hnst0n: You're right.  This was the one decision I took that I wasn't 100% about so I've amended it now.

api/f680/models.py
markj0hnst0n: Good idea.  Made these changes.

api/f680/enums.py
```diff
@@ -41,29 +55,29 @@ class SecurityGrading:
...
 
     product_choices = [
-        (UNCLASSIFIED, "Unclassified"),
-        (OFFICIAL, "Official"),
-        (OFFICIAL_SENSITIVE, "Official - sensitive"),
-        (RESTRICTED, "Restricted"),
-        (CONFIDENTIAL, "Confidential"),
-        (SECRET, "Secret"),
-        (TOP_SECRET, "Top-secret"),
+        (UNCLASSIFIED, "UNCLASSIFIED"),
```
depsiatwal, line 58: Are these changing to capital on the UI ?

api/f680/models.py
```diff
@@ -88,12 +88,23 @@ def on_submit(self, application_data):
...
+
         product = Product.objects.create(
             name=application_data["sections"]["product_information"]["fields"]["product_name"]["raw_answer"],
             description=application_data["sections"]["product_information"]["fields"]["product_description"][
                 "raw_answer"
             ],
             organisation=self.organisation,
+            security_grading_prefix=(
+                product_information_fields["prefix"]["raw_answer"]
+                if "security_classification" in product_information_fields
```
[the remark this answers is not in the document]
depsiatwal: sorry actually an easier way may be else ignore me

item_fields.get("other_prefix", {}).get("raw_answer", None)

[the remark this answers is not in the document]
markj0hnst0n: Added this and agree.  It looks a bit 'DRY'er now

api/f680/utils.py
```diff
@@ -0,0 +1,2 @@
+def json_to_security_grading_field_helper(json_object, field_name):
```
currycoder, line 1: Could this just be called `get_application_answer()`?

markj0hnst0n: Probably better.  That's changed

depsiatwal (approved):
lgtm

api/f680/caseworker/serializers.py
```diff
@@ -31,7 +32,7 @@ class Meta:
             "name",
             "description",
             "security_grading",
-            "security_grading_other",
```
currycoder, line 34: Would also recommend keeping `security_grading_other` in the responses until a final cleanup task where we go through all frontend code and make sure nothing is using these old fields.

api/f680/caseworker/tests/test_views.py
```diff
@@ -211,7 +211,7 @@ def test_f680_cases_list(self, get_hawk_client, get_f680_application, team_case_
             assert release_request.recipient.country.name == item["recipient"]["country"]["name"]
             assert release_request.recipient.name == item["recipient"]["name"]
             assert release_request.recipient.type == item["recipient"]["type"]["key"]
-            assert release_request.security_grading == item["security_grading"]["key"]
```
currycoder, line 214: For now it's probably good to leave this check in too

currycoder (approved):
A couple of small comments which I'll leave with you

PR #2567, opened 2025-05-15
LTD-6222: Update HCSAT Survey response to include case_type

saruniitr:
## Change description

Add `case_type` to survey response and also migrate data for existing responses.
Relevant FE PR is https://github.com/uktrade/lite-frontend/pull/2511

[LTD-6222](https://uktrade.atlassian.net/browse/LTD-6222)

[LTD-6222]: https://uktrade.atlassian.net/browse/LTD-6222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/survey/tests/test_crud.py
```diff
@@ -1,56 +1,97 @@
...
-        )
 
-    def test_create_survey(self):
-        url = reverse("survey:surveys")
+class TestSurveyResponseTests:
+
+    @pytest.mark.parametrize(
+        "request_data",
+        (
+            {},
```
saruniitr, line 43: Covers backward compatible case as currently we don't provide `case_type` from the FE

api/survey/models.py
```diff
@@ -16,6 +18,7 @@
 class SurveyResponse(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False, db_index=True)
     feedback_submission_date = AutoCreatedField(_("feedback_submission_date"))
+    case_type = models.ForeignKey(CaseType, on_delete=models.DO_NOTHING, null=True, blank=True)
```
currycoder, line 21: I don't think this field should be `null=True, blank=True`.  At the point that we have run the data migration, I believe we can drop these - right?  As every SurveyResponse created from then will have a value set

saruniitr: correct but when we create migrations using `makemigrations` Django complains,
`It is impossible to add a non-nullable field 'case_type' to surveyresponse without specifying a default. This is because the database needs something to populate existing rows.`

saruniitr: you mean specify a default of `siel` or `''` ? in this migration and in another migration make it non-nullable?
Once the FE changes are merged there will always be a value, also if the value is. not set we don't want to fail in this case as user is unlikely to submit the response again.

saruniitr: For this migration we need `null=True` to cover existing rows.
Added one more migration to drop this as we are populating this field where it is empty in this migration.

currycoder (approved):

api/survey/migrations/0003_alter_surveyresponse_case_type.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 4.2.20 on 2025-05-21 09:24
```
marcuspp, line 1: Can the migrations be combined into 1?

marcuspp (approved):

api/survey/migrations/0003_alter_surveyresponse_case_type.py
saruniitr: actually no because we want `null=True` as we have existing rows.
Once the data migration is applied then we can drop this, hence separated into multiple migrations.

PR #2568, opened 2025-05-15
LTD-6233-f680-submit-audit-events

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6233

Add Update on submission Audit logs and back populate the previous cases in prod.

### DIFF
--- a/api/f680/exporter/views.py
+++ b/api/f680/exporter/views.py
@@ -57,6 +57,7 @@ class F680ApplicationViewSet(viewsets.ModelViewSet):
 
         # TODO: some sort of validation that we have everything we need on the application -
         #   this may duplicate frontend validation in some way so needs some consideration.
+        old_application_status = application.status.status
         application.status = get_case_status_by_status(CaseStatusEnum.SUBMITTED)
         application.submitted_at = timezone.now()
         application.sla_remaining_days = get_application_target_sla(application.case_type.sub_type)
@@ -66,7 +67,7 @@ class F680ApplicationViewSet(viewsets.ModelViewSet):
         application.foi_reason = application_declaration_serializer.data.get("foi_reason", "")
 
         application.save()
-        application.on_submit(application_json_serializer.data)
+        application.on_submit(old_application_status, application_json_serializer.data)
 
         apply_flagging_rules_to_case(application)
         queues_assigned = run_routing_rules(application)
--- a/api/f680/models.py
+++ b/api/f680/models.py
@@ -82,9 +82,10 @@ class F680Application(BaseApplication, Clonable):
             return None
         return self.security_release_requests.first().product
 
-    def on_submit(self, application_data):
+    def on_submit(self, old_application_status, application_data):
         self.name = application_data["sections"]["general_application_details"]["fields"]["name"]["raw_answer"]
         self.save()
+        super().on_submit(old_application_status)
 
         product_information_fields = application_data["sections"]["product_information"]["fields"]
         # Create the Product for this application - F680s just have the one
--- a/api/f680/exporter/tests/test_views.py
+++ b/api/f680/exporter/tests/test_views.py
@@ -1,3 +1,6 @@
+from api.audit_trail.enums import AuditType
+from api.audit_trail.models import Audit
+from api.staticdata.statuses.enums import CaseStatusEnum
 import pytest
 from parameterized import parameterized
 from uuid import uuid4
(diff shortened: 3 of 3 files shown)


api/f680/exporter/views.py
```diff
@@ -66,7 +67,7 @@ def submit(self, request, **kwargs):
         application.foi_reason = application_declaration_serializer.data.get("foi_reason", "")
 
         application.save()
-        application.on_submit(application_json_serializer.data)
+        application.on_submit(old_application_status, application_json_serializer.data)
```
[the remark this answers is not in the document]
marcuspp: Another alternative is that the above logic is moved from the view into the `on_submit` function.

api/f680/models.py
```diff
@@ -82,9 +82,11 @@ def get_product(self):
             return None
         return self.security_release_requests.first().product
 
-    def on_submit(self, application_data):
+    def on_submit(self, old_application_status, application_data):
+
         self.name = application_data["sections"]["general_application_details"]["fields"]["name"]["raw_answer"]
         self.save()
+        super().on_submit(old_application_status)
```
marcuspp, line 89: Again this feels odd. Maybe rename the parent function to `audit_on_submit`

api/f680/exporter/views.py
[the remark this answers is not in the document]
depsiatwal: agree with all these comments the core issue is the the super class ATM expects the status and the model has no knowledge of the previous status. 

Hence why the view is passing it down , one option may be to override the save function on the base model but I do want to limit the impact here.

api/f680/models.py
depsiatwal: yeah as pairing how changed the base method to make this cleaner, also shifted some of the application settings  to the model.

api/f680/migrations/0013_add_missing_f680_case_submitted_events.py
```diff
@@ -0,0 +1,49 @@
...
+
+    with transaction.atomic():
+        # Audit UPDATED_STATUS was missed as initial release when submitting an f680 applicaion
+        # Reporting relies on this event so we added the event back but for the
+        # missing data this migration retrospectively adds for the f680 submittion events
+
+        F680Applications = F680Application.objects.all()
+
+        for f680_application in F680Applications:
+            Audit.objects.create(
```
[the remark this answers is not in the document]
depsiatwal: was just wondering about that when I see this comment changed

api/f680/exporter/views.py
```diff
@@ -57,15 +53,11 @@ def submit(self, request, **kwargs):
 
         # TODO: some sort of validation that we have everything we need on the application -
         #   this may duplicate frontend validation in some way so needs some consideration.
-        application.status = get_case_status_by_status(CaseStatusEnum.SUBMITTED)
-        application.submitted_at = timezone.now()
-        application.sla_remaining_days = get_application_target_sla(application.case_type.sub_type)
-        application.sla_days = 0
-        application.submitted_by = request.user.exporteruser
+        #
```
currycoder, line 56: Why are we getting rid of this and the below call to `save()`?

depsiatwal: it does a save in the model , whole thing is in transaction so assume no harm
issue with doing it outside is the status gets changes so we have to start tricking it down which looks a bit odd. 

.previous_status becomes submitted

currycoder: OK I see that it's been relocated, yeah.  This seems fine.  Would just recommend dropping this line with the `#` on it - assume that's overhanging from some debug

currycoder (approved):
Approved pending one comment about removing a `#`

api/f680/migrations/0013_add_missing_f680_case_submitted_events.py
```diff
@@ -4,35 +4,45 @@
...
 
-        F680Applications = F680Application.objects.all()
+        F680Applications = F680Application.objects.all().exclude(status__status=CaseStatusEnum.DRAFT)
 
         for f680_application in F680Applications:
-            Audit.objects.create(
-                created_at=f680_application.updated_at,
-                updated_at=f680_application.updated_at,
-                actor_object_id=SystemUser.id,
+            system_user = BaseUser.objects.all().first()
```
marcuspp, line 26: system_user = BaseUser.objects.get(id=SystemUser.id)

api/f680/migrations/0013_add_missing_f680_case_submitted_events.py
```diff
@@ -4,35 +4,45 @@
...
+    Case = apps.get_model("cases", "Case")
+    BaseUser = apps.get_model("users", "BaseUser")
 
     with transaction.atomic():
         # Audit UPDATED_STATUS was missed as initial release when submitting an f680 applicaion
         # Reporting relies on this event so we added the event back but for the
         # missing data this migration retrospectively adds for the f680 submittion events
 
-        F680Applications = F680Application.objects.all()
+        F680Applications = F680Application.objects.all().exclude(status__status=CaseStatusEnum.DRAFT)
```
marcuspp, line 23: `.all()` is not required

api/f680/migrations/0013_add_missing_f680_case_submitted_events.py
marcuspp: My comment was regarding how its selected. I'd use the id to select the system base user rather than presuming its the first base user.

PR #2574, opened 2025-05-20
LTD-6226 add security classification prefix to recommendation flows

markj0hnst0n:
### Aim

WIP

[LTD-6226](https://uktrade.atlassian.net/browse/LTD-6226)


api/f680/caseworker/read_only_serializers.py
```diff
@@ -39,16 +39,15 @@ class F680RecommendationViewSerializer(serializers.ModelSerializer):
...
+    security_grading_final = serializers.CharField(allow_blank=True, allow_null=True, required=False)
 
     class Meta:
         model = Recommendation
         fields = (
             "id",
             "created_at",
             "type",
-            "security_grading",
-            "security_grading_other",
```
currycoder, line 51: Can I suggest that we keep these in.  I think you've probably changed all of the references to security_grading in the frontend, but we may need the granularity of these fields for the outcome grouping work.

currycoder (approved):

